PR #36, opened 2015-01-28
Test OVAL details in HTML reports

jan-cerny:
Tests check that OVAL details are present in report when the rule fails. Tests use most popular objects from RHEL6 SSG.


isimluk:
You have split it to so many patches! that is so nice and cute and lovely. I am very happy.

isimluk:
Oh my, that was premature judgement. The first two patches does not work without the others. So if the separation to patches does not go strict (so after each there is workable tree) there is a little value in separating them. Please squash them.

tests/oval_details/test_oval_details.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
+# OpenSCAP test suite
+# Tests OVAL details in OpenSCAP reports
+# Last change: 29.1.2015
```
mpreisler: Why list Last change in the file? This will inevitably be the wrong value when somebody changes this file and doesn't notice this is there. If people want this info they can just do `git log $FILE`.

isimluk:
Thanks!

PR #40, opened 2015-02-04
Show info about objects and states in OVAL details in HTML report when o...

jan-cerny:
...bject could not be found


xsl/xccdf-report-oval-details.xsl
```diff
@@ -128,11 +132,110 @@ Authors:
...
+                            <tr>
+                        <xsl:variable name='variable_id' select='$state_info/*/@var_ref'/>
+                        <xsl:if test='$variable_id'>
+                            <td>
+                            <xsl:apply-templates select='ovalres:tested_variable'/>
+                            <xsl:apply-templates select='key("ovalsys-object",$object_id)'/>
+                            </td>
+                        </xsl:if>
+                            <xsl:apply-templates select='$state_info[1]'/>
+                        </tr>
```
mpreisler: Any reason why you aren't using @mode when applying those templates even though you use it elsewhere? Using modes is safer, it prevents templates eating up excess content.

The indentation seems off in the entire patch. And one of the hunks should be dropped because it just changes whitespace.

PR #45, opened 2015-02-09
Fix a double free.

jan-cerny:


mpreisler:
Doesn't this issue also exist on maint-1.0?

jan-cerny:
Yes, it should go to maint-1.0. It seems that "oid" should be set to NULL also.

PR #46, opened 2015-02-10
trca#352: <connection_string> should be exported before <sql> within <sq...

jan-cerny:
...l57_item>


mpreisler:
https://fedorahosted.org/openscap/ticket/352

mpreisler:
We need a regression test for this.

mpreisler:
After a brief discussion we found out that writing a proper regression test is quite hard and user would need to install sqlite or some other database. For this reason no test will be written for this.

PR #47, opened 2015-02-10
Fix a double free.

jan-cerny:


mpreisler:
Review done in #45 . This is the same patch targeted at maint-1.0.

PR #48, opened 2015-02-17
trac#447: Table of Contents added to generated HTML guide.

jan-cerny:


isimluk:
LGTM. @mpreisler, what do you think?

xsl/xccdf-guide-impl.xsl
```diff
@@ -351,6 +351,11 @@ Authors:
 
                 <td style="padding-left: {$indent * 19}px">
                     <h3>
+                        <xsl:if test="$indent=1 or $indent=2">
+                            <xsl:attribute name="id">
```
mpreisler: Why use title as id attribute when the id attribute would be much more suitable? It requires no translation.

xsl/xccdf-guide-impl.xsl
```diff
@@ -411,6 +416,42 @@ Authors:
     </xsl:if>
 </xsl:template>
 
+<xsl:template name="table-of-contents-items">
+    <xsl:param name="item"/>
+    <xsl:param name="level"/>
+    <ol>
+        <xsl:for-each select="$item/cdf:Group">
+            <xsl:if test="cdf:title">
```
mpreisler: In the rest of the guide and report we show ID when title is not available. I think we should do the same here instead of skipping that section.

xsl/xccdf-guide-impl.xsl
```diff
@@ -411,6 +416,42 @@ Authors:
...
+                    <a>
+                        <xsl:attribute name="href">
+                            <xsl:text>#</xsl:text>
+                            <xsl:value-of select="translate(cdf:title, ' ', '-')"/>
+                        </xsl:attribute>
+                        <xsl:value-of select="cdf:title"/>
+                    </a>
+                </li>
+            </xsl:if>
+            <xsl:if test="cdf:Group and $level=1">
```
mpreisler: Slightly changing the logic here would make the template way more reusable.

I suggest passing a parameter of how many levels are remaining, as well as current level. That way you can generate TOC of any depth when calling the template.

jan-cerny:
Issues fixed.

PR #51, opened 2015-02-17
Avoids possible dereferencing a null pointer.

jan-cerny:


isimluk:
Good.

But this error (benchmark missing in policy) is very unlikely and it will be hard to debug/spot when occurs. This function is called only from oscap tool and if it fails it will just return 1 to parent shell and not seying anything.

Please include error reporting like:

```
oscap_seterr(OSCAP_EFAMILY_OSCAP, "Could not find benchmark model for policy when generating fix for %s", ...
```

Thanks!

jan-cerny:
Error message added.

mpreisler:
Adding the xccdf_policy_get_id(policy) to the error message might help when somebody encounters this issue.

Could you please add it to the oscap_seterr call?

oscap_seterr(OSCAP_EFAMILY_OSCAP, "Could not find benchmark model for policy id='%s' when generating fixes.", xccdf_policy_get_id(policy));

policy_id is probably the best identifier we have in this error message.

jan-cerny:
added xccdf_policy_get_id

PR #55, opened 2015-03-03
trac#265: substituing xccdf:fix/xccdf:instance elements + test

jan-cerny:


isimluk:
Very good! I have had only minor comments.

isimluk:
LGTM!

The question is only to which branch we wanna push this. It applies well on the maint-1.0. Hence the target will not be maint-1.1 it will be either master or maint-1.0.

@mpreisler thoughts?

jan-cerny:
I tried to cherry-pick this commit to maint-1.0 and it works. Make check and make distcheck passes.

mpreisler:
I think we should merge this to maint-1.0. It doesn't break public API and it fixes issues, I see no reason why we shouldn't merge it to maint-1.0.

Jan: Nice work on this patch!

isimluk:
Merged to maint-1.0 as 17ccfedab76be86a16dffb36ce4dbb4bdbc46532.

Thanks Jan for the patch. Thanks Martin for the maint-1.0 ack!

PR #56, opened 2015-03-06
Tests that fixes containing unresolved elements can't be executed.

jan-cerny:


mpreisler:
Any reason why this is targeted at maint-1.1 instead of maint-1.0? Is this behavior broken in maint-1.0?

isimluk:
Lgtm, merged to maint-1.0 as 86016ae0ab0266d04279a00a3bc0663bde9aaa27.

Thanks Jan!

PR #57, opened 2015-03-06
OVAL details in HTML report for passed rules.

jan-cerny:


jan-cerny:
Shall I add a heading over "The tested object ... of type ... could not be found."? This doesn't look good to me.

snimek z 2015-03-09 09 26 01

mpreisler:
Yes, this isn't self documenting at all. It's confusing to me and I know what's going on :-D

mpreisler:
The first thing I would do is to group the message "The tested object xyz could not be found" and the metadata that is related to it - object details, state details. That way it will be obvious that all 3 are related.

But even then it's confusing. This will need some deeper UX thought.

isimluk:
There are three items in the screen shot.
 + Font size of the other two should be the same as the first.
 + blue banner helps me understand, perhaps we can use the very same blue banner
 + Let's use word `item` everywhere
 + Object and state details could be hidden by default and showed only when user hits `+details` or whatever (can go to other pr)

jan-cerny:
I think this looks better.
snimek z 2015-03-10 16 10 34

isimluk:
This is so much better!

isimluk:
What about expand (`+`), collapse (`-`) button? Perhaps in next PR?

mpreisler:
I think the titles of the sections are not very self documenting. This is what I propose:

1) Items found violating $condition
2) Items not found satisfying $condition
3) Items found satisfying $condition - for passed rules, not visible on this screenshot
4) ? Items not found violating $condition - also for passed rules, I believe we never show this with the current code

Considering that the user already clicked on the rule to see rule result details I don't think more expand / collapse buttons are needed.

I think indenting the sections may better emphasize what belongs together.

mpreisler:
The commit you added improves the situation but still falls short of my suggestion. 1) is not addressed by it.

jan-cerny:
now it writes found/not found everywhere

mpreisler:
We discussed the 50 limit with @jan-cerny and the last thing he will do before this is merged is test file sizes with limit 0, 50 and 100. If 100 limit has little impact on file size, we agreed to bump the limit to 100.

The results will be posted here.

jan-cerny:
The limit has only a little impact on the report size - only a few kilobytes. I generated report from three ARFs. Here is the list of generated files:

[jcerny@t440s arf]$ ll *.html
-rw-rw-r--. 1 jcerny jcerny 2171277 16. bře 15.59 ssg-rhel6-cscf-report0.html
-rw-rw-r--. 1 jcerny jcerny 2186715 16. bře 16.01 ssg-rhel6-cscf-report100.html
-rw-rw-r--. 1 jcerny jcerny 2186715 16. bře 15.59 ssg-rhel6-cscf-report50.html
-rw-rw-r--. 1 jcerny jcerny 2050179 16. bře 15.58 ssg-rhel6-report0.html
-rw-rw-r--. 1 jcerny jcerny 2094997 16. bře 16.00 ssg-rhel6-report100.html
-rw-rw-r--. 1 jcerny jcerny 2086276 16. bře 15.59 ssg-rhel6-report50.html
-rw-rw-r--. 1 jcerny jcerny 2212383 16. bře 15.58 ssg-rhel6-stig-report0.html
-rw-rw-r--. 1 jcerny jcerny 2247648 16. bře 16.01 ssg-rhel6-stig-report100.html
-rw-rw-r--. 1 jcerny jcerny 2241334 16. bře 16.00 ssg-rhel6-stig-report50.html

I think we can set the limit to 100.

jan-cerny:
Limit has been set to 100.

PR #58, opened 2015-03-11
Support for OVAL 5.11 for generating datastreams

jan-cerny:


mpreisler:
One thing that may be relevant here is that David Solin found issues with the new OVAL 5.11 schemas that break old valid content! @iankko also participated in that discussion so I think we should consult him about this.

isimluk:
@jan-cerny , have you tried to `make clean` and `make` of scap-security-guide master after this? (Not just Fedora content, but all)?

iankko:
@mpreisler 
From checking this proposal on Fedora 20 there truly will be issues (both when specifying '5.11' as OVAL version in XCCDF benchmark directly or when using this patch):
```
oscap ds sds-compose output/ssg-rhel6-xccdf-1.2.xml output/ssg-rhel6-ds.xml
File '/root/scap-security-guide/rpmbuild/BUILD/scap-security-guide-0.1.21/RHEL/6/output/ssg-rhel6-ds.xml' line 9487: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5}note': This element is not expected. Expected is ( {http://oval.mitre.org/XMLSchema/oval-common-5}note ).
File '/root/scap-security-guide/rpmbuild/BUILD/scap-security-guide-0.1.21/RHEL/6/output/ssg-rhel6-ds.xml' line 10539: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#linux}filepath': This element is not expected. Expected is ( {http://oval.mitre.org/XMLSchema/oval-definitions-5#linux}epoch ).
File '/root/scap-security-guide/rpmbuild/BUILD/scap-security-guide-0.1.21/RHEL/6/output/ssg-rhel6-ds.xml' line 12431: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#linux}filepath': This element is not expected. Expected is ( {http://oval.mitre.org/XMLSchema/oval-definitions-5#linux}epoch ).
File '/root/scap-security-guide/rpmbuild/BUILD/scap-security-guide-0.1.21/RHEL/6/output/ssg-rhel6-ds.xml' line 12437: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#linux}filepath': This element is not expected. Expected is ( {http://oval.mitre.org/XMLSchema/oval-definitions-5#linux}epoch ).
File '/root/scap-security-guide/rpmbuild/BUILD/scap-security-guide-0.1.21/RHEL/6/output/ssg-rhel6-ds.xml' line 12443: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#linux}filepath': This element is not expected. Expected is ( {http://oval.mitre.org/XMLSchema/oval-definitions-5#linux}epoch ).
OpenSCAP Error: Invalid SCAP Source Datastream (1.2) content in /root/scap-security-guide/rpmbuild/BUILD/scap-security-guide-0.1.21/RHEL/6/output/ssg-rhel6-ds.xml. [oscap_source.c:205]
make[1]: *** [content] Error 1
make[1]: Leaving directory `/root/scap-security-guide/rpmbuild/BUILD/scap-security-guide-0.1.21/RHEL/6'
error: Bad exit status from /root/scap-security-guide/rpmbuild/rpm-tmp.46oJcq (%build)


RPM build errors:
    Bad exit status from /root/scap-security-guide/rpmbuild/rpm-tmp.46oJcq (%build)
make: *** [rpm] Error 1
```
but IMHO this shouldn't be considered a show stopper.

The ```rpmverifyfile_object / rpmverifypackage_object``` ```minOccurs=1``` issue above (the ```version, release, epoch``` , and ```arch``` issue above is known issue:
  [1] http://making-security-measurable.1364806.n2.nabble.com/FOR-REVIEW-OVAL-language-version-5-11-1-Revert-the-rpmverifyfile-object-minOccurs-1-and-rpmverifypacn-tp7586293.html
 [2] https://github.com/OVALProject/Language/issues/238

and since we can't rely if the proposed change will be merged / approved for inclusion into OVAL 5.11.1 language version, we will need to fix the content anyway (it's possible to do it in the way so it would work both on OVAL 5.10 & OVAL.5.11 versions with changing like ~2 objects).

For the ```note``` issue mentioned above it would need to be investigated yet.

But IMHO we should switch OVAL 5.11 support on already. If there are issues with the content, we should fix them (better to notice them now, than later).

So FWIW regarding my opinion WRT to inclusion of this change, I am all for it - no objections from me regarding its inclusion.

Jan.

mpreisler:
OK, I agree about having visible loud fails sooner rather than later.

This pull request should probably be accompanied by a pull request for SSG that makes it valid under both OVAL 5.10 and 5.11 schemas.

iankko:
@mpreisler 
> This pull request should probably be accompanied by a pull request for SSG that makes it valid under both OVAL 5.10 and 5.11 schemas.

I can make those proposed patches for SSG content. But first need to find out the reason for the ```note``` issue.

iankko:
> I can make those proposed patches for SSG content. But first need to find out the reason for the note issue.

The corresponding SSG PR has been now submitted under:
* https://github.com/OpenSCAP/scap-security-guide/pull/478

Tested & verified once that change applied the content builds & works correctly on both:
OVAL 5.10 and OVAL 5.11 versions.

schemas/sds/1.2/scap-source-data-stream_1.2.xsd
```diff
@@ -9,7 +9,7 @@
     </xs:appinfo>
   </xs:annotation>
   <xs:import namespace="http://checklists.nist.gov/xccdf/1.2" schemaLocation="../../xccdf/1.2/xccdf_1.2.xsd"/>
-  <xs:import namespace="http://oval.mitre.org/XMLSchema/oval-definitions-5" schemaLocation="../../oval/5.10/oval-definitions-schema.xsd"/>
+  <xs:import namespace="http://oval.mitre.org/XMLSchema/oval-definitions-5" schemaLocation="../../oval/5.11/oval-definitions-schema.xsd"/>
```
mpreisler: Please include an XML comment around this change to make it more obvious that we changed the schema.

This is the official schema. It imports 5.10.

http://scap.nist.gov/schema/scap/1.2/scap-source-data-stream_1.2.xsd

mpreisler:
Jan added the comment I requested, so I am merging this.

PR #59, opened 2015-03-16
Added misssing parametr --skip-valid for oscap ds sds-compose

jan-cerny:


mpreisler:
Is --skip-valid available in the other actions of `oscap ds`? Like oscap ds sds-add?

Is it available in `oscap ds rds-*` actions?

I am not saying this pull request should add support for everything. I would just like to see where things are :-) From what I see --skip-valid is not listed when I call `oscap ds rds-split -h`.

jan-cerny:
I look again at the code and It seems that --skip-valid is recognized by all actions. I must improve the code and also documentate it in the help.

mpreisler:
Discussed this with @jan-cerny 

There are still documentation and help string fixes to be made. That's what is keeping this pull request from being merged.

jan-cerny:
I added "--skip-valid" for other actions and I edited help and manual page.

PR #60, opened 2015-03-16
Added schematron files for OVAL 5.11

jan-cerny:


jan-cerny:
all schematrons were generated automatically

mpreisler:
The second commit simply can't go in. You commented <note> elements even in CPEs and old OVALs.

In my opinion we should patch the schematron and XSD (if necessary) of OVAL 5.11 to be more lenient. It requires less changes and is more manageable than changing all the content files.

jan-cerny:
I agree with you.

jan-cerny:
Reverted

mpreisler:
In this case I think it might be better to `git rebase -i` and just leave out the two last commits. Then `git push --force`.

While it's not a big deal to merge a revert commit it bloats the repository and is not a good practice.

As promised I will work to fix the schemas, will update you here when that's done.

jan-cerny:
OK

mpreisler:
I have changed the schema to be more lenient in https://github.com/OpenSCAP/openscap/commit/8e1f69c0bacb974f604f56a12da5386e1bfa8c8f

PR #61, opened 2015-03-23
If environmentvariable58 probe is unable to collect the environment vari...

jan-cerny:
...ables of another process, an error must be reported. See https://github.com/OVALProject/Language/issues/4

### DIFF
--- a/src/OVAL/probes/independent/environmentvariable58.c
+++ b/src/OVAL/probes/independent/environmentvariable58.c
@@ -63,7 +63,7 @@ extern char **environ;
 
 static int read_environment(SEXP_t *pid_ent, SEXP_t *name_ent, probe_ctx *ctx)
 {
-	int err = 0, pid, empty, fd;
+	int err = 1, pid, empty, fd;
 	size_t env_name_size;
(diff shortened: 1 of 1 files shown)


jan-cerny:
Discussed with @isimluk. Thi is not the right way. It should be done differently. OpenSCAP should write a message in OVAL results.

jan-cerny:
Done a message in OVAL results.

isimluk:
Jan, we agreed that first commit is not correct. I think it should not be part of this pull request.

Note I cannot cherry-pick second commit as it depends on the first.

isimluk:
> Jan, we agreed that first commit is not correct.
> I think it should not be part of this pull request.
>
> Note I cannot cherry-pick second commit as it depends on the first.

@jan-cerny, can you please amend this pull request -- so it does not contain the first of the commits? Thanks!

jan-cerny:
First commit removed.

isimluk:
Merged as 92542c8d98e5b61a9d6a684f8aa876c79267ab86.

Thank You!

PR #62, opened 2015-03-26
Fixed bug in rpm epoch.

jan-cerny:


mpreisler:
I would really appreciate a test that demonstrates what failed before this fix. That way we will also be sure we don't regress.

jan-cerny:
See https://github.com/OpenSCAP/openscap/pull/64 - this fails before the fix.

isimluk:
I think this can be merged to maint-1.0

isimluk:
Merged as 44e1ae71c680b14c5c78a0737edb626c03b181fe. Thank You!

PR #63, opened 2015-03-26
Don't generate deprecated elements in rpmverifypackage_item.

jan-cerny:


jan-cerny:
Please How does the probe know what version of OVAL is processing?

mpreisler:
The elements are deprecated. From my understanding that means that users of the results should slowly stop relying on them. Why do we want to stop generating them? That seems a bit excessive to me. They are just deprecated, they haven't been removed in OVAL 5.11.

We should remove them when they are no longer part of the OVAL spec. In the meantime we probably want to generate all the deprecated elements we can.

Or am I missing something?

jan-cerny:
Quotation from the OVAL 5.11 schema:
"This item entity has been deprecated and will be removed in version 6.0 of the language."

This probably means that we shouldn't remove this now.

isimluk:
Right, we should probably not merge this until OVAL 6.0.

PR #64, opened 2015-03-30
Test for rpmverifypackage probe.

jan-cerny:
Tests proper evaluating of the rpmverifypackage_object.


mpreisler:
Is this test specific to the maint-1.2 branch? Any reason why we shouldn't instead add it to the maint-1.0 branch?

jan-cerny:
There isn't OVAL 5.11 schema in maint-1.0, so the validation fails on maint-1.0.

isimluk:
I have no issue merging fix and test to different branches now. Shortly, I will do merges before upstream 1.2.2 release.

isimluk:
Merged as a8ceee8df8fde177c9ae6459fe402aa7cdb36afc.

PR #67, opened 2015-04-07
Glob2regex

jan-cerny:


mpreisler:
Please consider using the description field in future pull requests. It's much better than assuming others remember everything :-P

Should I review this right now? Do you think it is ready for merging? What works? What doesn't?

src/OVAL/oval_component.c
```diff
@@ -996,6 +1051,22 @@ static int _oval_component_parse_SPLIT_tag(xmlTextReaderPtr reader,
...
+					   struct oval_parser_context *context, struct oval_component *component)
+{
+
+	__attribute__nonnull__(component);
+
+	oval_component_GLOB_t *glob_to_regex = (oval_component_GLOB_t *) component;
+	glob_to_regex->glob_noescape = (char *)xmlTextReaderGetAttribute(reader, BAD_CAST "glob_noescape");
+    if (glob_to_regex->glob_noescape == NULL) {
+        glob_to_regex->glob_noescape = oscap_strdup("false");
+    }
```
mpreisler: bad indentation above

src/OVAL/oval_component.c
```diff
@@ -1973,6 +2051,159 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_ESCAPE_REGEX(oval
...
+        default:
+            break;
+        }
+    }
+finish:
+    _string_append(regex, '$');
+    result = regex->str;
+    oscap_free(regex);
+    return result;
+}
```
mpreisler: The entire section above is indented with spaces.

t_string makes sense but I don't like that it's isolated in this file like this. Are you sure there is nothing in openscap common to use for this? If not, wouldn't it be better to rename t_string to something else and put it in common?

src/OVAL/oval_component.c
```diff
@@ -1973,6 +2051,159 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_ESCAPE_REGEX(oval
...
+    if (s->length + 1 >= s->capacity) {
+        if(realloc(s->str, s->capacity + INITIAL_CAPACITY) == NULL)
+            return;
+        s->capacity += INITIAL_CAPACITY;
+    }
+    s->str[s->length++] = c;
+    s->str[s->length] = '\0';
+}
+
+static char *_glob_to_regex (const char *glob, int noescape)
```
mpreisler: Discussed with @jan-cerny. Found cases where _glob_to_regex reads invalid memory. jcerny is working on a fix.

tests/API/OVAL/unittests/test_glob_to_regex.xml
```diff
@@ -0,0 +1,351 @@
...
+    <local_variable id="oval:x:var:19" version="1" comment="Applying the glob_to_regex function to the glob string." datatype="string">
+      <glob_to_regex glob_noescape="true">
+        <literal_component datatype="string">\*</literal_component>
+      </glob_to_regex>
+    </local_variable>
+    <local_variable id="oval:x:var:20" version="1" comment="Applying the glob_to_regex function to the glob string." datatype="string">
+      <glob_to_regex glob_noescape="true">
+        <literal_component datatype="string">\?</literal_component>
+      </glob_to_regex>
+    </local_variable>
```
mpreisler: Please add the naughty invalid globs that cause buffer overrun in the _glob_to_regex function right now. It would be valuable to make sure we don't regress in the future.

Maybe add them as a separate definition.

src/OVAL/oval_component.c
```diff
@@ -1973,6 +2051,159 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_ESCAPE_REGEX(oval
...
+    }
+    _string_append(regex, '^');
+    while(1) {
+        c = glob[i++];
+        switch (state) {
+        case START:
+            if (c == '\0') {
+                goto finish;
+            } else if (c == '?') {
+                _string_append(regex, '.');
```
mpreisler: According to http://man7.org/linux/man-pages/man7/glob.7.html this should be [^/] instead of .

? shouldn't match forward slash.

src/OVAL/oval_component.c
```diff
@@ -1973,6 +2051,159 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_ESCAPE_REGEX(oval
...
+            } else if (c == '*') {
+                _string_append(regex, '.');
+                _string_append(regex, '*');
```
mpreisler: Same as above.

* should expand to [^/]* rather than .*

Yes, this whole thing is a can of worms :-D

src/OVAL/oval_component.c
```diff
@@ -1973,6 +2051,159 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_ESCAPE_REGEX(oval
...
+            } else if (c == '.' || c == '|' || c == '^' || c == '(' || c == ')'
+                    || c == '{' || c == '}' || c == '+' ) {
```
mpreisler: What about $, shouldn't that also be escaped? I haven't actually checked the spec, so maybe I am wrong.

mpreisler:
We need special treatment for dotfiles. This will most likely be a nightmare. See https://github.com/OVALProject/Sandbox/blob/b139c75050c5b23ec9d7f1e5da187234c5efafb9/resources/x-glob-to-regex/oval-function-proposal-form-v%233.txt#L108

Also, we need to never match . and .. - current directory and parent directory. I am not sure whether this should be part of the glob_to_regex functionality. In practice this may not be an issue because we may not even try to match . or .. against any regex anywhere. -- https://github.com/OVALProject/Sandbox/blob/b139c75050c5b23ec9d7f1e5da187234c5efafb9/resources/x-glob-to-regex/oval-function-proposal-form-v%233.txt#L101

I think the latter can be dropped for now because it will be very hard to actually find a case where it matters to the output. The former is a real issue though. And unfortunately the spec is a bit vague.

mpreisler:
Spent a few minutes discussing this with @jan-cerny and @theinric . We think we have a solution for the dotfiles. The solution doubles the amount of states in the glob_to_regex state machine. Jan is working on implementing it.

jan-cerny:
I fixed some issues:
- fixed possibility of buffer overrun
- fixed expansion of ? and * (also in test)
- treatment dot files
- indentation

src/OVAL/oval_component.c
```diff
@@ -1973,6 +2051,268 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_ESCAPE_REGEX(oval
...
+				_string_append_char(regex, c);
+			}
+			break;
+		case LEFT_BRACKET:
+			if (c == '!') {
+				_string_append_char(regex, '^');
+			} else if (c == '\0') {
+				free(regex->str);
+				free(regex);
+				return NULL;
```
mpreisler: This is an error, shouldn't we use oscap_seterr here to report it?

src/OVAL/oval_component.c
```diff
@@ -1973,6 +2051,268 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_ESCAPE_REGEX(oval
...
+			if (c == '\\') {
+				if (noescape) {
+					_string_append_char(regex, '\\');
+				}
+			} else if (c == ']') {
+				state = NORMAL;
+			} else if (c == '\0') {
+				free(regex->str);
+				free(regex);
+				return NULL;
```
mpreisler: Also error

src/OVAL/oval_component.c
```diff
@@ -1973,6 +2051,268 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_ESCAPE_REGEX(oval
...
+			}
+			_string_append_char(regex, c);
+			break;
+		case ESCAPE:
+			if (c == '?' || c == '*' || c == '[' || c == ']') {
+				_string_append_char(regex, '\\');
```
mpreisler: Please add the comment we discussed so that people can immediately figure out why these 4 characters are treated differently.

tests/API/OVAL/unittests/test_glob_to_regex.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash
+
+result=`mktemp`
+
+set -e
+set -o pipefail
+
+$OSCAP oval eval --results $result $srcdir/test_glob_to_regex.xml || exit 1
```
isimluk: `exit 1` seems overabundant with `set -e`

src/OVAL/oval_component.c
```diff
@@ -1973,6 +2051,270 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_ESCAPE_REGEX(oval
...
+		default:
+			break;
+		}
+	}
+finish:
+	_string_append_char(regex, '$'); // regex must match only whole string
+	result = regex->str;
+	free(regex);
+	return result;
+}
```
isimluk: Could you please move `_glob_to_regex` to a separate file?

This file (`oval_component.c`) already holds numerous responsibilities. Some of the core concepts of `oval_component.c` are getting buried into the sand of static functions.

The `_glob_to_regex` function together with its static dependencies form a nice and straight forward module (even with only a single accessible function).

src/OVAL/oval_component.c
```diff
@@ -1973,6 +2051,270 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_ESCAPE_REGEX(oval
...
+		case ESCAPE:
+			// Only ?, *, [ and ] are escaped by a backslash
+			// Meaning of other characters is not affected by a backslash
+			if (c == '?' || c == '*' || c == '[' || c == ']') {
+				_string_append_char(regex, '\\');
+				_string_append_char(regex, c);
+			} else if (c == '\0') {
+				free(regex->str);
+				free(regex);
+				return NULL;
```
mpreisler: I am not sure but I think this is not an error. We most likely just want to output the backslash in case \\ is the last character.

jan-cerny:
Added a small test written in C to test the implemented oval_glob_to_regex() function.

ac_probes/configure.ac.tpl
```diff
@@ -569,6 +569,7 @@ AC_CONFIG_FILES([Makefile
 		src/OVAL/adt/Makefile
 		src/OVAL/results/Makefile
                  tests/API/OVAL/Makefile
+                 tests/API/OVAL/glob_to_regex/Makefile
```
mpreisler: bad indents

configure.ac
```diff
@@ -1372,6 +1372,7 @@ AC_CONFIG_FILES([Makefile
 		src/OVAL/adt/Makefile
 		src/OVAL/results/Makefile
                  tests/API/OVAL/Makefile
+                 tests/API/OVAL/glob_to_regex/Makefile
```
mpreisler: bad indents

mpreisler:
We discussed that we need better error reporting. In case the glob is invalid the user should see some specific error message that will help them fix it. There are already good places in the code to print the messages. Look into https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/OVAL/oval_component.c#L1992

mpreisler:
ACK. Good work, Jan. Kudos!

PR #71, opened 2015-04-22
Test oscap oval validate on OVAL 5.11 content.

jan-cerny:
We need to test `oscap oval validate`
and `oscap oval validate --schematron` on OVAL 5.11 content
to avoid mistakes in OVAL 5.11 schematron.

### DIFF
--- a/tests/probes/systemdunitdependency/test_validation.sh
+++ b/tests/probes/systemdunitdependency/test_validation.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+# Copyright 2015 Red Hat Inc., Durham, North Carolina.
+# All Rights Reserved.
+#
+# OpenScap Probes Test Suite.
+#
+# Authors:
+#   Jan Černý <jcerny@redhat.com>
+
+. ../../test_common.sh
+set -e
+
+DF="${srcdir}/test_probes_systemdunitdependency.xml"
+
+$OSCAP oval validate --schematron $DF
+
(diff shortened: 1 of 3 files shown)


mpreisler:
Does this really fail without `set -e`? Just brain-compiling this I think this will exit with 0 exit code no matter what.

Have you tried if the test fails with purposely invalid file?

jan-cerny:
Yes, I have tried it, It fails. But now, I find that it's just because it is the last command of the script :-) There should be definitely set -e

mpreisler:
Quoting openscap 1.2.2 docs:
```
oscap -> oval -> validate

Validate OVAL XML content

Usage: oscap [options] oval validate [options] oval-file.xml

Options:
   --definitions		 - Validate OVAL Definitions
   --variables  		 - Validate external OVAL Variables
   --syschar    		 - Validate OVAL System Characteristics
   --results    		 - Validate OVAL Results
   --schematron 		 - Use schematron-based validation in addition to XML Schema
```

It says --schematron uses schematron in addition to XSD schema. Wouldn't we achieve the same results if we just called the second command?

jan-cerny:
I looked into source code of oscap oval validate. You are true.

jan-cerny:
Removed first command

tests/probes/systemdunitdependency/test_validation.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+# Copyright 2015 Red Hat Inc., Durham, North Carolina.
+# All Rights Reserved.
+#
+# OpenScap Probes Test Suite.
```
mpreisler: ?

PR #72, opened 2015-04-28
Test - evaluating invalid regex in pattern match operations

jan-cerny:
We had to verify that an invalid regex leads to error in
pattern mattern match operations.


mpreisler:
Why maint-1.2?

isimluk:
imho maint-1.2 is ok ---> it depends on 5.11, which is only in maint-1.2

mpreisler:
Maybe I am missing something but to me it looks like it should work with OVAL 5.10 equally well. One line has to be changed to achieve that.

I haven't tested it, this is just brain-compiling.

isimluk:
Ah, I see... @jan-cerny, can you please rework this on top of maint-1.0?

jan-cerny:
Yes, It can go to maint-1.0. We only need to change it to OVAL 5.10. I verified that it works well on maint-1.0

PR #73, opened 2015-05-04
String improved

jan-cerny:
We had oscap_string_free, but we didn't use it nowhere.
We hide the definition of struct oscap_string from other modules.

### DIFF
--- a/src/common/oscap_string.c
+++ b/src/common/oscap_string.c
@@ -27,9 +27,23 @@
 
 #define INITIAL_CAPACITY 64
 
-oscap_string *oscap_string_new() {
-	oscap_string *s;
-	s = malloc(sizeof(oscap_string));
+/**
+ * String with unlimited length
+ * contains:
+ * - pointer to data,
+ * - actual length of string,
+ * - capacity of allocated memory
+ */
+struct oscap_string {
+	char *str;
+	unsigned int length;
+	unsigned int capacity;
+};
+
+struct oscap_string *oscap_string_new()
+{
+	struct oscap_string *s;
+	s = malloc(sizeof(struct oscap_string));
 	if (s == NULL)
 		return NULL;
 	s->str = malloc(INITIAL_CAPACITY);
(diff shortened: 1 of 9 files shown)


jan-cerny:
changed "oscap_string_get_string" to better name "oscap_string_get_cstr" (+ comment)

src/common/oscap_string.c
```diff
@@ -79,3 +92,8 @@ bool oscap_string_append_string(oscap_string *s, const char *t)
 	s->str[s->length] = '\0';
 	return true;
 }
+
+const char *oscap_string_get_cstr(oscap_string *s)
```
mpreisler: shouldn't oscap_string also be const here?

src/common/oscap_string.h
```diff
@@ -27,19 +27,7 @@
...
- * - pointer to data,
- * - actual length of string,
- * - capacity of allocated memory
- */
-typedef struct {
-	char *str;
-	unsigned int length;
-	unsigned int capacity;
-} oscap_string;
+typedef struct string_struct oscap_string;
```
mpreisler: Look at the conventions we use elsewhere.

jan-cerny:
@mpreisler : Should I use "struct oscap_string ....." instead of definition of new data type "oscap_string". Should I change the headers of functions?

mpreisler:
@jan-cerny Look at oscap_list for an example. I am not saying the convention you used is wrong, I am just saying it's inconsistent with what openscap does. There is value in doing things consistently.

jan-cerny:
I changed this, I tried to follow the conventions used in oscap_list

src/common/oscap_string.c
```diff
@@ -22,60 +22,66 @@
...
+ * String with unlimited length
+ * contains:
+ * - pointer to data,
+ * - actual length of string,
+ * - capacity of allocated memory
+ */
+struct oscap_string {
+	char *str;
+	unsigned int length;
+	unsigned int capacity;
```
mpreisler: *size_t* would work better than *unsigned int* here.

PR #74, opened 2015-05-04
Test - evaluating invalid regex in pattern match operations

jan-cerny:
We had to verify that an invalid regex leads to error in
pattern mattern match operations.


mpreisler:
Is this exactly the same as #72 except for maint-1.0 branch and OVAL 5.11 changed to OVAL 5.10?

jan-cerny:
Yes, it is.

isimluk:
Merged as f10db8c1917e04c3767bc1a26041d87cfc737f01. Thanks!

PR #75, opened 2015-05-05
Checking for remount, bind and move mount options

jan-cerny:
In OVAL 5.11, when evaluating partition_object, we need to check for some mount options which aren't listed in /proc/mounts, but can be retrieved by statvfs().
Some info about this in this mailing list
http://making-security-measurable.1364806.n2.nabble.com/Partition-test-more-questions-tp6348424p7580867.html

I have also re-factored the code.


isimluk:
Merged as 2873b1e6cfd88db4dd5227e2e18a5bf75c801031. Thank You!

PR #76, opened 2015-05-12
rhbz#1220262: Modify OVAL-5.11 schema files to be backward compatible

isimluk:
Previously, OVAL-5.10 allowed for missing epoch/version/release/arch.
The upstream OVAL-5.11 schema is now more strict and not backward
compatible.

We need to make OVAL-5.11 schema backward compatible to support
DataStream content that includes such OVAL-5.10 content. Note that
DataStream 1.2 schema imports fixed version of OVAL schemas.

Addressing:
File '/tmp/ssg-rhel6-ds.xml' line 8755: Element
'{http://oval.mitre.org/XMLSchema/oval-definitions-5#linux}filepath':
This element is not expected. Expected is (
{http://oval.mitre.org/XMLSchema/oval-definitions-5#linux}epoch ).

### DIFF
--- a/schemas/oval/5.11/linux-definitions-schema.xsd
+++ b/schemas/oval/5.11/linux-definitions-schema.xsd
@@ -1110,7 +1110,8 @@
                                                   <xsd:documentation>This is the package name to check.</xsd:documentation>
                                              </xsd:annotation>
                                         </xsd:element>
-                                        <xsd:element name="epoch">
+                                        <!-- minOccurs/maxOccurs here is OpenSCAP modification to allow older contents to pass -->
+                                        <xsd:element name="epoch" minOccurs="0" maxOccurs="1">
                                              <xsd:annotation>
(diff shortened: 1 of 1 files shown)


mpreisler:
There are 2 instances of these elements (epoch, release, ...) in the XSD that is changed by this PR. Only one of them is affected. Is that expected?

The state ones are changed, the object ones are still required. In OVAL 5.10 the object ones and the state ones are both optional. But in OVAL 5.10.1 the object ones are all required.

Compare https://github.com/OpenSCAP/openscap/blob/81a0efbd86d66bd250e87efa5c52cc75ba08250b/schemas/oval/5.10/linux-definitions-schema.xsd#L1468 and https://github.com/OpenSCAP/openscap/blob/81a0efbd86d66bd250e87efa5c52cc75ba08250b/schemas/oval/5.10.1/linux-definitions-schema.xsd#L1468

Just something I noticed. Either way I am merging this because it fixes a regression. If further fixes are needed they can be done separately.

isimluk:
Reported upstream as OVALProject/Language#238.

PR #77, opened 2015-05-15
Symlink

jan-cerny:
New probe implementing symlink_object from OVAL 5.11 + test


mpreisler:
Doesn't build on RHEL6.

```
../../../../src/OVAL/probes/unix/symlink.c: In function ‘collect_symlink’:
../../../../src/OVAL/probes/unix/symlink.c:63: error: format not a string literal and no format arguments
../../../../src/OVAL/probes/unix/symlink.c:92: error: format not a string literal and no format arguments
../../../../src/OVAL/probes/unix/symlink.c:120: error: format not a string literal and no format arguments
```

jan-cerny:
Hopefully fixed these errors.

src/OVAL/probes/unix/symlink.c
```diff
@@ -0,0 +1,159 @@
...
+		probe_cobj_add_msg(probe_ctx_getresult(ctx), msg);
+		SEXP_free(msg);
+		probe_cobj_set_flag(probe_ctx_getresult(ctx), SYSCHAR_FLAG_ERROR);
+		oscap_free(linkname);
+		return 0;
+	}
+
+	item_sexp = probe_item_create(OVAL_UNIX_SYMLINK, NULL,
+			"filepath", OVAL_DATATYPE_STRING, pathname,
+			"canonical_path", OVAL_DATATYPE_STRING, linkname,
```
mpreisler: shouldn't canonical_path also be absolute?

mpreisler:
According to https://oval.mitre.org/language/version5.11/ovaldefinition/documentation/unix-definitions-schema.html symlink_object the canonical should point to the final destination, not just one step forward.

If there are a->b, b->c, d->e links and filepath is abs(a) it should return abs(e), not abs(b) as it does right now.

mpreisler:
Please research what we should when filepath is relative, I checked the spec for symlink test and it's quite vague. Check file test and possibly even other tests.

mpreisler:
Discussed this with @jan-cerny , we found realpath and oscap_acquire_guess_realpath. Jan will look into whether this can be used in the probe. If that is the case it will make the code much simpler.

jan-cerny:
I tried to re-implement this using realpath() and it seems working well. It definitely makes the code simpler. And it reaches the symlink to the final destination as we want.
The oscap_acquire_guess_realpath() is not suitable, because it handles the error on realpath() very different way than we need. We need to raise an error message, but oscap_acquire_guess_realpath() tries just to guess what the canonical path would be if the file would exist.
I will enlarge the test now and if it will work properly, I will push the commits here then.

jan-cerny:
Regarding the relative filepath in element <filepath>: Objects like file_object, filehash58_object, fileextendedattribute_object, textfilecontent54_object, have also element named <filepath>, and they require absolute path there:
> The filepath element specifies the absolute path for a file on the machine. A directory cannot be specified as a filepath.

I am not sure what the probe should do when there will be a relative path in <filepath> element. I think it should result in a state of "does not exist".

mpreisler:
576ca1d builds and checks finished successfully on RHEL6

PR #79, opened 2015-05-21
oscap-ssh: hacked to allow CPE, OVAL file uploads (RHEL7)

openprivacy:
Not expecting this to get merged as is - still a work in progress - but wanted to share what I've been up to so that I can run oscap-ssh against a RHEL7 server with a command like:
```
oscap-ssh root@example.com xccdf eval --profile stig-rhel7-server-upstream \
--results results.xml --report report.html \
--cpe scap/ssg-rhel7-cpe-dictionary.xml scap/ssg-rhel7-xccdf.xml \
scap/ssg-rhel7-oval.xml scap/ssg-rhel7-cpe-oval.xml
```
Note: the scap sub-directory is a symlink:
```
scap -> /home/fen/workspace/scap-security-guide/RHEL/7/dist/content
```
I have changed the argument processing to enable additional files to be scp'd to the remote server, and have changed the --cpe arg to preserve the filename. Besides some hackish ugliness, the new arg processing loses immediate error catching as there is no longer a reliable count of arguments, so any unknown flag causes drop-through to the LOCAL_CONTENT_PATH processing.

But I think the ability to pass additional (e.g. OVAL) files is very useful.

### DIFF
--- a/utils/oscap-ssh
+++ b/utils/oscap-ssh
@@ -91,24 +91,33 @@ fi
 
 SSH_HOST="$1"
 SSH_PORT="$2"
+shift 2
+rargs=''
 
-if [ "$3" == "--v" ] || [ "$3" == "--version" ]; then
-    true
-elif [ "$3" == "-h" ] || [ "$3" == "--help" ]; then
-    true
-elif [ "$3" == "info" ]; then
-    true
-elif [ "$3 $4" == "xccdf eval" ]; then
-    true
-elif [ "$3 $4" == "oval eval" ]; then
-    true
-elif [ "$3 $4" == "oval collect" ]; then
-    true
+IDX=1
+if [ "$1" == "--v" ] || [ "$1" == "--version" ]; then
+    rargs[0]="$1"
+elif [ "$1" == "-h" ] || [ "$1" == "--help" ]; then
+    rargs[0]="$1"
+elif [ "$1" == "info" ]; then
+    rargs[0]="$1"
+elif [ "$1 $2" == "xccdf eval" ]; then
+    rargs[0]="$1"
+    rargs[1]="$2"
+    ((IDX++)) ; shift
+elif [ "$1 $2" == "oval eval" ]; then
+    rargs[0]="$1"
+    rargs[1]="$2"
+    ((IDX++)) ; shift
+elif [ "$1 $2" == "oval collect" ]; then
+    rargs[0]="$1"
+    rargs[1]="$2"
+    ((IDX++)) ; shift
 else
     die "This script only supports '-h', '--help', '--v', '--version', 'info', 'xccdf eval', 'oval eval' and 'oval collect'."
 fi
 
-shift 2
+shift
 
 MASTER_SOCKET_DIR=$(mktemp -d)
 MASTER_SOCKET="$MASTER_SOCKET_DIR/ssh_socket"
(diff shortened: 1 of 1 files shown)


jan-cerny:
It looks interesting. I am not sure that we need functionality of just copying files, the only value I see is that this reduces typing on a keyboard. I think it's a question to discuss more.
Moreover, if we want to pass additional files, we should distinguish somehow which of them will be an argument of oscap, and which of them will be just scp'd, because it can be unclear to a user that it is the first of the files. The argument processing gets complicated and the code is not self-explaining. (There are many "shift"s, I have lost in it.) Also, I think error handling should be solved before we can merge this.

mpreisler:
Hi @openprivacy , just curious, why not scan with the source datastream? It contains everything needed and you completely avoid this problem.

openprivacy:
Jan - I was thinking that - just as oscap can have multiple oval-definition-files after the INPUT_FILE, oscap-ssh could do the same. And yes, the argument processing is ugly - was meant to be more of a proof of concept for comments - thank you for yours!

Martin - Thanks for the pointer to datastreams - I just read your blog about them at http://martin.preisler.me/2012/12/introduction-to-scap-datastreams-with-openscap/ but I am still a bit confused. Do I need to make a datastream file that contains the oval, xccdf, etc. sub-parts? When I try e.g.:

oscap-ssh root@52.6.222.3 22 xccdf eval --results-arf results.xml scap/ssg-rhel7-ds.xml

it does not appear to actually run the tests included in the CPE/OVAL files (that incidentally my hacked code does). I'd love to use datastreams for remote scanning - can they currently be used for RHEL7 (local or remote) or does the datastream format not yet exist? How can I help?

mpreisler:
@openprivacy , all of this is supported and has been supported for at least 2 years :-) All CPE dictionaries in the datastream get auto-registered so they should be evaluated without any extra effort. Maybe your datastream doesn't contain the CPE dicts or CPE OVAL files? Could you paste `oscap info scap/ssg-rhel7-ds.xml`? Or is it the official SSG RHEL7 datastream? This is what I get from the official datastream:
```
$ oscap info ssg-rhel7-ds.xml 
Document type: Source Data Stream
Imported: 2015-05-21T15:40:03

Stream: scap_org.open-scap_datastream_from_xccdf_ssg-rhel7-xccdf-1.2.xml
Generated: (null)
Version: 1.2
Checklists:
	Ref-Id: scap_org.open-scap_cref_ssg-rhel7-xccdf-1.2.xml
		Profiles:
			xccdf_org.ssgproject.content_profile_test
			xccdf_org.ssgproject.content_profile_pci-dss
			xccdf_org.ssgproject.content_profile_rht-ccp
			xccdf_org.ssgproject.content_profile_common
			xccdf_org.ssgproject.content_profile_stig-rhel7-server-upstream
		Referenced check files:
			ssg-rhel7-oval.xml
				system: http://oval.mitre.org/XMLSchema/oval-definitions-5
Checks:
	Ref-Id: scap_org.open-scap_cref_ssg-rhel7-oval.xml
	Ref-Id: scap_org.open-scap_cref_output--ssg-rhel7-cpe-oval.xml
	Ref-Id: scap_org.open-scap_cref_output--ssg-rhel7-oval.xml
Dictionaries:
	Ref-Id: scap_org.open-scap_cref_output--ssg-rhel7-cpe-dictionary.xml
```

As you can see the CPE dict and OVAL are both there.

If you are using SCAP Security Guide you don't have to create any datastreams, they are already created automatically.

openprivacy:
I got the datastream (I am using the SCAP Security Guide) to work - thank you for your help and pointers. I used the following incantation:

    oscap-ssh root@52.6.222.3 22 xccdf eval --results-arf results.xml \
    --profile xccdf_org.ssgproject.content_profile_stig-rhel7-server-upstream scap/ssg-rhel7-ds.xml

I'll close this and abandon my efforts to enable trailing OVAL files to be included. If however, you'd like to include such functionality, please reopen and indicate where you would like to see the effort concentrated.

PR #81, opened 2015-05-28
trac#461: We should populate child elements of OVAL external_variable.

jan-cerny:
Collect elements possible_value and possible_restriction
and propagate them to OVAL results.

I don't think that this is ready for merge. I have put it here because I would appreciate any suggestions or notes.


jan-cerny:
Build successful on Jenkins.

src/OVAL/oval_variable.c
```diff
@@ -84,6 +84,85 @@ typedef struct {
...
+	pr->hint = oscap_strdup(hint);
+	pr->restrictions = oval_collection_new();
+	return pr;
+}
+
+void oval_variable_possible_restriction_free(struct oval_variable_possible_restriction *pr)
+{
+	if (pr != NULL) {
+		oscap_free(pr->hint);
+		if (pr->restrictions != NULL) {
```
mpreisler: can this ever be NULL?

src/OVAL/oval_variable.c
```diff
@@ -182,6 +261,52 @@ struct oval_value_iterator *oval_variable_get_values(struct oval_variable *varia
...
+		oval_variable_EXTERNAL_t *var = (oval_variable_EXTERNAL_t *) variable;
+		possible_values = var->possible_values;
+	} else {
+		possible_values = NULL;
+	}
+	if (possible_values != NULL) {
+		return oval_collection_iterator(possible_values);
+	} else {
+		return oval_collection_iterator_new();
+	}
```
mpreisler: This entire function looks way more complex than it needs to be...

src/OVAL/oval_variable.c
```diff
@@ -182,6 +261,52 @@ struct oval_value_iterator *oval_variable_get_values(struct oval_variable *varia
...
+		oval_variable_EXTERNAL_t *var = (oval_variable_EXTERNAL_t *) variable;
+		possible_restrictions = var->possible_restrictions;
+	} else {
+		possible_restrictions = NULL;
+	}
+	if (possible_restrictions != NULL) {
+		return oval_collection_iterator(possible_restrictions);
+	} else {
+		return oval_collection_iterator_new();
+	}
```
mpreisler: Again, way more complex than it needs to be.

src/OVAL/oval_variable.c
```diff
@@ -419,6 +546,45 @@ struct oval_variable *oval_variable_clone(struct oval_definition_model *new_mode
...
+			if (oval_collection_iterator_has_more(old_pv_itr)) {
+				evar->possible_values = oval_collection_new();
+			}
+			while (oval_collection_iterator_has_more(old_pv_itr)) {
+				struct oval_variable_possible_value *old_pv, *new_pv;
+				old_pv = oval_collection_iterator_next(old_pv_itr);
+				new_pv = oval_variable_possible_value_new(old_pv->hint, old_pv->value);
+				oval_collection_add(evar->possible_values, new_pv);
+			}
+			oval_collection_iterator_free(old_pv_itr);
```
mpreisler: Please implement separate clone functions for both types and just use them there.

Chances are somebody else will need them and it will make this function simpler and easier to follow.

src/OVAL/oval_variable.c
```diff
@@ -340,6 +465,8 @@ struct oval_variable *oval_variable_new(struct oval_definition_model *model, con
 				return NULL;
 
 			evar = (oval_variable_EXTERNAL_t *) variable;
+			evar->possible_values = NULL;
```
mpreisler: Is it worth the trouble to keep track of whether this is or isn't NULL?

Would we be better off with always allocating a collection here?

src/OVAL/oval_variable.c
```diff
@@ -781,9 +1013,39 @@ static xmlNode *_oval_VARIABLE_CONSTANT_to_dom(struct oval_variable *variable, x
...
+		struct oval_variable_possible_value *pv = oval_collection_iterator_next(possible_values);
+		xmlNode *possible_value_node = xmlNewTextChild(variable_node, ns_definitions, BAD_CAST "possible_value", BAD_CAST pv->value);
+		xmlNewProp(possible_value_node, BAD_CAST "hint", BAD_CAST pv->hint);
+	}
+	oval_collection_iterator_free(possible_values);
+
+	struct oval_iterator *possible_restrictions = oval_variable_get_possible_restrictions(variable);
+	while (oval_collection_iterator_has_more(possible_restrictions)) {
+		struct oval_variable_possible_restriction *pr = oval_collection_iterator_next(possible_restrictions);
+		xmlNode *possible_restriction_node = xmlNewTextChild(variable_node, ns_definitions, BAD_CAST "possible_restriction", NULL);
```
mpreisler: we always create the possible_restrictions node even if it ends up being empty?

src/OVAL/public/oval_definitions.h
```diff
@@ -1612,6 +1612,49 @@ struct oval_variable *oval_variable_clone(struct oval_definition_model *new_mode
 void oval_variable_free(struct oval_variable *);
 
 /**
+ * Construct new instance of possible_value element.
+ * @param hint A short description of what the value means or represents.
+ * @param value An expected value of an external variable
+ * @memberof oval_variable
```
mpreisler: \@memberof is wrong

mpreisler:
The \@memberof tag is wrong in most of the function docs.

src/OVAL/oval_variable.c
jan-cerny: Yes, but such input would not pass the validation, because specification says that a <possible_restriction> element must have at least one <restriction> child element.

src/OVAL/oval_variable.c
jan-cerny: I think it can't, because I allocate it in the constructor. I think I can remove the if condition.

src/OVAL/oval_variable.c
jan-cerny: We should not have a function in our API which can create invalid OVAL output. I will fix this.

jan-cerny:
I have done some improvements following your comments.

jan-cerny:
Build successful on Jenkins

mpreisler:
I would really appreciate a few tests of the functionality. It looks like that should be simple to do reliably.

src/OVAL/oval_variable.c
```diff
@@ -182,6 +259,31 @@ struct oval_value_iterator *oval_variable_get_values(struct oval_variable *varia
...
+{
+	if (variable->type == OVAL_VARIABLE_EXTERNAL) {
+		oval_variable_EXTERNAL_t *var = (oval_variable_EXTERNAL_t *) variable;
+		return oval_collection_iterator(var->possible_restrictions);
+	} else {
+		return oval_collection_iterator_new();
+	}
+}
+
+struct oval_iterator *oval_variable_get_restrictions(struct oval_variable_possible_restriction *possible_restriction)
```
mpreisler: Takes struct oval_variable_possible_restriction* as first arg but is called oval_variable_get_restrictions.

src/OVAL/oval_variable.c
```diff
@@ -419,6 +523,39 @@ struct oval_variable *oval_variable_clone(struct oval_definition_model *new_mode
...
+				oval_collection_add(evar->possible_values, new_pv);
+			}
+			oval_collection_iterator_free(old_pv_itr);
+
+			struct oval_iterator *old_pr_itr;
+			old_pr_itr = oval_variable_get_possible_restrictions(old_variable);
+			while (oval_collection_iterator_has_more(old_pr_itr)) {
+				struct oval_variable_possible_restriction *old_pr, *new_pr;
+				old_pr = oval_collection_iterator_next(old_pr_itr);
+				new_pr = oval_variable_possible_restriction_new(old_pr->operator, old_pr->hint);
```
mpreisler: This allocates *new_pr->restrictions* and then line 546 possibly allocates it again, leaking memory.

src/OVAL/oval_variable.c
```diff
@@ -419,6 +523,39 @@ struct oval_variable *oval_variable_clone(struct oval_definition_model *new_mode
...
+				struct oval_iterator *old_r_itr;
+				old_r_itr = oval_variable_get_restrictions(old_pr);
+				if (oval_collection_iterator_has_more(old_r_itr)) {
+					new_pr->restrictions = oval_collection_new();
+				}
+				while (oval_collection_iterator_has_more(old_r_itr)) {
+					struct oval_variable_restriction *old_r, *new_r;
+					old_r = oval_collection_iterator_next(old_r_itr);
+					new_r = oval_variable_restriction_new(old_r->operation, old_r->value);
+					oval_collection_add(new_pr->restrictions, new_r);
```
mpreisler: Use *oval_variable_possible_restriction_add_restriction* instead.

src/OVAL/oval_variable.c
```diff
@@ -419,6 +523,39 @@ struct oval_variable *oval_variable_clone(struct oval_definition_model *new_mode
...
 			evar = (oval_variable_EXTERNAL_t *) new_variable;
 			old_evar = (oval_variable_EXTERNAL_t *) old_variable;
+
+			struct oval_iterator *old_pv_itr;
+			old_pv_itr = oval_variable_get_possible_values(old_variable);
+			while (oval_collection_iterator_has_more(old_pv_itr)) {
+				struct oval_variable_possible_value *old_pv, *new_pv;
+				old_pv = oval_collection_iterator_next(old_pv_itr);
+				new_pv = oval_variable_possible_value_new(old_pv->hint, old_pv->value);
+				oval_collection_add(evar->possible_values, new_pv);
```
mpreisler: Would be nice if you provided an API to add possible value and then used it here.

src/OVAL/oval_variable.c
```diff
@@ -670,6 +811,61 @@ static int _oval_variable_parse_local_tag(xmlTextReaderPtr reader, struct oval_p
...
+	oval_variable_EXTERNAL_t *var = (oval_variable_EXTERNAL_t *) variable;
+	char *tagname = (char *) xmlTextReaderLocalName(reader);
+	char *namespace = (char *) xmlTextReaderNamespaceUri(reader);
+	int return_code;
+	if (strcmp(tagname, "possible_value") == 0) {
+		char *hint = (char *)xmlTextReaderGetAttribute(reader, BAD_CAST "hint");
+		if (xmlTextReaderRead(reader) == 1) {
+			char *value = (char *)xmlTextReaderValue(reader);
+			struct oval_variable_possible_value *pv = oval_variable_possible_value_new(hint, value);
+			oval_collection_add(var->possible_values, (void *)pv);
```
mpreisler: Use API instead, same as above.

jan-cerny:
Some issues fixed. Build successful in Jenkins.

mpreisler:
Build and distcheck successful in Jenkins on RHEL6. ACK.

PR #82, opened 2015-05-29
Test for recurse_file_system="defined" feature.

jan-cerny:
If you are traversing the '/' file system and
recurse_file_system="defined", you will not traverse
into any other file systems that are mounted
to the '/' file system (e.g. /proc, etc.).

### DIFF
--- a/tests/probes/file/Makefile.am
+++ b/tests/probes/file/Makefile.am
@@ -10,4 +10,4 @@ TESTS_ENVIRONMENT= \
 
 TESTS = test_probes_file.sh
 
-EXTRA_DIST += test_probes_file.sh test_probes_file.xml
+EXTRA_DIST = test_probes_file.sh test_probes_file.xml test_recurse.xml
--- a/tests/probes/file/Makefile.am
+++ b/tests/probes/file/Makefile.am
@@ -10,4 +10,4 @@ TESTS_ENVIRONMENT= \
 
 TESTS = test_probes_file.sh
 
-EXTRA_DIST = test_probes_file.sh test_probes_file.xml test_recurse.xml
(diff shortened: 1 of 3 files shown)


jan-cerny:
Build sucessfull on Jenkins

tests/probes/file/Makefile.am
```diff
@@ -10,4 +10,4 @@ TESTS_ENVIRONMENT= \
 
 TESTS = test_probes_file.sh
 
-EXTRA_DIST += test_probes_file.sh test_probes_file.xml
+EXTRA_DIST = test_probes_file.sh test_probes_file.xml test_recurse.xml
```
mpreisler: keep += here

mpreisler:
Just noting here that Tomas Heinrich suggested a different approach to this test.

theinric:
> Just noting here that @theinric suggested a different approach to this test.

The commit message says "will not traverse into any other file systems" but the test actually checks only for /proc. The test should actually do what it's claiming to do.
Why is there comment="check all files in root filesystem" in the actual OVAL file?

Additionally, there's a bug:
The object doesn't collect files: <filename xsi:nil="true"/>
Yet the state checks for them:  <filename operation="pattern match">.*</filename>

Jan, has this test been actually verified to fail if nothing is mounted under /proc ?

mpreisler:
I suggest testing this with a different path to avoid the need to umount `/proc`.

Mount something to `/mnt/test`, make sure `/mnt/test` itself and anything inside isn't collected. Then unmount it, make sure it's a proper directory with contents and make sure it gets collected. After you verify it like that you can be reasonably sure you have a test that does something and you can change the path to `/proc`.

jan-cerny:
It would take some time to implement this test correctly and this is a low priority. Closing,

PR #83, opened 2015-05-29
[Enhancement] add Scientific Linux CPEs

redhatrises:
- Add Scientific Linux CPEs to inbuilt OpenSCAP CPE
- Part of https://github.com/openscap/scap-security-guide/pull/571

### DIFF
--- a/cpe/openscap-cpe-dict.xml
+++ b/cpe/openscap-cpe-dict.xml
@@ -29,6 +29,18 @@
             <title xml:lang="en-us">Community Enterprise Operating System 7</title>
             <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.rhel:def:1007</check>
       </cpe-item>
+      <cpe-item name="cpe:/o:scientificlinux:scientificlinux:5">
+            <title xml:lang="en-us">Scientific Linux 5</title>
+            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.rhel:def:1008</check>
+      </cpe-item>
+      <cpe-item name="cpe:/o:scientificlinux:scientificlinux:6">
+            <title xml:lang="en-us">Scientific Linux 6</title>
+            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.rhel:def:1009</check>
+      </cpe-item>
+      <cpe-item name="cpe:/o:scientificlinux:scientificlinux:7">
+            <title xml:lang="en-us">Scientific Linux 7</title>
+            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.rhel:def:1010</check>
+      </cpe-item>
       <cpe-item name="cpe:/o:fedoraproject:fedora:16">
             <title xml:lang="en-us">Fedora 16</title>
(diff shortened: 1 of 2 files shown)


mpreisler:
Hi, please use different OVAL def and test IDs, I think the CentOS ones are unfortunate but we won't change them now. For Scientific Linux I suggest:

* oval:org.open-scap.cpe.scientific:def:5
* oval:org.open-scap.cpe.scientific:def:6
* oval:org.open-scap.cpe.scientific:def:7

1008+ will hog the namespace when CentOS 8 comes out.

In the meantime I will install a Scientific Linux VM to test this.

mpreisler:
I am testing on Scientific Linux 6.6.

```
[root@localhost mpreisle]# cat /etc/system-release
Scientific Linux release 6.6 (Carbon)

[root@localhost mpreisle]# uname -a
Linux localhost.localdomain 2.6.32-504.el6.x86_64 #1 SMP Tue Oct 14 11:22:00 CDT 2014 x86_64 x86_64 x86_64 GNU/Linux
```

And the OVAL definition for Scientific Linux 6.6 doesn't evaluate to true for me.

```
Definition oval:org.open-scap.cpe.rhel:def:7: false
Definition oval:org.open-scap.cpe.rhel:def:6: false
Definition oval:org.open-scap.cpe.rhel:def:5: false
Definition oval:org.open-scap.cpe.rhel:def:1010: false
Definition oval:org.open-scap.cpe.rhel:def:1009: false
Definition oval:org.open-scap.cpe.rhel:def:1008: false
Definition oval:org.open-scap.cpe.rhel:def:1007: false
Definition oval:org.open-scap.cpe.rhel:def:1006: false
Definition oval:org.open-scap.cpe.rhel:def:1005: false
Definition oval:org.open-scap.cpe.rhel:def:1: false
Definition oval:org.open-scap.cpe.fedora:def:23: false
Definition oval:org.open-scap.cpe.fedora:def:22: false
Definition oval:org.open-scap.cpe.fedora:def:21: false
Definition oval:org.open-scap.cpe.fedora:def:20: false
Definition oval:org.open-scap.cpe.fedora:def:19: false
Definition oval:org.open-scap.cpe.fedora:def:18: false
Definition oval:org.open-scap.cpe.fedora:def:17: false
Definition oval:org.open-scap.cpe.fedora:def:16: false
Evaluation done.
```

I think the issue is that the OVAL is checking for `redhat-release` which is not available. Shouldn't it be checking for `sl-release` instead?

```
Name        : sl-release
Arch        : x86_64
Version     : 6.6
Release     : 1
Size        : 47 k
Repo        : installed
From repo   : base
Summary     : Scientific Linux release file
License     : GPLv2
Description : Scientific Linux release files
```

redhatrises:
> Hi, please use different OVAL def and test IDs, I think the CentOS ones are unfortunate but we won't change them now. For Scientific Linux I suggest:

>   * oval:org.open-scap.cpe.scientific:def:5
>   * oval:org.open-scap.cpe.scientific:def:6
>   * oval:org.open-scap.cpe.scientific:def:7

> 1008+ will hog the namespace when CentOS 8 comes out.

Done.
Hmmm.... Shouldn't this done for centos as well?

> I think the issue is that the OVAL is checking for `redhat-release` which is not available. Shouldn't it be checking for `sl-release` instead?

Done.

mpreisler:
ACK. Thanks a lot for your contribution!

Works fine on Scientific Linux 6.6 with your changes.

`scientific:def:6` evaluates to true, everything else to false. The `sl-release` package has the same name in SL 5, 6 and 7 so I believe it works everywhere although I only tested Scientific Linux 6.

I think it's too late to change the CentOS def and test IDs. We already released with those IDs so we will stick with them unless we have a really important reason to change them.

PR #87, opened 2015-06-22
Test the "does not exist" flag at collected objects.

jan-cerny:


isimluk:
Merged manually. Thx.

PR #89, opened 2015-06-23
Plug a memory leak (coverity scan).

jan-cerny:
Variable "pathname" going out of scope leaks the storage it points to.
Function SEXP_string_cstr(ent_val) allocates memory that must be freed.


isimluk:
Merged manually as fcff66613ba2e08d36ce1d1bd9fc7030366a2dcd to maint-1.2.

Thanks!

PR #90, opened 2015-06-23
Fix warnings

jan-cerny:


mpreisler:
I have no problems with commit fd881cbb48ebaea6897a2d2b0ffcae21abfbbbc0

Commit 7aae93efb73d8bf10e4bac73c1b46cce4fe35ecb is the right fix but we need to merge it to maint-1.0 because the same issue exists there. Please see https://github.com/OpenSCAP/openscap/blob/maint-1.0/src/OVAL/probes/independent/system_info.c#L121

isimluk:
Right!

Since I had my red merging hat already on my head. I have merge these to maint-1.0 and maint-1.2 respectively.

Thanks!

PR #91, opened 2015-06-24
Fix shellcheck warnings.

jan-cerny:
Use * instead of @ for arrays expansion.


isimluk:
Merged as 008b1ed94e3f5e1000cbcebcf8dce0a6d428c4be to maint-1.2. Thanks!

PR #92, opened 2015-06-24
Fix invalid dereference. (coverity)

jan-cerny:
Null-checking "session" suggests that it may be null,
but it has already been dereferenced on all paths leading to the check.


isimluk:
Merged manually as 060f16b7e67b50493a380cff2a19fba87993dba4 to maint-1.2.

Thanks!

PR #93, opened 2015-07-01
Changed masking behaviour.

jan-cerny:
According to the clarified specification, the "mask" attribute
should not affect oval_definitions section in oval_results.
It says: ... the"oval_definitions" section must not be altered
and must be an exact copy of the definitions evaluated.

### DIFF
--- a/src/OVAL/oval_entity.c
+++ b/src/OVAL/oval_entity.c
@@ -398,7 +398,10 @@ xmlNode *oval_entity_to_dom(struct oval_entity *entity, xmlDoc * doc, xmlNode * 
 	bool mask = oval_entity_get_mask(entity);
 
 	/* omit the value and operation used for testing in oval_results if mask=true */
-	if (mask && !xmlStrcmp(root_node->name, BAD_CAST OVAL_ROOT_ELM_RESULTS)) {
+	/* Omit it only for older versions of OVAL than 5.10 */
+	oval_version_t oval_version = oval_definition_model_get_schema_version(entity->model);
+	if (oval_version_cmp(oval_version, OVAL_VERSION(5.10)) < 0 &&
+		mask && !xmlStrcmp(root_node->name, BAD_CAST OVAL_ROOT_ELM_RESULTS)) {
 		entity_node = xmlNewTextChild(parent, ent_ns, BAD_CAST tagname, BAD_CAST "");
 	} else {
 		entity_node = xmlNewTextChild(parent, ent_ns, BAD_CAST tagname, BAD_CAST content);
--- a/tests/probes/maskattr/Makefile.am
+++ b/tests/probes/maskattr/Makefile.am
@@ -21,5 +21,7 @@ EXTRA_DIST = \
 	all.sh \
 	test_object_entity_mask.sh \
 	test_object_entity_mask.xml \
+	test_object_entity_mask_oval_5_9.sh \
+	test_object_entity_mask_oval_5_9.xml \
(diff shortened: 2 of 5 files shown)


mpreisler:
If specification was just clarified and not altered, why are we fixing this only in maint-1.2?

jan-cerny:
@mpreisler  : That is a good question. I have looked into specification and the OVAL issue https://github.com/OVALProject/Language/issues/67 one again, and I am totally confused about it :-(  It seems to me that this is clarified opposite to 5.10, but altered to some previous version. What shall I do with it? Should I allow this behavior only when OVAL 5.11 document is processed?

jan-cerny:
@mpreisler -
OVAL 5.11 specification says:
... When the mask attribute is set to 'true' on an OVAL Entity or an OVAL Field, the corresponding collected value of that OVAL Entity or OVAL Field MUST NOT be present in the "results" section of the OVAL Results document; the "oval_definitions" section must not be altered and must be an exact copy of the definitions evaluated. ...

OVAL 5.10 specification says:
If the mask attribute is set to 'true', then the observed value of this field must not be presented in the results file.

OVAL 5.9 specification says the opposite of 5.11:
..... In addition to the value being omitted from the copy of the system characteristics file, the copy of the definition file should also omit the value and operation used for testing. ..

Github issues says:
... we removed documentation stating that OVAL Definitions documents must be altered
when applying the mask attribute. At the same time, we removed the documentation
that stated "...the observed value of this field,...must not appear in the
results section of an OVAL Results document.". As a result, it is no
longer clear that the masked entities, in the OVAL Definitions document, should
not be altered. ...........

When I read those citations, I think that the behavior changed between 5.9 and 5.10.  In 5.11 it is just a clarification. I think We should implement this only for OVAL >= 5.10, and keep the old behavior for <= 5.9. 
Am I right?

jan-cerny:
It seems that it is possible to implement this depending on OVAL version.

mpreisler:
Discussed with @jan-cerny . Narrowed the problem down to:

5.11 and 5.10:
> OVAL Results document; the "oval_definitions" section must not be altered and must be an exact copy of the definitions evaluated.

5.9:
> the copy of the definition file should also omit the value and operation used for testing.

In other words, 5.11 and 5.10 require and exact copy, 5.9 requires something to be omitted so the result is not an exact copy.

So we need to have different behavior for OVAL <= 5.9 and OVAL >= 5.10

jan-cerny:
Re-worked the commit. It omits the values only for OVAL 5.9 and older. For 5.10 and newer, it makes an copy of values.

jan-cerny:
Test written.

jan-cerny:
I would use assert_exists also. The test_object_entity_mask_oval_5_9.sh is only copied test_object_etity_mask.sh.

jan-cerny:
Should I change the test scripts to use "assert_exists" ?

isimluk:
No. As I have said: "However, I see similar uses (of direct XPATH) in other tests as well. So this is just a comment." :-)

mpreisler:
Shoudn't we apply this to `maint-1.0` as well since 1.0.x supports OVAL 5.9 and 5.10?

It's a relatively isolated change.

jan-cerny:
Yes it can be applied to maint-1.0, I tried it, it works. The change in behavior can be considered as bugfix.

jan-cerny:
Created another PR against maint-1.0 https://github.com/OpenSCAP/openscap/pull/109

PR #94, opened 2015-07-03
Use OScap source for remote resources

ybznek:
#436 Downloading OVAL file only into memory not to temp file
#473 --fetch-remote-resources support bz2 now


isimluk:
It is fine. We just need to fix build without `HAVE_BZ2`.

Also, please consider fixing following minor issues:
 * there are 3 whitespace errors (3 lines of the commit end with `\t` character).
 * `write_to_memory_callback` should be static (without signature in `oscap_acquire.h`)

isimluk:
Corrected whitespace errors (3 trailing tabs) and merged manually. Thanks!

PR #99, opened 2015-07-09
trac#482: Fixed table of contents in HTML guide.

jan-cerny:
The table of contents now contains only selected groups.

### DIFF
--- a/xsl/xccdf-guide-impl.xsl
+++ b/xsl/xccdf-guide-impl.xsl
@@ -419,29 +419,39 @@ Authors:
 <xsl:template name="table-of-contents-items">
     <xsl:param name="item"/>
     <xsl:param name="levels"/>
+    <xsl:param name="profile"/>
     <ol>
         <xsl:for-each select="$item/cdf:Group">
-            <li>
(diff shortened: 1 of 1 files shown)


xsl/xccdf-guide-impl.xsl
```diff
@@ -419,29 +419,39 @@ Authors:
...
+                            <xsl:when test="cdf:title">
+                                <xsl:value-of select="cdf:title"/>
+                            </xsl:when>
+                            <xsl:otherwise>
+                                <xsl:value-of select="@id"/>
+                            </xsl:otherwise>
+                        </xsl:choose>
+                    </a>
+                </li>
+                <xsl:if test="cdf:Group and $levels&gt;1">
```
mpreisler: Will this branch ever print anything if `$selected_final` isn't `true`?

jan-cerny: No, That's exactly what we want, don't we?

mpreisler: Ah nevermind, I was just confused.

PR #102, opened 2015-07-10
trac#449 Fix process58 collect command_line not just command

ybznek:
Modify only part of code for Linux


jan-cerny:
I think we can use oscap_alloc() and oscap_free() instead of malloc() and free().

ybznek:
I used malloc() and free() because there was some previous usage of these functions in the file, so I supposed there was some reason.

isimluk:
LGTM.

I have no strong opinion on malloc&free vs oscap_malloc/oscap_free. I just see that many probes do not use it.

Zbynek, before I merge it, can you please create a test case for this?

src/OVAL/probes/unix/process58.c
```diff
@@ -351,12 +351,110 @@ static int get_exec_shield_status(int pid) {
...
+	return true;
+}
+
+/**
+ * Make "[%s] <defunct>" from cmd string - inplace
+ * @param cmd_buffer @see read_process() > cmd_buffer
+ * @return pointer to start of string
+ */
+static inline char *make_defunc_str(char* const cmd_buffer){
+	static const char DEFUNC_STR[] = "] <defunct>";
```
mpreisler: I think this syntax isn't portable. Is there any reason why we shouldn't just use `static const char*` instead?

mpreisler:
@jan-cerny already wrote something like dynamic_buffer, don't reinvent the wheel please.

jan-cerny:
@mpreisler : It is `oscap_string` and you can find it in src/common/oscap_string.h, but it is not available in maint-1.0 because we merged it to maint-1.2.

isimluk:
Ok, I think I will get oscap_string to maint-1.0 and file ticket for other pieces of code I identified to use it.

isimluk:
I have merged oscap_string to maint-1.0. It can now be used within this pr.

isimluk:
Seems ready to be merged, right?

mpreisler:
Not ready, Jenkins failing with this PR.

mpreisler:
This is still failing in Jenkins.

```
+ Ensure that command_line is collected                       [ FAIL ]
```

@ybznek, can you reproduce?

ybznek:
I have no issue with tests on computer, but I can reproduce it in Jenkins.

mpreisler:
Still failing in Jenkins, HEAD = 4acabd1

ybznek:
@mpreisler should be fixed http://10.3.9.142/job/openscap-pull-requests/69/

src/OVAL/probes/unix/process58.c
```diff
@@ -351,12 +352,91 @@ static int get_exec_shield_status(int pid) {
...
 static int read_process(SEXP_t *cmd_ent, SEXP_t *pid_ent, probe_ctx *ctx)
 {
 	int err = 1;
 	DIR *d;
 	struct dirent *ent;
 
+	struct oscap_buffer *cmdline_buffer = oscap_buffer_new();
+
 	d = opendir("/proc");
 	if (d == NULL)
```
jan-cerny: What about oscap_buffer_free() here? I suggest moving `struct oscap_buffer *cmdline_buffer = oscap_buffer_new();` after this.

ybznek: Ok.. You are right, again.

src/OVAL/probes/unix/process58.c
```diff
@@ -351,6 +352,83 @@ static int get_exec_shield_status(int pid) {
...
+
+	oscap_buffer_clear(buffer);
+
+
+
+	for(;;) {
+		static const int chunk_size = 1024;
+		char chunk[chunk_size];
+		// Read data, store to buffer
+		ssize_t read_size = read(fd, chunk, chunk_size );
```
mpreisler: `(read_size > 0 && read_size <= chunk_size)` indicates EOF, but `-1` indicates error. If error happened while reading we should react accordingly and not continue as if nothing happened.

src/OVAL/probes/unix/process58.c
```diff
@@ -351,6 +352,83 @@ static int get_exec_shield_status(int pid) {
...
+
+		/**
+		 * Program and args are separated by '\0' and '\n'
+		 * Replace them with spaces ' '
+		 */
+		while( i >= 0 ){
+			char chr = buffer_mem[i];
+			if ( ( chr == '\0') || ( chr == '\n' ) ) {
+				buffer_mem[i] = ' ';
+			} else if ( chr == '\033' ) {
```
mpreisler: This really needs a comment. Is the `escape` char the only thing we need to handle here?

src/OVAL/probes/unix/process58.c
```diff
@@ -351,6 +352,86 @@ static int get_exec_shield_status(int pid) {
...
+
+
+
+	for(;;) {
+		static const int chunk_size = 1024;
+		char chunk[chunk_size];
+		// Read data, store to buffer
+		ssize_t read_size = read(fd, chunk, chunk_size );
+		if (read_size < 0) {
+			return false;
```
mpreisler: This leaks `fd` file descriptor.

PR #103, opened 2015-07-10
trac#433 Bash autocompletion support .xml.bz2 files, not only .xml

ybznek:


isimluk:
Merged manually as 8fe81c71b450205e2e1cdf738786680949ccbc25.

Thank You!

PR #106, opened 2015-07-13
New schemas

jan-cerny:


mpreisler:
Is `oval-devel` aware of the wrong version in schemas?

We have to do several changes to all OVAL schemas to make sure they work smoothly with `libxml2`. It is very valuable to do exact the same changes every time. Did you take a look at `git log schemas/oval/5.11/*` and made sure you have applied all those changes to 5.11.1?

I recommend referencing the original commit that explains why that particular change had to be done.

isimluk:
Please apply 5b2c5f8c8d6665ffb84223f962ae20ec577bd7c8, de6d7d727be9100e4d22060912f4490553e5f1c7 and rest of de6d7d727be9100e4d22060912f4490553e5f1c7 to 5.11.1 as well.

jan-cerny:
Fixed

mpreisler:
Out of the suggestions, what exactly was fixed?

jan-cerny:
@mpreisler : The last two commits fix what @isimluk suggests - Unused xmldsig-core-schema.xsd was removed and the paths was set to use system-wide xmldsig-core-schema.xsd  schema in "common" directory.

mpreisler:
> Is oval-devel aware of the wrong version in schemas?
> 
> We have to do several changes to all OVAL schemas to make sure they work smoothly with libxml2. It is very valuable to do exact the same changes every time. Did you take a look at git log schemas/oval/5.11/* and made sure you have applied all those changes to 5.11.1?
> 
> I recommend referencing the original commit that explains why that particular change had to be done.

OK, but what about ^?

jan-cerny:
From the `git log schemas/oval/5.11/*` : 
* 8ba623120fc9f479285f9d6032cb925db420011d  - **Resolve parsing issues with OVAL 5.11 schemas** - is solved in this PR by 627a29aef8246eded6b663e49480e232aeb63dae
* 3a97aee622028ef9eee9b0946b2fcf2ff9fe8134  - **Use the shared xmldsig-core-schema.xsd even in OVAL 5.11 schemas** - is solved in this PR by a5fef7c6dee500cd86564e34021a5abde2ee7db3 and 27bc677
* 313650810d5727e1a259d13e9c454a1c3abbd73f -  **Updated OVAL 5.11 schemas to the officially released schemas** - I have already used the officially released schemas from the OVAL project github. 
* 5f188a985e285b022c8aca1b84409c30ddebb089 - **Use substitutionGroup change to the OVAL 5.11 schema from David Solin** - This is related to the notes element bug in 5.11. This is fixed in 5.11.1 official schemas. The change log says: Re-implemented Notes using substitution groups that allow for backwards compatibility with Notes elements defined from oval-def.
* 5c5539e0f56f4a4dfa61f749d46de676a162dbc5 - **Fixed bad object names in unix definitions schema and schematron** - This is fixed in 5.11.1 official schemas The change log says: Fixed multiple Schematron rules referencing an incorrect Object, or using an OVAL Sandbox XML namespace.
* ebc82944e9c3f436a57539ff118b78c04eb2b238 - **rhbz#1220262: Modify OVAL-5.11 schema files to be backward compatible** - this will be also fixed  in 55e94f0f129de49a1ec19bd185fa00cdc6f1deec

mpreisler:
Very well then, ACK.

PR #107, opened 2015-07-13
Parsing and populating multiple platform-specific schema_version elements

jan-cerny:


jan-cerny:
This can't be merged before the new OVAL 5.11.1 schemas will be merged. (https://github.com/OpenSCAP/openscap/pull/106) Otherwise, the test will fail, because the input oval document will not pass the validation.

mpreisler:
> Where is the oscap_htable defined for public API user?

+1

I recommend changing the API so that user can just query by key instead of getting a pointer to an internal htable. It's dangerous to give such internal structures via public API.

isimluk:
That's one point. However, second point is: I do not see 'oscap_htable' defined in our public headers.

```grep -r ' oscap_htable '  /usr/include/openscap/``` gives nothing.

**Comment by isimluk** [comment]
We have similar problem already in place (see https://fedorahosted.org/openscap/ticket/486) I wonder why it did never get caught.

**Comment by jan-cerny** [comment]
@isimluk  : oscap_htable is defined in common/list.h. I know it is confusing, but it has already been there before. Should I separate the hash table implementation to a separate file?
@mpreisler: User can query by key using this function: const char *oval_generator_get_platform_schema_version (struct oval_generator *generator, const char *platform); - key is the "platform" parameter.

**Comment by mpreisler** [comment]
> @isimluk : oscap_htable is defined in common/list.h. I know it is confusing, but it has already been there before. Should I separate the hash table implementation to a separate file?

The issue is that you cannot include it as a user of the public API.

> @mpreisler: User can query by key using this function: const char *oval_generator_get_platform_schema_version (struct oval_generator *generator, const char *platform); - key is the "platform" parameter.

Then we probably don't need to have a getter that returns the htable in public API.

**Comment by jan-cerny** [comment]
Thank you for your explanation, now I understand what you mean. But them problem is that I use this function in oval_results_model_new() to copy the information about schema versions from the definition model to the results model. So I need to re-implement oval_results_model_new().

**Comment by mpreisler** [comment]
> Thank you for your explanation, now I understand what you mean. But them problem is that I use this function in oval_results_model_new() to copy the information about schema versions from the definition model to the results model. So I need to re-implement oval_results_model_new().

Or you can move this function to private API.

**Comment by jan-cerny** [comment]
Build fails in Jenkins. I will have a look at it.

**Comment by jan-cerny** [comment]
I forgot to add new schemas to Makefile.am in the schemas directory, should be fixed in 2f02da9

**Comment by jan-cerny** [comment]
Build in Jenkins is successful now.

**Comment by mpreisler** [comment]
No time for review now so I will leave this unassigned. Maybe I will get to this later this week.

**Comment by mpreisler** [comment]
Discussed with @jan-cerny that the API needs to be changed to be more convenient for internal usage. Without going much into details in this comment we need an API that gives us guaranteed platform versions. In the current state the user of the API needs to check OVAL version and then check platform version separately based on whether OVAL version is higher or lower than 5.11.1.

**Comment by jan-cerny** [comment]
@mpreisler  You are right. I will have a look at this.
Moreover, we need to:
* add comments to the `oscap_generator` structure, or perhaps rename the `char *schema_version` and `struct oscap_htable *platform_schema_versions` to something more self-explaining and less confusing
* rename `oval_generator_set_timestamp_actual` - it's bad translation to English
* in `oval_generator_parse_tag`, we need to investigate behavior in OVAL <= 5.10 and don't parse the value if `platform` == NULL

**Comment by jan-cerny** [comment]
@mpreisler : I think that the functions `oval_generator_set_schema_version()` and `oval_generator_get_schema_version()` became having confusing names, because it is not clear from their name whether they work with core schema version or a platform schema version. I suggest deprecating these two functions. Then I think I can implement two new functions - `oval_generator_set_core_schema_version()` and `oval_generator_get_core_schema_version()` - I think It is clear what  would be their purpose :-) They will do basically the same as the former ones, it is just different names for clarity. What do you think about it? Is it worth doing?

**Comment by mpreisler** [comment]
@jan-cerny: Agreed. Make the old deprecated functions call the new ones.

**Comment by jan-cerny** [comment]
I think the proposed API is now suitable to use at other places of the OpenSCAP codebase.

File: src/OVAL/oval_generator.c [hunk 1]
```diff
@@ -114,6 +117,16 @@ char *oval_generator_get_timestamp(struct oval_generator *generator)
 	return generator->timestamp;
 }
 
+const char *oval_generator_get_platform_schema_version (struct oval_generator *generator, const char *platform)
+{
+	char *pschv = oscap_htable_get(generator->platform_schema_versions, platform);
```
ybznek: I personally prefer "char* platform_schema_version". It is long, but I see what does it mean. I need to see "oscap_htable_get(generator->platform_schema_versions, platform);" to understand pschv. But is is short fragment of code, maybe its ok.

mpreisler:
Could you please make sure we always use the correct function when getting the schema version? There is no way to do that automatically, it will need a manual review. At every use of `get_schema_version` we need to decide whether to use `get_core_schema_version` or `get_platform_schema_version`.

Once that's done I think this PR can be merged.

jan-cerny:
That requires some changes in oval_definition_model, oval_object and oval_state. Many commits coming soon.

jan-cerny:
I have found that the functions:
*  `oval_definition_model_get_schema_version`
* `oval_object_get_schema_version`
* `oval_state_get_schema_version`

return `oval_version_t`. This type is defined as: `typedef uint32_t oval_version_t;` in src/OVAL/public/oval_version.h. This type is not able to work with platform schema version. We can deprecate it and design new type that will be able to store information about extended version. But `oval_version_t`  occurs 55 times in OpenSCAP codebase.

src/OVAL/oval_generator.c
isimluk: +1

jan-cerny:
I have created a new module for comparing OVAL schema versions, called `oval_schema_version.h`. This module can work with platform extensions versions - parse it, compare it, convert to string. I have deprecated the old functions from `oval_version.h`. I replaced their usage with the new ones in our codebase. 

Note that this change also changes communication protocol between oscap and probes. Before, the schema version was converted to SEXP as a 32 bit integer. This is no longer possible because we need to pass the probe also the platform extension schema version. So now the same item is converted to a string. Also new function to convert schema version from SEXP was needed thanks to this, and it is also introduced this PR.

I have also written a small tests verifying schema version comparison that can be found in `tests/API/OVAL/schema_version/.

src/OVAL/oval_generator.c
```diff
@@ -39,11 +39,13 @@
 #include "common/debug_priv.h"
 #include "oval_agent_api_impl.h"
 #include "oval_definitions_impl.h"
+#include "common/list.h"
 
 struct oval_generator {
 	char *product_name;
 	char *product_version;
-	char *schema_version;
+	const char *core_schema_version;
```
jan-cerny: I am not sure about `const` here. Any suggestions?

src/OVAL/oval_generator.c
```diff
@@ -152,8 +191,23 @@ xmlNode *oval_generator_to_dom(struct oval_generator *generator, xmlDocPtr doc,
...
 		xmlNewTextChild(gen_node, ns_common, BAD_CAST "product_version", BAD_CAST generator->product_version);
-	if (generator->schema_version)
-		xmlNewTextChild(gen_node, ns_common, BAD_CAST "schema_version", BAD_CAST generator->schema_version);
+	if (generator->core_schema_version)
+		xmlNewTextChild(gen_node, ns_common, BAD_CAST "schema_version", BAD_CAST generator->core_schema_version);
+
+	const char *namespace_uri = "http://oval.mitre.org/XMLSchema/oval-definitions-5";
+	sv_itr = oscap_htable_iterator_new(generator->platform_schema_versions);
+	while (oscap_htable_iterator_has_more(sv_itr)) {
+		oscap_htable_iterator_next_kv(sv_itr, &platform, (void **) &version);
```
ybznek: (void **) typecasting is useless, isn't it? And it gives me no new information.

jan-cerny: the prototype from src/common/list.h requires void ** `void oscap_htable_iterator_next_kv(struct oscap_htable_iterator *hit, const char **key, void **value);`

src/OVAL/oval_schema_version.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+oval_schema_version_t oval_schema_version_from_cstr(const char *ver_str)
+{
+	oval_schema_version_t version;
+	memset(&version, 0, sizeof(oval_schema_version_t));
+	if (ver_str == NULL) {
+		return version;
+	}
+#if defined USE_REGEX_PCRE
```
ybznek: Isn't there any interface providing some abstraction to work with regex? This code contains duplicity...

src/OVAL/oval_generator.c
ybznek: Oh, sorry. I thought you can pass <anything>* to void* by standard, But gcc gives me some warning. I have learned a new thing :]

src/OVAL/oval_schema_version.c
jan-cerny: Good observation. I do not know about any interface like that. Everywhere in OpenSCAP where we are using regular expressions we implement it twice :-( I think we will drop support for POSIX regexes in future and have only PERL regexes, but we can't do this in maint-1.2

isimluk:
Hello Martin & Jan,

What is the status of this PR? Can this be merged?

jan-cerny:
Hello @isimluk ! The schema versions are parsed and populated in the content, there is a new code to work with them. Old functions have been deprecated and new functions and symbols are used instead the old code at many places at whole code (all probes and more).  There are also two small tests - first test tests parsing and comparing schema versions; second tests the OVAL file with a platform extension specific schema version. For more details, please see comments above. I think that can be merged, but @mpreisler maybe have been doing some reviews.

isimluk:
Perfect! thanks for the update!

src/OVAL/oval_generator.c
```diff
@@ -106,7 +104,12 @@ char *oval_generator_get_product_version(struct oval_generator *generator)
 
 char *oval_generator_get_schema_version(struct oval_generator *generator)
 {
-	return generator->schema_version;
+	return (char *) oval_generator_get_core_schema_version(generator);
```
mpreisler: This nasty const_cast deserves a comment. I know why it's being done but another code reader may not.

src/OVAL/oval_generator.c
```diff
@@ -114,6 +117,16 @@ char *oval_generator_get_timestamp(struct oval_generator *generator)
 	return generator->timestamp;
 }
 
+const char *oval_generator_get_platform_schema_version (struct oval_generator *generator, const char *platform)
+{
+	char *platform_schema_version = oscap_htable_get(generator->platform_schema_versions, platform);
```
mpreisler: I don't understand the logic here. The platform version is implicitly the core version if it was not specified in the OVAL file? Can you quote the spec about this?

src/OVAL/oval_generator.c
```diff
@@ -128,8 +141,13 @@ void oval_generator_set_product_version(struct oval_generator *generator, const
...
 void oval_generator_set_schema_version(struct oval_generator *generator, const char *schema_version)
 {
-	oscap_free(generator->schema_version);
-	generator->schema_version = oscap_strdup(schema_version);
+	oval_generator_set_core_schema_version(generator, schema_version);
+}
+
+void oval_generator_set_core_schema_version(struct oval_generator *generator, const char *schema_version)
+{
+	oscap_free(generator->core_schema_version);
```
mpreisler: free taking `const char*`?

src/OVAL/oval_generator.c
```diff
@@ -138,9 +156,30 @@ void oval_generator_set_timestamp(struct oval_generator *generator, const char *
 	generator->timestamp = oscap_strdup(timestamp);
 }
 
+void oval_generator_update_timestamp(struct oval_generator *generator)
+{
+	time_t et;
+	struct tm *lt;
+	char timestamp[] = "yyyy-mm-ddThh:mm:ss";
```
mpreisler: this can be const I think

src/OVAL/oval_generator.c
```diff
@@ -138,9 +156,30 @@ void oval_generator_set_timestamp(struct oval_generator *generator, const char *
...
+{
+	oscap_htable_add(generator->platform_schema_versions, platform, oscap_strdup(schema_version));
+}
+
+
 xmlNode *oval_generator_to_dom(struct oval_generator *generator, xmlDocPtr doc, xmlNode *parent)
 {
-	xmlNode *gen_node;
+	struct oscap_htable_iterator *sv_itr;
+	const char *platform, *version;
```
mpreisler: You declare `platform` up here, it is only used in the while loop. At the same time you declare `platform_uri` and `namespace_uri_length` just in the while loop.

Opinions differ on this but I prefer to bind variables to as small scopes as possible. Either way, choose one way and stick to it please.

src/OVAL/oval_generator.c
```diff
@@ -138,9 +156,30 @@ void oval_generator_set_timestamp(struct oval_generator *generator, const char *
...
+	xmlNode *gen_node, *sv_node;
```
mpreisler: same thing as above.

src/OVAL/oval_generator.c
```diff
@@ -187,9 +241,19 @@ int oval_generator_parse_tag(xmlTextReader *reader, struct oval_parser_context *
 		val = (char *) xmlTextReaderValue(reader);
 		oval_generator_set_product_version(gen, val);
 	} else if (!strcmp("schema_version", tagname)) {
+		platform = (char *) xmlTextReaderGetAttribute(reader, BAD_CAST "platform");
```
mpreisler: and once more. See `platform_name` below.

src/OVAL/oval_schema_version.c
```diff
@@ -0,0 +1,157 @@
...
+#include "common/debug_priv.h"
+#include "public/oval_schema_version.h"
+
+#if defined USE_REGEX_PCRE
+#include <pcre.h>
+#elif defined USE_REGEX_POSIX
+#include <regex.h>
+#endif
+
+enum {
```
mpreisler: why anonymous enum?

src/OVAL/oval_generator.c
jan-cerny: Yes, exactly, the logic here is that the platform version is implicitly the core version if it was not specified in the OVAL file. The specification says nothing about what to do if some platform extension is used in an OVAL file but its schema version was not specified in that OVAL file.

I decided to do this this way, because then you will not need to use everywhere constructions like this (pseudocode):
```
if (oval_core_schema_version >= 5.11.1)
    get_oval_core_schema_version()
else
    get_oval_platform_schema_version()
```

src/OVAL/oval_generator.c
jan-cerny: I don't think so, because the timestamp variable is used in sprintf function three lines below.
"Const" means that the string is immutable, but snpritntf modifies it.
Manual page describes its prototype:
`int snprintf(char *str, size_t size, const char *format, ...);`
it requires `char *str`.
Moreover this is not my implementation, it is only refactored from `oval_generator_new()`.

src/OVAL/oval_generator.c
jan-cerny: OK, I will declare these inside while loop.

src/OVAL/oval_generator.c
jan-cerny: Thanks for showing up this because I wasn't sure about `const` there.

jan-cerny:
I have dealt with your remarks.  But build fails in Jenkins on test_probes_environmentvariable.sh. I will have a look at it soon.

jan-cerny:
It was some strange error on Jenkins, perhaps it did not clean properly before the build. New build on same machine was successful, also build on another slave was successful.

src/OVAL/oval_generator.c
jan-cerny: Fixed by removing `const` type of core_schema_version.

jan-cerny:
Is there anything that needs to be improved ?

src/OVAL/oval_generator.c
mpreisler: Yeah, you are right. I think I was assuming that it is just a format description.

jan-cerny:
@mpreisler @isimluk : I have reviewed this once again, and I can't find any blocker, so I think it can be merged.

mpreisler:
Hi, I have this on my radar but I haven't had time to review it yet. I am sorry about that, please be patient.

src/OVAL/oval_schema_version.c
```diff
@@ -0,0 +1,157 @@
...
+		format = "%1$d.%2$d";
+	}
+	snprintf(buf, buf_len, format,
+		version.component[OVAL_SCHEMA_VERSION_CORE_MAJOR],
+		version.component[OVAL_SCHEMA_VERSION_CORE_MINOR],
+		version.component[OVAL_SCHEMA_VERSION_CORE_UPDATE],
+		version.component[OVAL_SCHEMA_VERSION_PLATFORM_MAJOR],
+		version.component[OVAL_SCHEMA_VERSION_PLATFORM_MINOR],
+		version.component[OVAL_SCHEMA_VERSION_PLATFORM_UPDATE]);
+	return oscap_strdup(buf);
```
mpreisler: I don't understand the reasoning behind the strdup here. We create a buf on the stack, we strdup it on the heap and destroy it on the stack upon return. Why not allocate it on the heap in the first place and just return it?

Maybe I am missing something.

mpreisler:
Great contribution! Kudos!

PR #109, opened 2015-07-14
Changed masking behaviour

jan-cerny:


mpreisler:
See #93 for review notes.

PR #110, opened 2015-07-22
First five commits on the new OVAL Session and one commit that solves a ticket.

msrubar:


msrubar:
I forgot to initialize some members of the oval struct. I'll repair that in following commits.

PR #111, opened 2015-07-23
Add OVAL 5.11.1 schematrons

jan-cerny:


mpreisler:
PR looks clean.

Have you tested the schematron validation?

jan-cerny:
Yes, I have tested it on OVAL content from RHEL6 and Fedora SCAP Security Guide - I have changed the schema_version element in OVAL files to 5.11.1 and then I have run using `oscap oval validate --schematron`. The test was sucessful.

PR #112, opened 2015-07-24
trac#495: Remove oscap_string from public API and fix test

jan-cerny:


isimluk:
I think I will put this to master & to maint-1.0.

The maint-1.2 will be the only branch where we leave OSCAP_HIDDEN_? missing.

Issue #104, opened 2015-07-13
Remote Assessment Report does not get copied back on the assessor machine


praving5:
I have setup oscap 1.2.5 on the assessor machine and my target machine has oscap 1.1.0. I can evaluate the results remotely but the report.html file is not getting copied back to the assessor machine. 

I did some investigation and found that report.html file isn't even getting created on the target machine under /tmp/tmp.<>/ directory. I can however see the oval results at console of the assessor machine.
<img width="960" alt="assessor machine" src="https://cloud.githubusercontent.com/assets/11082039/8641877/691442e2-2939-11e5-9279-70bdf7553f86.PNG">
<img width="363" alt="target machine" src="https://cloud.githubusercontent.com/assets/11082039/8641878/691ac266-2939-11e5-8fc5-ba3625daf0a6.PNG">

Any pointers? Please check out the attached snapshot.

Please help.

Thanks and regards,
Pravin Goyal

mmsrubar:
Hi Pravin Goyal,

you're right that the --report in eval module doesn't work at the moment but you can use the following workaround:
oscap oval eval --results oval-results.xml oval_defs.xml | oscap oval generate report oval-results.xml

isimluk:
Hello Pravin,

Michal has just fixed this maint-1.2 branch. Next openscap-1.2.x upstream release will contain the fix.

Thank You for your report.

PR #115, opened 2015-07-29
OVAL Session

msrubar:


mpreisler:
I think it's questionable to want to merge this to `maint-1.2`. This is not a bugfix and the changes are extensive.

Perhaps we could reconsider and merge this to master instead?

mpreisler:
`make distcheck` has a lot fails. I haven't looked closely but they seem related to these changes. Can you reproduce?

msrubar:
Yes I can. I did a few fixes and it should be alright now.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,657 @@
...
+	struct oval_probe_session *pb_sess;
+
+	struct {
+		struct oscap_source *definitions;
+		struct oscap_source *variables;
+		struct oscap_source *syschars;
+		struct oscap_source *directives;
+	} oval;
+
+	const char *oval_id;
```
isimluk: This is never set. Can be dropped.

utils/oscap-oval.c
```diff
@@ -333,148 +334,53 @@ int app_collect_oval(const struct oscap_action *action)
...
 
-	/* validate inputs */
-	if (action->validate) {
-		if (!valid_inputs(action)) {
-			goto cleanup;
-		}
-	}
+	/* set validation level */
+	val = getenv("OSCAP_FULL_VALIDATION");
+	oval_session_set_validation(session, action->validate, (val != NULL && strncmp(val, "1", 1) == 0) ? true : false);
```
isimluk: This strncmp changes behavior. If you believe it is helpful, please file a separate PR which fixes all occurrences of O_F_V

utils/oscap-oval.c
```diff
@@ -333,148 +334,53 @@ int app_collect_oval(const struct oscap_action *action)
...
+	/* Evaluation */
+	if (action->id) {
+		if ((oval_session_evaluate_id(session, action->probe_root, action->id, &eval_result)) != 0)
 			goto cleanup;
-		}
-
-		if (oval_definition_model_bind_variable_model(def_model, var_model)) {
-			fprintf(stderr, "Failed to bind Variables to Definitions\n");
+		printf("Definition %s: %s\n", action->id, oval_result_get_text(eval_result));
+	}
```
isimluk: Extra space at the end of line.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,657 @@
...
+
+	struct oval_directives_model *dir_model = NULL;
+	const char *filename = NULL;
+	int ret = 0;
+
+	/* This result can represent OVAL Results or OVAL System Characteristics based
+	 * on what kind of an action the session performed:
+	 *
+	 * OVAL System Characteristics: analyse
+	 * OVAL Results: 								evaluation or collect
```
isimluk: This is unique indentation of a commentary.

utils/oscap-oval.c
```diff
@@ -333,148 +334,53 @@ int app_collect_oval(const struct oscap_action *action)
...
-		}
-
-		/* generate report */
-	        if (action->f_report != NULL)
-        		oval_gen_report(action->f_results, action->f_report);
-	}
+	oval_session_set_directives(session, action->f_directives);
+	oval_session_set_results_export(session, action->f_results);
+	oval_session_set_report_export(session, action->f_report);
+	oval_session_export(session, reporter);
```
isimluk: Check return value, do not set `ret to OSCAP_OK` when export fails.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,657 @@
...
+	return 0;
+}
+
+int oval_session_evaluate_id(struct oval_session *session, char *probe_root, const char *id, oval_result_t *result)
+{
+	__attribute__nonnull__(session);
+
+	if (probe_root) {
+		if (setenv("OSCAP_PROBE_ROOT", probe_root, 1) != 0) {
+			oscap_seterr(OSCAP_EFAMILY_OSCAP, "Failed to set the OSCAP_PROBE_ROOT environment variable.\n");
```
isimluk: Please remove trailing `\n`. oscap_seterr adds `[filename:linenum]\n` to the error message.

(multiple occurrences)

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,657 @@
...
+
+	struct {
+		const char *results;
+		const char *syschars;
+		const char *report;
+	} export;
+
+	bool validation;
+	bool full_validation;
+};
```
isimluk: I would put xml_reporter to the session as well (just my $0.02).

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,657 @@
...
+	__attribute__nonnull__(session->source);
+
+	struct ds_sds_session *sds_session;
+
+	if (!oval_session_validate(session->source, fn, OSCAP_DOCUMENT_SDS)) {
+		/* the source can still be OVAL Definitions */
+		oscap_clearerr();
+		if (!oval_session_validate(session->source, fn, OSCAP_DOCUMENT_OVAL_DEFINITIONS))
+			return 1;
+	}
```
isimluk: Drop `oval_session_validate`.

Just ensure that ```
    oscap_source_get_scap_type(definitions_source) != OSCAP_DOCUMENT_OVAL_DEFINITIONS &&
    oscap_source_get_scap_type(definitions_source) != OSCAP_DOCUMENT_SDS
``` and then run oscap_source_validate.

File: src/OVAL/oval_session.c [hunk 9]
```diff
@@ -0,0 +1,657 @@
...
+
+	return true;
+}
+
+static int oval_session_load_definitions(struct oval_session *session, xml_reporter fn)
+{
+	__attribute__nonnull__(session);
+	__attribute__nonnull__(session->source);
+
+	struct ds_sds_session *sds_session;
```
isimluk: sds_session leaks from this function.

isimluk: It needs to be stored in the oval_session.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,657 @@
...
+			oscap_seterr(OSCAP_EFAMILY_OVAL, "Internal error: OVAL file was not found in "
+					"Source DataStream session cache!");
+			ds_sds_session_free(sds_session);
+			return 1;
+		}
+	}
+	else {
+		oscap_source_free(session->oval.definitions);
+		session->oval.definitions = session->source;
+		session->source = NULL;		/* don't free the same thing twice */
```
isimluk: Don't NULL the session->source. Just never free `session->oval.definition`.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,657 @@
...
+int oval_session_evaluate(struct oval_session *session, char *probe_root, agent_reporter fn, void *arg)
+{
+	__attribute__nonnull__(session);
+
+	if (probe_root) {
+		if (setenv("OSCAP_PROBE_ROOT", probe_root, 1) != 0) {
+			oscap_seterr(OSCAP_EFAMILY_OSCAP, "Failed to set the OSCAP_PROBE_ROOT environment variable.\n");
+			return 1;
+		}
+	}
```
isimluk: Duplicate code to what I have seen in `oval_session_evaluate_id`

isimluk: Maybe it doesn't matter a much. Dunno.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,657 @@
...
+	else {
+		oscap_source_free(session->oval.definitions);
+		session->oval.definitions = session->source;
+		session->source = NULL;		/* don't free the same thing twice */
+	}
+
+	/* import OVAL Definitions */
+	if (session->def_model) oval_definition_model_free(session->def_model);
```
isimluk: We now prefer a new line after the `if` statement.

But this is just a nitpick. No big need to get this fixed. Just remember for the next time. :)

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,657 @@
...
+	return 0;
+}
+
+static int oval_session_load_variables(struct oval_session *session, xml_reporter fn)
+{
+	__attribute__nonnull__(session);
+
+	if (session->oval.variables == NULL) {
+		oscap_seterr(OSCAP_EFAMILY_OVAL, "No OVAL Variables set.\n");
+		return 1;
```
isimluk: This is basically unreachable code.

In `oval_session_load` you only call `oval_session_load_variables` when the oval.variables are *not* NULL. What about calling `_load_variables` any time and doing no-op when os->oval.variables is NULL.

The result will be a less code to maintain.

src/OVAL/oval_sysModel.c
```diff
@@ -411,6 +411,22 @@ xmlNode *oval_syschar_model_to_dom(struct oval_syschar_model * syschar_model, xm
...
+	xmlDocPtr doc = xmlNewDoc(BAD_CAST "1.0");
+	if (doc == NULL) {
+		oscap_setxmlerr(xmlGetLastError());
+		return NULL;
+	}
+
+	oval_syschar_model_to_dom(model, doc, NULL, NULL, NULL);
+	return oscap_source_new_from_xmlDoc(doc, file);
+}
+
```
isimluk: This is very similar to `oval_syschar_model_export`. Basically duplicate.

Can you please make `oval_syschar_model_export` to call `oval_syschar_model_export_source`?

src/OVAL/oval_session.c
msrubar: I guess I can do the same for oval_session_load_syschars, right?

isimluk: yep.

mpreisler:
Bumping:
> I think it's questionable to want to merge this to maint-1.2. This is not a bugfix and the changes are extensive.
> 
> Perhaps we could reconsider and merge this to master instead?

I think it's more of a quirk or trait than a bug. If it's really a bug we should merge it to an older maintenance branch like `maint-1.0`. I feel like `maint-1.2` is not the right branch for this and it should go to `master`.

jan-cerny:
@mpreisler : I agree that this is quite a big change. @msrubar : What advantages would it bring us? . Does this break or change the current behavior? Try to persuade us that we need it it maint-1.2 branch! Otherwise I incline to merging this to master.

msrubar:
I wanted to use rebase to squash the commits but there were too many conflicts so I decided to take it as a whole and break it into separate commits in a different branch. Then I pushed it to this branch. The result is less commits fot the review. ''isimluk'' already did a review of commit from ''fc75519'' to   ''cb372fd'' and also of ''e805e89''. Changes are already included.

Anyway if we would merge to the ''maint-1.2'' it could help to catch bugs inside the OVAL Session. It doesn't change any API nor break it. It only refactors evaluation module in ''oscap-oval.c''. In the future both ''app_collect_oval()'' and ''app_analyse_oval()'' can use. It would also help us to get rid of things like ''app_xslt'' in oscap-tool.c in future.

mpreisler:
From what I understand this fixes the `--report` issue when doing `oscap oval eval`. Without this change you *need* to have `--results` in the command line for `--report` to work. See issue #104

msrubar:
Yes, that's true it also fixes the issue #104.

src/OVAL/Makefile.am
```diff
@@ -123,6 +124,7 @@ liboval_testing_la_LIBADD = $(liboval_la_LIBADD)
 pkginclude_HEADERS  = \
 			public/oval_adt.h \
 			public/oval_agent_api.h \
+			public/oval_session.h \
```
ybznek: The list of files was in alphabetical order, I think.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+#include "common/util.h"
+#include "common/_error.h"
+#include "common/oscapxml.h"
+#include "source/xslt_priv.h"
+#include "public/oval_agent_api.h"
+#include "public/oval_session.h"
+#include "../DS/public/ds_sds_session.h"
+#include "oscap_source.h"
+
+static const char *oscap_productname = "cpe:/a:open-scap:oscap";
```
ybznek: I personally prefer uppercase "constants", but I don't know coding style for this. When I see "oscap_productname" in code, I will look for declaration of VARIABLE.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+		const char *results;
+		const char *report;
+	} export;
+
+	struct {
+		/* it's called when there is something invalid in input/output files */
+		xml_reporter xml_fn;
+	} reporter;
+
+	bool validation;
```
ybznek: Why do you use two bool variables? One enum looks like {NO, STANDARD, FULL} would be better, maybe. There is fragments of code looks like "if (result && session->validation && session->full_validation) " - you do not need value of "full_validation" if "validation" is false.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+	bool full_validation;
+};
+
+struct oval_session *oval_session_new(const char *filename)
+{
+	oscap_document_type_t scap_type;
+	struct oval_session *session;
+
+	session = (struct oval_session *) oscap_calloc(1, sizeof(struct oval_session));
```
ybznek: I think "(struct oval_session *)" is useless in C99. - You have declaration near the allocation, so you can see type of pointer without typecasting.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+
+	session->source = oscap_source_new_from_file(filename);
+	if ((scap_type = oscap_source_get_scap_type(session->source)) == 0) {
```
ybznek: You can use OSCAP_DOCUMENT_UNKNOWN instead of "0".

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+
+void oval_session_set_validation(struct oval_session *session, bool validate, bool full_validation)
+{
+	__attribute__nonnull__(session);
+
+	session->validation = validate;
+	session->full_validation = full_validation;
+}
+
+void oval_session_set_datastream_id(struct oval_session *session, const char *id)
```
ybznek: There are many "setters" with similar structure. Maybe you can use some macros to generate these functions.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+}
+
+static int oval_session_load_definitions(struct oval_session *session)
+{
+	__attribute__nonnull__(session);
+	__attribute__nonnull__(session->source);
+
+	oscap_document_type_t type = oscap_source_get_scap_type(session->source);
+	if (type != OSCAP_DOCUMENT_OVAL_DEFINITIONS && type != OSCAP_DOCUMENT_SDS) {
+		oscap_seterr(OSCAP_EFAMILY_OVAL, "Type mismatch: %s. Expecting %s "
```
ybznek: I have problem with reading such long function call. I personally prefer something like.

const char* readable_origin = oscap_source_readable_origin(session->source);
const char* expected = oscap_document_type_to_string(OSCAP_DOCUMENT_OVAL_DEFINITIONS);
const char* expected2 = oscap_document_type_to_string(OSCAP_DOCUMENT_SDS);
const char* found = oscap_document_type_to_string(type));

oscap_seterr(OSCAP_EFAMILY_OVAL, "Type mismatch: %s. Expecting %s "
       "or %s but found %s.", readable_origin, expected, expected2, found):

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	/* free the previous Agent session if this function was already called */
+	if (session->sess)
+		oval_agent_destroy_session(session->sess);
+
+	session->sess = oval_agent_new_session(session->def_model, basename(path_clone));
+	if (session->sess == NULL) {
+		oscap_seterr(OSCAP_EFAMILY_OVAL, "Failed to create a new agent session.");
+		free(path_clone);
```
ybznek: Do you have some reason to using free() instead of oscap_free()?

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+
+	session = (struct oval_session *) oscap_calloc(1, sizeof(struct oval_session));
+
+	session->source = oscap_source_new_from_file(filename);
+	if ((scap_type = oscap_source_get_scap_type(session->source)) == 0) {
+		oval_session_free(session);
+		return NULL;
+	}
+
+	if (scap_type != OSCAP_DOCUMENT_OVAL_DEFINITIONS && scap_type != OSCAP_DOCUMENT_SDS) {
```
jan-cerny: Are you sure there should be && ?

mpreisler: I am puzzled because this seems correct to me.

`if (!(A || B))` is equivalent to `if (!A and !B)`

src/OVAL/Makefile.am
msrubar: If you take a look inside the file then is no alphabetical order.

src/OVAL/oval_session.c
msrubar: The product name itself is common and used in both xccdf and oval sessions. Maybe it would be usefull to put it into the ```oscap.h``` but lots of functions is DEPRECATED in there. Do we plan to get rid of it in the future? The string ```"cpe:/a:open-scap:oscap"``` is already defined as macro in ```oscap-tool.h```

src/OVAL/oval_session.c
msrubar: That's probably true but I tried to change as less as possible during the refactoring.

src/OVAL/oval_session.c
msrubar: That's actualy a very good question. I went through some code that was commited lately and I saw both approaches with and without typecast.

src/OVAL/oval_session.c
msrubar: Definitely. I remember when I pointed you to a similar issue in your pull request last week and now I did the same thing :-).

src/OVAL/oval_session.c
msrubar: Nice catch! I overlooked this one. Fixed.

src/OVAL/oval_session.c
jan-cerny: Sorry for this. It is correct. I have misread the error message and got confused.

src/OVAL/oval_session.c
jan-cerny: My opinion is not use macros to generate any functions, because with macros I always got confused while debugging. I think it is not a problem that there are many setters with similar code.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+	if (oscap_err()) {
+		return 1;
+	}
+
+	session->res_model = oval_agent_get_results_model(session->sess);
+
+	return 0;
+}
+
+int oval_session_export(struct oval_session *session)
```
msrubar: May be it would be better to break this function into smaller pieces.

src/OVAL/Makefile.am
isimluk: :-))

src/OVAL/oval_session.c
isimluk: +1 to enum
+10 to not changing much during refactoring

src/OVAL/oval_session.c
isimluk: Yep, oscap.h is good candidate for dropping, but we will see.

The constant can be made uppercase, and consolidated. But does not need to be in this pr.

src/OVAL/oval_session.c
isimluk: typecast is not needed imho. but it is harmless.

isimluk:
Re discussion maint-1.2 vs master.

I agree that this might be a good fit for master branch. However, I see pragmatic reasons for merging this to maint-1.2. We don't do any radical change, we just move code from the tool to library here. Doing it in maint-1.2 means earlier adoption.

Note that it cannot be merged to maint-1.0 because it heavily depends on oscap_source. That being said, I see this change as a logical conclusion of oscap_source efforts -> interface is getting simpler thanks to that.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+}
+
+void oval_session_set_xml_reporter(struct oval_session *session, xml_reporter fn)
+{
+	__attribute__nonnull__(session);
+
+	session->reporter.xml_fn = fn;
+}
+
+static bool oval_session_validate(struct oval_session *session, struct oscap_source *source, oscap_document_type_t type)
```
isimluk: I think you should be able to drop `type` from `oval_session_validate`. But this can be done any time later. I think we can merge it like it is.

src/OVAL/oval_session.c
isimluk: not needed imho

There is too many variables to pass along anyway.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+	struct oscap_source *result = NULL;		/* OVAL Results */
+	const char *filename = NULL;
+	int ret = 0;
+
+	/* Import OVAL Directives if any */
+	if (session->oval.directives && session->res_model) {
+		/* OVAL Directives can be used only if there are OVAL Resutls and these are
+		 * only available if there is a Results model which mean that either
+		 * evaluation or analyse was performed. */
+		if (dir_model)
```
isimluk: always false. dead code.

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+			oval_directives_model_free(dir_model);
+		oval_directives_model_import_source(dir_model, session->oval.directives);
```
isimluk: What about checking return value -- directives model may be malformed.

(perhaps this has been always broken, not sure)

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+		if (oscap_source_apply_xslt_path(result, oval_results_report,
+				session->export.report, stdparams, oscap_path_to_xslt()) == -1) {
+			oscap_source_free(result);
+			return 1;
+		}
+	}
+
+	oscap_source_free(result);
+	if (dir_model)
+		oval_directives_model_free(dir_model);
```
isimluk: Move this ^^^ upper.

Otherwise all the `return 1;` situations will potentially leak `dir_model`

src/OVAL/oval_session.c
```diff
@@ -0,0 +1,463 @@
...
+		result = oval_results_model_export_source(session->res_model, dir_model, NULL);
+		filename = session->export.results;
+	}
+
+	/* Validate OVAL Results. The 'result' in condition will make sure that there is
+	 * something to validate */
+	if (result && session->validation && session->full_validation) {
+		if (!oval_session_validate(session, result, OSCAP_DOCUMENT_OVAL_RESULTS)) {
+		/* FIXME: if invalid output than add suffix ".invalid" to file name but
+		 * still export and set some error about it ?*/
```
isimluk: FIXMEs are like God. They are everlasting.

We could at least return 1 at the bottom when this happens.

isimluk: `ret = 1`

isimluk:
@msrubar , it seems fine. Please just correct `export` function and we can merge it.

src/OVAL/oval_session.c
msrubar: Hmm, I'm wonder why would we do that?

isimluk:
Well done!

Thanks Michal!

PR #124, opened 2015-08-03
oscap_string & oscap_buffer

ybznek:
Move body of oscap_string to oscap_buffer.
oscap_buffer will offer some extra features.

### DIFF
--- a/src/common/oscap_buffer.c
+++ b/src/common/oscap_buffer.c
@@ -80,11 +80,10 @@ void oscap_buffer_append_char(struct oscap_buffer *s, char c)
 }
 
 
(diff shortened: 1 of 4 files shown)


src/common/oscap_buffer.c
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Buffer with unlimited length
+ * contains:
+ * - pointer to data,
+ * - actual length of string,
+ * - capacity of allocated memory
+ */
+struct oscap_buffer {
+	char *str;
```
mpreisler: Could we rename this to `data`? `str` suggests it's supposed to be a zero-terminated string.

PR #129, opened 2015-08-05
Groupby

jan-cerny:
This resolves ticket trac#484.
It is not ready to merge although it is working in Firefox and Chrome.
I opened this PR as a place for discussion about the code. It definitely needs to improve, since I am a JavaScript newbie.

One question - to make this working, we need to run xccdf-resources-build.sh and commit here also the minified script. Should I commit it here? Or I think the resource could be generated by a makefile.

### DIFF
--- a/xsl/xccdf-resources/openscap.js
+++ b/xsl/xccdf-resources/openscap.js
@@ -195,6 +195,8 @@ function GroupBy(key) {
 	/* We must process grouping upon the original table.
 	 * Otherwise, we would have unwanted duplicties in new table. */
 	Reset();
+	if (key == "default")
+		return;
 
 	var lines = {};
 	$(".rule-overview-leaf").each(function() {
--- a/xsl/xccdf-resources/openscap.js
+++ b/xsl/xccdf-resources/openscap.js
@@ -122,7 +122,7 @@ $(document).ready( function() {
     is_original = true;
 });
 
-function Reset() {
+function resetTreetable() {
 	if (!is_original) {
 		$(".treetable").remove();
 		$("#rule-overview").append(original_treetable.clone());
@@ -134,7 +134,7 @@ function Reset() {
 	}
 }
 
-function NewGroupLine(group_name)
+function newGroupLine(group_name)
 {
 	return "<tr class=\"rule-overview-inner-node\" data-tt-id=\"" + group_name + "\">" +
 		"<td colspan=\"3\"><strong>" + group_name + "</strong></td></tr>";
@@ -148,17 +148,17 @@ var KeysEnum = {
 	DISA: "data-disa-id"
 };
 
-function GetTargetGroupsList(rule, key)
+/* This function returns an array of target groups indentifiers */
+function getTargetGroupsList(rule, key)
 {
-	/* This function returns an array of target groups indentifiers */
-	if (key == KeysEnum.SEVERITY) {
+	switch(key) {
+	case KeysEnum.SEVERITY:
 		var severity = rule.children("td:nth-child(2)").text();
 		return [severity];
-	} else if (key == KeysEnum.RESULT) {
+	case KeysEnum.RESULT:
 		var result = rule.children(".rule-result").text();
 		return [result];
-	}
-	else {
+	default:
 		var ref_list = rule.attr(key);
 		if (!ref_list) {
 			ref_list = "unknown";
@@ -167,13 +167,14 @@ function GetTargetGroupsList(rule, key)
 	}
 }
 
-function SortGroups(groups, key)
+function sortGroups(groups, key)
 {
-	if (key == KeysEnum.SEVERITY) {
+	switch(key) {
+	case KeysEnum.SEVERITY:
 		return ["high", "medium", "low"];
-	} else if (key == KeysEnum.DISA) {
+	case KeysEnum.DISA:
 		return groups.sort(function(a, b){return a-b});
-	} else if (key == KeysEnum.NIST) {
+	case KeysEnum.NIST:
 		return groups.sort(function(a, b){
 			var regex = /(\w\w)-(\d+)(.*)/;
 			var a_parts = regex.exec(a);
@@ -194,28 +195,29 @@ function SortGroups(groups, key)
 				}
 			}
 		});
-	} else {
+	default:
 		return groups;
 	}
 }
 
-function GroupBy(key) {
-	/* We must process grouping upon the original table.
-	 * Otherwise, we would have unwanted duplicties in new table. */
-	Reset();
+function groupRulesBy(key) {
+	/* We must reset the treetable and process grouping upon the original
+	 * table to avoid unwanted duplicties in new table. */
+	resetTreetable();
 	if (key == KeysEnum.DEFAULT)
 		return;
 
+	/* Browse the rules and sort them into groups */
 	var lines = {};
 	$(".rule-overview-leaf").each(function() {
 		$(this).children("td:first").css("padding-left","0px");
 		var id = $(this).attr("data-tt-id");
-		var target_groups = GetTargetGroupsList($(this), key);
+		var target_groups = getTargetGroupsList($(this), key);
 		for (i = 0; i < target_groups.length; i++) {
 			var target_group = target_groups[i];
 			if (!lines.hasOwnProperty(target_group)) {
 				/* Create a new group */
-				lines[target_group] = [NewGroupLine(target_group)];
+				lines[target_group] = [newGroupLine(target_group)];
 			}
 			var clone = $(this).clone();
 			clone.attr("data-tt-id", id + "copy" + i);
@@ -224,8 +226,10 @@ function GroupBy(key) {
 			lines[target_group].push(new_line);
 		}
 	});
+
+	/* Remove old treetable and replace it with a new one */
 	$(".treetable").remove();
-	var groups = SortGroups(Object.keys(lines), key);
+	var groups = sortGroups(Object.keys(lines), key);
 	var html_text = "";
 	for (i = 0; i < groups.length; i++) {
 		html_text += lines[groups[i]].join("\n");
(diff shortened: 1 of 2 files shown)


xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +108,131 @@ function ruleSearch()
...
+			toggleRuleDisplay(this);
+		});
+		is_original = true;
+	}
+}
+
+function NewGroupLine(group_name)
+{
+	return "<tr class=\"rule-overview-inner-node\" data-tt-id=\"" + group_name + "\">" +
+		"<td colspan=\"3\"><strong>" + group_name + "</strong></td></tr>";
```
ybznek: jQuery has some functions which can create new elements. Maybe it would be cleaner way.

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +108,131 @@ function ruleSearch()
...
+{
+	/* This function returns an array of target groups indentifiers */
+	if (key == KeysEnum.SEVERITY) {
+		var severity = rule.children("td:nth-child(2)").text();
+		return [severity];
+	} else if (key == KeysEnum.RESULT) {
+		var result = rule.children(".rule-result").text();
+		return [result];
+	}
+	else {
```
ybznek: IMHO "} else {" would looks better, but replacing "if else" with "switch" would be the best way.

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +108,131 @@ function ruleSearch()
...
+		if (!ref_list) {
+			ref_list = "unknown";
+		}
+		return ref_list.split(",");
+	}
+}
+
+function SortGroups(groups, key)
+{
+	if (key == KeysEnum.SEVERITY) {
```
ybznek: Why don't you use "switch"?

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +108,131 @@ function ruleSearch()
...
+				}
+			}
+		});
+	} else {
+		return groups;
+	}
+}
+
+function GroupBy(key) {
+	/* We must process grouping upon the original table.
```
ybznek: I think the comment describes whole function. Why is it inside of? http://stackoverflow.com/a/10126365

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +108,131 @@ function ruleSearch()
...
+			lines[target_group].push(new_line);
+		}
+	});
+	$(".treetable").remove();
+	var groups = SortGroups(Object.keys(lines), key);
+	var html_text = "";
+	for (i = 0; i < groups.length; i++) {
+		html_text += lines[groups[i]].join("\n");
+	}
+	new_table ="<table class=\"treetable table table-bordered\"><thead><tr><th>Group</th> <th style=\"width: 120px; text-align: center\">Severity</th><th style=\"width: 120px; text-align: center\">Result</th></tr></thead><tbody>" + html_text + "</tbody></table>";
```
ybznek: Again, you can use jQuery's way to create some element.

mpreisler:
`openscap.js` uses camel-case with first letter lowercase convention for function names. Please stick to it.

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +108,131 @@ function ruleSearch()
...
         toggleRuleDisplay(this);
     });
-
-    $(".treetable").treetable({ column: 0, expandable: true, initialState : "expanded", indent : 0 });
+    original_treetable = $(".treetable").clone();
+    $(".treetable").treetable({ column: 0, expandable: true, clickableNodeNames: true, initialState : "expanded", indent : 0 });
+    is_original = true;
 });
+
+function Reset() {
```
mpreisler: The name of this function is horribly generic. There are real risks of clashes. Keep in mind that your functionality is not alone in the report, there may be other things that need resetting!

xsl/xccdf-resources/openscap.js
jan-cerny: No, it doesn't. It describes why do we call Reset() here. Maybe the comment is bad when someone don't understand it :-)

jan-cerny:
Done some polishing and improved the table appearance.

mpreisler:
Apart from some minor code quality issues the only issue remaining is that the group-by combobox is hardcoded to some common choices. Some of these will be in every content -- severity, result. But some will be there even for content where they don't apply at all -- DISA or NIST IDs.

This code needs to figure out what to populate the combobox with on the spot. This can be achieved either by javascript or in XSLT, both options have their pros and cons.

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +109,152 @@ function ruleSearch()
...
+	}
+	new_table ="<table class=\"treetable table table-bordered\"><thead><tr><th>Group</th> <th style=\"width: 120px; text-align: center\">Severity</th><th style=\"width: 120px; text-align: center\">Result</th></tr></thead><tbody>" + html_text + "</tbody></table>";
+	$("#rule-overview").append(new_table);
+	is_original = false;
+	$(".treetable").treetable({
+		column: 0,
+		expandable: true,
+		clickableNodeNames: true,
+		initialState : "expanded",
+		indent : 0 });
```
mpreisler: code style...

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +109,152 @@ function ruleSearch()
...
+	}
+}
+
+function sortGroups(groups, key)
+{
+	switch(key) {
+	case KeysEnum.SEVERITY:
+		return ["high", "medium", "low"];
+	case KeysEnum.DISA:
+		return groups.sort(function(a, b){return a-b});
```
mpreisler: Shouldn't this use parseint or something like that?

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +109,152 @@ function ruleSearch()
...
+	default:
+		var ref_list = rule.attr(key);
+		if (!ref_list) {
+			ref_list = "unknown";
+		}
+		return ref_list.split(",");
+	}
+}
+
+function sortGroups(groups, key)
```
mpreisler: `key` is not a good name here I think.

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +109,152 @@ function ruleSearch()
...
     });
-
-    $(".treetable").treetable({ column: 0, expandable: true, initialState : "expanded", indent : 0 });
+    original_treetable = $(".treetable").clone();
+    $(".treetable").treetable({ column: 0, expandable: true, clickableNodeNames: true, initialState : "expanded", indent : 0 });
+    is_original = true;
+	stripeTreeTable();
 });
+
+function resetTreetable() {
```
mpreisler: Is this function used / usable outside of group-by functionality? If not then why doesn't the name signify that?

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +109,152 @@ function ruleSearch()
...
     $("#result-details").hide();
     $(".js-only").show();
+	$(".form-group select").val("default");
     $(".toggle-rule-display").each(function(){
         toggleRuleDisplay(this);
     });
-
-    $(".treetable").treetable({ column: 0, expandable: true, initialState : "expanded", indent : 0 });
+    original_treetable = $(".treetable").clone();
+    $(".treetable").treetable({ column: 0, expandable: true, clickableNodeNames: true, initialState : "expanded", indent : 0 });
```
mpreisler: code style...

xsl/xccdf-resources/openscap.js
jan-cerny: It is usable universally. If someone will change the tree-table in his code, he can set `is_original = false` and then use this function.

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +109,152 @@ function ruleSearch()
...
+	NIST: "data-nist-id",
+	DISA: "data-disa-id"
+};
+
+/* This function returns an array of target groups indentifiers */
+function getTargetGroupsList(rule, key)
+{
+	switch(key) {
+	case KeysEnum.SEVERITY:
+		var severity = rule.children("td:nth-child(2)").text();
```
mpreisler: This is not a very robust way to get severity. Could you decorate the severity element in some way to make this more robust? Similar to what we do with rule-result.

At some point somebody will slightly change the report and break this without noticing...

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +109,152 @@ function ruleSearch()
...
+		return [severity];
+	case KeysEnum.RESULT:
+		var result = rule.children(".rule-result").text();
+		return [result];
+	default:
+		var ref_list = rule.attr(key);
+		if (!ref_list) {
+			ref_list = "unknown";
+		}
+		return ref_list.split(",");
```
mpreisler: What if the references contain `,` in them? Can that happen according to the XCCDF spec?

xsl/xccdf-resources/openscap.js
```diff
@@ -108,12 +109,157 @@ function ruleSearch()
...
     $(".js-only").show();
+	$(".form-group select").val("default");
     $(".toggle-rule-display").each(function(){
         toggleRuleDisplay(this);
     });
-
+    original_treetable = $(".treetable").clone();
     $(".treetable").treetable({ column: 0, expandable: true, initialState : "expanded", indent : 0 });
+    is_original = true;
+	stripeTreeTable();
```
mpreisler: tab indent

mpreisler:
Discussed this with @jan-cerny, only minor changes remain - indentation needs to be fixed to be consistent. XSLT and javascript files should use 4 spaces as indents.

mpreisler:
The PR is ready to be merged but is targeted at `maint-1.2`. Discussed with @jan-cerny and he can't remember why he targeted it at `maint-1.2` instead of `maint-1.1`. He said he will create a new PR for `maint-1.1`.

mpreisler:
Replaced by #138

PR #130, opened 2015-08-06
Add PID to error message when probe has been killed

ybznek:
It is easier to trace probe's logfile.


mpreisler:
According to `/usr/include/bits/typesizes.h`, pid_t is a signed 32bit integer on x86 and x86_64

```
#define __PID_T_TYPE            __S32_TYPE
```

PR #131, opened 2015-08-06
Turn the search engine on in Doxygen documentation

msrubar:
This will enable us to easily search through the Doxygen documentation.

See: https://fedorahosted.org/openscap/ticket/493

### DIFF
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -1414,4 +1414,4 @@ DOT_CLEANUP            = YES
 # The SEARCHENGINE tag specifies whether or not a search engine should be 
 # used. If set to NO the values of all tags below this one will be ignored.
 
-SEARCHENGINE           = NO
+SEARCHENGINE           = YES


ybznek:
It is very good idea! :+1:

mpreisler:
Why `maint-1.2`?

msrubar:
Oh, you're right, wrong branch, didn't think this through too much. I guess it should go to ```maint-1.0```. Is it correct?

mpreisler:
Agreed, search engine is not a new feature in doxygen, lets fix this in `maint-1.0`.

I tried it and the search bar appears but typing stuff into it doesn't do anything. I am using `Chrome 44.0.2403.125 (64-bit)`.

msrubar:
That's interesting. It works just fine with firefox 39.0. Can someone else also try it out?

mpreisler:
The functionality is glitchy in Chrome but I decided to merge it anyway. It's a bug in Doxygen and even glitchy search is better than no search.

Applied manually as 8f2df6dd9646fb5d9e81298bbfe5277421e0021b

PR #137, opened 2015-08-11
oscap-ssh - enable remote sudo operation by non-privileged users

openprivacy:
Apologies if you see this twice (I had committed to the wrong branch initially). This patch allows both root and a non-root user (via sudo) to use oscap-ssh.

### DIFF
--- a/utils/oscap-ssh
+++ b/utils/oscap-ssh
@@ -76,9 +76,13 @@ function usage()
     echo "  --variables"
     echo "  --skip-valid"
     echo
+    echo "specific option for oscap-ssh (must be first argument):"
+    echo "  --sudo"
+    echo
     echo "See \`man oscap\` to learn more about semantics of these options."
 }
 
+OSCAP_SUDO=""
 if [ $# -lt 1 ]; then
     echo "No arguments provided."
     usage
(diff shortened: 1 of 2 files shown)


jan-cerny:
Hi, I think that generally it is not a good idea to force users to scan with root privileges. I propose having a new option in oscap-ssh, something like --enable-sudo, that would enable remote sudo operation by non-privileged users. I would not change the default behavior.

mpreisler:
Hi, thanks for the contribution! We were looking to implement something like this in the future :-)

I agree with @jan-cerny but I vote for just `--sudo` instead of `--enable-sudo`. We can mix it with the oscap arguments and flip the OSCAP_COMMAND variable when parsing them on line `138`.

openprivacy:
I think best to parse `--sudo` first that enables a command line like `oscap-ssh --sudo oscap-user@192.168.1.17` which seems more logical than placing the `--sudo` later in the line. Also, if parsed with the `args[j]` array, the somewhat ugly `let j=j-1` would need to be included.

mpreisler:
I agree, having that as the first arg makes sense.

Does removing the result files on remote host work with your changes? I would expect permission issues because they are removed with normal user but created as root.

openprivacy:
@mpreisler, good point though in my testing the temp files are removed just fine (maybe since the tmp dir is created by an unpriv user, the unpriv user can remove it even though it has root-owned files?). So it seems to work as is (on rhel7 and centos7) but I can add the 'sudo' in front of the `rm -r` on line 241 if you'd like.

mpreisler:
Hmm, I think a better solution is to `chown` the resulting files to the original user and group. I will test whether this is an issue at all later today.

mpreisler:
Turns out it's not a problem. I have merged this manually into `maint-1.2` so that this feature reaches users sooner. See 9f7cdb002d0a5f0f8442d6b446cb632337174c5b

I have also done some very minor changes on top to enable pseudo-tty for users that need to retype their passwords. See 7f6dac2e2b2d4dc9db6943b786ce0c312a644e21

Thanks a lot for your contribution!

PR #138, opened 2015-08-13
Groupby2

jan-cerny:
This is PR #129 targeted to maint-1.1.

Added updated xccdf-resources.xsl


mpreisler:
See #129 for review notes

PR #139, opened 2015-08-13
oscap_acquire use oscap_buffer

ybznek:
Some refactoring...

### DIFF
--- a/src/common/oscap_acquire.c
+++ b/src/common/oscap_acquire.c
@@ -36,6 +36,7 @@
 
 #include "oscap_acquire.h"
 #include "common/_error.h"
+#include "oscap_buffer.h"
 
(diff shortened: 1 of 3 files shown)


ybznek:
I am thinking about it now... maybe using oscap_string instead of oscap_buffer would by cleaner.

isimluk:
Well, I think that the downloaded content may be bzip. Thus, it may contain zeros, that means it is not really string. Or not?

isimluk:
Lgtm.

Though, I did not run tests.

mpreisler:
I would prefer `oscap_string` to be used. The resulting code will be equivalent but easier to read.

mpreisler:
ACK

PR #141, opened 2015-08-14
Test all .c source files for config.h

ybznek:
https://fedorahosted.org/openscap/ticket/485


mpreisler:
Discussed with @ybznek that we want this in `maint-1.0` and in a test separate from test_codestyle because the codestyle test is not in `maint-1.0`.

PR #145, opened 2015-08-25
Solve --results-arf segfault with xccdf without oval

ybznek:
Before  commit c7b718b394d0e049e556002c295973e08a5a1d8d (sep 2014) function **ds_rds_create_source** was called through **ds_rds_create** which initialized **oval_result_sources**.

Now **ds_rds_create_source** is called from **xccdf_session_export_arf** directly and "sometimes" oval_result_sources stay uninitialized (==NULL);

Fix: https://fedorahosted.org/openscap/ticket/500, maybe fix https://bugzilla.redhat.com/show_bug.cgi?id=1255654


ybznek:
Replaced by https://github.com/OpenSCAP/openscap/pull/146

PR #147, opened 2015-08-25
Use a value specified by <set-value> element

msrubar:
This is a fix for an issue which was reported by Jacob Varughese.

See: https://fedorahosted.org/openscap/ticket/498

### DIFF
--- a/src/SCE/sce_engine.c
+++ b/src/SCE/sce_engine.c
@@ -372,7 +372,11 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
 
 		char* name = xccdf_value_binding_get_name(binding);
 		xccdf_value_type_t type = xccdf_value_binding_get_type(binding);
-		char* value = xccdf_value_binding_get_value(binding);
+		char* value = xccdf_value_binding_get_setvalue(binding);
+		if (value == NULL)
+		{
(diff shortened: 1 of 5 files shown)


isimluk:
The trac ticket has an exemplary content attached. Can you please re-use that content to create a test case?

src/SCE/sce_engine.c
```diff
@@ -372,7 +372,11 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
 
 		char* name = xccdf_value_binding_get_name(binding);
 		xccdf_value_type_t type = xccdf_value_binding_get_type(binding);
-		char* value = xccdf_value_binding_get_value(binding);
+		char* value = xccdf_value_binding_get_setvalue(binding);
+		if (value == NULL)
+		{
+				value = xccdf_value_binding_get_value(binding);
```
mpreisler: 2 tab indent?

msrubar:
To **isimluk**: thank you for the tips. I got inspired in the ```tests/sce/test_sce.sh``` script. Anyway it should be all fine now.

isimluk:
Perfect.

Thank You!

mpreisler:
One question, doesn't this change break the refine-value functionality with SCE? What if there is no `set-value` but `refine-value` instead?

PR #148, opened 2015-08-31
trac#501: Do not print tooltips in printed form of HTML report

jan-cerny:
Tooltips will not be printed because it was repeating redundant information.
It caused lines to span to a half of page.


jan-cerny:
If this will be approved, it also requires to push here updated xccdf-resources.xsl

xsl/xccdf-resources/openscap.css
```diff
@@ -40,3 +40,10 @@ div.top-spacer-10
 {
     margin-top: 10px;
 }
+
+@media print
+{
+    .container { width: 100% }
```
mpreisler: what is the reason behind the `width: 100%` rule?

jan-cerny: It is to do not have very big unused white edges of paper on left and on right.

mpreisler:
`::after` is the correct standard CSS syntax, `:after` is done for compatibility with IE8, would be great to have a short comment about that in the CSS.

jan-cerny:
Is it really needed to write there comments like that? Is there anything that blocks merging this?

mpreisler:
It is not strictly necessary but it is important. When the next person glances over the CSS they want to know why we are using invalid syntax.

jan-cerny:
Reworked - less generic selectors + a comment

mpreisler:
Thx

PR #153, opened 2015-09-04
oscap_string fix visibility

ybznek:
Some functions were not visible from public API in 1.2.5 and we don't want to make them public now.


ybznek:
test this please

mpreisler:
Thanks for spotting this, this was most likely a merging mess-up by me.

PR #154, opened 2015-09-07
Support check_existence attribute in state entities + test

jan-cerny:
This is a new feature in OVAL 5.11.1


src/OVAL/oval_stateContent.c
```diff
@@ -242,6 +266,12 @@ xmlNode *oval_state_content_to_dom(struct oval_state_content * content, xmlDoc *
 	oval_check_t ent_check = oval_state_content_get_ent_check(content);
 	if (ent_check != OVAL_CHECK_ALL)
 		xmlNewProp(content_node, BAD_CAST "entity_check", BAD_CAST oval_check_get_text(ent_check));
+	// TODO: serialize check_existence only if OVAL core schema version >= 5.11.1
```
jan-cerny: blocked by #107

src/OVAL/oval_sexp.c
```diff
@@ -719,6 +720,12 @@ int oval_state_to_sexp(void *sess, struct oval_state *state, SEXP_t **out_sexp)
 			SEXP_free(r0);
 		}
 
+		oext = oval_state_content_get_check_existence(content);
+		if (oext != OVAL_EXISTENCE_UNKNOWN) {
```
msrubar: What if the oval_state_content_get_check_existence() will return OVAL_EXISTENCE_UNKNOWN? Can it happen?

src/OVAL/public/oval_definitions.h
```diff
@@ -2450,6 +2450,11 @@ void oval_state_content_set_varcheck(struct oval_state_content *, oval_check_t);
  * @memberof oval_state_content
  */
 void oval_state_content_set_entcheck(struct oval_state_content *, oval_check_t);
+/**
```
msrubar: Since it's public API it would be good to provide more documentation for doxygen like comment the parameter and return value. Even if it's obvious to us it won't be obvious to someone who will want to try build something on top of it using only doxygen doc.

src/OVAL/public/oval_definitions.h
```diff
@@ -2476,6 +2481,11 @@ oval_check_t oval_state_content_get_var_check(struct oval_state_content *);
  * @memberof oval_state_content
  */
 oval_check_t oval_state_content_get_ent_check(struct oval_state_content *);
+/**
+ * Get "check_existence" attribute of an OVAL state entity
```
msrubar: Comment the parameter and return value.

src/OVAL/results/oval_resultTest.c
```diff
@@ -465,6 +465,132 @@ static inline oval_result_t _evaluate_sysent(struct oval_syschar_model *syschar_
 	}
 }
 
+struct oval_item_statuses {
```
msrubar: Struct in the middle of the file? I would look for it at the beginning but it's just my opinion.

src/OVAL/results/oval_resultTest.c
```diff
@@ -465,6 +465,132 @@ static inline oval_result_t _evaluate_sysent(struct oval_syschar_model *syschar_
...
+		is->exists_cnt++;
+		break;
+	case SYSCHAR_STATUS_DOES_NOT_EXIST:
+		is->does_not_exist_cnt++;
+		break;
+	case SYSCHAR_STATUS_NOT_COLLECTED:
+		is->not_collected_cnt++;
+		break;
+	default:
+		oscap_seterr(OSCAP_EFAMILY_OSCAP, "Invalid oval status type: %s.", oval_syschar_status_get_text(status));
```
msrubar: Should it be OSCAP_EFAMILY_OVAL?

src/OVAL/results/oval_resultTest.c
```diff
@@ -465,6 +465,132 @@ static inline oval_result_t _evaluate_sysent(struct oval_syschar_model *syschar_
...
+			&& is->not_collected_cnt >= 0) {
+			result = OVAL_RESULT_ERROR;
+		} else if ((is->exists_cnt == 0 || is->exists_cnt == 1)
+			&& is->does_not_exist_cnt >= 0 && is->error_cnt == 0
+			&& is->not_collected_cnt >= 1) {
+			result = OVAL_RESULT_ERROR;
+		}
+		break;
+	default:
+		oscap_seterr(OSCAP_EFAMILY_OSCAP, "Invalid check_existence value: %s.",
```
msrubar: EFAMILY_OVAL?

src/OVAL/results/oval_resultTest.c
```diff
@@ -465,6 +465,132 @@ static inline oval_result_t _evaluate_sysent(struct oval_syschar_model *syschar_
...
+				oval_existence_get_text(check_existence));
+		result = OVAL_RESULT_ERROR;
```
msrubar: If non of the cases is hit then the result is already set to ```OVAL_RESULT_ERROR```.

tests/API/OVAL/unittests/test_state_check_existence.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+result=`mktemp`
+
+set -e
+set -o pipefail
+
+cat > /tmp/foo.txt <<EOF
```
msrubar: What if somebody else decided to use same file? I would consider to use the ```mktemp``` here

src/OVAL/oval_sexp.c
jan-cerny: I this case, the block after the condition will not be executed, that means the check_existence attribute will not affect the probe at all.

src/OVAL/public/oval_definitions.h
jan-cerny: You are right that our doxygen documentation of source code is bad. There are even some functions that do not have any documentation, and most of the functions have only very short description.

src/OVAL/results/oval_resultTest.c
jan-cerny: Perhaps move all connected with this to a separate C file.

Regarding the name of this structure, I would like to ask someone who speaks English better than me, because "stats" would mean for me "statistics", which is not this case.

src/OVAL/results/oval_resultTest.c
jan-cerny: Yes, you are right, thank you.

tests/API/OVAL/unittests/test_state_check_existence.sh
jan-cerny: I always try to use `mktemp` where possible, but this is not the case. The name of the file appears in the attached OVAL content used in a textfilecontent54_test. ( `tests/API/OVAL/unittests/state_check_existence.xml` )

src/OVAL/results/oval_resultTest.c
jan-cerny: Yes, you are right, thank you.

src/OVAL/results/oval_resultTest.c
isimluk: I vote for moving this to separate file.

tests/API/OVAL/unittests/test_state_check_existence.sh
isimluk: I was in a similar situation in the past. In `test_remediation_simple.oval.xml` I used local file in the build directory for a similar purpose. What do you think about such solution?

jan-cerny: @isimluk Unfortunately, a similar solution here is not possible, because it is not possible to use a relative path in //textfilecontent54_test/filepath element. The OVAL specification says regarding filepath: 
> The filepath element specifies the absolute path for a file on the machine. A directory cannot be specified as a filepath.

jan-cerny: @msrubar : My only idea is to generate the OVAL file dynamically using a bash script. The temporary file will be crated by mktemp and then the file name will be put in all possible occurences in the OVAL file. Have you got a better idea?

msrubar: This one come to my mind after I read the @isimluk note. I think if would be better then the first solution. May @mpreisler or @ybznek could give us their opinions.

ybznek: I don't have better idea than https://github.com/OpenSCAP/openscap/pull/154/files#r39490287  ...
I prefer something like that pseudo-bash:
cat \<oval file with relative paths\> | sed replace \<relative path =\> absolute path\> | \`mktemp\`

IMHO generating of whole oval file in bash isn't good solution.

jan-cerny:
I have discussed the naming issue with @isimluk and so I have renamed "Oval_item_statuses" to "oval_item_status_counter".

isimluk:
Jan, one of my remarks has been that that `oval_item` name, is not yet present in OpenSCAP. The oval:item is a term that is used in OVAL standard. However, in OpenSCAP we have structure called oval_sysitem for representation of oval:item.

What is the relation of oval_item_status_counter to `oval:item` or `oval_sysitem`?

jan-cerny:
@isimluk In specification of this newly implemented check_existence attribute they always use word *"item"*.  Those *items* are called `struct oval_sysent` in our implementation and are members of the `struct oval_sysitem`.

isimluk:
Can you please provide a link to a documentation which uses the word "item"?

In other words, I am trying to understand, why have you choosen `oval_item_status_counter` over `oval_status_counter`? I remember we have agreed `oval_status_counter` before.

jan-cerny:
I was thinking about the reason why I choose different name than we agreed, and now I finally got it. The problem was that I wrote down to my notes `rename it to oval_item_status_counter` by mistake instead of `oval_status_counter` that we agreed about, so I followed my notes. I am sorry for this, it was not intended.
`oval_item_status_counter` also makes sense, but it is not consistent with other names.

I am going to rename it again.

P.S.: 5.11.1 HTML documentation is not online, I have them offline, there are only schemas on OVAL github.

jan-cerny:
Renamed to oval_status_counter

isimluk:
Thank You very much!

PR #157, opened 2015-09-09
Port oscap-docker from bash to python

baude:
This supports two key changes in scanning docker images
    and containers where now both atomic and openscap-daemon
    are capable of scanning.  The port to python should allow
    more utilities to leverage the existing openscap-docker
    function and create less diversion.


mpreisler:
@ybznek : Could you help me check that this passes through our `oscap-docker` integration test? If I remember correctly you wrote the test.

ybznek:
@mpreisler test is for beaker. I can now run tests manually with downloaded files

utils/oscap-docker.py
```diff
@@ -0,0 +1,82 @@
+''' oscap docker command '''
```
ybznek: Probably missing shebang

ybznek:
Can you please provide some list of dependecies? Thanks
```
  File "./oscap-docker.py", line 4, in <module>
    from oscap_docker_python.oscap_docker_util import oscapScan
  File "/mnt/testarea/test/oscap_docker_python/oscap_docker_util.py", line 6, in <module>
    from Atomic.mount import DockerMount
ImportError: No module named Atomic.mount
```

mpreisler:
> Traceback (most recent call last):
>   File "./oscap-docker.py", line 4, in <module>
>     from oscap_docker_python.oscap_docker_util import oscapScan
>   File "/mnt/testarea/test/oscap_docker_python/oscap_docker_util.py", line 6, in <module>
>     from Atomic.mount import DockerMount
> ImportError: No module named Atomic.mount

`dnf install atomic` should be enough if you are testing on Fedora 22.

baude:
You may need a newer atomic depending on where and how you test.  Some of the mount function was present but you really need atomic with -> https://github.com/projectatomic/atomic/pull/121 to fully use this.

I don't see new packages built yet. Ill circle back with Dan.

mpreisler:
This has been merged manually as is not part of maint-1.2.

See 68db46beef9885b044b1b884fc7d21d25cc13f46

PR #159, opened 2015-09-13
Show process name and thread name in debug logs

jan-cerny:
I was playing a bit with threads and processes. For my better understanding, I gave threads some names and I showed the process and thread names in debug logs. I thought that this might be useful also for other people, so I propose it as a pull request.

Debug log before:
`(22827:7f9815527700) [I:seap-packet.c:903:SEAP_packet_recv] Received packet`
And after:
`(lt-probe_file(22827):input_handler(7f9815527700)) [I:seap-packet.c:903:SEAP_packet_recv] Received packet`


isimluk:
Jan, do we have any documentation on the topic of various threads here? What does the `icache_worker` do? What do the `command_worker` and `worker` do?

PR #160, opened 2015-09-17
Support OVAL 5.11.1 in rpmverifyfile probe + test

jan-cerny:
New attributes "nocaps" and "nodigest" in rpmverifyfile_object/behaviors were introduced and attribute "nomd5" was deprecated in OVAL 5.11.1. Also a new child element "filedigest_differs" within rpmverifyfile_item deprecates element "md5_differs" in OVAL 5.11.1. 

The test focuses on parsing and processing of these new elements. but we haven't any test of this probe before.


tests/probes/rpmverifyfile/test_probes_rpmverifyfile.sh
```diff
@@ -0,0 +1,85 @@
...
+. ../../test_common.sh
+
+set -e -o pipefail
+
+function test_probes_rpmverifyfile {
+    probecheck "rpmverifyfile" || return 255
+
+    DF="test_probes_rpmverifyfile.xml"
+    RF="results.xml"
+
```
ybznek: please, use whole words to make names of variables

src/OVAL/probes/unix/linux/rpmverifyfile.c
```diff
@@ -345,10 +345,22 @@ void probe_fini (void *ptr)
...
 #define VF_RESULT(f) (res->oflags & (f) ? "not performed" : (res->vflags & RPMVERIFY_FAILURES ? "not performed" : (res->vflags & (f) ? "fail" : "pass")))
 #define FF_RESULT(f) (res->fflags & (f) ? true : false)
 
+	oval_version = probe_obj_get_platform_schema_version(probe_ctx_getobject(ctx));
+	if (oval_schema_version_cmp(oval_version, OVAL_SCHEMA_VERSION(5.11.1)) >= 0) {
+		digest_attr_name = "filedigest_differs";
+		digest_attr_value = VF_RESULT(RPMVERIFY_FILEDIGEST);
+	} else {
+		digest_attr_name = "md5_differs";
+		digest_attr_value = VF_RESULT(RPMVERIFY_MD5);
```
ybznek: screenshot_2015-09-17_18-11-54
These lines aren't covered by your test ( I haven't started full test suite).
Why aren't they covered?

jan-cerny: That would require to add another test that will use content older versions of OVAL.

mpreisler:
The md5_digest was deprecated in OVAL 5.11.1 but not removed, right?

The code seems to suggest that we only handle filedigest in 5.11.1 and only md5 in older versions. I would expect to handle md5 everywhere and filedigest for 5.11.1 and later.

I agree with Zbynek that a test for the md5 functionality for both 5.11.1 and older OVAL would be helpful.

jan-cerny:
@mpreisler : Thank you for your feedback. You are true that **lin-sys:rpmverifyfile_item/lin-sys:md5_differs** element was not removed from the OVAL Linux System Characteristics specification. It was only deprecated. I am aware of it since I started working on this feature and I know that it means we must support it.

But I don't see how can my implementation break support for **nomd5** attribute of **lin-def:rpmverifyfile_object/lin-def:behaviors** element. If you could have a look to `src/OVAL/probes/unix/linux/rpmverifyfile.c`, line 400 in this pull request, you will see that **nomd5** attribute is still there, I have only put there a comment that it is deprecated. So if we parse some content where is this attribute used, it should not fail. 

When creating system characteristics in the probe, I am creating the elements depending on OVAL version, that is true. This is done by checking for the RPMVERIFY_FILEDIGEST flag (filedigest_differs element) and RPMVERIFY_MD5 flag (md5_differs element). Those flags are defined in `<rpm/rpmvf.h>` as follows:

```C
enum rpmVerifyAttrs_e {
    RPMVERIFY_NONE  = 0,        /*!< */
    RPMVERIFY_MD5   = (1 << 0), /*!< from %verify(md5) - obsolete */
    RPMVERIFY_FILEDIGEST= (1 << 0), /*!< from %verify(filedigest) */
...
```
So the user gets the same thing, only with a new cool name.

I would be happy if you would help me with finding where is the problem, because unfortunately I can't see it now. Thanks!

jan-cerny:
I will have a look why the build fails asap. I can also create a similar test with lets say OVAL 5.11 content. I only thought I have not so much time to do this because it seemed to me as a low priority.

tests/probes/rpmverifyfile/test_probes_rpmverifyfile.sh
```diff
@@ -0,0 +1,85 @@
...
+# OpenScap Probes Test Suite.
+
+. ../../test_common.sh
+
+set -e -o pipefail
+
+function test_probes_rpmverifyfile {
+    probecheck "rpmverifyfile" || return 255
+
+    DF="test_probes_rpmverifyfile.xml"
```
ybznek: maybe, something like "$srcdir/test_probes_rpmverifyfile.xml" can help.

tests/probes/rpmverifyfile/test_probes_rpmverifyfile_older.xml
```diff
@@ -0,0 +1,37 @@
...
+
+  <objects>
+    <lin-def:rpmverifyfile_object id="oval:x:obj:1" version="1" comment="Object">
+        <lin-def:behaviors nolinkto="true" nosize="true" nouser="true" nogroup="true" nomtime="true" nomode="true" nordev="true" noconfigfiles="true" noghostfiles="true" nomd5="true"/>
+        <lin-def:name operation="pattern match"/>
+        <lin-def:epoch operation="pattern match"/>
+        <lin-def:version operation="pattern match"/>
+        <lin-def:release operation="pattern match"/>
+        <lin-def:arch operation="pattern match"/>
+        <lin-def:filepath>/etc/redhat-release</lin-def:filepath>
```
mpreisler: I don't like that we are relying on `/etc/redhat-release` in an upstream test. Could we choose a different file? Could we skip the test if /etc/redhat-release is not available?

jan-cerny: @mpreisler Thank you for pointing this out, I have learned thanks to this that there are not-RedHat systems who use rpm. My suggestion how to solve this is using filepath `/usr/bin/sh` beacuse I think shell is on every system. Is it a good idea?
I don't like the idea about skipping the test, I would prefer to find a solution.

mpreisler: What about just `/bin/sh`? Not every distro has done the `usermove` thing.

jan-cerny:
@mpreisler  `/bin/sh` was the first thing that came to my mind when you were against `/etc/redhat-release`, but unfortunately it did not work for me with rpmverifyfile_test, so I moved my mind to `/usr/bin/sh`, because I did not know that it is not everywhere. When you explained why `/bin/sh`, I tried to investigate why it did not work. It seems to me as a bug in the probe.  I have filed an issue https://github.com/OpenSCAP/openscap/issues/170. Please see it for more details.

But I don't want to block this pull request till this issue is fixed and also I don't want to mix the of that issue to this pull request.

So I propose this solution: I will use `/usr/bin/sh` and I will use your idea about skipping the test when that file is not available.

isimluk:
/etc/redhat-release is often present on systems with rpm package manager. Otherwise, there is /etc/system-release which should be quite omnipresent. But no idea about Manjaro Linux (@ybznek :] ).

Anyway, /etc/passwd should be always present and owned by some rpm.

[test this please]

jan-cerny:
@isimluk Before, /etc/passwd did not work for me, I thought that it is also related to our issue so I decided to look for something else. But when you mentioned /etc/passwd again, I became eager to investigate why it did not work for me, and I found that I had `noconfigfile="true"` in my OVAL content, which caused skipping this file. So there is no issue with /etc/passwd, it works well. I have pushed here a commit which updates the test and replaces /etc/redhat-release with /etc/passwd.

isimluk:
LGTM.

mpreisler:
Yup. Kudos!

PR #161, opened 2015-09-17
Add test for loading OVAL using relative path

ybznek:
If you reference OVAL file from XCCDF using relative path, oscap will load OVAL file correctly.
But when your $PWD != dirname \`file.xccdf.xml\` oscap doesn't load it.

https://fedorahosted.org/openscap/ticket/507

Only non-passing test for now. 

Do not merge now.

[skip ci]


ybznek:
Replaced by https://github.com/OpenSCAP/openscap/pull/173

PR #162, opened 2015-09-18
Improve caching and fetching of CVE input data

baude:
Previously, we determined if new CVE data should be
    fetched by specificying the max age of the files in
    units of hours.

    The new implementation uses the following to determine
    if it should fetch the new file:

        * If the file isn't in the local cache dir, fetch it
        * If the file is local, check the mtime of the remote
          file and compare it to the local file.  If they
          differ by more than two seconds, fetch a new one.
        * Else, do not fetch a new one.

    We also now return the fully qualified file name in question
    from _fetch_single.  And in the case of fetch_dist_data, we
    return a list of all the filenames.  This is in preparation
    of the files being compressed upstream as we will want to know
    which file(s) we are dealing with.

    And finally, debug was added to the time related functions so
    they are easier to debug.  This will also come in handy when
    we contemplate handling of compressed and/or uncompressed
    files.

### DIFF
--- a/utils/oscap_docker_python/get_cve_input.py
+++ b/utils/oscap_docker_python/get_cve_input.py
@@ -17,8 +17,9 @@
 
 import urllib2
 import urlparse
-from os.path import join, exists, getmtime
-import time
+from os.path import join, exists
+from os import stat, utime
+import datetime
 
 
 class getInputCVE(object):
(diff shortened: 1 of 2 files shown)


isimluk:
[test]

ybznek:
ok to test

mpreisler:
sidenote: We can ignore failing Jenkins in this case, it's a false negative.

mpreisler:
@ybznek offered to review this, thanks!

mpreisler:
ok to test

utils/oscap_docker_python/get_cve_input.py
```diff
@@ -28,52 +29,106 @@ class getInputCVE(object):
...
 
-    def _is_recent_enough(self, hours, dist):
+        # Correct Last-Modified timestamp
+        remote_ts = dict(resp.info())['last-modified']
+        epoch = datetime.datetime.utcfromtimestamp(0)
+        remote_dt = datetime.datetime.strptime(remote_ts, self.remote_pattern)
+        seconds_epoch = (remote_dt - epoch).total_seconds()
+        utime(dest_file, (seconds_epoch, seconds_epoch))
+
+        return self.dest
```
ybznek: Return value is currently used only in "fetch_dist_data" function, which is not called from other part of code now. I think, it should be "return dest_file", or am I wrong?

baude: You are correct.  Do you want me to fix and ammend or do you want to fix with the merge?

ybznek: Ok. I will fix it with the merge

PR #163, opened 2015-09-22
tests: oval_details fix make distclean

ybznek:
make distclean was deleting important test files


mpreisler:
Why not `maint-1.1`?

ybznek:
Do we still have any reason to maintain 1.1?

mpreisler:
`maint-1.1` has not been announced as dead, so we maintain it for now.

PR #165, opened 2015-09-25
Distribute the html version of the oscap user manual

msrubar:
There is no asciidoctor in the RHEL5 so we have to distribute the oscap user manual in the HTML format.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -73,6 +73,7 @@ EXTRA_DIST =	m4/gnulib-cache.m4 \
 		docs/examples/oval_probes.py \
 		docs/examples/package-test.xml \
                 docs/oscap-scan.cron \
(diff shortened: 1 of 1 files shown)


msrubar:
Wrong branch it should go to the **maint-1.0**.

mpreisler:
Thx, no need to do another PR, I will just cherry-pick it into maint-1.0.

mpreisler:
Verified that manual.html gets included in tarball with `make dist`.

Merged, see https://github.com/OpenSCAP/openscap/commit/6d5303840b5c61c325df34f960f19cf5e4a5b1b1

One thing that kind of bothers me is that we don't check for asciidoctor as part of `./configure`. We assume it's in `PATH`. I think it's fine and doesn't block merging this PR but wanted to mention this here.

PR #169, opened 2015-09-28
Add `--verbose` option to OpenSCAP

jan-cerny:
Nowadays OpenSCAP contains some debugging information statements.
But they are not available to normal user, only to a developer who
compiles the program with `--enable-debug` option.
This commit will make them accessible to all users. They will be possible
get by run OpenSCAP with `--verbose` or `-v` option. Unlike previous state,
information from both library and probes will be printed out to a single file,
determined by a program argument.


jan-cerny:
This pull request was opened only as a place for discussion. I don't want to be merged now, because there are more possibilities how to implement this and there are pros and cons. Thanks for all feedback.

* Output can be either stored in an extra file (as proposed in this commit) or can be send to stderr (I tried it and it worked for me).
* I think that the debugging functions should not be part of public API, because I will probably change their declarations in future, so I would like to struggling with backward compatibility.

mpreisler:
Have you thought implementing this as a callback instead? This would work better with workbench where you want the output displayed in a GUI dialog instead of a file.

mpreisler:
Could you paste sample output here?

jan-cerny:
No, I didn't thought about it, thanks for an idea

isimluk:
Martin, indeed we have been thinking about callback before. Jan has designed a complex solution with threads, readers & writers to shared memory. I suggested to keep things simple in first step.

isimluk:
I have been talking with Jan about this. I think we first need to understand the debugging use-cases better, before he can decide a direction for some of the changes.

For instance, it is not clear
 * whether we want to log into the file or directly to stderr
 * whether we want to have verbosity levels or not
 * whether we want to have a complex machinery for log filtering (callback, threads, ...)

Jan plans to analyze a use-cases. And he tries to find a use-case which cannot be handled by verbosity levels.

mpreisler:
Currently many SSG developers (@iankko included) run SCAP Workbench repeatedly to test content. It would be of great help if they could see the verbose output in a window somewhere.

Making the callback itself reentrant is the job of the 3rd party -- for example SCAP Workbench, not of openscap. In SCAP Workbench I would surely add the messages in a thread-safe manner.

I don't think it's necessary to have this in the first iteration, just wanted to have this use-case out there.

jan-cerny:
@mpreisler @isimluk Thank you very much for your suggestions and ideas. @isimluk is completely true that I must analyze usecases and verbosity levels and how they fit together. Without it, I can't move ahead. I am going to write it down during this weekend.

jan-cerny:
Write-up is here: https://docs.google.com/document/d/1zToieqBXwCIUsVnRjgrK16_-0LXkCkQWWYej8StiWys/edit?usp=sharing

isimluk:
Jan, thank you! There is a lot of good thoughts in the write-up.

I am missing some kind of conclusion, what approach do you want to choose and why.

Also, I would welcome more clarity in the examples section of "Use-cases that can't be solved by selecting some level". There are some examples, however, I miss the purpose of each. As a content-developer, why would I choose to pursue each of the examples. With some examples, it is more clear then with others. I am trying to make my mind, if we can ignore these examples all together (and force people to run content_processing mode) or not.

jan-cerny:
@isimluk Thanks. I thought that conclusions will be added after your feedback. I now add conclusions and clarify examples now. I am afraid that I don't understand 
>(and force people to run content_processing mode)

Could you explain it more, please?

isimluk:
Thanks for writing the conclusion. You have convinced me towards the log file (as I have commented in the document). I am just not sure about the issue of command-line options:

Version A:
```--verbose /tmp/log --verbosity-level=WARNINGS```
Version B:
```--verbose WARNINGS --verbose-output /tmp/log```

I see some advantages of B, but what do you think?

>>(and force people to run content_processing mode)
>
> Could you explain it more, please?

I have put specific comments to the document. I trying to understand how important the "Use-cases that can't be solved by selecting some level" are. If we are sure, these use-case are important, we may want to implement them. However, if we found that anyone with grep can achieve the similar outcome, I would vote for doing just one thing (getting the logs out).

That is what I meant by "forcing people" to use level=info/content_processing or level=debug/devel if they need more info.

**Comment by iankko** [comment]
@isimluk 
> Thanks for writing the conclusion. You have convinced me towards the log file (as I have commented in the document). I am just not sure about the issue of command-line options:
> 
Version A:
--verbose /tmp/log --verbosity-level=WARNINGS
Version B:
--verbose WARNINGS --verbose-output /tmp/log
> 
I see some advantages of B, but what do you think?

Yet another proposal - what about to follow traditional Unix (e.g valgrind approach):
* use ```--log-file``` option to indicate filename where to store the information (or ```--verbose-log-file``` should this be desired),
* indicate desired verbosity level with count of ```-v``` / ```-verbose``` options (repeating the option increasing the verbosity)

The option ```log-file``` above has the advantage that if we would want logging information to go e.g. into fd or socket in the future, it could be easily enhanced and would still follow the format (```--log-fd``` and ```--log-socket```).

Just my 2cents.

**Comment by jan-cerny** [comment]
Thank you all for valuable feedback! Me and @isimluk discussed this PR today. The result is:
* Output will be to a log file (not stderr). 
* I will implement it using two options:  `--verbose` sets verbosity level,  `--verbose-log-file` sets output file. If there will be an idea with a better name, I'm all in, but I think these are acceptable. 
* I will change man page and bash completion.
* I will implement verbosity levels.
* My aim is to be able to remove the "do not merge" label. I don't want to introduce any big complex functionality here.

File: src/common/debug.h [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ *
+ * Authors:
+ *       Jan Černý <jcerny@redhat.com>
+ */
+ 
+#ifndef OSCAP_DEBUG_PRIV_H_
+#define OSCAP_DEBUG_PRIV_H_
```
jan-cerny: Sorry this needs to be fixed

jan-cerny:
The tasks from my previous comment seem to be done. @isimluk Could you kindly review?

src/common/debug.c
```diff
@@ -41,16 +40,27 @@
...
+    {DBG_UNKNOWN, NULL}
+};
+
 #  if defined(OSCAP_THREAD_SAFE)
 #   include <pthread.h>
 static pthread_mutex_t __debuglog_mutex = PTHREAD_MUTEX_INITIALIZER;
 #  endif
 static FILE *__debuglog_fp = NULL;
-int __debuglog_level  = -1;
+int __debuglog_level = DBG_UNKNOWN;
```
isimluk: Why the neighbors are `static` while the level is not `static`?

jan-cerny: It is because it is declared as `extern` in the header file. But after I have removed macro `debug(l)`, it is no longer needed to declare it as `extern`. Nobody uses it now. I think I can remove that declaration from header file, and add `static` modifier to it, because it is used only inside this module.

jan-cerny:
@isimluk Thank your very much for your review. I fixed the problem with "static" modifier on Friday in the latest commit. I started new build at Jenkins and it has just finished successfully.

utils/oscap-tool.c
```diff
@@ -35,6 +35,7 @@
 #include <errno.h>
 #include <limits.h>
 #include <cvss_score.h>
+#include "../src/common/debug.h"
```
isimluk: This needs to be changed to `#include <oscap_debug.h>`

The `oscap` tools is compiled only using public headers. So, when we introduce a new dependency we need to make it public. That way we ensure that anyone else (without the library sources can implement their own scanner).

Please move `debug.h` to `public/oscap_debug.h` and also update the `src/common/Makefile.am` appropriately.

isimluk:
Jan, for some reason, this PR breaks `configure --enable-debug && make` for me. Can you please confirm that on your system?

src/common/debug.h
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Turn on debugging information
+ * @param verbosity_level Verbosity level
+ * @param filename Name of file used as output file for store debugging
+ *                 and other additional information.
+ * @param is_probe Determines whether the function is called
+ *                 from a probe (true) or from the base library (false).
+ * @return When the verbose mode has been turned on it returns true, otherwise false.
+ */
+bool oscap_set_verbose(const char* verbosity_level, const char *filename, bool is_probe);
```
isimluk: This is just a nitpick, but please use ` *` instead of `* `.

I would be quiet, but there is an inconsistency on a single line.

src/common/debug.h
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Parse verbosity level from a string.
+ * @param level_name Verbosity level as a C string.
+ * @return Verbosity level number (or -1 on error).
+ */
+int oscap_verbosity_level_from_cstr(const char *level_name);
```
isimluk: I like you have put this public.

Please put `oscap_verbosity_levels` also public to this file. Then you can change prototype to `oscap_verbosity_levels oscap_verbosity_level_from_cstr(const char *level_name);`

isimluk: The other option would be to force `oscap_verbosity_level_from_cstr` to be static.

src/common/debug.c
```diff
@@ -68,6 +78,34 @@ static void __oscap_debuglog_close(void)
...
+		setenv("OSCAP_PROBE_VERBOSITY_LEVEL", verbosity_level, 1);
+		setenv("OSCAP_PROBE_VERBOSE_LOG_FILE", filename, 1);
+		__debuglog_fp = fopen(filename, "w");
+	}
+	if (__debuglog_fp == NULL) {
+		oscap_seterr(OSCAP_EFAMILY_OSCAP, "Failed to open file %s.", filename);
+		return false;
+	}
+	setbuf(__debuglog_fp, NULL);
+	__debuglog_level = oscap_verbosity_level_from_cstr(verbosity_level);
```
isimluk: I would prefer to move this line upper. That way you can check what user entered. If they enters 'BLA' you may want to stop proceeding (i.e. not open any file).

utils/oscap-oval.c
```diff
@@ -217,6 +224,9 @@ int app_collect_oval(const struct oscap_action *action)
 	struct oval_generator		*generator = NULL;
 	int ret = OSCAP_ERROR;
 
+	/* Turn on verbosity */
+	oscap_set_verbose(action->verbosity_level, action->f_verbose_log, false);
```
isimluk: I would wrap this by if (...) goto cleanup;

If the verbosity level is not set correctly. However, I have no strong opinion here. What do you think?

jan-cerny: Correct setting of verbosity level is checked by a function `check_verbose_options`. It is called during options processing. If a user runs `oscap` with verbosity level wrongly specified, he will get an error message and oscap will terminate.

However, we should wrap this by `if goto cleanup` to terminate when the output file can't be opened.

jan-cerny:
@isimluk I confirm that it happened to me. But after rebasing and adding new commits it unfortunately doesn't happen anymore. It's working now. Perhaps it was some conflict of header files.

isimluk:
`tests/API/XCCDF/report` in check and distcheck fail for me. Is that something you are experiencing as well or my borked set-up?

jan-cerny:
@isimluk yes, same to me, because of 6b278aad5943249dbc871ccca710306be4b76db2

isimluk:
This looks fine now. Thank You @jan-cerny!

I have been playing with the tool and I have found it quite useful!

PR #173, opened 2015-10-06
Fix loading oval results from relative path

ybznek:
https://fedorahosted.org/openscap/ticket/504

When you want to create HTML report from XCCDF which references OVAL using relative path using --oval-details, oscap tries to create file e.g. ./my/path/file.oval.xml.result.xml
- when the folder "./my/path/" doesn't exist, it was causing segfault - the folder usually exists (because it is the same directory as directory of existing oval file) - but when your cwd is different from directory of xccdf, the relative directories won't match

This pull requst:
- add tests (other tests are in beakerlib format)
- fix path when generating of oval-results files
- add OVAL details to HTML report from (internally generated) ARF not from oval result.files - which fix problem with loading oval-details from escaped path.

### DIFF
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -86,6 +86,7 @@ struct xccdf_session {
 		struct oval_agent_session **agents;	///< OVAL Agent Session
 		xccdf_policy_engine_eval_fn user_eval_fn;///< Custom OVAL engine callback
 		char *product_cpe;			///< CPE of scanner product.
+		struct oscap_source* arf_report;	///< ARF report
 		struct oscap_htable *result_sources;    ///< mapping 'filepath' to oscap_source for OVAL results
 	} oval;
 	struct {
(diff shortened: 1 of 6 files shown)


mpreisler:
This replaces #161 , am I right?

isimluk:
Well done.

Interestingly, jenkins new the sha of first commit, so he decided to not re-run the tests. I'll wait for the tests and then I'll merge this. Thanks!

[test this please]

ybznek:
Yes, it replaces #161 

I did rebase of these commits in pullrequest.
I've
- modified some commit messages
- fix whitespace
- fixed oscap_free() -> oscap_source_free
- and maybe fixed tests - after make distcheck oscap didn't have access to create oval results files

isimluk:
ybznek, I see environmentvariable58 test failing again.

Perhaps this is not related to the PR, however, would you mind to investigate more? What was the cause? Are the test passing for you locally?

ybznek:
I cannot take a look at this before friday afternoon. But probably logs from environment variable tests are replaced by other test now.

But still my test is not passing. I've started only my single test after ammending. There is usually some problem with different environment after make distcheck - which was OK when I have started only my single test. It's quicker for me to run run whole distcheck test only on jenkins not on my pc & after on jenkins.

isimluk:
Understood.

Please feel free to merge this yourself, once you feel you handled the test failures.

ybznek:
I've fixed problems, but there are still random environmentvariable errors. But this PR should be ok.

PR #177, opened 2015-10-09
Fix errors of sysctl probe

jan-cerny:


jan-cerny:
Build failed due to a segfault in runlevel probe, which must be unrelated to this pull request. I am going to try build it again, manually.

src/OVAL/probes/unix/sysctl.c
```diff
@@ -109,9 +112,22 @@ int probe_main(probe_ctx *ctx, void *probe_arg)
...
 
                 snprintf(mibpath, sizeof mibpath, "%s/%s", ofts_ent->path, ofts_ent->file);
 
+		/* Skip write-only files, eg. /proc/sys/net/ipv4/route/flush */
+		if (stat(mibpath, &file_stat) == -1) {
+			dE("Stat failed on %s: %u, %s\n", mibpath, errno, strerror(errno));
+			oval_ftsent_free(ofts_ent);
+			continue;
+		}
+		if ((file_stat.st_mode & S_IRUSR) == 0) {
```
mpreisler: I am not sure if it's even a real possibility but what if the file doesn't have the "owner read" bit set but has "group read" or "other read" bit set? For correctness it would make sense to handle that here.

src/OVAL/probes/unix/sysctl.c
```diff
@@ -149,9 +165,21 @@ int probe_main(probe_ctx *ctx, void *probe_arg)
...
 
                         if (ferror(fp)) {
-                                dE("An error ocured when reading from \"%s\" (fp=%p): l=%ld, %u, %s\n",
-                                    mibpath, fp, l, errno, strerror(errno));
-                                goto fail_item;
+				/* Linux 4.1.0 introduced a per-NIC IPv6 stable_secret file.
+				 * The stable_secret file cannot be read until it is set,
+				 * so we skip it.
+				 */
+				if (strncmp(ofts_ent->path, ipv6_conf_path, ipv6_conf_path_len) == 0 &&
```
mpreisler: Wouldn't it make sense to check for this path before we attempt to read it? If we are sure it will cause an error we want to avoid even attempting to read it.

src/OVAL/probes/unix/sysctl.c
jan-cerny: I am also not sure if it is real possibility. The `sysctl` utility uses the same condition as me.

mpreisler: > The sysctl utility uses the same condition as me.

This would be a perfect comment in the source code, along with citing the file.

src/OVAL/probes/unix/sysctl.c
jan-cerny: @mpreisler This was my original idea. Originally I have checked for this path before fopen(), I have had this block of code somewhere about line 118. I have already discussed this with @isimluk. We realized that it would skip the file also if it is possible to read from it. When someone has written an ipv6 address to this file, it is possible to read it, and I think in such case the probe should collect it. For this use-case, I decided to move this check here.

PR #179, opened 2015-10-14
Download bzipped variants of RHSA OVAL data in oscap-docker

mpreisler:
Saves bandwidth and (at least in most cases) time.

We don't do any unbzip-ing because oscap does that on its own.

This change requires recent `oscap` tool with bzip support but I think it's not an issue since we ship both in closely related packages. The only issue I can think of is when somebody compiles openscap with oscap-docker but without bzip support. If we agree that this is a serious issue I suggest we blacklist that combination in `configure`.

### DIFF
--- a/ac_probes/configure.ac.tpl
+++ b/ac_probes/configure.ac.tpl
@@ -501,6 +501,12 @@ else
 fi
 AC_SUBST([vgcheck])
 
+if test "x${util_oscap_docker}" = "xyes"; then
+	if test ! "x${HAVE_BZIP2}" = xyes; then
+		AC_MSG_FAILURE(oscap-docker requires bzip2! Either disable oscap-docker or install bzip2.)
+	fi
+fi
+
 if test "x${perl_bind}" = xyes; then
 	AC_PATH_PROG(PERL, perl)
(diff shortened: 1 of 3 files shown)


mpreisler:
Jenkins failed on file probe test. That code is in no way related to the patch in this PR.

```
FAIL: test_probes_file.sh
make[7]: Entering directory `/home/jenkins/workspace/openscap-pull-requests/openscap-1.2.7/_build/tests/probes/file'
make[7]: Nothing to be done for `all'.
make[7]: Leaving directory `/home/jenkins/workspace/openscap-pull-requests/openscap-1.2.7/_build/tests/probes/file'
============================================================================
Testsuite summary for openscap 1.2.7
============================================================================
# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
```

mpreisler:
I have added the explicit fail at `./configure`. It will now fail if oscap-docker is enable but bzip2 wasn't found.

ybznek:
Tested by oscap-docker tests on my fedora22 and all checks pass.

ybznek:
I think, we could merge it.

PR #180, opened 2015-10-14
this patch autodetects the SUSE Linux Enterprise versions (without SP)

msmeissn:
and openSUSE versions

it hooks this on presence of the various -release RPM packages and their versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: c
+
+os:
+ - linux
+ - osx
+
+compiler:
+ - clang
+ - gcc
+
+script:
+ - ./configure && make && make check
+
+notifications:
+  email:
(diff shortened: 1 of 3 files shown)


mpreisler:
The CPE patch manually applied as 5a8295d4520f50aa2b23714e9d3809059666b8b1

I don't see why we should add `.travis.yml`, we are using Jenkins for CI. If you feel the OpenSCAP project could really benefit from `.travis.yml` please open another pull request and let's discuss it.

msmeissn:
ah sorry, that should not have been in there. i only saw afrterwards the jenkins bot triggered

mpreisler:
No problem, thanks a lot for your contribution!

PR #181, opened 2015-10-15
Add versioning manual

ybznek:
Add versioning manual

Waiting for https://github.com/OpenSCAP/website/issues/84

Do not merge it now.

### DIFF
--- a/docs/contribute/versioning.adoc
+++ b/docs/contribute/versioning.adoc
@@ -1,0 +1,53 @@
+= Versioning
+
+openscap release versions have the common *$major.$minor.$patch* format.
+
+Most of bleeding edge unstable development happens in the *master* branch. This branch is likely to break API and ABI and is likely not suitable for anyone but the developers. Only release candidates or other test releases are tagged from the master branch.
+
+*$major.$minor* together (e.g.: 1.0) are called *feature releases*. As the name suggests, once a feature release is made, no new features are added to it. For example once 1.0.0 is released, 1.0.1, 1.0.2, 1.0.3, etc. shall not have more features than 1.0.0. Patch version 0 (e.g.: 1.0.0) establishes the feature set.
+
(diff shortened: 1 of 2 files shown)


mpreisler:
Why `maint-1.2`? This needs to go into `maint-1.0` IMO.

msrubar:
I'm confused about the distribution. I thought that the original idea was just to put this document from the old website to the asciidoc format into the contribute folder. Why do we want the html version of it? Do we want to have in on the static page?

mpreisler:
That's a good point, @msrubar. Consumers of the tarball most likely don't care about our versioning policy. Does it make sense for us to put this document into the tarball?

ybznek:
I've opened pull request to make some place to discusion, so thanks for your feedback.

@mpreisler We was talking about reference from README.md and it is only in maint-1.2.

So, should I move it to maint-1.0 and don't compile to html?

docs/contribute/versioning.adoc
```diff
@@ -0,0 +1,53 @@
...
+Bugfixes applied to maintenance branches must be merged up. For example a bugfix applied to maint-1.0 shall be merged to maint-1.2 which then shall be merged to master.
+
+This rule makes sure we always carry bugfixes to new feature releases.
+
+In cases where the bugfix does not make sense in newer feature releases, a noop merge shall be done. For example a fix for functionality that was removed in newer feature release.
+
+== Maintenance status
+
+	* 1.2.x: supported
+	* 1.1.x: unsupported
```
msrubar: maybe a note that it'll be removed soon?

mpreisler: please change this to "1.1.x: unsupported since 2015-10-16."

ybznek:
Replaced by https://github.com/OpenSCAP/openscap/pull/185

PR #182, opened 2015-10-15
oscap-docker: Download bz2 version of OVAL

ybznek:
Use packed .bz2 version of OVAL file instead of .xml

Tested by oscap-docker tests on my fedora22


ybznek:
eh.. duplicate of https://github.com/OpenSCAP/openscap/pull/179

PR #183, opened 2015-10-16
Add how to contribute

msrubar:
We decided to put it into the maint-1.0 after discussion with @isimluk so we can have the user manual, README and all the contribute documents in the maint-1.0.

Merge it together with #181 because the versioning is referenced in the contribute document.


isimluk:
Michal this is perfect write-up. Imho we could advertise this on open-scap mailing-list!

This can be merged once the versioning.adoc link is moved.

msrubar:
@isimluk Thank you for the review. The versioning was moved. I'll advertise it to the list when it gets merged.

I would also like to ask other for any suggestions that could be added into the document. Especially @ybznek.

isimluk:
Perfect thanks!

I just came across a github guide on how to create links from markdown. https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links The current README looks weird with these long links there. Does it make sense to use the markdown links?

Sorry for not catching this in the first review.

PR #201, opened 2015-10-29
Support  OVAL 5.11.1 in dpkginfo probe

jan-cerny:
Last thing missing to finish OVAL 5.11.1 support. Dpkginfo probe is used on Ubuntu and Debian systems to collect information about installed software packages. New data type *debian_evr_string*  describing epoch, version and revision of a package replaces *evr_string* type in specification. I have tested it on Ubuntu 14.04 LTS.


isimluk:
Except for the ABI break it looks good to me.

configure.ac
```diff
@@ -632,7 +632,7 @@ echo
 echo '* Checking for apt_pkg library used by: dpkginfo '
 PKG_CHECK_MODULES([apt_pkg], [libapt-pkg >= 0.0],[],[
 SAVE_LIBS=$LIBS
-AC_SEARCH_LIBS([pkgInitConfig],[apt-pkg],[
+AC_SEARCH_LIBS([pkgVersion],[apt-pkg],[
```
isimluk: You need to edit ac_probes/configure.ac.tpl as well. And make sure that they are in sync by running `./confgen.sh`

jan-cerny: @isimluk Have you got any idea where and how I could start with this, please?

isimluk: 1) I would edit `ac_probes/configure.ac.tpl` similarly as you did with `configure.ac`.
2) Then I would run `./confgen.sh`. It will replace/regenerate configure.ac
3) `git add -p` and you are done.

jan-cerny: @isimluk There is nothing similar like this in configure.ac.tpl.

isimluk: Ok, so I was confused a little bit. Let me start again.

When I take this PR and I run `./confgen.sh ; git diff` it is not clean.

What we perhaps need to do is to patch `ac_probes/libs/apt_pkg` like:
```
-functions_req=(pkgInitConfig pkgInitSystem)
+functions_req=(pkgVersion)
```

and then run ./confgen.sh on Fedora. That way we ensure that confgen is always clean.

However, it is not clear to me whether it is enough for Debian build to pass. Can you please re-check on your debian vm that build passes?

isimluk: Also please check on Debian system that dpkg probe builds and 
```
$ grep lapt-pkg Makefile
apt_pkg_LIBS = -lapt-pkg)
```

jan-cerny: @isimluk OK

jan-cerny:
I can't run `./confgen.sh` on Ubuntu. 
```
ac_probes.sh... Package libapt-pkg was not found in the pkg-config search path.
Perhaps you should add the directory containing `libapt-pkg.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libapt-pkg' found
failed: 123
```
I have libapt-pkg-dev installed.

I can't find any *.pc file for libapt-pkg because `apt-file search /usr/lib/x86_64-linux-gnu/pkgconfig | grep libapt` gives me no results.

Could you help me, please?

mpreisler:
@jan-cerny Which pkgconfig files does the package install? Perhaps they changed the name.

jan-cerny:
@mpreisler The libapt-pkg-dev package on Ubuntu contains no pkgconfig files. But as you told me, if we don't want to be able to run `./confgen.sh` on Ubuntu, we don't have to solve Ubuntu's pkgconfig problems. I will amend this PR to run `./confgen.sh` on Fedora in such way that generated `configure.ac` is able to build dpkg probe on Ubuntu.

isimluk:
That is perfect solution.

Thanks Jan and Martin.

Issue #203, opened 2015-11-02
oscap oval eval --skip-valid doesn't skip validating


jan-cerny:
Current upstream git maint-1.2 : `oscap oval eval --skip-valid` runs the validation the OVAL content. It shouldn't validate! In maint-1.0 it works as expected, so this is a regression from maint-1.0.

PR #204, opened 2015-11-02
Issue #203: oscap oval eval --skip-valid doesn't skip validating

jan-cerny:
When refactoring the code into the "OVAL session", we have probably
forgotten about the --skip-valid option.


isimluk:
Perfect. I'll just wait for jenkins before merging.

Issue #191, opened 2015-10-27
Segfault --oval-results


ybznek:
I have found some openscap segfault during work with openscap-daemon
```
$ oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_common --oval-results   /tmp/tmplTsI9g.txt 
```

I was suspecting my pullrequest https://github.com/OpenSCAP/openscap/pull/173, but when I revert its changes, problem is still here.

When I revert this commit https://github.com/OpenSCAP/openscap/commit/5a8295d4520f50aa2b23714e9d3809059666b8b1, problem disappear. BUT the wrong xml content should not be able to cause segfault, so the main problem is not in this content, but in openscap.

I have attached the content(Yes, it is xml, but with txt extension - github doesn't allow me to upload xml)
[tmplTsI9g.txt](https://github.com/OpenSCAP/openscap/files/21527/tmplTsI9g.txt)

Original content from https://github.com/OpenSCAP/openscap-daemon/blob/master/tests/data_dir_template/tasks/1.xml

```
Program received signal SIGSEGV, Segmentation fault.
strlen () at ../sysdeps/x86_64/strlen.S:106
106		movdqu	(%rax), %xmm12
Missing separate debuginfos, use: dnf debuginfo-install nss-3.20.0-1.0.fc22.x86_64 pcre-8.37-4.fc22.x86_64
(gdb) where
#0  strlen () at ../sysdeps/x86_64/strlen.S:106
#1  0x00007ffff7b4f82d in oval_state_to_dom (state=0x1128460, doc=0x12ea440, parent=0x132c590) at oval_state.c:344
#2  0x00007ffff7b43626 in oval_definition_model_to_dom (definition_model=0x10b5750, doc=0x12ea440, parent=<optimized out>)
    at oval_defModel.c:556
#3  0x00007ffff7b70aa6 in oval_results_to_dom (parent=0x0, doc=0x12ea440, directives_model=0x10b7710, results_model=0x10ba510)
    at oval_resModel.c:257
#4  oval_results_model_export_source (results_model=results_model@entry=0x10ba510, directives_model=directives_model@entry=0x0, 
    name=name@entry=0x12eb450 "/home/zmoravec/gitRepos/openscap/%2Fhome%2Fzmoravec%2FgitRepos%2Fopenscap%2Fcpe%2Fopenscap-cpe-oval.xml.result.xml") at oval_resModel.c:281
[...]

PR #207, opened 2015-11-03
Fix invalid CPE content

jan-cerny:
This OVAL file was invalid. There were missing objects and states.
It was one of causes of issue #191 (segfault on oscap xccdf eval).


mpreisler:
@msmeissn Could you please test the resulting CPE and CPE OVAL on SLES?

msmeissn:
I applied the patch to my SLES package and oscap xccdf eval of our default scap profile works correctly. 

so it is good from my side.

mpreisler:
Great, thanks for testing, @msmeissn !

PR #208, opened 2015-11-03
Refactor + fix a segfault

jan-cerny:
This code repeats 4 times in OpenSCAP. This commit refactors it out
to a single function. This functions converts oval family to a XML namespace.

### DIFF
--- a/src/OVAL/oval_definitions_impl.h
+++ b/src/OVAL/oval_definitions_impl.h
@@ -40,6 +40,7 @@
 OSCAP_HIDDEN_START;
 
 oval_family_t oval_family_parse(xmlTextReaderPtr);
+xmlNs *oval_family_to_namespace(oval_family_t family, const char *schema_ns, xmlDoc *doc, xmlNode *parent);
 oval_subtype_t oval_subtype_parse(xmlTextReaderPtr);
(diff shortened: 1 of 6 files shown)


jan-cerny:
@isimluk I am sorry, but this is different than we agreed yesterday. Let me explain why. I have thought about a convenient name for this new function. I have realized what it does: It converts an OVAL family to appropriate xml namespace. Therefore I think suitable name for this is `oval_family_to_namespace`. I couldn't find a better name. And here is my problem. If it's name starts with `oval_family`, I think the best place where to define it is where `oval_family*` functions are defined. So  I have put it to `oval_enumerations.c` instead of creating new file `oval_commons.c`. What do you think about it?

jan-cerny:
Also fixing segmentation fault caused by processing invalid OVAL file. This fixes issue #191 .

mpreisler:
Do we all remember what refactoring in stable branches has done for us in the past? :-D

isimluk:
Thanks Jan!

Martin, I don't fully understand. What is your concern with this particular refactoring?

Once this is resolved, it can be merged.

mpreisler:
I don't have an issue, just wanted to stress that we need to be extra careful with refactoring.

jan-cerny:
OK

PR #209, opened 2015-11-04
Fix and test openscap-cpe-oval.xml

jan-cerny:
This OVAL file did not pass validation by schematron.
A rpminfo_test should not reference a rpmverifyfile_object,
it can be referenced only from a rpmverifyfile_test.
We must change tests to rpmverifyfile_test and
also change respective states.

### DIFF
--- a/tests/CPE/all.sh
+++ b/tests/CPE/all.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# Copyright 2015 Red Hat Inc., Durham, North Carolina.
+# All Rights Reserved.
+
+set -e -o pipefail
+
+. ${srcdir}/../test_common.sh
+
+function test_cpe() {
+    $OSCAP oval validate --schematron ${srcdir}/../../cpe/openscap-cpe-oval.xml
+    $OSCAP cpe validate ${srcdir}/../../cpe/openscap-cpe-dict.xml
+}
+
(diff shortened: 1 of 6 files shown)


jan-cerny:
I have fixed this even though it is a low priority, because now I can create test which will check that `openscap-cpe-oval.xml` is valid. Such test will help us to avoid issues similar to #191 in future.

jan-cerny:
Added a test.

tests/CPE/all.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# Copyright 2015 Red Hat Inc., Durham, North Carolina.
+# All Rights Reserved.
+
+set -e -o pipefail
+
+. ${srcdir}/../test_common.sh
+
+function test_cpe() {
+    $OSCAP oval validate --schematron ${srcdir}/../../cpe/openscap-cpe-oval.xml
```
mpreisler: Use `$top_srcdir` for better readability

mpreisler:
Have you tested applicability of the CPEs on all affected platforms?

isimluk:
Except for the Martin's suggestion on `top_dir` this LGTM.

jan-cerny:
@mpreisler No, I haven't.

isimluk:
If the test passes on all jenkins slaves, it is sufficient testing for me.

PR #215, opened 2015-11-16
Fix opening a log file

jan-cerny:
There was a serious issue in verbosing feature. Some messages were
missing in log file. It is fixed by opening log file in append mode
in both library and probes.


isimluk:
Nice catcha!

PR #219, opened 2015-11-18
Small message improvements

jan-cerny:
This pull request brings two very small improvements to verbose output, This pull request contains two commits. First commit improves a message regarding URI probe that is going to be started. Second commit adds a message that a SCAP content validation will be proceeded. It can be quite useful when we would like to see in log which documents are evaluated and which schema version  is used.


jan-cerny:
@isimluk Thank you very much for the review. I have amended the last commit. I have moved the message to devel level.

jan-cerny:
@isimluk Is this OK now? Can it be merged?

PR #225, opened 2015-11-22
Avoid traceback on rpmbuild errors

isimluk:
Chances are that the rpmbuild failure is not a problem in our source,
but in the inputs given to the tool.

Otherwise, we receive unneeded Abrt reports like
https://retrace.fedoraproject.org/faf/reports/723484/


jan-cerny:
In think it is not a good idea that the message will be lost and will not be displayed to the user.

isimluk:
Good catch. Will fix.

PR #244, opened 2015-11-30
Improve a warning

jan-cerny:
Adding a missing end of line and change wording of this message.


jan-cerny:
Missing end of line caused the next message continued on same line and it happened often during scanning Fedora SSG.

PR #253, opened 2015-12-07
Oscap-docker: python3 fix

ybznek:


utils/oscap_docker_python/get_cve_input.py
```diff
@@ -70,7 +70,7 @@ def _fetch_single(self, dist):
             raise Exception("Unable to fetch CVE inputs due to {0}"
                             .format(url_error))
 
-        fh = open(dest_file, "w")
+        fh = open(dest_file, "wb")
```
jan-cerny: Why is binary mode required here?

ybznek: Main reason is that fix python3 issue. Second reason is that I think we should save .bz2 as binary

jan-cerny:
lgtm

PR #259, opened 2015-12-09
Test behavior of OVAL module with invalid OVAL content

jan-cerny:
This test tests "oscap oval eval" with "--skip-valid" option
on invalid OVAL content. The tested content references a not
existent state. This situation was one of two reasons of
a segmentation fault, described in issue #191.
The issue #191 has been fixed already. This commit wants to
test it to avoid regressions in future.

### DIFF
--- a/tests/API/OVAL/unittests/Makefile.am
+++ b/tests/API/OVAL/unittests/Makefile.am
@@ -58,6 +58,8 @@ EXTRA_DIST += \
 	test_float_comparison.syschar.xml \
 	test_oval_empty_variable_evaluation.sh \
 	test_oval_empty_variable_evaluation.xml \
+	test_skip_valid.sh \
+	test_skip_valid.oval.xml \
 	test_xmlns_missing.oval.xml \
 	test_xmlns_missing.sh \
 	test_xsinil_envv58_pid.oval.xml \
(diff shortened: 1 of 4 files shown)


jan-cerny:
@isimluk Thank you very much for pointing this out.

Firstly, let me explain why we generate results here, We must generate results to hit the segmentation fault described in issue #191. I have tried to run the test on code base without the fix for #191. The segmentation fault did happen only with `--results`. That means if we omit `--results` from this test we would not test whether that particular segmentation fault doesn't happen.

The result is not supposed to be valid OVAL document. The OVAL results documents always include the definition file. In this test, we have invalid definition file, so the result file should be invalid, too. The definition file has one state missing.  Actually, OpenSCAP handles the missing state and the results contain this strange thing:
```
    <states>
      <**INVALID**_state id="oval:x:ste:1" version="0" operator=""/>
    </states>
```
I think the elements of such name are not allowed by OVAL standard. This is confirmed by the fact that validation does not pass.
```
$ oscap oval validate results.xml
OpenSCAP Error: Unrecognized document type for: results.xml [oscap_source.c:258]
```
Does it make sense to you to test that that element mentioned above is present there ?

Regarding the testing for stderr: It is a good idea. It prints on stderr:
```
OpenSCAP Error: Warning: Zero family index [oval_enumerations.c:724]
```
But in my opinion this is quite strange message because "Error" and "Warning" are different things, Should we improve the message?

isimluk:
Ok, then we can assert there is the **INVALID**_state in the results file.

Then, we can assert that `Zero family index` or the improved message happens to appear in the stderr.

jan-cerny:
I agree with you. I have improved the test as you suggested.

isimluk:
Thanks!!

Issue #258, opened 2015-12-09
Sysctl probe is inconsistent with output of sysctl command


jan-cerny:
The sysctl probe does not collect variable `fs.binfmt_misc.status`, while `sysctl -a` does it.

jan-cerny:
Issue appears on both current maint-1.0 and maint-1.2

jan-cerny:
The /proc/sys/binfmt_misc/ directory is used to provide kernel support for miscellaneous binary formats.

jan-cerny:
After fixing this, please write a test for it in tests/probes/sysctl. I think we should test that sysctl probe is fully consistent with sysctl command - it means it collects all the objects that `sysctl -a` does.

ybznek:
On my system, when I've done diff between ```sysctl -a``` and our probe result.

- In our result is missing only "fs.binfmt_misc.status".
- In our results are values like
```
net.ipv6.neigh.virbr1-nic.base_reachable_time
net.ipv6.neigh.virbr1-nic.base_reachable_time_ms
```
but sysctl has only
```
net.ipv6.neigh.virbr1-nic.base_reachable_time_ms
```
and vice versa

ybznek:
Do you think we want to merge such _time values? Merge = More work, maybe worse use case, but output would be same as ```sysctl -a```

Binfmt issue:
```
$ stat ./nfs/ ./binfmt_misc/ | grep Device
Device: 4h/4d	Inode: 240780      Links: 1
Device: 3eh/62d	Inode: 1           Links: 2
```

All files of ```/proc/sys``` share same device, except ```binfmt```

Currently I don't know why it is so. (maybe different module manage this?)

Fix seems to be easy - enable different devices during recursive file-search.

But I am afraid that when I enable all devices, scan can get via symlink somewhere where we don't want.

I will try to get some information about ```/proc/sys``` filesystem. If you know anything, please save my time.

ybznek:
Only for record:

from sysctl source
```
static const char *DEPRECATED[] = {
	"base_reachable_time",
	"retrans_time",
	""
};
```

ybznek:
Fixed.

PR #260, opened 2015-12-10
Add a test for sysctl probe

jan-cerny:
This is a very simple test for sysctl probe.
We haven't had test the probe yet.


jan-cerny:
I tried to write some test for sysctl probe, because some bugs have been reported and fixed in the probe recently. However, this test is very simple. I wanted to have more complex test, but I have found issue https://github.com/OpenSCAP/openscap/issues/258. After fixing that issue, it would be nice to enlarge this test. Now we will at least run the probe , we don't run the probe anywhere in our test suite.

jan-cerny:
I have added check for empty stderr.

Issue #262, opened 2015-12-13
Docs: Missing build dependency in README.md


mzaoui:
When configuring OpenSCAP for a build, the oscar-docker requires bzip2 to be installed.

Build environment:
* CentOS 7.1: contents of /etc/centos_release is "CentOS Linux release 7.1.1503 (Core)"
* OpenSCAP: maint-1.2 branch or new branch from 1.2.7 tag
* Configure command line: ./configure --enable-dependency-tracking --enable-cce --enable-python --enable-regex-posix --enable-ssp --enable-sce

Error messages are the following:
"configure: error: in `/home/mzaoui/Projects/OpenSCAP/openscap':
configure: error: oscap-docker requires bzip2! Either disable oscap-docker or install bzip2."

Problem is solved by installing 'bzip2-devel' package.

Suggestion: add bzip2-devel as a build dependency in the README.md file.

PR #263, opened 2015-12-13
Fix issue #262: Docs: Missing build dependency in README.md

mzaoui:
Add bzip2-devel build dependency to the 'README.md' file.


jan-cerny:
Nice, thank you!

Issue #264, opened 2015-12-13
Path to 'oscap_source.h' header file is missing


mzaoui:
# Description
A fatal error is produced when building OpenSCAP from CentOS 7.1: path to 'oscar_source.h' header file is missing.

# Build environment
* CentOS 7.1: contents of /etc/centos_release is "CentOS Linux release 7.1.1503 (Core)"
* OpenSCAP: maint-1.2 branch or new branch from 1.2.7 tag
* Configure command line: ./configure --enable-dependency-tracking --enable-cce --enable-python --enable-regex-posix --enable-ssp --enable-sce

# Error messages
cce.c:40:26: fatal error: oscap_source.h: No such file or directory
 #include "oscap_source.h"
                          ^
compilation terminated.
make[3]: *** [libcce_la-cce.lo] Error 1
make[3]: Leaving directory `/home/mzaoui/Projects/OpenSCAP/openscap/src/CCE'
make[2]: *** [all-recursive] Error 1
make[2]: Leaving directory `/home/mzaoui/Projects/OpenSCAP/openscap/src'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/home/mzaoui/Projects/OpenSCAP/openscap'

PR #265, opened 2015-12-13
Fix issue #264: Path to 'oscap_source.h' header file is missing

mzaoui:
When configuring OpenSCAP with the --enable-cce flag and building OpenSCAP, the following error message is produced: "cce.c:40:26: fatal error: oscap_source.h: No such file or directory".

Solution to this problem consists in adding the '-I$(top_srcdir)/src/source/public' to the 'libcce_la_CPPFLAGS' directive of the "src/CCE/Makefile.am" file.


ybznek:
ok to test

jan-cerny:
LGTM. thanks!

PR #268, opened 2015-12-14
Plug a memory leak

jan-cerny:
Fixing a resource leak reported by coverity scan.
SEXP_string_cstr allocates memory which was not freed.
Pay attention that SEXP_string_cstr is a confusing name,
because it returns "char *", not "const char *".
Therefore the return value must be freed after use.


isimluk:
dear jenkins, test this please

isimluk:
[test this please]

PR #270, opened 2015-12-14
Remove dead code

jan-cerny:
Coverity scan has reported a dead code here.
Variable "dir_model" is assigned only once, to a constant
NULL value, and doesn't change throughout its scope.
This condition will never be evaluated as true, because
dir_model is always NULL and therefore this code is useless.

### DIFF
--- a/src/OVAL/oval_session.c
+++ b/src/OVAL/oval_session.c
@@ -440,8 +440,6 @@ int oval_session_export(struct oval_session *session)
 cleanup:
 	if (result)
 		oscap_source_free(result);
-	if (dir_model)
-		oval_directives_model_free(dir_model);
 	return ret;
 }
 


jan-cerny:
This is the second time today that Jenkins raises an exception. I have opened an issue https://github.com/OpenSCAP/jenkins/issues/6 against our Jenkins.

isimluk:
This is weird. I would like @msrubar to help here please.

In maint-1.0 we have
```
            dir_model = oval_directives_model_new();
```
at https://github.com/OpenSCAP/openscap/blob/c36da85a16453e2a38377efd08550b7790ea4f62/utils/oscap-oval.c#L550

However, in oval_session this is missing. Michal, do you remember the reason for that?

jan-cerny:
Thank you for analyzing this deeply. I think we forgot  `dir_model = oval_directives_model_new();` while we were re-factoring the code into OVAL session. I wonder whether we have a test for this.

isimluk:
I don't think we have test for this.
```
$ git grep 'oval eval .*--directives' tests/
```

msrubar:
I think @jan-cerny is right. I'll fix the code and create the test case for it.

jan-cerny:
Thank you @msrubar, that would be great!
@isimluk I think we can close this pull request, because the proposed commit is wrong fix.

isimluk:
Closing this in favour of 275.

PR #271, opened 2015-12-15
Fix shellcheck warning

jan-cerny:
SC2155: Declare and assign separately to avoid masking return values


jan-cerny:
This shellcheck warning was reported by coverity scan. Similar situation appears on lines 113 and 114, but I won't fix them, because there is a `TODO` comment. I expect that it would be fixed after the `TODO` will be solved.

mpreisler:
Please fix this even for the TODO variables, the same issue is there, the TODO is related to the contents of the variable, not to how it's assigned.

jan-cerny:
OK then

jan-cerny:
I have fixed also the two others.

isimluk:
Thanks!

PR #274, opened 2015-12-16
Add libcurl leak to valgrind suppression file

jan-cerny:
Valgrind reports memory leaks when using curl_easy_init(),
but we clean the memory correctly.

### DIFF
--- a/tests/suppressions.supp
+++ b/tests/suppressions.supp
@@ -18,3 +18,12 @@
    fun:dlerror
    ...
 }
+
+{
+   curl
+   Memcheck:Leak
+   match-leak-kinds: reachable
+   ...
+   fun:curl_easy_init
+   ...
+}


jan-cerny:
[test this please]

jan-cerny:
[test this please]

jan-cerny:
Anyone for review? :-)

mpreisler:
Sorry for the delays.

I think valgrind is actually right. Check the API docs:
http://curl.haxx.se/libcurl/c/curl_easy_init.html

We are not calling `curl_global_init` and we are not calling `curl_global_cleanup`!

jan-cerny:
@mpreisler I don't think that your idea would solve the problem. Please let me explain why. Before I proposed this proposed this pull request, I spent a lot of time investigating this this, reading documentation of libcurl and playing with valgrind.

Firstly, the documentation for curl_easy_init states 
> If you did not already call curl_global_init, curl_easy_init does it automatically

which I verified that it it true statement. The specifiaction requires to use curl_global_init in multithreaded applications, but that is not our case.

Secondly, the memory leak is actually in `curl_global_init`. I tried to use the `curl_global_init` and `curl_global_cleanup` as you suggested, but it didn't help at all, the leak was still there. 

Finally, to verify that the leak is not our fault, I have created this simple C program:
http://ur1.ca/oehrt
It is `curl_experiment.c`. It definitely should not leak, but it does. See this:
```
[jcerny@thinkpad ~]$ gcc curl_experiment.c -o curl_experiment -lcurl
[jcerny@thinkpad ~]$ ./curl_experiment
[jcerny@thinkpad ~]$ echo $?
0
[jcerny@thinkpad ~]$ valgrind ./curl_experiment
==4368== Memcheck, a memory error detector
==4368== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==4368== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==4368== Command: ./curl_experiment
==4368== 
==4368== 
==4368== HEAP SUMMARY:
==4368==     in use at exit: 22,141 bytes in 621 blocks
==4368==   total heap usage: 646 allocs, 25 frees, 61,369 bytes allocated
==4368== 
==4368== LEAK SUMMARY:
==4368==    definitely lost: 0 bytes in 0 blocks
==4368==    indirectly lost: 0 bytes in 0 blocks
==4368==      possibly lost: 0 bytes in 0 blocks
==4368==    still reachable: 22,141 bytes in 621 blocks
==4368==         suppressed: 0 bytes in 0 blocks
==4368== Rerun with --leak-check=full to see details of leaked memory
==4368== 
==4368== For counts of detected and suppressed errors, rerun with: -v
==4368== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```
The version without the `global*` functions http://ur1.ca/oehsd does the same leak.

So my conclusion from these observations was that there is a leak in libcurl and we in OpenSCAP are using the libcurl API correctly.

mpreisler:
Discussed this with @jan-cerny , even though the curl_global_init and curl_global_cleanup calls don't fix the leaks we want to have them in because documentation says we should call them. Jan will add this to the PR and then I will merge it.

jan-cerny:
Added the curl_global_init and curl_global_cleanup calls in https://github.com/OpenSCAP/openscap/pull/287

mpreisler:
ACK, merged. Thanks!

Issue #277, opened 2015-12-21
Segmentation fault in probe_variable


GaryGapinski:
The attached datastream induces a segmentation fault in OpenSCAP v1.2.8 built and run on Ubuntu 15.10 while evaluating OVAL definition oval:gov.nasa.ascs:def:1 (which uses a http://oval.mitre.org/XMLSchema/oval-definitions-5#independent variable_test).

[example-ds.zip](https://github.com/OpenSCAP/openscap/files/68185/example-ds.zip)

jan-cerny:
Hello @GaryGapinski,
thank you very much for reporting this issue. I confirm that I can reproduce it with your datastream.

I have tried to run OpenSCAP in verbose mode.
```
myoscap xccdf eval --verbose DEVEL --verbose-log-file logfile example-ds.xml
```
In log I have found some warnings:
````
(oscap(25753):oscap(7f58255ec800)) [E:error.c:57:oscap_err_new] (oval_entity.c:329:oval_entity_parse_tag()) The var_ref attribute for the var_ref entity of a variable_object is prohibited since OVAL 5.6. Use plain var_ref instead.
(oscap(25753):oscap(7f58255ec800)) [E:error.c:57:oscap_err_new] (oval_entity.c:334:oval_entity_parse_tag()) Could not found variable 'oval:gov.nasa.ascs:var:1' referenced by var_ref element.
```

It means that syntax that you used is deprecated, According to OVAL specification, you should put variable id as a value of `var_ref` element. like in this example:
```
         <objects>
            <ind:variable_object id="oval:gov.nasa.ascs:obj:1" version="1">
               <ind:var_ref>oval:gov.nasa.ascs:var:1</ind:var_ref>
            </ind:variable_object>
         </objects>
```

This should fix your problem.

However, OpenSCAP should not segfault on any content. The segfault needs to be fixed.

GaryGapinski:
Hello @jan-cerny :

Thanks for pointing out the deprecated syntax. I have corrected it and the segfault no longer occurs.

Regards,

Gary

Corrected data stream: 
[example-ds.zip](https://github.com/OpenSCAP/openscap/files/68441/example-ds.zip)

isimluk:
I had a good time reviewing the datastream. Thank You Gary and Merry Christmas!

PR #278, opened 2015-12-21
Issue #277: Fix a segmentation fault

jan-cerny:
When a variable is not found in OVAL document, then OpenSCAP should
write an error message and finish correctly.


isimluk:
Thanks! 

Next time, please don't be shy to include backtrace in the commit message. It allows our users to found a fix to their problem using the search engines.

PR #280, opened 2015-12-31
Report errors on evaluation of object component of a variable + TEST

jan-cerny:
This pull request wants to introduce two commits to improve experience of indirect referencing of OVAL object within an OVAL variable, which is often considered as difficult part of the OVAL language.

First commit removes TODOs in _oval_component_evaluate_OBJECTREF() function by implementing error reporting as required by OVAL specification:

> If an entity is not found with a name that matches the value of the item_field an error should be reported when determining the value of an variable.

Similar statement applies also for entities of the record data type.

Second commit implements a test for this feature.  It tests whether The ObjectComponentType complex type is processed correctly.  It contains one correct case and two cases which should lead to report an error. For details see comments in supplied OVAL content. However the test does not test fields of the record data type, because the record data type is used only within `ldap_item` and `sql57_item` which are generated by probes which we don't test in upstream.

### DIFF
--- a/src/OVAL/oval_component.c
+++ b/src/OVAL/oval_component.c
@@ -45,6 +45,7 @@
 #include "oval_agent_api_impl.h"
 #include "oval_parser_impl.h"
 #include "oval_probe_session.h"
+#include "oval_probe_impl.h"
 #include "common/util.h"
 #include "common/debug_priv.h"
 #include "common/_error.h"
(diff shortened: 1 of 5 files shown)


jan-cerny:
Anyone for review ? :-)

isimluk:
I like the prudent and helpful error messages.

Just, please make the test a little bit more rigorous too.

jan-cerny:
@isimluk Thank you very much for the review. I have amended the pull request.

PR #282, opened 2016-01-08
Remove EOLs from messages and use macros everywhere

jan-cerny:
To allow to change or redesign the format of verbose log we have to remove
line breaks at the end of some messages. Then it will be possible to move
the message to the begining or middle of a line without any hacks.
The commit also enforces usage of dI, dE, dW macros instead of
ocassionaly appearing oscap_dlprintf macro. This change makes
codebase more consistent regarding the debugging.
To avoid possible wrong usage of debugging macros or wrong formating
by developers in future, first part of this change has to be done
in the maint-1.0 branch.


isimluk:
Merging. However, please review multipline calls of `dX` like

```
dW("bla"
  "..."
  "bla\n".
```

PR #283, opened 2016-01-08
RHBZ#1285757: OpenSCAP pattern match test wrongly results with pass

jan-cerny:
The textfilecontent54 probe currrently works only with UTF8 strings.
It is unable to assess files in different encodings (eg. ISO 8859-1),
because pcre_exec returns an error code. However, the probe reports
that assessed file does not exist. This commit fixes this wrong
behavior in a way that the probe will report an error message
in OVAL results and set the object flag to error when regular
expression matching failed and pcre_exec() returned an error code.


isimluk:
Perfect. Thanks!

PR #284, opened 2016-01-08
Always strip the source file path

jan-cerny:
Do not depend on an environment variable.
Now we will use behavior which used to be default.
Showing the full filepath is not as helpful to slow down the program
by reading an environment variable.


isimluk:
Nice catch. I didn't knew we had this feature. However, I also vote for dropping it. :-)

There is however one subtle example of this being useful. See
```
$ find . -name elements.c 
src/XCCDF/elements.c
src/common/elements.c
```
:-)

PR #286, opened 2016-01-11
Remove linebreaks from debug messages in maint-1.2

jan-cerny:
To allow to change or redesign the format of verbose log we have to remove
    line breaks at the end of some messages. Then it will be possible to move
    the message to the begining or middle of a line without any hacks.
    The commit also enforces usage of dI, dE, dW macros instead of
    ocassionaly appearing oscap_dlprintf macro. This change makes
    codebase more consistent regarding the debugging.
    This is a continuing of work done in PR#282, which has already removed
    the line breaks from maint-1.0 branch.


isimluk:
lgtm. Thanks!

PR #293, opened 2016-01-19
Fix scope of a global variable

jan-cerny:
Static modifier made the pointer to log file unreachable
from other modules. It caused that some debug messages from some
functions (eg. SEXP_list_sort()) could not be printed into log.
This commit fixes the bug by removing the 'static' modifier.


isimluk:
```
+ linux-def_rpmverify_test.xml                                [ FAIL ]
```
@jan-cerny, we haven't seen this test fail in jenkins before. Do you think it is related to the pr?

jan-cerny:
@isimluk I don't think so, I'm trying to investigate, doesn't fail on my machine.

jan-cerny:
@isimluk Run again on same node is successful

isimluk:
Achjo :-)

PR #296, opened 2016-01-20
Use { NULL } instead of NULL as envp when executing remediations

mpreisler:
Using just NULL is discouraged and nonportable, see `man 2 execve`.
This commit doesn't change behavior, just improves portability.


mpreisler:
bad branch

PR #297, opened 2016-01-20
Use { NULL } instead of NULL as envp when executing remediations

mpreisler:
Using just NULL is discouraged and nonportable, see `man 2 execve`.
This commit doesn't change behavior, just improves portability.

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -330,7 +330,11 @@ static inline int _xccdf_fix_execute(struct xccdf_rule_result *rr, struct xccdf_
 				NULL
 			};
 
-			execve(interpret, argvp, NULL);
+			char *const envp[1] = {
+				NULL
+			};
+
+			execve(interpret, argvp, envp);
 			/* Wow, execve returned. In this special case, we failed to execute the fix
 			 * and we return 0 from function. At least the following error message will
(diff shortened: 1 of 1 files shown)


iankko:
@mpreisler 
> Using just NULL is discouraged and nonportable, see man 2 execve.
> This commit doesn't change behavior, just improves portability.

Though based on:
&nbsp; &nbsp; [1] https://bugzilla.kernel.org/show_bug.cgi?id=8408#c2
according to Alan (Cox?) there might be cases when address zero is mapped (and in that case it is valid the ```envp``` argument to contain either environment or argument), the corresponding glibc discussion:
&nbsp; &nbsp; [2] https://sourceware.org/ml/libc-help/2013-06/msg00020.html

concludes:
```
...
> However, we ran into a case where we had some old code that used execve such as:
> 
>   char *cmd[] = { "/path/to/command", NULL };
>   execve(cmd[0], cmd, NULL);
> 
> And it works fine.  However, this obviously violates the documented
> usage. 
...

The above code you quote is not portable since it violates the standard. 

We are not going to change the behaviour of execve on Linux to return
EINVAL if envp is NULL since that might break non-portable programs 
on Linux that currently work. We would break these programs for no
reward.
...
```

So for me (and also after https://sourceware.org/ml/libc-help/2013-06/msg00025.html ) this is a no brainer. ACK for merging this.

jan-cerny:
ack

PR #298, opened 2016-01-21
Use canonical PATH when executing remediations

mpreisler:
This makes content author's life a little bit easier. The change doesn't
break any existing remediations but new remediations relying on these
PATHs won't work with old openscap!


mpreisler:
Do not merge, this is buggy, will fix in a minute :-)

iankko:
E.g. sudo uses the very same 4-tuple as default search path to be set
```
# grep 'secure_path' /etc/sudoers
Defaults    secure_path = /sbin:/bin:/usr/sbin:/usr/bin
```
So IMHO the proposal should be safe enough / reasonable default for all use cases. ACK from me for merging this once Jenkins test finished successfully (but someone else has to approve).

PR #299, opened 2016-01-25
Fix SCE tests

jan-cerny:
This commit adds missing files to the EXTRA_DIST section of makefile.


jan-cerny:
We haven't seen Jenkins failing due to this issue, because Jenkins runs `make distcheck` but `make distcheck` does not configure with SCE support, so SCE tests are not run in  `make distcheck`

isimluk:
I have just checked all our jenkins job and they include `--enable-sce`. What I am missing?

isimluk:
test this please

PR #301, opened 2016-01-26
oscap-vm cosmetic changes

dahaic:
Hello,
these are more of cosmetic changes, self-explanatory I guess, except for the introduction of `fusermount -u`. Originally used `guestunmount` is wrapper of `fusermount -u`, which provides option to wait for unmounting. But usecase in `oscap-vm` does not necessitate this, thus both commands should be interchangeable.

And reasoning is, that some distributions, for example RHEL6, does not have `libguestfs` up-to-date enough to provide `guestunmount` command :)

### DIFF
--- a/utils/oscap-vm
+++ b/utils/oscap-vm
@@ -22,9 +22,6 @@ function die()
     exit 1
 }
 
-which guestmount > /dev/null || die "Cannot find guestmount, please install libguestfs utilities."
-which guestunmount > /dev/null || die "Cannot find guestunmount, please install libguestfs utilities."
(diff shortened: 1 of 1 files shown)


jan-cerny:
[ok to test]

mpreisler:
Hi, it's not clear to me from the docs that `fusermount -u $MOUNTPOINT` is equivalent to `guestunmount $MOUNTPOINT`. It seems to me that guestunmount does something first before running fusermount -u. See https://github.com/libguestfs/libguestfs/blob/master/fuse/guestunmount.c

mpreisler:
ACK, thank you for your contribution!

Issue #304, opened 2016-01-27
oscap-docker gives a weird message on vulenrability scan


jan-cerny:
I have openscap-contaners 1.2.8 on Fedora 23 minimal with Python3.
I have run the vulnerability scan of Fedora container. I am aware of fact it does not make nay sense to run vulnerability scan of Fedora for RHEL 7 vulnerabilitites. However, I am surprised with a weird error message at the end of the output. 

Please see the last line:

```
[root@localhost ~]# docker run -d fedora sleep 3600
c47eb2130910301e577e5c9758c1546eb83c2c4bac0a6b157b4671c58e18846f
Usage of loopback devices is strongly discouraged for production use. Either use `--storage-opt dm.thinpooldev` or use `--storage-opt dm.no_warn_on_loop_devices=true` to suppress this warning.
[root@localhost ~]# oscap-docker container-cve c47eb2130910301e577e5c9758c1546eb83c2c4bac0a6b157b4671c58e18846f
Definition oval:com.redhat.rhsa:def:20160071: false

[ .................................................. ]

Definition oval:com.redhat.rhsa:def:20140675: false
Evaluation done.
Response header of HTTP doesn't contain"last-modified" field. Cannot determine version of remote file "https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2"[root@localhost ~]# 
````
It seems to me an error of communication with rhsa server.

**Comment by jan-cerny** [comment]
Hello! I have pulled the RHEL7 container and I experience the same message also for scanning the image.
Steps to reproduce:
1) docker pull rhel7
2) oscap-docker image-cve rhel7

The vulnerability scan works OK, return code is 0, but the suspicious message is here again: 
```
Response header of HTTP doesn't contain"last-modified" field. Cannot determine version of remote file "https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2"
```

**Comment by isimluk** [comment]
I think @baude and @mpreisler have seen the "last-modified" field missing previously.

Do you now for sure this is not just a warning? Does the scan compelete or not? If this is harmless, does it make sense to include `Warning: ` at the start of the message?

**Comment by mpreisler** [comment]
Last time we've seen this the https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2 regressed by not sending the headers. Could be that the same situation is happening again... :-(

I did some digging to confirm and it seems the header is there. See
```
[...]

PR #308, opened 2016-02-02
oscap docker http header fix

ybznek:
Fix case sensitivity issue while getting HTTP header dict.
+ refactoring
+ add "warning" before message

https://github.com/OpenSCAP/openscap/issues/304

I will test it later.

[SKIP_CI]

### DIFF
--- a/utils/oscap_docker_python/get_cve_input.py
+++ b/utils/oscap_docker_python/get_cve_input.py
@@ -45,7 +45,7 @@ class getInputCVE(object):
     def __init__(self, fs_dest, DEBUG=False):
         ''' Simple init declaration '''
         self.dest = fs_dest
-        self.DEBUG = DEBUG
+        self.DEBUG = True
 
     def _fetch_single(self, dist):
         '''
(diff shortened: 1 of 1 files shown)


ybznek:
Test passed.

utils/oscap_docker_python/get_cve_input.py
```diff
@@ -84,11 +84,23 @@ def _fetch_single(self, dist):
             seconds_epoch = (remote_dt - epoch).total_seconds()
             utime(dest_file, (seconds_epoch, seconds_epoch))
         except KeyError:
-            stderr.write("Response header of HTTP doesn't contain" \
-                  "\"last-modified\" field. Cannot determine version" \
-                  " of remote file \"{0}\"".format(dist_url))
+            self._print_no_last_modified_warning(dist_url)
```
mpreisler: Don't we also want to check for `self.DEBUG` here? Or even check for it inside the method.

ybznek: Yes, you are right. I will check it inside the method.

utils/oscap_docker_python/get_cve_input.py
```diff
@@ -45,7 +45,7 @@ class getInputCVE(object):
     def __init__(self, fs_dest, DEBUG=False):
         ''' Simple init declaration '''
         self.dest = fs_dest
-        self.DEBUG = DEBUG
+        self.DEBUG = True
```
jan-cerny: Is this intentional or just a oversight? I think if we set this to true, we don't have to check for it (`if self.DEBUG:`) further in the code. I can't find any other place in the code where this variable is set.

mpreisler: Definitely an oversight. Good catch.

ybznek: Eh, sorry for that. I have to more check more my pull requests.

PR #311, opened 2016-02-07
Adjust the valgrind suppression file

jan-cerny:
After merging c97417a328687ebf49f4da0ac54d6104d15297c7, we must suppress
the memory leak in curl_global_init directly, not via curl_easy_init.


jan-cerny:
This should fix the failing Jenkins openscap-valgrind job.

PR #314, opened 2016-02-09
Provide basic information about evaluating test results in verbose log

jan-cerny:
In this pull request I would like to introduce some very basic informational messages that can help user of the verbose mode to figure out the reasons of OVAL test results.

(More will follow, but I think it will be in another pull request.)

### DIFF
--- a/src/OVAL/public/oval_definitions.h
+++ b/src/OVAL/public/oval_definitions.h
@@ -253,6 +253,7 @@ const char *oval_operator_get_text(oval_operator_t);
 const char *oval_subtype_get_text(oval_subtype_t);
 const char *oval_family_get_text(oval_family_t);
 const char *oval_check_get_text(oval_check_t);
+const char *oval_check_get_description(oval_check_t);
 const char *oval_existence_get_text(oval_existence_t);
 const char *oval_affected_family_get_text(oval_affected_family_t);
 const char *oval_datatype_get_text(oval_datatype_t);
--- a/src/OVAL/public/oval_definitions.h
+++ b/src/OVAL/public/oval_definitions.h
@@ -1184,12 +1184,6 @@ struct oval_object *oval_test_get_object(struct oval_test *);
  * @memberof oval_test
  */
 struct oval_state_iterator *oval_test_get_states(struct oval_test *);
-/**
- * Get a nicely formated list of states referenced by test.
- * @return String containing state names delimited by a comma.
- * @memberof oval_test
- */
-char *oval_test_get_state_names(struct oval_test *test);
 
 /** @} */
 
--- a/src/OVAL/public/oval_definitions.h
+++ b/src/OVAL/public/oval_definitions.h
@@ -253,7 +253,6 @@ const char *oval_operator_get_text(oval_operator_t);
 const char *oval_subtype_get_text(oval_subtype_t);
 const char *oval_family_get_text(oval_family_t);
 const char *oval_check_get_text(oval_check_t);
-const char *oval_check_get_description(oval_check_t);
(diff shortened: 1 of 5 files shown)


jan-cerny:
Test failed in process58 probe test, but it should be unrelated to this PR. I have failed a new issue https://github.com/OpenSCAP/openscap/issues/315

isimluk:
test this please

src/OVAL/public/oval_definitions.h
```diff
@@ -1183,6 +1184,12 @@ struct oval_object *oval_test_get_object(struct oval_test *);
  * @memberof oval_test
  */
 struct oval_state_iterator *oval_test_get_states(struct oval_test *);
+/**
+ * Get a nicely formated list of states referenced by test.
+ * @return String containing state names delimited by a comma.
+ * @memberof oval_test
+ */
+char *oval_test_get_state_names(struct oval_test *test);
```
isimluk: It seems this needs not to be public.

Can we move this into `oval_definitions_impl.h`?

src/OVAL/results/oval_resultTest.c
```diff
@@ -618,6 +621,13 @@ static oval_result_t eval_check_state(struct oval_test *test, void **args)
 	syschar_model = oval_result_system_get_syschar_model(SYSTEM);
 	ores_clear(&item_ores);
 
+	char *state_names = oval_test_get_state_names(test);
+	dI("To get result of '%s', collected items will be compared with these oval states: %s.",
+		oval_test_get_id(test), state_names);
+	dI("In test '%s' %s of the collected items must satisfy these states: %s.",
+		oval_test_get_id(test), oval_check_get_description(ste_check), state_names);
```
isimluk: I am confused here and I need a little of persuasion here. :-)

Imagine you have a `test` without any `state`s and `OVAL_CHECK_AT_LEAST_ONE`. The resulting sentence will be like `In test 'bla' at least one of the collected items must satisfy these states: '',"` While majority of the people with good engineering training may understand, it will not be 100% clear message. You can get wild philosophical discussion in kitchen about the meaning of `must satisfy none`.

Then imagine next day, you have a `test` without any `state`s and `OVAL_CHECK_NONE_EXIST`. The resulting sentence will be like `In test 'bla' none of the collected items must satisfy these states: '',"` If I apply the same philosophical reasoning as the previous day in kitchen, it works, but I am still confused. Does none satisfy none?

There has to be more convoluted sentences, if you think of the possible combinations. And then, at least one of the convoluted sentences satisfies none. :))

What I want to say is that `satisfy` and `exists` are two different verbs and it would be better for us to have the correct verb in the sense.

Does it make sense to amend the sentences based on the existence of state_names?

src/OVAL/public/oval_definitions.h
jan-cerny: Yes, you are true, I will move it to `oval_definitions_impl.h`.

src/OVAL/results/oval_resultTest.c
jan-cerny: I understand you. The fact is that in current code the situation that there will be empty string (`satisfy states: ''`) will never happen, because this function `eval_check_state` is called only when the test contains at least one state, so this message will be displayed only when there is some state defined.

However this is not obvious from the code at all. I will improve it and also more ensure that this situation will not happen.

jan-cerny:
Thanks for the review. I have improved the PR and pushed new commits here.

isimluk:
Well, my comments are now gone. However, it seems they have not been addressed. :-(

Issue #303, opened 2016-01-27
oscap-docker tracebacks with no argument


jan-cerny:
Hello!
I am have installed openscap-containers-1.2.8-1.fc23.noarch on Fedora 23 minimal with Python 3.
I run oscap-docker command with NO argument. I got a traceback. Please see the command that I am running and its result:
```
[root@localhost ~]# oscap-docker
Traceback (most recent call last):
  File "/usr/bin/oscap-docker", line 108, in <module>
    args.func()
AttributeError: 'Namespace' object has no attribute 'func'
```

isimluk:
Oh. That's a regression, previously we have shown the help: https://github.com/OpenSCAP/openscap/blob/ad719afdadfa4175c00e012e38836cb2db4fcf82/utils/oscap-docker#L144

ybznek:
I have seen the issue. Same issue had Atomic, but
https://github.com/projectatomic/atomic/issues/242#issuecomment-162646407
```This was a bug in f23 python, but newer atomic clients should fix it.```

I hoped that it will be fixed early without any our action.

jan-cerny:
@ybznek I don't think so. I think we must fix it in oscap-docker. I think it is badly implemented  arguments parsing.

PR #316, opened 2016-02-11
Issue #303: Fix oscap-docker traceback

jan-cerny:
The oscap-docker utility failed with a traceback with Python 3
whe invoked with no argument.
It will print the help instead.


jan-cerny:
test this please

Issue #315, opened 2016-02-10
Process58_probe test failed


jan-cerny:
Process58_probe test failed in Jenkins build https://jenkins.open-scap.org/job/openscap-pull-requests/333/ on RHEL6.

Secifically, the test `tests/probes/process58/command_line.sh` failed in test case where it is testing a zombie process. It was definitiely a good idea that we previously added `set -x` in this script, so now we can get more information.

In the script on line 89 (https://github.com/OpenSCAP/openscap/blob/maint-1.2/tests/probes/process58/command_line.sh#L89) a function `get_zombie_pid_from_ppid` is called.
However, in this case, it did not returned any value.
That function actively waits for the process for 100*0.1 = 10 seconds to get its PID by grepping the output  of `ps` command. In this case it was unable to get the PID for 10 seconds.

The empty value of variable `$ZOMBIE_PID` then broke following assertions => The test failed on this comand:
`assert_match_command_line "${ZOMBIE_PID}" "${ZOMBIE_CMDLINE_REGEX}"`,  because `$ZOMBIE_PID` was an empty string.

jan-cerny:
@isimluk amended the test to get more information when it happens in future. See https://github.com/OpenSCAP/openscap/issues/319 Closing for now.

PR #319, opened 2016-02-11
Desperate grave digger

isimluk:
Voodoo: Tell us where is zombie

We think she left dead, but the priest (ps) conceals where it is.

This is not a black magic really. We think the zombie process should be
listed in cemetery, but sometimes it is missing. Good thing is to put
more details into the log. Next time we need to record the moon phase.

See github issue #315 for more details.


jan-cerny:
Thank you. I hope it will be helpful. ACK.

PR #325, opened 2016-02-14
Improve a message about opening file

jan-cerny:
We can show a more understandable message
when a probe opens some file (eg. configuration file)
to collect data.

### DIFF
--- a/src/OVAL/probes/oval_fts.c
+++ b/src/OVAL/probes/oval_fts.c
@@ -817,7 +817,7 @@ OVAL_FTS *oval_fts_open(SEXP_t *path, SEXP_t *filename, SEXP_t *filepath, SEXP_t
 		return NULL;
 	}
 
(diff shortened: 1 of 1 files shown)


jan-cerny:
I am not sure, but it look quite nice in log when you see that the probe is doing something related to the test that is evaluated :-)
```
...................
I: oscap: Evaluating textfilecontent54 test 'oval:ssg-test_accounts_no_uid_except_root:tst:1': test that there are no accounts with UID 0 except root in the /etc/passwd file.
I: oscap: Querying textfilecontent54 object 'oval:ssg-object_accounts_no_uid_except_root:obj:1', flags: 0.
I: oscap: Creating new syschar for textfilecontent54_object 'oval:ssg-object_accounts_no_uid_except_root:obj:1'.
I: lt-probe_textfilecontent54: Opening file '/etc/passwd'.
...................
```

Issue #292, opened 2016-01-18
Support offline mode in rpmverifyfile


mpreisler:
```
(probe_rpmverifyfile(11249):input_handler(7f722da45700)) [D:input_handler.c:114:probe_input_handler] offline_mode=00000001
(probe_rpmverifyfile(11249):input_handler(7f722da45700)) [D:input_handler.c:115:probe_input_handler] offline_mode_supported=00000000
(probe_rpmverifyfile(11249):input_handler(7f722da45700)) [W:input_handler.c:119:probe_input_handler] Requested offline mode is not supported by probe_rpmverifyfile.
```

Offline mode is important for container and virtual machine scanning. Lack of support in rpmverifyfile is *most likely* an oversight since offline mode is supported in other RPM probes.

ybznek:
Fixed.

PR #327, opened 2016-02-16
Enable offline scan of some probes 

ybznek:
Should fix also
https://github.com/OpenSCAP/openscap/issues/292

### DIFF
--- a/src/OVAL/probes/unix/linux/partition.c
+++ b/src/OVAL/probes/unix/linux/partition.c
@@ -59,6 +59,8 @@
 #include <limits.h>
 #include <sys/statvfs.h>
 #include <probe-api.h>
+#include <probe/probe.h>
+#include <probe/option.h>
 #include <mntent.h>
 #include <pcre.h>
 
@@ -235,6 +237,13 @@ static int collect_item(probe_ctx *ctx, oval_schema_version_t over, struct mnten
         oscap_free(mnt_opts);
 
         return (0);
+}
+
+void *probe_init(void)
+{
+	// Intentionally commented-out, see commit message
+	// probe_setoption(PROBEOPT_OFFLINE_MODE_SUPPORTED, PROBE_OFFLINE_CHROOT);
+	return (NULL);
 }
 
 int probe_main(probe_ctx *ctx, void *probe_arg)
--- a/src/OVAL/probes/unix/password.c
+++ b/src/OVAL/probes/unix/password.c
@@ -58,6 +58,8 @@
 #include "probe/entcmp.h"
 #include "alloc.h"
 #include "common/debug_priv.h"
+#include <probe/probe.h>
+#include <probe/option.h>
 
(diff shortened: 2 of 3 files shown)


isimluk:
I don't know. This is hardish to review. It would be great to split this into many pull-requests and review one offline probe per pr.

It would help to have a tests written where applicable or perhaps have the test report for others.

For example probe partition uses `statvfs` and depends on `libblkid`. The `statfs` will imho need proper /proc bindmounted to get the information about current mount points. I don't know what happens when we run this on some chroot directory without proper bindmounts. Actually partition/main depends on correct `/proc`. Perhaps we should include checks like `if offline_mode_on; then assert for bindmounted /proc`

The shadow and password probes should perhaps work. And it should be fairly easy to create a test for them in our upstream test suite. (1) Create /mnt/shadow_offline.XXXXXX/etc/passwd (2) PROBE_CHROOT=/mnt/shadow_offline.XXXXXX. (3) collect and assert for something in that dummy /etc/passwd.

The rpmXXX probes might well work like they are. However, testing (or at least report) would he highly helpful.

I know @dkopecek have done the initial porting of offline probes and he might be good point of contact to discuss this further.

mpreisler:
Have you tested the probes in offline mode?

isimluk:
For the record: passwd & shadow probes are not that easy. They depend on what nss switch says. The `getpwent` queries /etc/passwd, NIS, or LDAP.

Perhaps we should get an ldap into an image set-up the nss switch and then run the scan.

ybznek:
I am investigating partition probe

1) **when I use atomic mount**
in mount dir isn't mounted ```/proc``` - we cannot read mounts
2) **when find PID of container process**
`/proc/$PID/mounts` has same content as `docker exec -t -i a3d74642cf93  cat /proc/mounts ` - it is what we want to read
3) **after 'chroot'** to `/proc/$PID/root` I cannot `cat /proc/mounts `, I can do `cat /proc/1/mounts` - but it will work only if docker container don't use *--pid host*

So for running docker container we can do some work-around to get container process PID and read `/proc/$PID/mounts`. But it requires some new code and provide partition offline scan only for one of many use cases.


Any tips how to deal with it?

mpreisler:
@ybznek, perhaps you could create a few smaller PRs that are easier to review? Let's start with the easiest probes, make sure they work and get them in first. The RPM probes look like they are both very useful and should be easy to support in offline mode.

mpreisler:
Just a note for a future review: @ybznek has split this PR off and this PR now only has the problematic probes. The probes that were tested and work were merged in #331

ybznek:
Unfortunately, it takes quite lot of time without any result.

Partition probe can be possibly available if we/atomic change behavior of offline scanning.

Link to description if in commit messages.

mpreisler:
I am merging this so that the comments are in the code-base. We can later work on each probe separately.

PR #331, opened 2016-02-24
Maint 1.2 offline scan tested probes

ybznek:
I have done simple beaker test using oscap-docker for these probes, so these probes should really work.


mpreisler:
ACK, thx!

PR #332, opened 2016-02-25
Added the oscap-chroot tool - scan filesystems on arbitrary paths

mpreisler:
A very very small command-line tool with no dependencies outside `oscap`. The only thing it does is setup environment variables correctly for `oscap` for running it for a chroot.

I could do all of this inside the daemon but I'd prefer to have the logic neat and simple so I created it as a separate script. We could also decide we don't want this in OpenSCAP and just ship it with the daemon but I think that would be a shame. This makes a few use-cases simpler.

Let me know what you think!

### DIFF
--- a/utils/oscap-chroot
+++ b/utils/oscap-chroot
@@ -1,0 +1,106 @@
+#!/bin/bash
+
+# Copyright 2016 Red Hat Inc., Durham, North Carolina.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


isimluk:
Does it make sense to call oscap-chroot from oscap-vm?

mpreisler:
> Does it make sense to call oscap-chroot from oscap-vm?

I did think about that but then I would have to do a lot more magic around the configuration switches so I decided against it. I want it to be cleanly possible to enable `oscap-chroot` and disable `oscap-vm`. Without any surprises at runtime.

utils/oscap-chroot
```diff
@@ -0,0 +1,106 @@
...
+
+function die()
+{
+    echo "$*" >&2
+    exit 1
+}
+
+function usage()
+{
+    echo "oscap-chroot -- Tool for offline SCAP evaluation of filesystems mounted in arbitrary paths."
```
ybznek: Personally, I would use "here document" than multiple echos, but doesn't matter.

isimluk:
lgtm & thanks!

PR #333, opened 2016-02-28
Evaluate results while collecting objects

jan-cerny:
This pull request proposes a small change to process of evaluation of OVAL definition.
Currently, we do two steps to get the results of a definition:
1) Querying probes for objects for all the definitions
2) After all items are collected, we compute results separately.
It causes some confusion in the verbose log in cases one definition contains multiple criteria,
the messages are not in logical order here.
I think that messages describing computing results of a test should be displayed just after messages  describing evaluation of the same test, it should not be mixed as is now.
We can fix this by merging aforementioned steps into single step, which is done in this PR by some refactoring.
It would basically mean that each test is evaluated separately, no other querying starts before a result of single test is done.
Therefore we would have messages in verbose mode in expected order.

Since this is not trivial change, I have done some testing. I have compared OVAL results produced by evaluating RHEL6, RHEL7, and Fedora OVAL from SSG with and without these patches and reviewed the result files using meld. Results from both were same (except item IDs and time stamps, of course).

### DIFF
--- a/src/OVAL/oval_agent.c
+++ b/src/OVAL/oval_agent.c
@@ -86,10 +86,13 @@ static const struct oval_result_to_xccdf_spec XCCDF_OVAL_RESULTS_MAP[] = {
 	{0, 0, 0}
 };
 
+static struct oval_result_system *_oval_agent_get_first_result_system(oval_agent_session_t *ag_sess);
+
 oval_agent_session_t * oval_agent_new_session(struct oval_definition_model *model, const char * name) {
 	oval_agent_session_t *ag_sess;
 	struct oval_sysinfo *sysinfo;
 	struct oval_generator *generator;
+	struct oval_result_system *result_system;
 	int ret;
 
 	dI("Started new OVAL agent.", name);
@@ -102,7 +105,16 @@ oval_agent_session_t * oval_agent_new_session(struct oval_definition_model *mode
 	ag_sess->def_model = model;
 	ag_sess->cur_var_model = NULL;
 	ag_sess->sys_model = oval_syschar_model_new(model);
-	ag_sess->psess     = oval_probe_session_new(ag_sess->sys_model);
+
+	/* one system only */
+	ag_sess->sys_models[0] = ag_sess->sys_model;
+	ag_sess->sys_models[1] = NULL;
+	ag_sess->res_model = oval_results_model_new(model, ag_sess->sys_models);
+	generator = oval_results_model_get_generator(ag_sess->res_model);
+	oval_generator_set_product_version(generator, oscap_get_version());
+
+	result_system = _oval_agent_get_first_result_system(ag_sess);
+	ag_sess->psess = oval_probe_session_with_result_system_new(ag_sess->sys_model, result_system);
 
 	/* probe sysinfo */
 	ret = oval_probe_query_sysinfo(ag_sess->psess, &sysinfo);
@@ -114,14 +126,6 @@ oval_agent_session_t * oval_agent_new_session(struct oval_definition_model *mode
 	}
 	oval_syschar_model_set_sysinfo(ag_sess->sys_model, sysinfo);
 	oval_sysinfo_free(sysinfo);
-
-	/* one system only */
-	ag_sess->sys_models[0] = ag_sess->sys_model;
-	ag_sess->sys_models[1] = NULL;
-	ag_sess->res_model = oval_results_model_new(model, ag_sess->sys_models);
-	generator = oval_results_model_get_generator(ag_sess->res_model);
-	oval_generator_set_product_version(generator, oscap_get_version());
-
 
 	ag_sess->product_name = NULL;
 
--- a/src/OVAL/oval_agent.c
+++ b/src/OVAL/oval_agent.c
@@ -245,7 +245,7 @@ int oval_agent_reset_session(oval_agent_session_t * ag_sess) {
 	}
 
 	oval_probe_session_destroy(ag_sess->psess);
-	ag_sess->psess = oval_probe_session_new(ag_sess->sys_model);
+	ag_sess->psess = oval_probe_session_with_result_system_new(ag_sess->sys_model, _oval_agent_get_first_result_system(ag_sess));
 
 	return 0;
 }
--- a/src/OVAL/oval_agent.c
+++ b/src/OVAL/oval_agent.c
@@ -165,7 +165,6 @@ int oval_agent_eval_definition(oval_agent_session_t *ag_sess, const char *id)
 {
 	int ret;
 	const char *title = NULL;
-	struct oval_result_system *rsystem;
 	struct oval_definition *oval_def;
 
 	oval_def = oval_definition_model_get_definition(ag_sess->def_model, id);
@@ -176,13 +175,6 @@ int oval_agent_eval_definition(oval_agent_session_t *ag_sess, const char *id)
 
 	/* probe */
 	ret = oval_probe_query_definition(ag_sess->psess, id);
-	if (ret == -1)
-		return ret;
-
-	rsystem = _oval_agent_get_first_result_system(ag_sess);
-	/* eval */
-	ret = oval_result_system_eval_definition(rsystem, id);
-
 	return ret;
 }
 
--- a/src/OVAL/oval_probe.c
+++ b/src/OVAL/oval_probe.c
@@ -433,12 +433,17 @@ static oval_result_t oval_probe_query_criterion(oval_probe_session_t *sess, stru
 	return oval_result_test_eval(result_test);
 }
 
-static int oval_probe_query_extend_definition(oval_probe_session_t *sess, struct oval_criteria_node *cnode)
+static oval_result_t oval_probe_query_extend_definition(oval_probe_session_t *sess, struct oval_criteria_node *cnode, struct oval_result_criteria_node *result_cnode)
 {
 	struct oval_definition *oval_def = oval_criteria_node_get_definition(cnode);
 	const char *def_id = oval_definition_get_id(oval_def);
 	dI("Criteria are extended by definition '%s'.", def_id);
-	return oval_probe_query_definition(sess, def_id);
+	int ret = oval_probe_query_definition(sess, def_id);
+	if (ret != 0) {
+		return OVAL_RESULT_ERROR;
+	}
+	struct oval_result_definition *extends = oval_result_criteria_node_get_extends(result_cnode);
+	return oval_result_definition_get_result(extends);
 }
 
 static oval_result_t oval_probe_query_criteria(oval_probe_session_t *sess, struct oval_criteria_node *cnode, struct oval_result_criteria_node *result_cnode)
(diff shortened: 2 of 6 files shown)


src/OVAL/oval_probe.c
```diff
@@ -329,14 +331,13 @@ int oval_probe_query_sysinfo(oval_probe_session_t *sess, struct oval_sysinfo **o
 	return(0);
 }
 
-static int oval_probe_query_criteria(oval_probe_session_t *sess, struct oval_criteria_node *cnode);
+static oval_result_t oval_probe_query_criteria(oval_probe_session_t *sess, struct oval_criteria_node *cnode, struct oval_result_criteria_node *rslt_cnode);
```
ybznek: (Personal opinion.)

In generally, I don't like "acronyms" in variable names rather than "rslt_cnode" I prefer something like "result_cnode", especially when it is not much longer. When I see "rslt" - I need to think what is it.
There are other similar variable names...

ybznek: Maybe I have oversight something. But e.g. cnode can be CONST.

src/OVAL/oval_probe.c
```diff
@@ -350,84 +351,118 @@ int oval_probe_query_definition(oval_probe_session_t *sess, const char *id) {
...
+
+	/* probe objects referenced like this: test->state->variable->object */
+	ste_itr = oval_test_get_states(test);
+	while (oval_state_iterator_has_more(ste_itr)) {
+		struct oval_state *state = oval_state_iterator_next(ste_itr);
+		struct oval_state_content_iterator *contents = oval_state_get_contents(state);
+		while (oval_state_content_iterator_has_more(contents)) {
+			struct oval_state_content *content = oval_state_content_iterator_next(contents);
+			struct oval_entity * entity = oval_state_content_get_entity(content);
+			if (oval_entity_get_varref_type(entity) == OVAL_ENTITY_VARREF_ATTRIBUTE) {
```
ybznek: I would consider to move IF block to extra function. Very hard to read, many variables, long function.

jan-cerny: Although this is not my code (I only decreased indent during refactoring here), I agree with you that this very long and should be refactored out to a separate function. But it is still better than it used to be before :-)

jan-cerny:
@ybznek @isimluk Could you review, please?

isimluk:
I review this every other day. I felt, there is unanswered note by @ybznek.

> Maybe I have oversight something. But e.g. cnode can be CONST.

I might be missing something though.

jan-cerny:
Thanks, I have renamed variables. Now I am looking into possibility of adding `const` modifier.

src/OVAL/oval_probe.c
jan-cerny: @ybznek I think it can't be `const`, because subesequently we call functions `oval_criteria_node_get_test` and `oval_criteria_node_get_definition`. These functions are defined in public API but they don't have `const` in their prototypes. I have tried to add const modifier as you suggested, but it introduced some new compiler warnings. For example:

```
oval_probe.c:454:71: warning: passing argument 1 of ‘oval_criteria_node_get_definition’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  struct oval_definition *oval_def = oval_criteria_node_get_definition(cnode);
                                                                       ^
In file included from ./public/oval_probe.h:37:0,
                 from oval_probe.c:36:
./public/oval_definitions.h:2220:25: note: expected ‘struct oval_criteria_node *’ but argument is of type ‘const struct oval_criteria_node *’
 struct oval_definition *oval_criteria_node_get_definition(struct oval_criteria_node *);
```

jan-cerny:
This comment should be copied here, because it has been hidden when GitHub outdated the diff.
> @ybznek I think it can't be const, because subesequently we call functions oval_criteria_node_get_test and oval_criteria_node_get_definition. These functions are defined in public API but they don't have const in their prototypes. I have tried to add const modifier as you suggested, but it introduced some new compiler warnings. For example:

> ```
oval_probe.c:454:71: warning: passing argument 1 of ‘oval_criteria_node_get_definition’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  struct oval_definition *oval_def = oval_criteria_node_get_definition(cnode);
                                                                       ^
In file included from ./public/oval_probe.h:37:0,
                 from oval_probe.c:36:
./public/oval_definitions.h:2220:25: note: expected ‘struct oval_criteria_node *’ but argument is of type ‘const struct oval_criteria_node *’
 struct oval_definition *oval_criteria_node_get_definition(struct oval_criteria_node *);
```

File: src/OVAL/public/oval_results.h [hunk 3]
```diff
@@ -724,6 +724,10 @@ void oval_result_criteria_node_iterator_free(struct oval_result_criteria_node_it
  * @name Evaluators
  * @{
  */
+/**
+ * @memberof oval_result_criteria_node
+ */
+oval_result_t oval_result_criteria_node_negate(struct oval_result_criteria_node *node, oval_result_t result);
```
isimluk: Please do not expose this to public. You only need this in `src/OVAL/oval_probe.c`. However, in another commit, you are making the `oval_probe.c` to depend on `results/oval_results_impl.h`. So, I think `oval_result_criteria_node_negate` should go to `oval_results_impl.h` as well.

src/OVAL/oval_agent.c
```diff
@@ -169,13 +170,14 @@ int oval_agent_eval_definition(oval_agent_session_t *ag_sess, const char *id)
 		title = oval_definition_get_title(oval_def);
 	}
 	dI("Evaluating definition '%s': %s.", id, title);
+	rsystem = _oval_agent_get_first_result_system(ag_sess);
+	oval_probe_session_set_result_system(ag_sess->psess, rsystem);
```
isimluk: We really should think about handling this differently.

Please draw a diagram of relations between `agent_session`, `results_model`, `result_system`, `probe_session`, and `syschar_model`. You will see the madness. :-)

From the diagram. I think we should never allow to `probe_session_set_result_system`. It make sense only for `result_system` assigned with the same `syschar_model`.

What about creating new oval_probe_session_new which would take `result_system` and deprecating the other one?

isimluk:
I wish the `refactor: extract function` commits were done before changing the functionality. It would allow us to have smaller commits that are easier to read. :-)

isimluk:
Look @jan-cerny, this PR is extremely hard to review.

The existing code is convoluted and there is a lot of assumptions build into it. It is very hard to understand broader implications of the changes.

I would love to see smaller commits that would help me to understand what is the change about. I am reading https://github.com/OpenSCAP/openscap/pull/333/commits/93e46fb6fda0d0f47a5456550de802dc49bed173 again and again and my mind is about to explode. 

Can you please make the refactoring commits first and then the actual changes?

I know this will be extra work. However, I am not yet convinced what are the side effects of this change. And I am trying to visualize what exactly are the changes. And what other parts of the code we can drop because of this merge.

(I have manually merged commit 1, commit 2, and commit 13 into maint-1.2).

jan-cerny:
@isimluk I understand. Thanks. I will surely do it.

src/OVAL/oval_probe.c
jan-cerny: Addressed in 5f0205abc938624d0391821ca45239c0bd285f95

jan-cerny:
@isimluk I have reworked this pull request. I have made smaller commits. I hope it is easier to review.

jan-cerny:
I have repeated the testing again with RHEL7 and Fedora SSG OVAL. I haven't found any issue. It gives same results as OpenSCAP 1.2.8.

src/OVAL/oval_probe.c
```diff
@@ -355,6 +355,43 @@ int oval_probe_query_definition(oval_probe_session_t *sess, const char *id) {
...
+			}
+
+			flag = oval_variable_get_collection_flag(var);
+			switch (flag) {
+			case SYSCHAR_FLAG_COMPLETE:
+			case SYSCHAR_FLAG_INCOMPLETE:
+				break;
+			default:
+				oval_state_content_iterator_free(contents);
+				return -1;
```
isimluk: This seems to be a change of behavior in refactoring commit. I think Martin Fowler, Kent Beck, and Robert C. Martin wouldn't be happy here. :-)

Please keep return value 0 or -1 the same. And forward it to the caller.

The exact change you introduced is that: error (`-1`) from `oval_probe_query_variable` is no longer propagated.

isimluk:
I have again merged first two commits in this PR. And commented on third. I am continuing with the review.

Feel free to rebase on current maint-1.2 when you see a fit.

src/OVAL/oval_probe.c
jan-cerny: OK, I see, I will fix it.

src/OVAL/oval_probe.c
```diff
@@ -433,13 +433,7 @@ static int oval_probe_query_extend_definition(oval_probe_session_t *sess, struct
...
 	const char *def_id = oval_definition_get_id(oval_def);
 	dI("Criteria are extended by definition '%s'.", def_id);
-	struct oval_criteria_node *node =  oval_definition_get_criteria(oval_def);
-	if (node == NULL) {
-		oscap_seterr(OSCAP_EFAMILY_OSCAP, "Could not find extended definition: %s.",
-			def_id);
-		return -1;
-	}
-	return oval_probe_query_criteria(sess, node);
+	return oval_probe_query_definition(sess, def_id);
```
isimluk: Lovely! Thanks!

isimluk:
Merged commit 11 from this PR.

jan-cerny:
Rebased on current maint-1.2.

jan-cerny:
Test for selinuxboolean probe again failed. It should be unrelated to this PR.
[test this please]

isimluk:
Merged commit 5.

isimluk:
You have merged the more high-level code (result calculation) into the more low-level code (probe of the system). Why did you choose this way and not the other way?

Look, I am not arguing for this or against it, however I am more concerned whether that option was considered.

As I dig deeper and deeper, it seems that having the probe session in results model, would do the click. If there is a session, do probe, if there is no session, just eval the syschar. No big extra logic, no breakage of API.

Btw, the current PR will break API for anyone using `oval_probe_query_definition` which will now return (-1). I am afraid we will end-up with more convoluted code, if we want to keep API.

jan-cerny:
Thanks. I confess that I didn't consider it. I hadn't such idea. 

I had a look at it yesterday. Firstly I thought that your proposal would be more difficult, but fortunately I realized that the structures in results model contain pointers to their siblings from definition model (oval_test, oval_object, ...) It means that we are capable to call probing form oval_result_test_*. Yesterday I tried to code that and it seems working. Moreover it looks much easier. After I will review and test it, I will push the code here.

isimluk:
Feel free to open new PR if you see fit.

jan-cerny:
Replaced by https://github.com/OpenSCAP/openscap/pull/360. Closing.

PR #334, opened 2016-02-28
Indent messages in verbose log

jan-cerny:
This commit adds two functions that either increase or
decrease indent of messages in verbose log.

### DIFF
--- a/src/OVAL/results/oval_resultDefinition.c
+++ b/src/OVAL/results/oval_resultDefinition.c
@@ -151,7 +151,9 @@ oval_result_t oval_result_definition_eval(struct oval_result_definition * defini
 	if (definition->result == OVAL_RESULT_NOT_EVALUATED) {
 		struct oval_result_criteria_node *criteria = oval_result_definition_get_criteria(definition);
 		if (criteria != NULL) {
+			dIndent(1);
 			definition->result = oval_result_criteria_node_eval(criteria);
+			dIndent(-1);
 		}
 	}
 	dI("Definition '%s' evaluated as %s.", oval_result_definition_get_id(definition), oval_result_get_text(definition->result));
--- a/src/OVAL/results/oval_resultTest.c
+++ b/src/OVAL/results/oval_resultTest.c
@@ -1045,7 +1045,9 @@ oval_result_t oval_result_test_eval(struct oval_result_test *rtest)
 		if ((oval_independent_subtype_t)oval_test_get_subtype(oval_result_test_get_test(rtest)) != OVAL_INDEPENDENT_UNKNOWN ) {
 			struct oval_string_map *tmp_map = oval_string_map_new();
 			void *args[] = { rtest->system, rtest, tmp_map };
+			dIndent(1);
 			rtest->result = _oval_result_test_result(rtest, args);
+			dIndent(-1);
 			oval_string_map_free(tmp_map, NULL);
 
 			if (!rtest->bindings_initialized) {
--- a/src/common/debug.c
+++ b/src/common/debug.c
@@ -139,7 +139,7 @@ static const char *__oscap_path_rstrip(const char *path)
 }
 
 
-static void debug_message_start(int level)
+static void debug_message_start(int level, int indent)
 {
 	char  l;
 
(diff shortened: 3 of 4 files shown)


jan-cerny:
Process58_probe test failed on Jenkins. Should be unrelated to this PR.

```
make[5]: Entering directory `/home/jenkins/workspace/openscap-pull-requests/openscap-1.2.9/_build/tests/probes/process58'

----------------------------------------------------------------------
+ Ensure that selinux_domain_label is collected               [ PASS ]
+ Ensure that tty number is translated into name              [ PASS ]
+ Ensure loguid return unsigned int (cat /proc/ID/loginuid)   [ PASS ]
+ Ensure sessionid is correct                                 [ PASS ]
+ Ensure capabilities with OVAL 5.11                          [ PASS ]
+ Ensure that command_line is collected                       [ FAIL ]
--------------------------------------------------
See tests/probes/process58/tests_probes_process88.log.
```

jan-cerny:
[test this please]

mpreisler:
I don't like the API at all but to be honest I can't think of a way to improve it. This API will be very easy to misuse. We need something like C++ RAII here... I will think a bit more about this.

ybznek:
Maybe we can use something like this http://stackoverflow.com/a/368731 (I have no personal experience)
We can use some macros to ignore it if code is not compiled by gcc.

mpreisler:
I thought about this and I still really dislike the API.

I think the indent value shouldn't be global. It disallows us from ever doing parallel processing in a clean way. What about passing the value when logging?

jan-cerny:
Thanks all for ideas. I will get back to this PR after finishing #333.

mpreisler:
Another alternative is to make the indent value `thread_local` instead of `global`. It definitely is more finicky than passing variables and will be harder to read for developers. But I am mentioning it here as an idea.

isimluk:
I am not sure what would the thread local indentation give us.

I have an idea though. It may perhaps help us design likable  (:rabbit:) API.  ;-)

What if we set in stone the base indentation for anything from probe? Current proposal allows for the probe log to be indented by any number of characters. That may seem to be needed because of the possible tree of criterions, extended definitions etc. Can we come-up with some probe indentation that would make sense in most cases? Or am I totally wrong (:pig:)?

jan-cerny:
We can either:
1) For probes set fixed indentation to 0 or whatever that looks "nice". Current format of the log is `level:name_of_process: message`. Because each probe has a name of different length, it would not fit the indentation. We can have indentation only for oscap process. This is very simple option.
2) change the format of the log by moving the name of the process behind the text of the message.
Then it makes sense to solve indent also for messages from probes. In such case it would be more difficult to design API.

jan-cerny:
I have thought about the possibilities and I have reworked this pull request.

In my opinion, passing the value (indentation level) as a parameter of macros dI, dW, dE is not a way because:
* we would have to add its parameter everywhere it is used, either directly (horrible idea) or via an wrapper
* we need to store somewhere the current value - perhaps in results_system, add functions for operation with this, and a lot of code (but likely simple) in all places where we would like to use it
* most likely someone in future when he will want to add a new message, he would forgot about it and just use dI() without indentation parameter and break the visual of the log. 

If we would like to do this 100% correctly, we would have to store value of indentation in some shared memory and use semaphores to control access from all processes and threads. This would be too complicated in comparison with small benefit that it brings.

Therefore I suggest messages from probes without possibility to indent. And solve this only for the library.

jan-cerny:
Test selinuxboolean probe again failed. It should be unrelated to this PR. But it happens more and more often recently. I think we should invest some time into investigating it.

mpreisler:
This pull request adds a central singleton global value to the system. We don't have many of those and that's why I am so opposed to this.

The problem of de-indending when errors happen is still unsolved in your new iteration. Developer has to de-indent in all the error paths and as we know one of us will screw this up sooner rather than later :-D

mpreisler:
I have looked at it again and even though I have reservations about how it's done I don't think it makes sense to hold this up any further. There is no easy way to make the API hard to misuse, so let's merge this.

PR #335, opened 2016-02-28
Change messages category to DEVEL

jan-cerny:
While browsing verbose log produced on Fedora SSG,
I have discovered several messages which disturb in INFO level.
They should be in DEVEL category instead.


mpreisler:
ACK

PR #337, opened 2016-03-03
Fix compiler errors on older systems

jan-cerny:
Functions pthread_getname_np and pthread_setname_np were introduced
in glibc 2.12. With older versions of glibc used on RHEL5 or SLES11
OpenSCAP failed to compile. This commit adds an autoconf rule
to detect the presence of these functions.

See mail thread on our mailing list https://www.redhat.com/archives/open-scap-list/2016-March/msg00000.html

### DIFF
--- a/ac_probes/configure.ac.tpl
+++ b/ac_probes/configure.ac.tpl
@@ -149,7 +149,7 @@ fi
 SAVE_CFLAGS=$CFLAGS
 CFLAGS="$CFLAGS -D_GNU_SOURCE"
 LIBS="$pthread_LIBS"
-AC_CHECK_FUNCS([pthread_timedjoin_np clock_gettime])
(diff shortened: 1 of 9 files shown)


mpreisler:
Have you checked that it compiles on RHEL5?

jan-cerny:
@mpreisler This patch fixes the issue with functions pthraed_getname_np and pthread_setname_np. There are other issues that make compiling on RHEL5 impossible. They were discovered by S. Gautam - see https://www.redhat.com/archives/open-scap-list/2016-March/msg00006.html

jan-cerny:
I have added another patch here. With this patch it compiles successfully on my RHEL 5.11, but only if I do `./configure --enable-python=no` because swig gives me tons of errors.

mpreisler:
OK, thank you. ACK

PR #341, opened 2016-03-07
Seap: Fix double free

ybznek:
The free caused "double free or corruption (fasttop)" when I was using
probe from commandline.

Free is already in **SEXP_list_free** https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/OVAL/probes/SEAP/sexp-manip.c#L941

### DIFF
--- a/src/OVAL/probes/SEAP/seap-packet.c
+++ b/src/OVAL/probes/SEAP/seap-packet.c
@@ -730,7 +730,6 @@ eloop_exit:
                         } else {
                                 SEXP_list_free (sexp_buffer);
                                 SEXP_psetup_free (psetup);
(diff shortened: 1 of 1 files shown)


mpreisler:
Shouldn't we be fixing this in maint-1.0?

mpreisler:
The answer is of course we should fix this in `maint-1.0`, I have applied the patch there manually.

https://github.com/OpenSCAP/openscap/commit/61edbbfb6d1cd12f023d3f0d65d256dccce7688e

Please pay closer attention to the branches in the future!

ybznek:
Eh, thanks. Sorry. I was updating some code in maint-1.2 and this commit was only side-effect.

Issue #295, opened 2016-01-20
Can't execute sce from Datastream


xCXDx:
I'm using openscap.x86_64 1.2.7-0.3.el7.centos on Centos 7.2 and I want to execute a Datastream witch includes OVAL and SCE. When I execute this with the command

    oscap xccdf eval datastream.xml

i get the error

    OpenSCAP Error: SCE couldn't find script file 'sce_true.sh'. Expected location: './sce_true.sh'. [sce_engine.c:329]

When I copy the script I use, manually to this location, it works, but thats not the way Datastream should work. I tried also the Version openscap.x86_64 1.0.10-3.el6.centos on Centos 6.7 and Ubuntu 14.04 and there it works the way it should work. openScap extracts the scripts there from datastream and executes them.

The configuration is similar on both hosts.

Does anybody have a solution for sce in datastream with openscap 1.2?

isimluk:
Thanks for reporting this. I have found that we do not have tests for this. We will need to look further into this.

ybznek:
I confirm the issue, with maint-1.0 is it ok, but with maint-1.2 doesn't work

Error message
```
OpenSCAP Error: Error while creating script component (id='scap_org.open-scap_ecomp_services--obsolete--disable_nfs_exports.sh') to file '/tmp/oscap.2PicNM/././services/obsolete/disable_nfs_exports.sh'. [sds.c:165]
SCE couldn't find script file 'services/obsolete/disable_nfs_exports.sh'. Expected location: './services/obsolete/disable_nfs_exports.sh'. [sce_engine.c:329]
```

PR #347, opened 2016-03-09
tests/sce: Add test for SCE in DS

ybznek:
Test for SCE in DS, will FAIL in maint-1.2 branch.
Wais for https://github.com/OpenSCAP/openscap/pull/349

PR #349, opened 2016-03-10
SDS SCE fix

ybznek:
Should fix:
- fix https://github.com/OpenSCAP/openscap/issues/295
- refactoring/don't create temp directory for SCE when we don't have any SCE

### DIFF
--- a/src/DS/sds.c
+++ b/src/DS/sds.c
@@ -159,6 +159,10 @@ static xmlNodePtr _lookup_component_in_collection(xmlDocPtr doc, const char *com
 static int ds_sds_dump_component_sce(xmlNode *script_node, const char *component_id, const char *filename)
 {
 	if (script_node) {
+		if (oscap_acquire_ensure_parent_dir(filename) < 0) {
+			oscap_seterr(OSCAP_EFAMILY_XML, "Error while creating script parent directory for file '%s' of (id='%s')", filename, component_id);
+			return -1;
+		}
 		// TODO: should we check whether the component is extended?
 		int fd;
 		xmlChar* text_contents = xmlNodeGetContent(script_node);
--- a/src/DS/sds.c
+++ b/src/DS/sds.c
@@ -7,13 +7,13 @@
  * License as published by the Free Software Foundation; either
  * version 2.1 of the License, or (at your option) any later version.
  *
- * This library is distributed in the hope that it will be useful, 
+ * This library is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  * Lesser General Public License for more details.
  *
  * You should have received a copy of the GNU Lesser General Public
- * License along with this library; if not, write to the Free Software 
+ * License along with this library; if not, write to the Free Software
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  *
  * Authors:
--- a/src/DS/sds.c
+++ b/src/DS/sds.c
@@ -197,7 +197,7 @@ static int ds_sds_dump_component_sce(xmlNode *script_node, const char *component
 	}
 }
 
-static int ds_sds_dump_component(const char* component_id, struct ds_sds_session *session, const char *sce_filename, const char *relative_filepath)
+static int ds_sds_dump_component(const char* component_id, struct ds_sds_session *session, const char *target_filename_dirname, const char *relative_filepath)
 {
 	xmlDoc *doc = ds_sds_session_get_xmlDoc(session);
 	xmlNodePtr component = _lookup_component_in_collection(doc, component_id);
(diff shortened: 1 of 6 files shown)


ybznek:
I don't have time to review it by myself today. I will review it and remove potentially WIP label tomorrow.

src/DS/sds_priv.h
```diff
@@ -40,7 +40,7 @@ xmlNode *containter_get_component_ref_by_id(xmlNode *container, const char *comp
 
 int ds_sds_dump_component_ref(xmlNodePtr component_ref, struct ds_sds_session *session);
```
jan-cerny: You added const in definition of this function, but not in header file,

src/DS/sds_priv.h
```diff
@@ -40,7 +40,7 @@ xmlNode *containter_get_component_ref_by_id(xmlNode *container, const char *comp
...
 
-int ds_sds_dump_component_ref_as(xmlNodePtr component_ref, struct ds_sds_session *session, const char *target_dir, const char *relative_filepath);
+int ds_sds_dump_component_ref_as(xmlNodePtr component_ref, struct ds_sds_session *session, const char *sub_dir, const char *relative_filepath);
```
jan-cerny: same here (const xmlNodePtr)

ybznek: Thanks!

src/XCCDF/xccdf_session.c
```diff
@@ -821,7 +821,14 @@ int xccdf_session_load_check_engine_plugin(struct xccdf_session *session, const
 
 	oscap_list_add(session->check_engine_plugins, plugin);
 
-	return check_engine_plugin_register(plugin, session->xccdf.policy_model, oscap_source_readable_origin(session->xccdf.source));
+	if (xccdf_session_get_datastream_id(session)!=NULL){
+		return check_engine_plugin_register(plugin, session->xccdf.policy_model, ds_sds_session_get_target_dir(session->ds.session));
+	}else{
```
jan-cerny: Please stick to our codestyle.

ybznek:
Thanks @jan-cerny  for feedback. I've fixed ```const``` and tried to fix coding style, but maybe I didn't understand what exactly do you mean.

jan-cerny:
@ybznek What about how we will test this? I am asking because @isimluk said in https://github.com/OpenSCAP/openscap/pull/347 :
> iirc, it works on maint-1.2. We perhaps need different test for #349. :)

ybznek:
Test failed due to process58, not related to this PR.

test this please

ybznek:
My oversight.
Test "pass" in maint-1.2 because it wasn't started during make check 

https://github.com/OpenSCAP/openscap/pull/358

jan-cerny:
@ybznek Do I understand it well that this pull request is blocked by #358 ?

ybznek:
Not really blocked. Rather to add fix before tests than tests before fix.

jan-cerny:
@ybznek OK, understood, I will do review now.

src/DS/sds.c
```diff
@@ -255,16 +266,11 @@ int ds_sds_dump_component_ref_as(xmlNodePtr component_ref, struct ds_sds_session
...
-
-	const char* target_filename_dirname = oscap_sprintf("%s/%s", target_dir, file_reldir);
-	const char* sce_filename = oscap_sprintf("%s/%s/%s", target_dir, file_reldir, file_basename);
-	ds_sds_dump_component(component_id, session, sce_filename, relative_filepath);
-	oscap_free(sce_filename);
+	const char* target_filename_dirname = oscap_sprintf("%s/%s",sub_dir, file_reldir);
 	oscap_free(filename_cpy);
 
+	ds_sds_dump_component(component_id, session, target_filename_dirname, relative_filepath);
+
```
jan-cerny: I think that there are some needless operations.
* `filename_cpy` is always `./relative_filepath` - in other words it always starts with `./`
* therefore calling `dirname(filename_cpy)` will always return `.`
* then `file_reldir` will always be `.`
* then `target_filename_dirname` will be `sub_dir/.`. I find it weird that this path will always finish with `.`
* Conslusion: `target_filename_dirname` equals `sub_dir` (from the point of view of unix paths)

I think you can remove lines 266-271 and call the function `ds_sds_dump_component` like this: 
```
ds_sds_dump_component(component_id, session, sub_dir, relative_filepath)
```

The variable `target_filename_dirname` is used also in recursive call of `ds_sds_dump_component_ref_as()` at line 327. I think we can also use `sub_dir` there instead of `target_filename_dirname`.

jan-cerny:
Other parts of this LGTM.

Just a reminder that we should merge maint-1.0 into maint-1.2 after merging this because of fix for test for this feature (commit 609ccc7cf64f63901d174c1f7295f33739065874)

src/DS/sds.c
ybznek: I don't agree.
When I write
```
printf("relative_filepath: %s| reldir %s \n", relative_filepath, file_reldir);
```
I will get
```
relative_filepath: xccdf.xml| reldir . 
relative_filepath: services/obsolete/disable_nfs_exports.sh| reldir ./services/obsolete 
```
So I expect, that your premise is wrong.







That path is ends with '.' isn't generally bad I think.
You can use **/home** or **/home/**, or when you build path from parts, you can get **/home/./././././** - strange, but as I know, should be always valid unix path.



Such building of paths already was in the code, but I tried to refactor it.
So If I didn't understand you well, please comment, we can improve it together.

jan-cerny: Thanks for explanation.

jan-cerny:
ACK.

PR #350, opened 2016-03-11
Issue #348: Fix offline mode in rpmverifyfile probe

jan-cerny:
If we chroot into filesystem where is not a RPM database,
we should not fail to init the probe.

### DIFF
--- a/src/OVAL/probes/unix/linux/rpminfo.c
+++ b/src/OVAL/probes/unix/linux/rpminfo.c
@@ -303,16 +303,12 @@ ret:
 
 void *probe_init (void)
 {
-	probe_offline_flags offline_mode = PROBE_OFFLINE_NONE;
+	probe_setoption(PROBEOPT_OFFLINE_MODE_SUPPORTED, PROBE_OFFLINE_CHROOT|PROBE_OFFLINE_RPMDB);
+	addMacro(NULL, "_dbpath", NULL, getenv("OSCAP_PROBE_RPMDB_PATH"), 0);
 
         if (rpmReadConfigFiles ((const char *)NULL, (const char *)NULL) != 0) {
                 dI("rpmReadConfigFiles failed: %u, %s.", errno, strerror (errno));
                 return (NULL);
-        }
-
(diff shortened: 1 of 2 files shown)


jan-cerny:
@ybznek You have a great experience with offline modes. Please could you test this in your environment?

src/OVAL/probes/unix/linux/rpminfo.c
```diff
@@ -303,18 +303,14 @@ static int get_rpminfo (struct rpminfo_req *req, struct rpminfo_rep **rep)
...
+	probe_setoption(PROBEOPT_OFFLINE_MODE_SUPPORTED, PROBE_OFFLINE_CHROOT|PROBE_OFFLINE_RPMDB);
+	addMacro(NULL, "_dbpath", NULL, getenv("OSCAP_PROBE_RPMDB_PATH"), 0);
 
         if (rpmReadConfigFiles ((const char *)NULL, (const char *)NULL) != 0) {
                 dI("rpmReadConfigFiles failed: %u, %s.", errno, strerror (errno));
                 return (NULL);
         }
 
-        probe_getoption(PROBEOPT_OFFLINE_MODE_SUPPORTED, NULL, &offline_mode);
-        if (offline_mode & PROBE_OFFLINE_RPMDB) {
```
mpreisler: @jan-cerny What distinguishes the 2 modes after you remove this logic check?

ybznek:
It fix https://github.com/OpenSCAP/openscap/issues/348 issue with "all true", but I don't get "not applicable" when I want to check presence of package.

ybznek:
I will merge it, we can fix minor issues later.

src/OVAL/probes/unix/linux/rpminfo.c
jan-cerny: @mpreisler Environment variable OSCAP_PROBE_RPMDB_PATH distinguishes between 2 modes.

mpreisler: But there is no check for the mode inside the probe and there was one before. I don't think it makes sense to use an implementation detail when we can use openscap API to get the mode...

PR #351, opened 2016-03-11
Fix offline mode in file probe

jan-cerny:
I have found a logical error in file probe.
The offline mode in file probe is enabled on only when a mutex cannot be initialized,
which is not very likely and also it is definitely not what we want.
This commit enables the offline mode in file probe.


mpreisler:
ACK

Issue #230, opened 2015-11-24
OpenSCAP prints many error messages when building sce-community-content


jan-cerny:
I have OpenSCAP 1.2.6 installed from Fedora 22 repository. I have installed openscap-sce-engine package. I have cloned the sce-community-content repository from https://github.com/OpenSCAP/sce-community-content The problem is I am not able to build it.
when I do
```
cd sce-community-content
cd src
make
```
I got the following error:
```
[jcerny@thinkpad src{master}]$ make
../utils/xccdf_compose.py .
Rule 'xccdf_org.open-scap.sce-community-content_rule_services_obsolete-disable_finger_server' contains no proper list of affected file(s)
Rule 'xccdf_org.open-scap.sce-community-content_rule_services_obsolete-disable_rsh_server' contains no proper list of affected file(s)
Rule 'xccdf_org.open-scap.sce-community-content_rule_services_obsolete-disable_telnet_server' contains no proper list of affected file(s)
oscap xccdf validate-xml all-xccdf.xml
oscap xccdf resolve -o all-resolved-xccdf.xml all-xccdf.xml
oscap xccdf generate guide all-resolved-xccdf.xml > guide.html
../utils/xml_append.py all-resolved-xccdf.xml dummy-oval-rule.xml > all-resolved-xccdf-dummyoval.xml
oscap ds sds-compose all-resolved-xccdf-dummyoval.xml sds-datastream-unformated.xml
OpenSCAP Error: Start tag expected, '<' not found [oscap_source.c:230]
Entity: line 1: parser error : Start tag expected, '<' not found
#!/usr/bin/env sh
^
Unable to parse XML at: 'system/authsettings/passwd_file_permissions.sh' [oscap_source.c:232]
Start tag expected, '<' not found [oscap_source.c:230]
Entity: line 1: parser error : Start tag expected, '<' not found
#!/usr/bin/env sh
^
Unable to parse XML at: 'system/authsettings/passwd_file_owner.sh' [oscap_source.c:232]
Start tag expected, '<' not found [oscap_source.c:230]
Entity: line 1: parser error : Start tag expected, '<' not found
#!/usr/bin/env sh
^
Unable to parse XML at: 'system/authsettings/passwd_file_group.sh' [oscap_source.c:232]
Start tag expected, '<' not found [oscap_source.c:230]
Entity: line 1: parser error : Start tag expected, '<' not found
#!/usr/bin/env sh
^
[...]

PR #354, opened 2016-03-11
Stop printing of XML errors during SCE loading

ybznek:
Should fix https://github.com/OpenSCAP/openscap/issues/230

Feedback is welcome.


mpreisler:
Epic stuff. I will note that it doesn't work in the general case for ELF binaries but for the most common things used as SCE this will work nicely.

ybznek:
I though that we don't use ELF binaries. But now it is easy to make function more general.

mpreisler:
@ybznek The SCE spec allows them but AFAIK nobody uses them in practice.

Issue #342, opened 2016-03-08
The group-by feature in HTML report should sort the references by ID


mpreisler:
Else it's very hard to find the requirement ID I am looking for. See attached screenshot.

The order seems implementation defined or undefined.

![image](https://cloud.githubusercontent.com/assets/753153/13613237/4a7f5d5a-e539-11e5-87c7-526fe199c83a.png)

Please consider sorting in a way that makes sense for most of the IDs. Most of the IDs are sets of numbers delimited by some delimiter. Just sorting them as strings will not provide the best results. Splitting the IDs by known delimiters and then sorting the lists will work better.

jan-cerny:
@mpreisler Thanks for bringing this up. I am looking into it now. I found that sorting for PCI-DSS IDs has been already implemented before, but that time we did not have the granularity (eg. 10.5.2). I would like to implement it in a more robust way.

jan-cerny:
Fixed in https://github.com/OpenSCAP/openscap/pull/357

PR #357, opened 2016-03-15
Fix grouping in HTML report

jan-cerny:
This pull request fixes issue #342 by a new, more generic sorting function.

Also this pull request fixes some minor issues that I observed in report from RHEL7 SSG.


xsl/xccdf-resources/openscap.js
```diff
@@ -179,44 +180,27 @@ function sortGroups(groups, key)
...
-            if (b_parts == null)
-                return -1;
-            var result = a_parts[1].localeCompare(b_parts[1]);
-            if (result != 0) {
-                return result;
-            } else {
-                result = a_parts[2] - b_parts[2];
-                if (result != 0) {
-                    return result;
+            var a_parts = a.split(/[.()-]/);
```
ybznek: I think it doesn't cause real-world performance issue, but If I understand it well
you need to compile these regexps every time you need to compare some two elements.
I don't know if JS can optimize this. I personally prefer move initialization of these regexps out of function what is call many times.

jan-cerny: From what I found here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions this is a so-called "regular expression literal". It is compiled when the script is loaded, not at runtime, so it will be compiled only once. It think we can leave it as is.

jan-cerny:
I have tested it on Google Chrome 49.0.2623.75 and Firefox 44.0.2 with report from RHEL7 and Fedora SSG.

xsl/xccdf-resources/openscap.js
ybznek: Good, thanks.

mpreisler:
Please attach an HTML report for us to test.

jan-cerny:
@mpreisler I provided example report to @ybznek on Tuesday. I think he had already done some testing. However I agree that it is much more better to post the example publicly. So here it is:

[report.html.zip](https://github.com/OpenSCAP/openscap/files/179345/report.html.zip)

mpreisler:
ACK

PR #358, opened 2016-03-15
tests/sce-in-ds: Add to Makefile.am

ybznek:
I forgot to append test to 
Append forgotten test

### DIFF
--- a/tests/sce/Makefile.am
+++ b/tests/sce/Makefile.am
@@ -9,7 +9,8 @@ TESTS_ENVIRONMENT= \
 
 TESTS = test_sce.sh \
 		test_passing_vars.sh \
-		test_check_engine_results.sh
+		test_check_engine_results.sh \
+		test_sce_in_ds.sh
 
 EXTRA_DIST +=	test_sce.sh \
 		sce_xccdf.xml \


mpreisler:
Why is this going to TESTS and not EXTRA_DIST?

ybznek:
I think, only one line is missing (see whole file).

tests/sce/Makefile.am
```diff
@@ -9,7 +9,8 @@ TESTS_ENVIRONMENT= \
 
 TESTS = test_sce.sh \
 		test_passing_vars.sh \
-		test_check_engine_results.sh
+		test_check_engine_results.sh \
+		test_sce_in_ds.xml
```
mpreisler: .xml ?!

jan-cerny:
@ybznek test_sce_in_ds.xml should be in extra_dist
Please verify that you can successfully run and pass `make distcheck` with your changes.
The `make distcheck` is different than `make check`.

ybznek:
@jan-cerny 
https://github.com/ybznek/openscap/blob/maint-1.0-fix-tests/tests/sce/Makefile.am#L28
Isn't it what you want?

mpreisler:
@jan-cerny It is in EXTRA_DIST and it was there already.

Will merge this after Jenkins finishes.

jan-cerny:
@mpreisler Yes, I see it, thanks.

PR #359, opened 2016-03-15
tests/sce_parse_errors: Add test to Makefile.am

ybznek:
Add forgotten test


ybznek:
environmentvariable FAIL should not be related

test this please

PR #360, opened 2016-03-23
Query test and evaluate its results in one step

jan-cerny:
Currently, we do two steps to get the results of a definition:
1) Querying probes for objects for all the definitions
2) After all items are collected, we compute results separately.
It causes some confusion in the verbose log in cases one definition contains multiple criteria,
the messages are not in logical order here.
I think that messages describing computing results of a test should be displayed just after messages describing evaluation of the same test, it should not be mixed as is now.
We can fix this by merging aforementioned steps into single step.
It would basically mean that each test is evaluated separately, no other querying starts before a result of single test is done.
Therefore we would have messages in verbose mode in expected order.

This pull request is a rework of #333. It tries to call probing from the results model.

### DIFF
--- a/src/OVAL/oval_agent.c
+++ b/src/OVAL/oval_agent.c
@@ -172,9 +172,6 @@ int oval_agent_eval_definition(oval_agent_session_t *ag_sess, const char *id)
 	dI("Evaluating definition '%s': %s.", id, title);
 
 	/* probe */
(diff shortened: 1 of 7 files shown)


isimluk:
I have no words. This is one is more than perfect.

Thank You, Jan!!

jan-cerny:
Great! Thanks for review.

PR #362, opened 2016-03-24
Fix openscap-cpe-oval.xml: Remove check to unix family

ybznek:
When we cannot get RPM database, we should return "NOT APPLICABLE".
Problem is so, that if we get
isUnix(TRUE) AND isFedora(not applicable) we get TRUE -> so for all systems we get TRUE

### DIFF
--- a/cpe/openscap-cpe-oval.xml
+++ b/cpe/openscap-cpe-oval.xml
@@ -196,12 +196,6 @@
             </definition>
       </definitions>
       <tests>
-            <family_test check_existence="at_least_one_exists" id="oval:org.open-scap.cpe.rhel:tst:1" version="1" check="only one" 
-                  comment="installed operating system is part of the Unix family" 
-                  xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent">
-                  <object object_ref="oval:org.open-scap.cpe.unix:obj:1"/>
-                  <state state_ref="oval:org.open-scap.cpe.unix:ste:1"/>
-            </family_test>
             <rpmverifyfile_test check_existence="at_least_one_exists" id="oval:org.open-scap.cpe.rhel:tst:2" version="1" check="at least one" comment="/etc/redhat-release is provided by redhat-release package"
                   xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5#linux">
                   <object object_ref="oval:org.open-scap.cpe.redhat-release:obj:3"/>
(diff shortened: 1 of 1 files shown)


jan-cerny:
@ybznek What needs to be done before merging this?

ybznek:
@jan-cerny Sorry, I am currently busy. After I got advice how to deal with it I've done quick commits from my personal laptop when I am not able to test it. So I need to test it now.

ybznek:
I've done changes according to discussion in https://github.com/OpenSCAP/openscap/issues/348.

It is working with chroot offline scan, too.

jan-cerny:
@ybznek thank you. The patch looks good to me. Merging.

However, I discussed this with @iankko today. It seems that we can't achieve the right solution, because these definitions does not actually check CPE. They only check RPM packages, which is a different thing. To do it correctly, we would need to have some "CPE probe", in other words, specify and propose a `cpe_test` to OVAL standard.

mpreisler:
> However, I discussed this with @iankko today. It seems that we can't achieve the right solution, because these definitions does not actually check CPE. They only check RPM packages, which is a different thing. To do it correctly, we would need to have some "CPE probe", in other words, specify and propose a cpe_test to OVAL standard.

AFAIK, we check for redhat-release and based on that we declare whether a specific CPE is applicable. Can you elaborate how a cpe_test would be used?

iankko:
> AFAIK, we check for redhat-release and based on that we declare whether a specific CPE is applicable. Can you elaborate how a cpe_test would be used?

The issue is right now we are "misusing" rpminfo test to do this service for us (checking presence of 'redhat-release' or whatever name that specific RPM is called {seems to be e.g. ```redhat-release-server-6Server-6.7.0.3.el6.x86_64``` on RHEL-6 Server system}).

But the very same package provides also:

/etc/system-release-cpe

where we can directly obtain CPE id (on RHEL / Fedora from). I would assume RHEL- derivative 
OSes to have similar / the very same location (just different CPE listed there). For Debian, Gentoo, 
Ubuntu I expect the CPE information to be provided under some filesystem location too (the only 
distribution I am aware of that doesn't provide CPE information seems to be ArchLinux right now).

So basically the proposal discussed with Honza today @ lunch was instead of using ```rpminfo``` 
probe / test / check, rather to introduce ```cpe_probe``` into the standard that would provide 
particular information about underlying OS:

* on RHEL / Fedora it would look into ```/etc/system-release-cpe``` and obtain CPE from there,
* on Debian, Gentoo, Ubuntu it would look into (the concrete location would need to be found out 
yet),
* not sure what to do on ArchLinux (but maybe in the meantime we could get ArchLinux upstream to 
supply / provide the CPE information),
* for Microsoft Windows I would assume they to provide the CPE information somewhere within the 
system too.

mpreisler:
AFAIK vendors don't always follow this convention. Even on one distribution the paths differ between versions. Checking for CPEs in a reliable generic way is a big challenge which is why we work with rpminfo instead.

If we were to rely on this path we could just use textfilecontent. I guess I am missing something because I still don't see the value in proposing `cpe_test`. Keep in mind that it would take a long time to push it through and in the meantime we have to rely on our current solution anyway. I would dare say that the current solution has some issue but works well overall.

Issue #329, opened 2016-02-19
oscap-docker mount container in current directory


ybznek:
I think, we should mount container in /tmp or in some directory which is intended to mounting.

```
zmoravec: /proc 13:37:41
$ sudo oscap-docker image-cve rhel7
Traceback (most recent call last):
  File "/bin/oscap-docker", line 108, in <module>
    args.func()
  File "/bin/oscap-docker", line 44, in cve_scan
    result = OS.scan_cve(self.args.scan_target, self.unknown_args)
  File "/usr/lib/python2.7/site-packages/oscap_docker_python/oscap_docker_util.py", line 193, in scan_cve
    _tmp_mnt_dir = DM.mount(image)
  File "/usr/lib/python2.7/site-packages/Atomic/mount.py", line 322, in mount
    driver_mount_fn(identifier, options)
  File "/usr/lib/python2.7/site-packages/Atomic/mount.py", line 368, in _mount_devicemapper
    raise MountError(e)
Atomic.mount.MountError: [Errno 2] No such file or directory: '/proc/024f294ea48f3e3b79f9'
```
I cannot run oscap-docker in ```/proc``` directory, because oscap-docker want to mount the image to ```/proc/024f294ea48f3e3b79f9```.

mpreisler:
Is this an issue only in `oscap-docker` CLI tool or does the same thing apply to other tools using the `oscap_docker_python` package?

ybznek:
It seems, that problem is probably here https://github.com/OpenSCAP/openscap/blob/maint-1.2/utils/oscap_docker_python/oscap_docker_util.py#L180 ```mnt_dir=os.getcwd()```

**Daemon:**
Daemon doesn't use ```OscapScan```
openscap_daemon/cve_scanner/scan.py:        self.DM = DockerMount("/tmp", mnt_mkdir=True)
TLDR: daemon doesn't have this problem.

In my opinion, easiest solution is replace ```os.getcwd()``` with ```/tmp```, so oscap_docker will use same directory for mount as daemon.

mpreisler:
Alright. Are we sure this is safe? Can I guess the directory that will be created in `/tmp` and break it by creating it?

What I am getting at is that maybe we should be using private temp dir. It's very little added effort and we are guaranteed a directory that is private to whoever created it.

ybznek:
Sorry, you are right. It is better to have private directory for mount in oscap-docker and daemon.

mpreisler:
[...]

PR #363, opened 2016-03-24
oscap-docker: Mount in safe temporary container

ybznek:
https://github.com/OpenSCAP/openscap/issues/329

If no directory to mount is passed, it create temporary directory before mount and remove it after operation is done.

### DIFF
--- a/utils/oscap_docker_python/oscap_docker_util.py
+++ b/utils/oscap_docker_python/oscap_docker_util.py
@@ -178,19 +178,39 @@ class OscapHelpers(object):
 
 
 class OscapScan(object):
-    def __init__(self, tmp_dir=tempfile.gettempdir(), mnt_dir=os.getcwd(),
+    def __init__(self, tmp_dir=tempfile.gettempdir(), mnt_dir=None,
                  hours_old=2):
         self.tmp_dir = tmp_dir
         self.helper = OscapHelpers(tmp_dir)
         self.mnt_dir = mnt_dir
         self.hours_old = hours_old
 
+    def _get_mnt_dir(self):
+        '''
+        Get directory for mount container
+        '''
+        if self.mnt_dir is None:
+            return tempfile.mkdtemp()
+        else:
+            return self.mnt_dir
+
+    def _remove_mnt_dir(self, mnt_dir):
+        '''
+        Remove temporary directory if it was not passed by __init__
+        We don't want to remove directory what was created by used
+        '''
+        if self.mnt_dir != mnt_dir:
+            os.rmdir(mnt_dir)
+
     def scan_cve(self, image, scan_args):
         '''
         Wrapper function for scanning a container or image
         '''
+
+        mnt_dir = self._get_mnt_dir()
+
         # Mount the temporary image/container to the dir
-        DM = DockerMount(self.mnt_dir, mnt_mkdir=True)
+        DM = DockerMount(mnt_dir, mnt_mkdir=True)
         try:
             _tmp_mnt_dir = DM.mount(image)
         except MountError as e:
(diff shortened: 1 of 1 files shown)


utils/oscap_docker_python/oscap_docker_util.py
```diff
@@ -178,19 +178,39 @@ def _cleanup_by_path(self, path):
...
+        '''
+        if self.mnt_dir is None:
+            return tempfile.mkdtemp()
+        else:
+            return self.mnt_dir
+
+    def _remove_mnt_dir(self, mnt_dir):
+        '''
+        Remove temporary directory if it was not passed by __init__
+        We don't want to remove directory what was created by used
```
mpreisler: I don't understand this comment.

utils/oscap_docker_python/oscap_docker_util.py
```diff
@@ -178,19 +178,39 @@ def _cleanup_by_path(self, path):
...
+        '''
+        if self.mnt_dir != mnt_dir:
```
mpreisler: why not `is None`? that's way more readable

ybznek: I wanted to have more explicit that we want to remove only directory created by us and we don't want to remove directory created by user. I can change if you want.

utils/oscap_docker_python/oscap_docker_util.py
ybznek: Thanks for feedback. typo, I will try to improve it.

utils/oscap_docker_python/oscap_docker_util.py
mpreisler: I think `is None` with a comment is way more readable. The whole logic is a bit confusing, so feel free to rename variables and move things around a bit.

utils/oscap_docker_python/oscap_docker_util.py
```diff
@@ -178,19 +178,39 @@ def _cleanup_by_path(self, path):
...
                  hours_old=2):
         self.tmp_dir = tmp_dir
         self.helper = OscapHelpers(tmp_dir)
         self.mnt_dir = mnt_dir
         self.hours_old = hours_old
 
+    def _get_mnt_dir(self):
+        '''
+        Get directory for mount container
+        '''
```
jan-cerny: According to its name and comment, this method looks like a getter, but actually it can create a directory. Could we have a better name or better comment?

ybznek: Yes, good point. I will try to improve name, but I don't know how to comment it better. Comment message longer than method code isn't good way. Especially in so short non public method.

I will amend last commit with the changes.

utils/oscap_docker_python/oscap_docker_util.py
ybznek: Personally I think we don't need to have temporary directory what can be defined by user. Especially when we don't leave anything in the directory. I didn't want to break some usecase.

ybznek: If I should do more refactoring, I would
- merge common code of scan/scan_cve - maybe use __enter__/__exit__
- remove tmp_dir and mnt_dir from __init__ signature - it will break some use case, but as I know, nowhere it is used now.

(there is still some quite confusing parts like "_cleanup_by_path" Create new DockerMount instance)

Please, share your expectation.

mpreisler: I don't think we can do either of those in `maint-1.2`. It's public API after all.

mpreisler:
ACK

PR #365, opened 2016-03-26
If verbose log file is not specified, write messages on stderr

jan-cerny:
This pull request proposes a small enhancement in OpenSCAP verbose mode, based on feedback from developers team.

The `--verbose-log-file` will no longer be obligatory with verbose mode. When this option won't be specified, the messages will be written on stderr.

I hope this will make our developer's life a little easier.

Also edited the user manual to reflect this change.


isimluk:
Removal of that `dup2` in `NDEBUG` has potential to give us some surprises in future. However, I am ok accepting the risk.

Thanks a lot for this change @jan-cerny!!

PR #367, opened 2016-03-30
Fix openscap-cpe-oval.xml: Remove check to unix family - 1.2 only

ybznek:
Related to
https://github.com/OpenSCAP/openscap/pull/362

But I don't know reason why we don't share same cpe-oval.xml in both branches.

### DIFF
--- a/cpe/openscap-cpe-oval.xml
+++ b/cpe/openscap-cpe-oval.xml
@@ -78,7 +78,6 @@
                         <description>The operating system installed on the system is Community Enterprise Operating System 5</description>
                   </metadata>
                   <criteria>
-                        <criterion comment="Installed operating system is part of the unix family" test_ref="oval:org.open-scap.cpe.rhel:tst:1"/>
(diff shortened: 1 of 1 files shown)


jan-cerny:
This is interesting. `openscap-cpe-oval.xml` and `openscap-cpe-dict.xml` in maint-1.2 contain CPEs for CentOS and (Open)SUSE systems, whereas in maint-1.0 these are not included. Otherwise the two files are the same.

@isimluk Is there any reason why CentOS and SUSE CPEs can't be in maint-1.0?
I haven't found any,
If there is not any reason, I suggest merging them into maint-1.0 and have these two files same in all branches.

ybznek:
If you agree, I will merge it - (including changes in this PR)

jan-cerny:
@ybznek Personally I am OK with it. Please double check that OVALs in both branches are valid.

ybznek:
@mpreisler FIxed this during merge. We don't need this PR now.

PR #369, opened 2016-03-31
Adding implementation for runlevel probe on SUSE

GautamSatish:
This check-in implements the run-level probe for SUSE as was discussed on the mailing list.

If have implemented the function separate from get_runlevel_sysv because the file paths to init.d and the rc directories are different on SUSE and there was no GCC macros to distinguish between RHEL and SUSE.

Also included the linux/limits.h header file in oval_session.c to resolve PATH_MAX.

### DIFF
--- a/src/OVAL/oval_session.c
+++ b/src/OVAL/oval_session.c
@@ -27,6 +27,7 @@
 #include <libgen.h>
 #include <unistd.h>
 #include <string.h>
+#include <linux/limits.h>
 
 #include "common/alloc.h"
 #include "common/debug_priv.h"
--- a/src/OVAL/probes/unix/runlevel.c
+++ b/src/OVAL/probes/unix/runlevel.c
@@ -85,15 +85,10 @@ struct runlevel_rep {
 static int get_runlevel (struct runlevel_req *req, struct runlevel_rep **rep);
 
 #if defined(__linux__) || defined(__GLIBC__) || (defined(__SVR4) && defined(__sun))
-static int get_runlevel_sysv (struct runlevel_req *req, struct runlevel_rep **rep)
+static int get_runlevel_sysv (struct runlevel_req *req, struct runlevel_rep **rep, bool suse, const char *init_path, const char *rc_path)
 {
 	const char runlevel_list[] = {'0', '1', '2', '3', '4', '5', '6'};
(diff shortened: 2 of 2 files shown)


isimluk:
Test this please

(message from our jenkins) :bath:

jan-cerny:
@GautamSatish Thanks for your contribution.

I don't like that most of the code is a copy from `get_runlevel_sysv()`. The two functions differs only a little. We should avoid repeating code. If there is a bug in the code, we will have to fix it in 2 places (which we will definitely forget to do). :-) Could you re-factor this in a way that you don't copy the code, please? Maybe a good idea would be to move the common code to a new static function.

GautamSatish:
@jan-cerny , forgive me, you had already given me this comment on the mailing list, it slipped my mind.

I can add a wrapper to the `get_runlevel_sysv()` function for each platform and have them initialize the init_path and rc_path variables and then pass them over to it along with a flag to determine if platform is suse. So that this check will be more of a run-time logic. 

Or, I could add a new define option in CFLAGS say -D_SUSE=0 and the value needs to be set to 1 when building for SUSE so that we can simply add  #if defined(_SUSE_) in the current `get_runlevel_sysv()` module. I don't see any existing gcc pre-processor macros that can distinguish between Redhat and SUSE. 

What do you think?

jan-cerny:
@GautamSatish First option sounds better to me.

GautamSatish:
@jan-cerny , I have made the modifications in the second commit. Could you please take a look?

jan-cerny:
@GautamSatish Thanks! Now the code looks good to me, Please squash the two commits, do `push -f` and then I think it can be merged,

GautamSatish:
@jan-cerny I think I have merged the way you wanted.

jan-cerny:
@GautamSatish Thank you very much. We highly appreciate all efforts that make OpenSCAP usable on other platforms. There are still some minor issues in your code -- an unused variable in `get_runlevel_redhat`, and some trailing whitespace. We also try to follow the [Linux Kernel Coding Style](https://www.kernel.org/doc/Documentation/CodingStyle).  However, I accept the patch and merge it upstream. Please keep those remarks in mind. We are looking forward to your further contribution.

PR #373, opened 2016-04-03
Move a message from unused code

jan-cerny:
Since we have merged PR #360, we miss a message reporting extended
definition in our verbose mode. This commit moves a message from
function which is not used now to the right place.


mpreisler:
ACK.

PR #376, opened 2016-04-04
openscap-cpe-*.xml: update from maint-1.2

ybznek:
Merge CPE oval from maint-1.2.

After this commit should be these files same - byte by byte.


ybznek:
Short discussion there https://github.com/OpenSCAP/openscap/pull/367

jan-cerny:
@ybznek Thanks. ACK.

mpreisler:
Discussed this with @jan-cerny on IRC. I don't think this PR makes sense. It's completely fine that the CPE OVALs are different between branches. We don't add support for new platforms into legacy stable branches.

PR #377, opened 2016-04-04
Revert "openscap-cpe-*.xml: update from maint-1.2"

mpreisler:
Reverts OpenSCAP/openscap#376


jan-cerny:
Discussed this with @mpreisler on IRC. This commit that is going to be reverted should no go to maint-1.0. I will wait for CI and then merge.

PR #382, opened 2016-04-07
Remove information about spec files from user manual

jan-cerny:
The spec files are not present in maint-1.2 anymore.
They were removed in commits
3fe208f3caf1f3a46cd2ee2ea7baa88154600674
41fe1106a680885cd684313d306c12eb54d4353d

### DIFF
--- a/docs/manual/manual.adoc
+++ b/docs/manual/manual.adoc
@@ -1169,13 +1169,6 @@ installed on your system:
 
 . Run the installation procedure by executing the following command: # make install
 
-If you want to create a package for Fedora or Red Hat Enterprise Linux
-distribution, you will need the respective spec files. These spec files are
-available under the following directories:
-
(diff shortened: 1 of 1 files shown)


ybznek:
I think we definitely should not have .spec files in this repo.

Don't you think It can be useful to refer another/right way to get spec files?

jan-cerny:
@ybznek Thanks for idea. Do you think that something like I answered on mailing list would be enough? https://www.redhat.com/archives/open-scap-list/2016-April/msg00030.html

ybznek:
@jan-cerny Sorry. I was busy.

I prefer to have mentioned  ```fepkg``` way too. So both - .srpm and fedpkg.

It can be separated pull request. Deletion if invalid information is one independent step forward, so I will merge it.

Issue #368, opened 2016-03-30
Suspicious error messages with chroot


jan-cerny:
When testing #361 I have discovered that running a scan with chroot to a directory where can emit suspicious error messages on stderr.
```
[root@thinkpad openscap]# utils/oscap-chroot /tmp oval eval cpe/openscap-cpe-oval.xml
error: Failed to initialize NSS library
Definition oval:org.open-scap.cpe.rhel:def:7: not applicable
Definition oval:org.open-scap.cpe.rhel:def:6: not applicable
error: Failed to initialize NSS library
Definition oval:org.open-scap.cpe.rhel:def:5: not applicable
Definition oval:org.open-scap.cpe.rhel:def:1: not applicable
Definition oval:org.open-scap.cpe.fedora:def:25: not applicable
Definition oval:org.open-scap.cpe.fedora:def:24: not applicable
Definition oval:org.open-scap.cpe.fedora:def:23: not applicable
Definition oval:org.open-scap.cpe.fedora:def:22: not applicable
Definition oval:org.open-scap.cpe.fedora:def:21: not applicable
Definition oval:org.open-scap.cpe.fedora:def:20: not applicable
Definition oval:org.open-scap.cpe.fedora:def:19: not applicable
Definition oval:org.open-scap.cpe.fedora:def:18: not applicable
Definition oval:org.open-scap.cpe.fedora:def:17: not applicable
Definition oval:org.open-scap.cpe.fedora:def:16: not applicable
Evaluation done.
libgcc_s.so.1 must be installed for pthread_cancel to work
libgcc_s.so.1 must be installed for pthread_cancel to work
libgcc_s.so.1 must be installed for pthread_cancel to work
```

I think that these messages were hidden before, because probes used to have redirected stderr to /dev/null. I have removed this recently in https://github.com/OpenSCAP/openscap/pull/365, so these error messages came into light.

jan-cerny:
Actually @isimluk has predicted future. 
> Removal of that dup2 in NDEBUG has potential to give us some surprises in future. However, I am ok accepting the risk.

ybznek:
I've successfully handled
```error: Failed to initialize NSS library``` I will add it do ***dI*** output

**Comment by ybznek** [comment]
```
libgcc_s.so.1 must be installed for pthread_cancel to work
```

[...]

PR #386, opened 2016-04-11
probes: Load libgcc_s.so.1 before chroot

ybznek:
Should help to avoid messages like
'libgcc_s.so.1 must be installed for pthread_cancel to work'
When target root doesn't contain this library

### DIFF
--- a/src/OVAL/probes/probe/main.c
+++ b/src/OVAL/probes/probe/main.c
@@ -43,6 +43,7 @@
 #include "input_handler.h"
 #include "probe-api.h"
 #include "option.h"
+#include <dlfcn.h>
 
 static int fail(int err, const char *who, int line)
 {
@@ -257,6 +258,7 @@ int main(int argc, char *argv[])
 			if (chdir(rootdir) != 0) {
 				fail(errno, "chdir", __LINE__ -1);
 			}
+			dlopen("libgcc_s.so.1", RTLD_NOW | RTLD_GLOBAL);
 			if (chroot(rootdir) != 0) {
 				fail(errno, "chroot", __LINE__ - 1);
 			}


ybznek:
https://github.com/OpenSCAP/openscap/issues/368

src/OVAL/probes/probe/main.c
```diff
@@ -257,6 +258,7 @@ int main(int argc, char *argv[])
 			if (chdir(rootdir) != 0) {
 				fail(errno, "chdir", __LINE__ -1);
 			}
+			dlopen("libgcc_s.so.1", RTLD_NOW | RTLD_GLOBAL);
```
mpreisler: I don't think this is very portable. We should be detecting the path as part of `./configure` instead of using a hard-coded path. Do all other platforms we support have this library with the same name? What if you compile with `clang`?

jan-cerny:
@ybznek I would like to have some things clarified. What happens due to this? Does it have an impact on the scan? Will your fix just remove some message on stderr? Or is this necessary for correct functionality of offline probe?

src/OVAL/probes/probe/main.c
ybznek: It is quite complicated scenario.
I was trying to find some other function to force to load right library before chroot, but without success.

It fail during pthread_cancel() I don't know clean way how to find what does the function need.
And we can have similar functions.

Code compiled using clang gives me same error ```libgcc_s.so.1 must be installed for pthread_cancel to work
```
But after fix it isn't able to find dlopen.

@mpreisler What do you recommend to set during configure?

ybznek:
@jan-cerny It practically only remove error messages.

I don't know behavior of chroot to system with different architecture/os from x86_64 to x86, arm, or bsd. Nobody is interested in it, I think.

In real we get these messages when we do chroot to wrong directory and threads cannot be cancelled using pthread_ functions. As I know these errors are only during termination of scan and probably don't influence results.

src/OVAL/probes/probe/main.c
mpreisler: That's probably because pthread depends on libgcc_s.so. On other platforms pthread may depend on something else.

I suggest we don't attempt to fix this and tolerate the error.

mpreisler:
I am against merging this in `maint-1.0`. It's a hacky workaround at best.

ybznek:
Replaced by https://github.com/OpenSCAP/openscap/pull/389

PR #387, opened 2016-04-11
probes/sysctl: Enable scan 'files' through different devices

ybznek:
/proc/sys/fs/binfmt_misc is on different device than rest of /proc/sys
(stat /proc/sys/fs/binfmt_misc). This commit enable to scan it too

More info https://github.com/OpenSCAP/openscap/issues/258


ybznek:
I am going to write some test, later.

jan-cerny:
@ybznek 
This  pull request addresses issue #258. It would be nice if the commit message references the issue.
The commit message could look like this, eg.:
```
Issue #258: Fix sysctl probe output inconsistency

The sysctl probe does not collect variable fs.binfmt_misc.status, while sysctl -a does it.
This commit fixes this problem by enabling scanning through different devices.
```
You will find many similar commits in the `git log`. In my opinion it helps orientation in the git.

The code change itself looks good to me. Thanks for investigating & fixing. I have tested it and it fixes issue #258 for me. ACK.

Please propose the test in a new pull request,

PR #388, opened 2016-04-12
sysctl - test

ybznek:
I've added check for deprecated parameters according to sysctl command.
Sysctl command check it in the same way since Mon Feb 13 21:21:43 - check only filename, not path.
https://github.com/OpenSCAP/openscap/issues/258

### DIFF
--- a/tests/probes/sysctl/Makefile.am
+++ b/tests/probes/sysctl/Makefile.am
@@ -17,9 +17,13 @@ TESTS = all.sh
 endif
 endif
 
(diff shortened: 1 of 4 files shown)


mpreisler:
We can't drop something we collected in the past in `maint-1.0`. What is the reason for removing what we collect? Does it hurt anything if we include deprecated things in the results?

Furthermore, you checked with latest sysctl, these things are probably not deprecated on sysctl versions on some of the platforms we support.

ybznek:
According to @mpreisler comment I have removed the commit and updated test.

PR #389, opened 2016-04-13
probes dynamic library preloading

ybznek:
Replaces https://github.com/OpenSCAP/openscap/pull/386
(It could be possible to divide it to more pull request, but it would delay merging.)

- I have deleted callback log function of rpm, because I didn't understand well function interface.
- libgcc_s is loaded before chroot due to call of pthread_cancel()
- rpm dynamic libraries are loaded from host, not from container as before
- ```disable_library_loading()``` I wasn't able so far to find right way how to implement it.
I've asked in mailing list and hope I will get some answer.

### DIFF
--- a/src/OVAL/probes/unix/linux/rpm-helper.c
+++ b/src/OVAL/probes/unix/linux/rpm-helper.c
@@ -27,7 +27,8 @@
 int rpmErrorCb (rpmlogRec rec, rpmlogCallbackData data)
 {
 	dE("RPM: %s", rpmlogRecMessage(rec));
-	return RPMLOG_EXIT; // don't perform default logging
+	return RPMLOG_DEFAULT;
+}
 
 void rpmLibsPreload()
 {
--- a/src/OVAL/probes/unix/linux/rpm-helper.c
+++ b/src/OVAL/probes/unix/linux/rpm-helper.c
@@ -28,4 +28,8 @@ int rpmErrorCb (rpmlogRec rec, rpmlogCallbackData data)
 {
 	dE("RPM: %s", rpmlogRecMessage(rec));
 	return RPMLOG_EXIT; // don't perform default logging
+
+void rpmLibsPreload()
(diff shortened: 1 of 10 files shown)


mpreisler:
Let's separate the preload and preventing shared library loading.

I am happy to merge the pthread and rpm library preloads, not so sure about the shared library loading yet. Could you please strip those commits? Let's do that in another PR.

ybznek:
I've removed last commit.

src/OVAL/probes/unix/linux/rpm-helper.c
```diff
@@ -24,8 +24,11 @@
 #include <config.h>
 #endif
 
-int rpmErrorCb (rpmlogRec rec, rpmlogCallbackData data)
+void rpmLibsPreload()
 {
-	dE("RPM: %s", rpmlogRecMessage(rec));
-	return RPMLOG_EXIT; // don't perform default logging
```
mpreisler: This is very drastic, could we keep the Cb and set it after we are done with the preload? We still want those messages, imagine the pain of debugging any problems with the RPM probes without those messages...

ybznek: API of the rpmlib is quite strange. We cannot replace handler by our handler. As I remember there is RPMLOG_DEFAULT - perform default logging or RPMLOG_EXIT. I understood the second value as don't perform default logging, but reality is not so. It exit everything.

We use something like:
```
if (rpmReadConfigFiles ((const char *)NULL, (const char *)NULL) != 0) {
                 dI("rpmReadConfigFiles failed: %u, %s.", errno, strerror (errno));
```

Problem is that rpm errors aren't always errors - when you scan debian - it is quite expected that we cannot read rpm database.

mpreisler:
OK, I am happy with the code now. Let's wait until Jenkins finishes, then I will merge this.

PR #404, opened 2016-05-02
RHBZ#1250072: Fix regression in loading DS session

jan-cerny:
Loading session for a DS file multiple times with different XCCDF IDs
didn't work. When a session for a data stream is created using Python API
and then it is loaded twice with different XCCDF IDs being set,
the session doesn't reflect the change of the XCCDF ID.
This commit fixes the regression by resetting the DS session
when a new XCCDF session is loaded. Therefore the IDs are resetted.


jan-cerny:
Sorry, wrong branch.

Issue #195, opened 2015-10-27
MS Windows Support 


isimluk:
Rationale: Current OpenSCAP community are POSIX enthusiasts. There is a demand for OpenSCAP on MS Windows platform.

Deliverable: OpenSCAP runs on Windows, executables are distributed, tests are set-up.

Note: Consider implementing Cygwin support in the ./run script

This is a big idea.

isimluk:
Some more info can be found in #196

jan-cerny:
I think some Windows support has been implemented in master branch. :smiley_cat:

isimluk:
That's beautiful sound to my old ears. :1st_place_medal: :+1: :tada:

PR #406, opened 2016-05-02
[Bugfix][WIN] Fix terminal output and colors for Windows systems

redhatrises:
- /r was creating weird output in Windows Command Prompt
- Windows Command Prompt does not use ANSI so use correct non-ANSI terminal colors
- Part of #195

### DIFF
--- a/utils/oscap-xccdf.c
+++ b/utils/oscap-xccdf.c
@@ -46,6 +46,10 @@
 #include <syslog.h>
 #endif
 
+#if defined(_WIN32)
+#include <windows.h>
+#endif
+
 #include "oscap-tool.h"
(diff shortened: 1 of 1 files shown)


redhatrises:
Just an FYI. I did get openscap to compile and run on windows without probes, but I realized that I hadn't updated my master with the latest commits. Once updated, openscap fails to build on Windows. 

This PR is a simple one that just makes the output look nice for Windows. To however is reviewing, I am very rusty at C.

mpreisler:
I'd prefer if the printing was abstracted out so that the `#ifdef` only appears once but this is good enough as an initial fix.

PR #407, opened 2016-05-02
Windows build fixes

redhatrises:
- Implement OSS version of flock and setenv that can be used only by MinGW.
- Use custom get_program_name function for WIN32.
- Fix various issues when using --disable-probes.
- Update Windows cross-compilation instructions.
- Update #195

### DIFF
--- a/docs/manual/manual.adoc
+++ b/docs/manual/manual.adoc
@@ -1385,6 +1385,8 @@ OpenSCAP for Windows:
 
 1) Install the cross-compiler & dependencies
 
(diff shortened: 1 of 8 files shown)


redhatrises:
Only issue I am now running into when compiling is this:

`oval_resModel.c:70:28: error: conflicting types for 'oval_results_model_new_with_probe_session'`

I actually did get openscap to compile and run on an older code version and only then realized that I was not running on an up-to-date master branch. Any recommendations for dealing with the above error? I am very rusty on C.

mpreisler:
ACK

PR #412, opened 2016-05-11
Add GConf2-devel to build dependencies list

jan-cerny:
This package is needed to build gconf probe.


jan-cerny:
All checks have failed ... This is very weird, because a change in README should not break any tests :-) This is another of many random fails. We encounter them quite often recently.

Issue #394, opened 2016-04-19
SCE results aren't embedded in XCCDF reports


AndreasKrause:
When evaluating xccdf with `--oval-results` OVAL result files (`*-oval.xml.result.xml`) are generated in current work dir. Similarily, when called with `--check-engine-results` SCE result files (`*.-sce.sh.result.xml`) appear.

However, when finally generating XCCDF reports, OVAL results are embedded in the HTML report file while SCE results aren't. This is reproducable on RHEL5+6, SLES11+12 (all with openscap-1.2.5).

mpreisler:
Are you using the <check-import> feature in the SCE checks? See https://www.open-scap.org/features/other-standards/sce/

AndreasKrause:
No, I'm referring to

> 2. use SCE result file
> In case there is no  <check-import import-name=”stdout” />  in the
> XCCDF (or the XSLT can’t find it for whatever reason), it will try to look for
> SCE result files and get their stdout element content.

Can't test with < check-import > right now since I'm on vacation, but I think I tested this as well and the result wasn't embedded either.

jan-cerny:
I confirm this issue still persists on current maint-1.2.
When I run it in two steps,
```
oscap xccdf eval --profile default --check-engine-results --results results.xml  sce_xccdf.xml
oscap xccdf generate report results.xml > report.html
```
 the SCE results are missing, as reported.

However, When I generate report directly during XCCDF evaluation, the SCE results are shown in report. Try this:
````
oscap xccdf eval --profile default --check-engine-results --results results.xml  --report report.html sce_xccdf.xml
```
The SCE results are there.

My opinion is that there is a bug in generating HTML reports from XCCDF in `oscap xccdf generate report` command. I think that it simply does not try to find whether `*.result.xml` files are available.

jan-cerny:
@AndreasKrause 
Does
`oscap xccdf generate report --sce-template "%.result.xml" results.xml > report.html`
work for you?

AndreasKrause:
@jan-cerny 
Ahoj, Jan! Strangely enough, this also only embeds oval results, no sce results. The Option --sce-template works for me as a synonym for --oval-template.

jan-cerny:
Hello, patches proposed in pull request https://github.com/OpenSCAP/openscap/pull/414 should fix this issue.

jan-cerny:
Fixed in https://github.com/OpenSCAP/openscap/pull/414

PR #414, opened 2016-05-18
Fix Issue #394 (SCE results aren't embedded in XCCDF reports) + test

jan-cerny:
This pull request tries to fix issues related to generating a HTML report form XCCDF with SCE checks.

This pull request contains three commits:
1) Add into man page a description for already existing, but undocumented option "--sce-template"
2) Add missing default template for SCE results
3) A new simple test checking whether SCE script stdout is shown in the HTML report

Thanks for review && comments.

### DIFF
--- a/tests/sce/bash_list.sh
+++ b/tests/sce/bash_list.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+ls -al /
+RET=$?
+
+if [[ $RET != 0 ]] ; then
+    exit $XCCDF_RESULT_FAIL
+else
+    exit $XCCDF_RESULT_PASS
+fi
--- a/tests/sce/bash_list.sh
+++ b/tests/sce/bash_list.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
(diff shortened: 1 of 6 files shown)


tests/sce/bash_list.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+ls -al /
+RET=$?
+
+if [[ $RET != 0 ]] ; then
+    exit $XCCDF_RESULT_FAIL
+else
+    exit $XCCDF_RESULT_PASS
+fi
```
mpreisler: This test  can be much simpler and not depend on the machine we are testing. Why is it listing the root dir?

tests/sce/bash_list.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
mpreisler: just `/bin/bash` should be fine

tests/sce/bash_list.sh
jan-cerny: It is just to have some strings on stdout. There can be whatever.

Issue #345, opened 2016-03-09
Improve error message when <oval:generator> is missing


isimluk:
## Reproducer
Have oval.xml file without <oval:generator> element.

```
$ oscap oval validate /tmp/oval.xml 
OpenSCAP Error: Invalid OVAL Definition ((null)) content in /tmp/oval.xml. [oscap_source.c:268]
```

jan-cerny:
Fixed in https://github.com/OpenSCAP/openscap/pull/416/

PR #416, opened 2016-05-19
Issue #345: Fix error message when <generator> is missing

jan-cerny:
When oscap reports that the document is invalid, it reports
against which schema it is invalid. However if the generator
element is missing the schema version is unknown.
This commit handles a situation when schema version is missing.


isimluk:
:cherry_blossom: :hibiscus: :bee:

PR #418, opened 2016-05-26
add opensuse 42.1

nevion:


mpreisler:
[test this]

mpreisler:
Any reason why this is targeted at OpenSCAP master and not `maint-1.2`?

nevion:
I am on maint-1.2 branch locally and pushed that to github... for some reason the PR becomes 77 commits when I have maint-1.2 selected as the local and remote targets.  It went down to one with master.

redhatrises:
IMO this also needs to map to the office NIST CPE which is `cpe:/o:novell:leap:42.1`

mpreisler:
I agree with @redhatrises 

Could you please fix that @nevion ? Don't worry about the target branch, I can just merge it manually to `maint-1.2`.

ybznek:
Jenkins, re[test this please].

ybznek:
```+ ind-def_environmentvariable58_test.xml                      [ FAIL ]```
After manual rerun of tests it PASS.

nevion:
@redhatrises @mpreisler  I don't understand why this would be the only one using the NIST CPE.  Looking at all neighboring entries, it is different.  Why should it change now?  Why not change the others to use that CPE as well?  Would these CPE variants break compatibility with anything?  Please provide a justification as I don't like breaking flow as a newcomer, with only one entry using the "new way"  with all these "old way" entries laying around.

Never the less I've incorporated your desired change and fixed the issue with the branch (botched rebase maybe at some point, I guess - not that I recall)

mpreisler:
Merged manually as f3c324d168c0cd850e82fc05db2244adc8feb2d6

If the other CPEs are not using the NIST CPE scheme that is a bug. It is hard to change that because other tools rely on those CPE IDs. We still should adhere to the NIST CPE scheme for newer additions.

Issue #419, opened 2016-05-31
oscap-docker - python2, ignore parameters, print always help


ybznek:
I've just tried oscap-docker on RHEL7.3 (Python 2.7.5)

It ignore every parameters and print help always.

It is broken due to this change
https://github.com/OpenSCAP/openscap/commit/c00f3c44b7a64b814e765265fe6941c3f6f90cf5
but it doesn't seem that problem is in python/strange python API.

Before any deeper investigation; quick fix could be - run this line only with python3 and not with python2.

**We should also check why we haven't noticed it earlier.**

dahaic:
Hello, the reason is change of defaults priority in python during 2.7 evolution :)

https://bugzilla.redhat.com/show_bug.cgi?id=1340036

ybznek:
@dahaic Ah, thank you :)

ybznek:
Fixed by https://github.com/OpenSCAP/openscap/commit/fa334865b1eb24c6eabfe4c80a339c84ffc2d250

PR #420, opened 2016-05-31
oscap-docker: Fix argparse.set_default()

ybznek:
This is workaround. Argparse has different behavior with different python versions.

We have to call *parser.set_defaults()* only with python 3.x.
Across python2.x can have the function different behavior.

Issue: https://github.com/OpenSCAP/openscap/issues/419


ybznek:
After discussion with @dahaic, I will improve this PR. Do not merge, now.

ybznek:
@mpreisler will have better solution.

mpreisler:
Just for the record this was fixed by fa334865b1eb24c6eabfe4c80a339c84ffc2d250

Issue #426, opened 2016-06-10
*-cve scan on non-rhel target returns HTTP Error 404: Not Found


ybznek:
```
Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/oscap_docker_python/get_cve_input.py", line 67, in _fetch_single
    resp = urllib.urlopen(_url)
  File "/usr/lib64/python3.4/urllib/request.py", line 161, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python3.4/urllib/request.py", line 469, in open
    response = meth(req, response)
  File "/usr/lib64/python3.4/urllib/request.py", line 579, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib64/python3.4/urllib/request.py", line 507, in error
    return self._call_chain(*args)
  File "/usr/lib64/python3.4/urllib/request.py", line 441, in _call_chain
    result = func(*args)
  File "/usr/lib64/python3.4/urllib/request.py", line 587, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 404: Not Found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/oscap-docker", line 109, in <module>
    args.func()
  File "/usr/bin/oscap-docker", line 44, in cve_scan
    result = OS.scan_cve(self.args.scan_target, self.unknown_args)
  File "/usr/lib/python3.4/site-packages/oscap_docker_python/oscap_docker_util.py", line 231, in scan_cve
    fetch._fetch_single(dist)
  File "/usr/lib/python3.4/site-packages/oscap_docker_python/get_cve_input.py", line 71, in _fetch_single
    .format(url_error))
Exception: Unable to fetch CVE inputs due to HTTP Error 404: Not Found
```

I think, we should return something like "not based on rhel".

PR #427, opened 2016-06-10
oscap-docker: *-cve scan of non-rhel fix

ybznek:
Fix:
https://github.com/OpenSCAP/openscap/issues/426

Changes:
- Print error message when target is not rhel based
- unmount container when exception error occur

### DIFF
--- a/utils/oscap_docker_python/oscap_docker_util.py
+++ b/utils/oscap_docker_python/oscap_docker_util.py
@@ -220,22 +220,25 @@ class OscapScan(object):
         # Remeber actual mounted fs in 'rootfs'
         chroot = os.path.join(_tmp_mnt_dir, 'rootfs')
 
-        # Figure out which RHEL dist is in the chroot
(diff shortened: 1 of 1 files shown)


utils/oscap_docker_python/oscap_docker_util.py
```diff
@@ -220,22 +220,29 @@ def scan_cve(self, image, scan_args):
...
-        # Scan the chroot
-        sys.stdout.write(self.helper._scan_cve(chroot, dist, scan_args))
+            # Fetch the CVE input data for the dist
+            fetch = getInputCVE(self.tmp_dir)
 
-        # Clean up
-        self.helper._cleanup_by_path(_tmp_mnt_dir)
-        self._remove_mnt_dir(mnt_dir)
+            # TODO
+            # This should probably be in a try/except
```
mpreisler: Fix the comments.

Issue #424, opened 2016-06-08
During check of fileextendedattribute oscap freezes


ybznek:
**How to reproduce:**
- `cp probe_test_fileextendedattribute.oval.xml to /`
- `setfattr /probe_test_fileextendedattribute.oval.xml -n user.foo -v barbar` - important
- `oscap oval eval --results /dev/stdout --verbose DEVEL --verbose-log-file /dev/stderr ./probe_test_fileextendedattribute.oval.xml`

I haven't checked it with different paths & etc...

**mount**
```
$ mount | grep ' / '
/dev/mapper/fedora_dhcp--10--40--3--88-root on / type ext4 (rw,relatime,seclabel,data=ordered)
```

**oscap-version**
1.2.9 from package or upstream

**probe_test_fileextendedattribute.oval.xml**
```
<?xml version="1.0" encoding="UTF-8"?>
<oval_definitions
	xsi:schemaLocation="http://oval.mitre.org/XMLSchema/oval-definitions-5"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5"
	xmlns:oval-def="http://oval.mitre.org/XMLSchema/oval-definitions-5"
	xmlns:oval="http://oval.mitre.org/XMLSchema/oval-common-5"
	xmlns:linux="http://oval.mitre.org/XMLSchema/oval-definitions-5#linux"
	xmlns:ind="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent"
	xmlns:unix="http://oval.mitre.org/XMLSchema/oval-definitions-5#unix" >

	<generator>
		<oval:schema_version>5.11</oval:schema_version>
		<oval:timestamp>2016-06-19T22:19:00+02:00</oval:timestamp>
	</generator>

	<definitions>
		<definition id="oval:my:def:1" version="1" class="miscellaneous">
			<metadata>
				<title>Test fileextended</title>
				<description></description>
			</metadata>
			<criteria>
				<criterion test_ref="oval:my:tst:1" />
			</criteria>
		</definition>
	</definitions>

	<tests>
		<fileextendedattribute_test version="1" id="oval:my:tst:1" check_existence="only_one_exists" check="all" comment="true" xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5#unix">
			<object object_ref="oval:1:obj:1"/>
			<state state_ref="oval:1:ste:1"/>
[...]

PR #433, opened 2016-06-13
probe extended attribute fix

ybznek:
https://github.com/OpenSCAP/openscap/issues/424
Should fix:
- `while (i < xattr_buflen && xattr_buf[i] != '\0') ++i;`  - When loop was going through `\0` separeted buffer, one time index `i` started to point to character `\0` and it never went after that character
- attributes in results didn't contain trailing \0
- replace goto by do while


ybznek:
missing dD() in maint-1.0 uaa...

ybznek:
test this please again

PR #442, opened 2016-06-21
Fix autoconf issues

jan-cerny:
Changes made recently in src/OVAL/probes/Makefile.am caused that
./confgen.sh removed check macros for header files that are needed
for rpm probes from configure.ac.
This commit fixes the syntax of respective Makefile.am and also
includes freshly generated configure.ac.


ybznek:
Doesn't dpkginfo need fix too?

jan-cerny:
@ybznek it looks like that

Issue #408, opened 2016-05-03
openscap compile problems on RHEL5 and SLES11


AndreasKrause:
When compiling openscap on RHEL5 or SLES11 problems occur with rpm related probes:

    [  313s]   CC       probe_rpminfo-rpminfo.o
    [  313s] In file included from unix/linux/rpminfo.c:63:
    [  313s] unix/linux/rpm-helper.h:77: error: expected declaration specifiers or '...' before 'rpmlogCallbackData'
    [  313s] unix/linux/rpminfo.c: In function 'probe_init':
    [  313s] unix/linux/rpminfo.c:275: warning: passing argument 1 of 'rpmlogSetCallback' from incompatible pointer type
    [  313s] unix/linux/rpminfo.c:275: error: too many arguments to function 'rpmlogSetCallback'
    [  313s] unix/linux/rpminfo.c: In function 'collect_rpm_files':
    [  313s] unix/linux/rpminfo.c:346: warning: implicit declaration of function 'rpmfiNew'
    [  313s] unix/linux/rpminfo.c:346: warning: assignment makes pointer from integer without a cast
    [  313s] unix/linux/rpminfo.c:348: warning: implicit declaration of function 'rpmfiNext'
    [  313s] unix/linux/rpminfo.c:350: warning: implicit declaration of function 'rpmfiFN'
    [  313s] unix/linux/rpminfo.c:350: warning: assignment makes pointer from integer without a cast
    [  313s] unix/linux/rpminfo.c:361: warning: implicit declaration of function 'rpmfiFree'
    [  313s] make[3]: *** [probe_rpminfo-rpminfo.o] Error 1
    [  313s] make[3]: Leaving directory `/usr/src/packages/BUILD/openscap-1.2.9/src/OVAL/probes'
    [  313s] make[2]: *** [all-recursive] Error 1
    [  313s] make[2]: Leaving directory `/usr/src/packages/BUILD/openscap-1.2.9/src/OVAL/probes'
    [  313s] make[1]: *** [all-recursive] Error 1
    [  313s] make[1]: Leaving directory `/usr/src/packages/BUILD/openscap-1.2.9'
    [  313s] make: *** [all] Error 2
    [  313s] error: Bad exit status from /var/tmp/rpm-tmp.20198 (%build)

Seems to be related to commits earlier this year (in April: https://github.com/OpenSCAP/openscap/commits/maint-1.2/src/OVAL/probes/unix/linux) when these probes were restructured.

[...]

PR #443, opened 2016-06-21
Issue #408: Fix compilation on older systems

jan-cerny:
Function rpmLogSetCallback has changed its prototype in RPM 4.6 API.
Before it had only one parameter, now it has two parameters.
Older version (RPM 4.4.x) is shipped in RHEL5, therefore we must use
conditional compilation to get the rpm probes compiled on RHEL5.


mpreisler:
Fixes #408

PR #449, opened 2016-06-22
Valgrind_test: Add free-fill, malloc-fill

ybznek:
If we use memory in bad way, this should increase possibility of test fail during valgrind test. I hope it will helps us to find problems.

Memory after malloc and free will be set to "UUUU..."


ybznek:
test this please

jan-cerny:
@ybznek Great idea! Thanks! ACK
:smile:  :cake:

PR #450, opened 2016-06-23
Remove callback on RPM < 4.6

jan-cerny:
Recently we have merged a fix to compile the rpm probes
on maint-1.0 on systems with older versions of rpmlib.
However, the reporting function  is not present there,
so it does not make any sense to set the logging callback,
because it has no effect. This commit removes the callback
on those systems.


jan-cerny:
Compilation fo maint-1.0 was supposed to be fixed in PR #443 . Unfortunately it didn't work in `./configure --enable-debug`. Sorry for that and thanks @ybznek for discovering the issue,

ybznek:
I think, we could use `rpmlogMessage()` with old librpm. But it is very old version, and oscap was working years without special log callback function.

I will merge it.

ybznek:
Oh, I have clicked on wrong button Test was passing before I have accidentally closed it.

Issue #447, opened 2016-06-22
Segmentation fault in file probe


jan-cerny:
We have seen a segmentation fault in file probe while testing pull request #440 on Jenkins.
The segfault happened on tests tests/API/OVAL/unittests/test_deprecated_def.sh.

Backtrace:

```
#0  SEXP_ID_v_callback (sexp=sexp@entry=0x9898989898989898, pair=pair@entry=0x7f0332ec0d30) at ../../../../../src/OVAL/probes/SEAP/sexp-ID.c:74
#1  0x00007f033a5267b2 in SEXP_ID_v (s=s@entry=0x9898989898989898) at ../../../../../src/OVAL/probes/SEAP/sexp-ID.c:110
#2  0x0000000000408b0b in probe_icache_worker (arg=0x193f670) at ../../../../../src/OVAL/probes/probe/icache.c:179
#3  0x00007f0338299dc5 in start_thread () from /lib64/libpthread.so.0
#4  0x00007f0337990ced in clone () from /lib64/libc.so.6
```

Please notice the very weird address 0x9898989898989898

ybznek:
https://github.com/OpenSCAP/openscap/pull/454 I hope it will help us at least a little bit

ybznek:
https://jenkins.open-scap.org/job/openscap-pull-requests/532/console

verbose output
[verbose.txt](https://github.com/OpenSCAP/openscap/files/339016/verbose.txt)

ybznek:
Another backtrace
```
Core was generated by `/home/jenkins/workspace/openscap-pull-requests/openscap-1.0.11/_build/src/OVAL/'.
Program terminated with signal 11, Segmentation fault.
#0  SEXP_ID_v_callback (sexp=0x3434343434343434, pair=pair@entry=0x7f81ddb6ad40) at ../../../../../src/OVAL/probes/SEAP/sexp-ID.c:74
74	        SEXP_val_dsc(&v_dsc, sexp->s_valp);
(gdb) where
#0  SEXP_ID_v_callback (sexp=0x3434343434343434, pair=pair@entry=0x7f81ddb6ad40) at ../../../../../src/OVAL/probes/SEAP/sexp-ID.c:74
#1  0x00007f81e51acc22 in SEXP_ID_v (s=<optimized out>) at ../../../../../src/OVAL/probes/SEAP/sexp-ID.c:110
[...]

PR #454, opened 2016-06-23
probes/file: Add debug output & modify test

ybznek:
We have often randomly failed file probe test.
It should help us to find the issue

### DIFF
--- a/src/OVAL/probes/probe/icache.c
+++ b/src/OVAL/probes/probe/icache.c
@@ -176,6 +176,8 @@ static void *probe_icache_worker(void *arg)
                         /*
                          * Compute item ID
                          */
+			dD("pair address: %"PRIu64, (uint64_t) pair);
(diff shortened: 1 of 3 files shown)


mpreisler:
Looks like a minor change, why `maint-1.2`?

ybznek:
@mpreisler Verbose mode doesn't work with maint-1.0 (dD doesn't exist there), and we would need to compile with --enable-debug to see logged messages - it could possibly change tests results. Reading of these log debug files  is more complicated.

I hope we will revert these changes after fixing of issue, currently we don't know where is the problem with file probe.

jan-cerny:
Yes, the verbose mode is not present in the maint-1.0 branch. The file probe has been crashing more and more often recently. Hopefully this will help us.
@mpreisler To clarify, @ybznek created this because we caught a segmentation fault recently in this piece of code,  https://github.com/OpenSCAP/openscap/issues/447

jan-cerny:
LGTM

PR #455, opened 2016-06-23
Fix build on master with disabled probes

jan-cerny:


src/OVAL/public/oval_definitions.h
```diff
@@ -251,6 +251,7 @@ typedef enum {
 oval_family_t oval_subtype_get_family(oval_subtype_t);
 const char *oval_operator_get_text(oval_operator_t);
 const char *oval_subtype_get_text(oval_subtype_t);
+const char *oval_subtype_to_str(oval_subtype_t);
```
mpreisler: Let's mark this as deprecated and replace all occurrences with the get_text variant.

mpreisler:
We can do the deprecation later.

src/OVAL/public/oval_definitions.h
jan-cerny: I agree, perhaps we could deprecate and replace this function in maint-1.2.
I have noticed that there is  also `oval_subtype_parse` which looks like it can share its code with  `oval_str_to_subtype`.

PR #458, opened 2016-06-24
Deprecate oval_subtype_to_str

jan-cerny:
Occurrences of oval_subtype_to_str are replaced by oval_subtype_to_text

These functions are two implementations of the same thing.

### DIFF
--- a/src/OVAL/oval_enumerations.c
+++ b/src/OVAL/oval_enumerations.c
@@ -740,6 +740,13 @@ const char *oval_subtype_get_text(oval_subtype_t subtype)
 	}
 }
 
+bool oval_subtype_is_valid(oval_subtype_t subtype)
+{
+	const char *str = oval_subtype_get_text(subtype);
+	return (str != NULL && strcmp(str, _invalid) != 0);
+}
+
+
 static const struct oscap_string_map OVAL_RESULT_MAP[] = {
 	{OVAL_RESULT_TRUE, "true"},
 	{OVAL_RESULT_FALSE, "false"},
--- a/src/OVAL/oval_probe.c
+++ b/src/OVAL/oval_probe.c
@@ -175,14 +175,7 @@ static void __init_once(void)
 
 const char *oval_subtype_to_str(oval_subtype_t subtype)
 {
-        oval_subtypedsc_t *d;
-
-        __init_once();
-
-        d = oscap_bfind(OSCAP_GSYM(__s2n_tbl), __s2n_tbl_count, sizeof(oval_subtypedsc_t), &subtype,
-                        (int(*)(void *, void *))__s2n_tbl_cmp);
-
-        return (d == NULL ? NULL : d->name);
+	return oval_subtype_get_text(subtype);
 }
 
 oval_subtype_t oval_str_to_subtype(const char *str)
--- a/src/OVAL/oval_sexp.c
+++ b/src/OVAL/oval_sexp.c
@@ -671,13 +671,16 @@ int oval_state_to_sexp(void *sess, struct oval_state *state, SEXP_t **out_sexp)
 	size_t buflen;
 	const char *subtype_name;
 	struct oval_state_content_iterator *contents;
+	oval_subtype_t subtype;
 
-        subtype_name = oval_subtype_to_str(oval_state_get_subtype(state));
+	subtype = oval_state_get_subtype(state);
 
-	if (subtype_name == NULL) {
-		dI("FAIL: unknown subtype: %d", oval_state_get_subtype(state));
+	if (!oval_subtype_is_valid(subtype) && subtype != OVAL_SUBTYPE_SYSINFO) {
+		dI("FAIL: unknown subtype: %d", subtype);
 		return (-1);
 	}
+
+	subtype_name = oval_subtype_get_text(subtype);
 
 	buflen = snprintf(buffer, sizeof buffer, "%s_state", subtype_name);
 	_A(buflen < sizeof buffer);
--- a/src/OVAL/probes/probe-api.c
+++ b/src/OVAL/probes/probe-api.c
@@ -1405,12 +1405,12 @@ SEXP_t *probe_item_create(oval_subtype_t item_subtype, probe_elmatr_t *item_attr
         bool    multiplied;
         int     value_i, multiply;
 
-        subtype_name = oval_subtype_to_str(item_subtype);
-
-        if (subtype_name == NULL) {
+        if (!oval_subtype_is_valid(item_subtype) && item_subtype != OVAL_SUBTYPE_SYSINFO) {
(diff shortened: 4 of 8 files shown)


ybznek:
It seems, it is related https://github.com/OpenSCAP/openscap/pull/455, isn't it?

src/OVAL/probes/probe-api.c
```diff
@@ -1405,7 +1405,7 @@ SEXP_t *probe_item_create(oval_subtype_t item_subtype, probe_elmatr_t *item_attr
         bool    multiplied;
         int     value_i, multiply;
 
-        subtype_name = oval_subtype_to_str(item_subtype);
+        subtype_name = oval_subtype_get_text(item_subtype);
```
ybznek: If `item_subtype` is really invalid, `oval_subtype_get_text()` will return `"**INVALID**"` and condition bellow will not recognize invalid subtype and will continue

```
const char *oval_subtype_get_text(oval_subtype_t subtype)
{
...
	if (map) {
		return oval_enumeration_get_text(map, subtype);
	} else {
		oscap_seterr(OSCAP_EFAMILY_OVAL, "Invalid OVAL family.");
		return _invalid;
	}
```

Please, tell me why it cannot happen or why the behavior wasn't changed with your changes.
If are there any hidden assumptions, maybe I would like to have some `assert()` in the code if it is possible.

Similar "issue" https://github.com/OpenSCAP/openscap/pull/458/files#diff-4fcc494830030cae0c83ae842a047a49R675

jan-cerny: You are true, I haven't considered this.

I have looked more into this. I have noticed that `oval_enumeration_get_text(map, subtype)` can return NULL, so if item_subtype is invalid, it can return either `NULL` or `"**INVALID**"` (both of the possibilities dependent on the reason of invalidity).

I suggest returning NULL always

jan-cerny:
@ybznek  Yes, it is.

src/OVAL/probes/probe-api.c
ybznek: Always NULL is a good idea, but please check if it isn't in public API. If it is, we shouldn't modify it.

We could also introduce function like this `isValid(str) {return  str != NULL && str!=_invalid}`

jan-cerny: Yes, that could be a problem, the  function is used also in oval_object_get_name and other places, so it might be difficult to identify what everything needs to changed to not introduce a regression.

I think I will try to introduce oval_subtype_is_valid and use the new function in all the relevant places.

jan-cerny:
I have amended this pull request to deal with possible `"**INVALID**"` value returned by `oval_subtype_get_text()`.

src/OVAL/oval_enumerations.c
```diff
@@ -740,6 +740,13 @@ const char *oval_subtype_get_text(oval_subtype_t subtype)
 	}
 }
 
+bool oval_subtype_is_valid(oval_subtype_t subtype)
+{
+	const char *str = oval_subtype_get_text(subtype);
+	return (str != NULL && strcmp(str, _invalid) != 0);
```
ybznek: I think, you don't need to use strcmp, you only need compare pointers.

src/OVAL/oval_sexp.c
```diff
@@ -671,14 +671,17 @@ int oval_state_to_sexp(void *sess, struct oval_state *state, SEXP_t **out_sexp)
...
 	const char *subtype_name;
 	struct oval_state_content_iterator *contents;
+	oval_subtype_t subtype;
 
-        subtype_name = oval_subtype_to_str(oval_state_get_subtype(state));
+	subtype = oval_state_get_subtype(state);
 
-	if (subtype_name == NULL) {
-		dI("FAIL: unknown subtype: %d", oval_state_get_subtype(state));
+	if (!oval_subtype_is_valid(subtype) && subtype != OVAL_SUBTYPE_SYSINFO) {
```
ybznek: Why `subtype != OVAL_SUBTYPE_SYSINFO`?

src/OVAL/oval_probe.c
```diff
@@ -175,14 +175,7 @@ static void __init_once(void)
...
 {
-        oval_subtypedsc_t *d;
-
-        __init_once();
-
-        d = oscap_bfind(OSCAP_GSYM(__s2n_tbl), __s2n_tbl_count, sizeof(oval_subtypedsc_t), &subtype,
-                        (int(*)(void *, void *))__s2n_tbl_cmp);
-
-        return (d == NULL ? NULL : d->name);
+	return oval_subtype_get_text(subtype);
```
ybznek: Maybe you should add something like this
```
return oval_subtype_is_valid(subtype) ? oval_subtype_get_text(subtype) : NULL;
```
because currently somebody who use public API can rely on `NULL value`.

src/OVAL/oval_enumerations.c
ybznek: The function forces you have to resolve subtype text usually twice.

Why not to check only `char* subtype_text` ?

mpreisler:
@jan-cerny Any ETA on wrapping this up? There is a risk of this going stale.

jan-cerny:
@mpreisler Hi Martin. this was intended only as a minor enhancement and there were others more priority tasks.

The problem here is that there are 2 similar functions, both of them public, however with a slightly different behavior. `oval_subtype_to_str` is located in `oval_probe.c`, which is not suitable location and causes probles when building with disabled probles. On the other hand, the `oval_subtype_get_text` sometimes returns "**INVALID**"  and sometimes NULL, whereas the other returns always NULL.
There are some other built-in assumptions.

We don't need to fix it now, but we should handle that before we start using master.

Possible solution could be to deprecate both of the function, introduce a new one which will not have any side effect, and remove all the occurrences of them.

ybznek:
`oval_subtype_to_str()` is not in my `/usr/include/openscap`, so I expect that it isn't published and we can delete it. Or am I wrong?

According to my assumption, I would move `oval_subtype_to_str()` to right place in the code (make it public) and deprecate `oval_subtype_get_text()`. - I don't like "INVALID" as return value.

Inner code for convertion from subtype -> str  is very different between `oval_subtype_to_str()` and `oval_subtype_get_text()`, we should choose better one.

ybznek:
What are our plans? We could merge it somehow to master. What do you think?

jan-cerny:
Yeah, master seems a better place for this kind of fix or improvement.
I think we should close it and I will be able to propose the fix to master branch. I think that it will be easier to do because I will not have to keep the API and implement workarounds to not break the API.
@ybznek Thanks for the idea!

mpreisler:
@jan-cerny You suggested closing this but your comment sounds like you are still thinking about it. Are there any updates? Would be great to clean the PR queue.

jan-cerny:
Ok

Issue #281, opened 2016-01-04
Remediation broken when non standard characters captured on the scripts output


isimluk:
After the remediation is applied, oscap puts `stdout` of the remediation script back to XCCDF file. The problem is when the `stdout` contains non standard characters. Then the XCCDF XML creation fails like

```
Title^M Enable the NTP Daemon
Rule^M  service_ntpd_enabled
Ident^M CCE-27093-4
Ident^M DISA FSO RHEL-06-000247
/tmp/oscap.86u1ve/xccdf-result.xml:27283: parser error : PCDATA invalid Char value 27
      <message severity="info">Starting ntpd: ^[[60G[^[[0;32m  OK  ^[[0;39m]&#13;
                                              ^
/tmp/oscap.86u1ve/xccdf-result.xml:27283: parser error : PCDATA invalid Char value 27
      <message severity="info">Starting ntpd: ^[[60G[^[[0;32m  OK  ^[[0;39m]&#13;
                                                    ^
/tmp/oscap.86u1ve/xccdf-result.xml:27283: parser error : PCDATA invalid Char value 27
      <message severity="info">Starting ntpd: ^[[60G[^[[0;32m  OK  ^[[0;39m]&#13;
                                                                 ^
OpenSCAP Error: Conversion of the string "" to an integer (64 bits) failed: Invalid argument [oval_cmp.c:113]
Conversion of the string "" to an integer (64 bits) failed: Invalid argument [oval_cmp.c:113]
Conversion of the string "" to an integer (64 bits) failed: Invalid argument [oval_cmp.c:113]
Conversion of the string "" to an integer (64 bits) failed: Invalid argument [oval_cmp.c:113]
Conversion of the string "" to an integer (64 bits) failed: Invalid argument [oval_cmp.c:113]
Could not parse the XML document '/tmp/oscap.86u1ve/xccdf-result.xml' [oscapxml.c:429]
Result^M        fixed
```

See #251 for more info.

szechyjs:
:+1: I get this problem all the time when I try to generate a report from my result xml.

```
[root@localhost ~]# oscap xccdf generate report localhost.localdomain-ssg-results.xml > localhost.localdomain-ssg-results.html
localhost.localdomain-ssg-results.xml:22270: parser error : PCDATA invalid Char value 27
      <message severity="info">Stopping automount:         [  OK  ]&#
                                                   ^
[...]

PR #459, opened 2016-06-27
Issue #281: Fix invalid characters in remediation results

jan-cerny:
We include stdout of a remediation script to the XCCDF results.
But the output of the remediation script can contain characters
that are forbidden in XML documents, eg. most of the non-printable
characters. We need to escape or remove them from the string
before we include it to the XML document. Libxml provides a
function to get rid of those characters.

### DIFF
--- a/tests/API/XCCDF/unittests/test_remediation_invalid_characters.xccdf.xml
+++ b/tests/API/XCCDF/unittests/test_remediation_invalid_characters.xccdf.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Benchmark xmlns="http://checklists.nist.gov/xccdf/1.2" id="xccdf_moc.elpmaxe.www_benchmark_test">
+  <status>accepted</status>
+  <version>1.0</version>
+  <Rule selected="true" id="xccdf_moc.elpmaxe.www_rule_1">
+    <title>Ensure that file exists and it is not executable</title>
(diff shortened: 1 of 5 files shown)


jan-cerny:
I'll try provide a simple test.

jan-cerny:
I have created a very simple test for remediation producing non-printable characters on stdout. The test tests whether the generated XCCDF results document is a valid XCCDF document (it wasn't before this fix).

tests/API/XCCDF/unittests/test_remediation_invalid_characters.xccdf.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Benchmark xmlns="http://checklists.nist.gov/xccdf/1.2" id="xccdf_moc.elpmaxe.www_benchmark_test">
+  <status>accepted</status>
+  <version>1.0</version>
+  <Rule selected="true" id="xccdf_moc.elpmaxe.www_rule_1">
+    <title>Ensure that file exists and it is not executable</title>
+    <fix system="urn:xccdf:fix:script:sh">
+      echo -e "\a \b \r \x1F"
+      touch test_file
```
ybznek: I think, you should remove this file using Makefile.am, because it left there. We will see it in jenkins result.

jan-cerny:
test_file should be fixed

ybznek:
LGTM, Thank you!

Issue #461, opened 2016-06-27
New perl-XML-XPath support


ybznek:
We use `perl-XML-XPath` in tests for reading xml files using xpath. From version 1.34 (Fedora24 & other distros) it quietly started to use different interface. This package doesn't provide pkgconfig `.pc` file.

We can provide some simple wrapper to make calls compatible, but how to recognize which version system use? We could guess version from output. Any better ideas?

```
/tmp/XML-XPath-1.33
$ ./blib/script/xpath
Usage:
./blib/script/xpath [filename] query

	If no filename is given, supply XML on STDIN.
```

```
/tmp/XML-XPath-1.34
$ ./blib/script/xpath
Name "XML::XPath::ParseParamEnt" used only once: possible typo at ./blib/script/xpath line 40.
Usage:
./blib/script/xpath [options] -e query [-e query...] [filename...]

If no filenames are given, supply XML on STDIN. You must provide at
least one query. Each supplementary query is done in order, the
previous query giving the context of the next one.

Options:

-q quiet, only output the resulting PATH.
-s suffix, use suffix instead of linefeed.
-p postfix, use prefix instead of nothing.
-n Don't use an external DTD.
```

iankko:
@ybznek 
> ...
We can provide some simple wrapper to make calls compatible, but how to recognize which version system use? We could guess version from output. Any better ideas?
...

You mean how to find the version of the XML::XPath perl module, that is installed on the system?
If that's the question, each (well defined) Perl module (on CPAN) should have ```VERSION``` defined, thus
something like the following should work (below is the output on my F22 system. Didn't try on F24, but this should work there too).

```
$ perl -MXML::XPath -e 'print $XML::XPath::VERSION."\n";'
1.13
```

Just my 2c.

ybznek:
I haven't know that way to get version. It seems that it could be useful, I will try it.

Thank you much, @iankko!

jan-cerny:
Fixed in https://github.com/OpenSCAP/openscap/pull/464

PR #462, opened 2016-06-28
tests/process58: Use specified xpath

ybznek:
Part of solution of https://github.com/OpenSCAP/openscap/issues/461.


jan-cerny:
Nice catch! Thanks!

PR #463, opened 2016-06-28
Fix systemd test on fedora24

ybznek:
"-.mount Wants" isn't presented on Fedora24, so part of test
was separated and now it should be tested only on "supported systems"

mostly copy & paste


jan-cerny:
Reviewed and tested. ACK

PR #464, opened 2016-06-28
xpath wrapper

ybznek:
Should fix https://github.com/OpenSCAP/openscap/issues/461

If newer version of xpath is installed, wrapper will translate parameters from old format

Similar solution is in arch linux package https://aur.archlinux.org/packages/openscap


ybznek:
@NonerKao If this will be merged, it could remove need of xpath patch in AUR package.
Any suggestions?

NonerKao:
Great!
I think we can leave the AUR package as is for a while before the next release.

jan-cerny:
@ybznek Great! Thank you! ACK.

Issue #467, opened 2016-06-29
Add test: file probe & unicode filenames


ybznek:
I would like to see behavior of file probe/sexp/oscap with very crazy filenames touch `"$(echo -e '\x01\x20')"`.

Linux filesystems allow usage of almost every filename (except \0,...).

We should create these files during test, I don't know if git can handle them.

Maybe it isn't too important to can work with such files, but oscap should crash during their processing.

https://github.com/OpenSCAP/openscap/issues/447#issuecomment-229316426

jan-cerny:
Test was introduced in https://github.com/OpenSCAP/openscap/pull/468/commits/2bd56625acdc6ef5ae431b84b722d18c2370b01a

PR #468, opened 2016-06-29
test/file: Add test to strange filenames (missing fix)

ybznek:
Only test - without fix - should fail currently.

It should be test for https://github.com/OpenSCAP/openscap/issues/467, but it doesn't crash.

But I have found another problem; we don't escape correctly filenames in xml result

```
Entity: line 3457: parser error : PCDATA invalid Char value 27
            <unix-sys:filepath>/tmp/tmp.5mogjwJMK2/filename_</unix-sys:filepa
                                                             ^
```

### DIFF
--- a/tests/probes/file/Makefile.am
+++ b/tests/probes/file/Makefile.am
@@ -10,4 +10,7 @@ TESTS_ENVIRONMENT= \
 
 TESTS = test_probes_file.sh
 
-EXTRA_DIST += test_probes_file.sh test_probes_file.xml
+EXTRA_DIST += test_probes_file.sh \
+	test_probes_file.xml \
+	test_probes_file_filename.xml
+
--- a/tests/probes/file/test_probes_file.sh
+++ b/tests/probes/file/test_probes_file.sh
@@ -31,11 +31,87 @@ function test_probes_file {
     if [ -f $RF ]; then
 	verify_results "def" $DF $RF 13 && verify_results "tst" $DF $RF 204
 	ret_val=$?
-    else 
+    else
 	ret_val=1
     fi
 
     return $ret_val
+}
+
+function test_probes_file_filenames {
+
+	probecheck "file" || return 255
+
+	local ret_val=0
+	local DF="$srcdir/test_probes_file_filename.xml"
+	result="results.xml"
+	files_dir=$(mktemp -d)
+	DF_INJECTED=$(mktemp)
+
+	echo "Files dir:	${files_dir}"
+	echo "Content file:	${DF_INJECTED}"
+
+	# create ascii files - should be processed correctly
+	pushd "$files_dir" && {
+		for i in {1..127};
+		do
+			[ $i == 47 ] && continue ; # '/' cannot be in filename
+			hex=\\x$(printf "%x" "$i")
+			touch "$(printf filename_ok_${i}_$hex$hex$hex)"
+		done
+		popd
+	}
+
+	# inject real path to content
+	sed "s;<!--injected-path -->;${files_dir};" "$DF" > $DF_INJECTED
+
+	$OSCAP oval eval --results $result $DF_INJECTED || ret_val=1
+	$OSCAP oval validate $result || ret_val=1
+
+	# some of files maybe weren't successfully created
+	# (possibly not supported filenames by filesystem)
+	# get real count of them
+	filesCount=$(find "$files_dir" -type f | wc -l)
+
+	assert_exists 1 '//results//criterion' || ret_val=1
+	assert_exists 1 '//results//criterion[@result="true"]' || ret_val=1
+	assert_exists "$filesCount" '//unix-sys:file_item' || ret_val=1
+
+	rm $DF_INJECTED
+	rm -rf "$files_dir"
+
+	return $ret_val
+}
+
+function test_probes_file_invalid_utf8 {
+
+	probecheck "file" || return 255
+
+	local ret_val=0
+	local DF="$srcdir/test_probes_file_filename.xml"
+	result="results.xml"
+	files_dir=$(mktemp -d)
+	DF_INJECTED=$(mktemp)
+
+	echo "Files dir:	${files_dir}"
+	echo "Content file:	${DF_INJECTED}"
+
+	# known invalid utf-8 sequence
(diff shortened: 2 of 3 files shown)


ybznek:
```
E: lt-probe_file: Unable to match regex pattern, pcre_exec() returned error: -10.
 [lt-probe_file(30608):probe_worker(7f08e7fff700):oval_cmp_basic.c:150:strregcomp]
E: lt-probe_file: Unable to match regex pattern, pcre_exec() returned error: -10.
 [lt-probe_file(30608):probe_worker(7f08e7fff700):oval_cmp_basic.c:150:strregcomp]
E: lt-probe_file: Unable to match regex pattern, pcre_exec() returned error: -10.
 [lt-probe_file(30608):probe_worker(7f08e7fff700):oval_cmp_basic.c:150:strregcomp]
E: lt-probe_file: Unable to match regex pattern, pcre_exec() returned error: -10.
```
error code:
PCRE_ERROR_BADUTF8        (-10)

mpreisler:
Do not merge, this currently fails.

mpreisler:
#473 was merged so this should now pass

test this please

ybznek:
https://github.com/OpenSCAP/openscap/pull/473 escapes output in xml result, but current test fails due to pcre issues. Some filenames aren't matched by due to non-ascii characters - we get `E: lt-probe_file: Unable to match regex pattern, pcre_exec() returned error: -10.`

mpreisler:
I don't want to merge this until the tests are passing. Merging something that causes all tests to fail would make merging anything else difficult.

ybznek:
I was trying to investigate  `E: lt-probe_file: Unable to match regex pattern, pcre_exec() returned error: -10`.

- `-10` means `PCRE_ERROR_BADUTF8 or PCRE_ERROR_BADUTF16 or PCRE_ERROR_BADUTF32`  - all these constants has same value
- option PCRE_NO_UTF8_CHECK exists - but behavior of pcre with the option and invalid string is undefined - it doesn't help us
- I don't know how/if we can fix it

I could remove check from the test, but the same issue could be with another probes.

Maybe possible issue, could be hiding some configuration values:
```
option = very_dangerous # <invalid char>
```
and our regexp will not match it

Any ideas?

mpreisler:
This tells us that we are passing invalid utf8 in, looks like a real bug. I think the solution is never passing invalid utf8 around rather than providing some option that disables the checks.

mpreisler:
> We don't have currently standard way to do something like TDD. We can merge only passing test or skip test using "return 255" - there is a risk that nobody will enable it, later. Any ideas how to deal with such situations in future?

We don't do TDD, we do a lot of regression testing instead. In our model we make sure the tests don't start failing when we introduce new features, code, fixes, ... Introducing TDD into a project that wasn't started as a TDD project is very challenging, I suggest we keep our current workflow.

ybznek:
@mpreisler We cannot ensure valid utf8 input text what comes from users data. After discussion with @jan-cerny We should rather add some message to output and set result to error.

Due to complicated debugging, I will append backtrace here
```
#0  strregcomp (test_str=0x7fffe00037d0 "filename_\377\377\377", pattern=<optimized out>) at oval_cmp_basic.c:151
#1  oval_string_cmp (state=<optimized out>, syschar=0x7fffe00037d0 "filename_\377\377\377", operation=<optimized out>) at oval_cmp_basic.c:194
#2  0x00007ffff7b4cadb in probe_ent_cmp_string (val1=val1@entry=0x7fffe0003790, val2=val2@entry=0x7fffe0000de0, op=op@entry=OVAL_OPERATION_PATTERN_MATCH) at probes/probe/entcmp.c:161
#3  0x00007ffff7b4cfcf in probe_ent_cmp_single (op=OVAL_OPERATION_PATTERN_MATCH, sysent=0x7fffe0000de0, state_data_type=OVAL_DATATYPE_STRING, state_ent=0x7fffe0003790) at probes/probe/entcmp.c:207
#4  probe_ent_cmp (ent=ent@entry=0x7fffe0000920, val2=val2@entry=0x7fffe0000de0) at probes/probe/entcmp.c:266
#5  0x00007ffff7b4d2bb in probe_entobj_cmp (ent_obj=0x7fffe0000920, val=val@entry=0x7fffe0000de0) at probes/probe/entcmp.c:444
#6  0x00007ffff7b4aa93 in oval_fts_read_recurse_path (ofts=0x7fffe0000ef0) at probes/oval_fts.c:1102
#7  oval_fts_read (ofts=ofts@entry=0x7fffe0000ef0) at probes/oval_fts.c:1293
#8  0x0000000000405ad1 in probe_main (ctx=ctx@entry=0x7fffee700da0, mutex=<optimized out>) at ../../../../../src/OVAL/probes/unix/file.c:508
#9  0x0000000000407880 in probe_worker (probe=0x7fffffffdad0, msg_in=<optimized out>, ret=0x7fffee700e1c) at ../../../../../../src/OVAL/probes/probe/worker.c:952
#10 0x00000000004073e7 in probe_worker_runfn (arg=0x7fffe8000d50) at ../../../../../../src/OVAL/probes/probe/worker.c:56
#11 0x00007ffff564b5ca in start_thread (arg=0x7fffee701700) at pthread_create.c:333
#12 0x00007ffff4f6cead in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```

ybznek:
I have push --force'ed the test

tests/probes/file/Makefile.am
```diff
@@ -10,4 +10,7 @@ TESTS_ENVIRONMENT= \
 
 TESTS = test_probes_file.sh
 
-EXTRA_DIST += test_probes_file.sh test_probes_file.xml
+EXTRA_DIST = test_probes_file.sh \
```
jan-cerny: should be += because the variable is already set on line 1

ybznek: Thank you.

jan-cerny:
Sysctl probe failed, should be unrelated to this PR, previous builds runs OK.

Great job! LGTM. ACK.

jan-cerny:
test this please

ybznek:
Backtrace https://github.com/OpenSCAP/openscap/issues/447#issuecomment-232098532

test this please

ybznek:
test this please

Issue #478, opened 2016-07-04
valgrind-check - tests/API/XCCDF/variable_instance failure


ybznek:
https://jenkins.open-scap.org/job/openscap-maint-valgrind-check/58/consoleFull
```
10279 Segmentation fault      (core dumped) valgrind $valgrind_args $oscap_program "$@"
```

ybznek:
PR fix some issue, if we will found something using valgrind, we can open new issue

PR #481, opened 2016-07-04
Fix usage of probe_session after free

ybznek:
https://github.com/OpenSCAP/openscap/issues/478

probe_session were referenced from two places. 
- oval_agent_session
- oval_agent_session->res_model (https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/OVAL/results/oval_resultTest.c#L933)

When we create new session in oval_agent_reset_session, res_model will still use old pointer to free'ed memory.

`oval_probe_session_reinit` could be in OSCAP_HIDDEN in maint-1.2oval_probe_session_reinit

---
Another solution is add setter to res_model to set probe_session, but IMHO, my solution is ~~clean~~ cleaner.

### DIFF
--- a/src/OVAL/oval_probe_session.c
+++ b/src/OVAL/oval_probe_session.c
@@ -118,13 +118,11 @@ static void __init_once(void)
         return;
 }
 
-oval_probe_session_t *oval_probe_session_new(struct oval_syschar_model *model)
+static void oval_probe_session_init(oval_probe_session_t *sess, struct oval_syschar_model *model)
 {
-        oval_probe_session_t *sess;
         void *handler_arg;
         register size_t i;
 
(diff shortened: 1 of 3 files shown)


src/OVAL/oval_probe_session.c
```diff
@@ -148,19 +146,37 @@ oval_probe_session_t *oval_probe_session_new(struct oval_syschar_model *model)
...
+void oval_probe_session_free(oval_probe_session_t *sess)
 {
 	if (sess == NULL) {
 		dE("Invalid session (NULL)");
 		return;
 	}
 
-        oval_phtbl_free(sess->ph);
-        oval_pext_free(sess->pext);
-        oscap_free(sess);
```
mpreisler: why was the `oscap_free(sess)` line removed here?

PR #483, opened 2016-07-11
Add a missing build dependency to README

jan-cerny:
The header file dbus.h provided by the dbus-devel package is required
to build systemd probes, we should list the package in our build
dependencies list.


ybznek:
@jan-cerny Could you please check/ensure, that our jenkins slaves have this package?

jan-cerny:
@ybznek I have checked and ensured all our slaves.

PR #484, opened 2016-07-11
probes/sysctl.c: Fix typo

ybznek:
ftp -> fts


jan-cerny:
Nice catch :grinning:  ACK

PR #485, opened 2016-07-11
probes/oval_fts: Return error if regex issue occurs

ybznek:
When we cannot match filename by regex (due to non-utf8-characters) we used to ignore it, but now we will set result to error.

Maybe we could return `message` to oval result, but I haven't found any current probe what does it. So I don't know if it is a good idea. What do you think?

---
Error: `For example, if there was an error returned by an api when trying to determine if an object exists on a system.` https://oval.mitre.org/language/version5.6/ovalresults/documentation/oval-results-schema.html

### DIFF
--- a/src/OVAL/probes/independent/filehash.c
+++ b/src/OVAL/probes/independent/filehash.c
@@ -262,7 +262,7 @@ int probe_main (probe_ctx *ctx, void *mutex)
 		return (PROBE_EFATAL);
         }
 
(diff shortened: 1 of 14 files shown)


mpreisler:
> Maybe we could return message to oval result, but I haven't found any current probe what does it. So I don't know if it is a good idea. What do you think?

I agree, even this patch alone is an improvement so I will merge it but we should look into reporting the issue so it can be debugged.

PR #491, opened 2016-07-13
call signal in mutex

ybznek:
I have replaced goto -> do{}while without fixing of indentation. I expect more our changes and don't want make conflicts.


jan-cerny:
lgtm ack
Thank you.

PR #496, opened 2016-07-15
oval_probe_session.c: Add missing 'static'

ybznek:


jan-cerny:
Nice catch! Thank you.

PR #498, opened 2016-07-18
Refactoring: Remove misleading indexing

ybznek:
I think it is misleading, because "+1" means automatically `+1 * size array element`, Previous code will not work e.g. with array of int.

Don't you think?


mpreisler:
Yeah, you are right. I have no idea what I was thinking when I wrote that...

PR #501, opened 2016-07-19
Fix format string of debug message

ybznek:


ybznek:
Sorry, I always forget to assign milestone.

PR #503, opened 2016-08-04
Initial Wind River Linux support

radzy:

### DIFF
--- a/cpe/openscap-cpe-oval.xml
+++ b/cpe/openscap-cpe-oval.xml
@@ -442,6 +442,20 @@
                         <criterion comment="openSUSE 42.1 is installed" test_ref="oval:org.open-scap.cpe.opensuse:tst:421"/>
                   </criteria>
             </definition>
+            <definition class="inventory" id="oval:org.open-scap.cpe.wrlinux:def:8" version="1">
+                  <metadata>
+                        <title>WR Linux 8</title>
+                        <affected family="unix">
+                            <platform>WR Linux 8</platform>
+                        </affected>
+                        <reference ref_id="cpe:/o:windriver:wrlinux:8" source="CPE"/>
+                        <description>The operating system installed on the system is WR Linux 8</description>
+                  </metadata>
+                  <criteria>
+                        <criterion comment="Installed operating system is part of the unix family" test_ref="oval:org.open-scap.cpe.wrlinux:tst:1"/>
+                        <!-- <criterion comment="WR Linux 8 is installed" test_ref="oval:org.open-scap.cpe.wrlinux:tst:8"/> -->
+                  </criteria>
+            </definition>
       </definitions>
       <tests>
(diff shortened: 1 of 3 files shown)


cpe/openscap-cpe-oval.xml
```diff
@@ -442,6 +442,20 @@
...
+                        <title>WR Linux 8</title>
+                        <affected family="unix">
+                            <platform>WR Linux 8</platform>
+                        </affected>
+                        <reference ref_id="cpe:/o:windriver:wrlinux:8" source="CPE"/>
+                        <description>The operating system installed on the system is WR Linux 8</description>
+                  </metadata>
+                  <criteria>
+                        <criterion comment="Installed operating system is part of the unix family" test_ref="oval:org.open-scap.cpe.wrlinux:tst:1"/>
+                        <!-- <criterion comment="WR Linux 8 is installed" test_ref="oval:org.open-scap.cpe.wrlinux:tst:8"/> -->
```
mpreisler: Why is this commented out?

As it is, all operating systems of unix family will behave as if they were WR Linux in openscap CPE.

src/OVAL/probes/unix/runlevel.c
```diff
@@ -327,6 +332,15 @@ static int is_solaris (void)
         return (stat ("/etc/release", &st)   == 0);
 }
 
+static int is_wrlinux (void)
+{
+	struct stat st;
+	int retval;
+	if ( retval = stat ("/etc/os-release", &st) != 0 ) return(retval);
```
mpreisler: what is the logic behind this? WR Linux must not have /etc/os-release? Isn't the second line enough to make a good guess?

mpreisler:

radzy:
@mpreisler : The logic behind the /etc/os-release was that I wanted it to require /etc/os-release to be present.  This was for historical reasons, and I'm happy to remove it.  (Done in my local in-progress copy, so it will show up in the next upload.)  The check for /etc/wrlinux-release was added later (because we didn't already have /etc/wrlinux-release, so I had to create it.  But since I did that, os-release is no longer relevant.)

radzy:
@mpreisler : The commented out test is an error.  I'm re-validating the latest version, with it in place.

radzy:
@mpreisler : Incidentally, w.r.t. os-release/wrlinux-release, my internal reviewers suggest that I open and parse the /etc/wrlinux-release file to get the version number, instead of using the RPM metadata.  Nobody else does that.  Would this be an acceptable strategy ?

mpreisler:
> @mpreisler : Incidentally, w.r.t. os-release/wrlinux-release, my internal reviewers suggest that I open and parse the /etc/wrlinux-release file to get the version number, instead of using the RPM metadata. Nobody else does that. Would this be an acceptable strategy ?

I am not sure how it works on WR Linux but on RHEL based distros we trust the rpm metadata more than contents of the file. However if you think it's preferable to parse the file you can. Check out the textfilecontent OVAL test.

radzy:
Will upload a new PR when ready.

PR #505, opened 2016-08-11
[V2] Initial Wind River Linux support

radzy:
Changes since [V1] : 
use file_test to check for the presence of /etc/wrlinux-release and textfilecontent_test to check the WR Linux version.
For runlevel check, use stat("/etc/wrlinux-release"...) as the only check.


radzy:
Hmmm.  Looks like I did something wrong here.

PR #506, opened 2016-08-11
[v2] Initial Wind River Linux support

radzy:
Changes since [V1] :
use file_test to check for the presence of /etc/wrlinux-release and textfilecontent_test to check the WR Linux version.
For runlevel check, use stat("/etc/wrlinux-release"...) as the only check.

### DIFF
--- a/cpe/openscap-cpe-dict.xml
+++ b/cpe/openscap-cpe-dict.xml
@@ -142,4 +142,12 @@
                  We assume that the productivity is equal to RHEL5. We are unable to write a better OVAL check. -->
             <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.rhel:def:5</check>
       </cpe-item>
+      <cpe-item name="cpe:/o:windriver:wrlinux">
+            <title xml:lang="en-us">Wind River Linux all versions</title>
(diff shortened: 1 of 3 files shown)


radzy:
> mpreisler commented 6 days ago (in PR #503)
>  
> > @mpreisler : Incidentally, w.r.t. os-release/wrlinux-release,
> > my internal reviewers suggest that I open and parse the
> > /etc/wrlinux-release file to get the version number, instead of
> > using the RPM metadata. Nobody else does that. Would this be an
> > acceptable strategy ?
>  
> I am not sure how it works on WR Linux but on RHEL based distros
> we trust the rpm metadata more than contents of the file. However
> if you think it's preferable to parse the file you can. Check
> out the textfilecontent OVAL test.

If I read the logs correctly, it looks like the textfilecontent
test is causing a test failure because it's deprecated.  I wasn't
able to get textfilecontent54 working.  Any suggestions about
what needs to be done differently for textfilecontent54 than for
textfilecontent ?

jan-cerny:
@radzy You are right, the textfilecontent_test has been deprecated by OVAL standard and replaced by textfilecontent54_test.
In the newer version, there is no `line` element, neither in textfilecontent54_object nor textfilecontent54_state. 
You should use `pattern` element in the object and `text` element in the state.

Hopefully the specification can explain that better than me :grinning: See http://oval.mitre.org/language/version5.11/ovaldefinition/documentation/independent-definitions-schema.html#textfilecontent54_test

radzy:
Thanks @jan-cerny !  That's a very helpful link!  I hadn't realized that there was an element dictionary or that it had this level of detail.  Much easier than trying to discover the usage from examples in the test cases and reading C code.

PR #509, opened 2016-08-15
DO NOT REVIEW OR MERGE : Initial Wind River Linux support

radzy:


radzy:
I'm having trouble getting a local "make check" run to pass.  So I thought I would see if there's something different in my environment that prevents it.  No sense wasting time doing a review right now, but you're welcome to if you want.  :-)

radzy:
Sigh.  Old commit.

PR #510, opened 2016-08-15
Initial Wind River Linux support

radzy:

### DIFF
--- a/cpe/openscap-cpe-dict.xml
+++ b/cpe/openscap-cpe-dict.xml
@@ -142,4 +142,12 @@
                  We assume that the productivity is equal to RHEL5. We are unable to write a better OVAL check. -->
             <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.rhel:def:5</check>
       </cpe-item>
+      <cpe-item name="cpe:/o:windriver:wrlinux">
+            <title xml:lang="en-us">Wind River Linux all versions</title>
+            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.wrlinux:def:1</check>
+      </cpe-item>
+      <cpe-item name="cpe:/o:windriver:wrlinux:8">
+            <title xml:lang="en-us">WR Linux 8</title>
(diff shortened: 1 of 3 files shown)


radzy:
I'm having trouble getting a local "make check" run to pass. So I thought I would see if there's something different in my environment that prevents it. No sense wasting time doing a review right now, but you're welcome to if you want. :-)

radzy:
The failing test on my system is marked as "skip" in the jenkins output.  "test sysctl probe that collects everything".

radzy:
Looks good to go.  :-)

cpe/openscap-cpe-dict.xml
```diff
@@ -142,4 +142,12 @@
                  We assume that the productivity is equal to RHEL5. We are unable to write a better OVAL check. -->
             <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.rhel:def:5</check>
       </cpe-item>
+      <cpe-item name="cpe:/o:windriver:wrlinux">
+            <title xml:lang="en-us">Wind River Linux all versions</title>
+            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.wrlinux:def:1</check>
```
jan-cerny: Here you reference `oval:org.open-scap.cpe.wrlinux:def:1`, but there is no definition with that ID in `cpe/openscap-cpe-oval.xml`. Every check referenced from the cpe_items must be defined in `cpe/openscap-cpe-oval.xml`

jan-cerny:
You can:
* either remove that generic cpe_item for all versions (that's not so good solution)
* or implement oval:org.open-scap.cpe.wrlinux:def:1 (that's much more better solution)

Hint: You can reuse your "generic" definition in the "version-8-specific" definition. OVAL language has a `extend_definition` element, which can be used instead of a `criterion` element and which allows you to include one definition in another.
See http://oval.mitre.org/language/version5.11/ovaldefinition/documentation/oval-definitions-schema.html#ExtendDefinitionType

Then, `oval:org.open-scap.cpe.wrlinux:def:1` will consist of these criteria:
* test installed os is part of the unix family
* test WR Linux is installed

And  `oval:org.open-scap.cpe.wrlinux:def:8` will consist of:
* `<extend_definition definition_ref="oval:org.open-scap.cpe.wrlinux:def:1"/>`
* test WR linux is version 8

I hope this helps.

radzy:
Thanks!  That's very helpful.

PR #517, opened 2016-09-05
xccdf_session.c: Fix free of dirname

ybznek:
https://bugzilla.redhat.com/show_bug.cgi?id=1367896

According to man page, dirname can return pointer to original
path. So we should not call free on returned value.

(But we should free result of oscap_strdup)


ybznek:
maint 1.0 doesn't segfault and I haven't found similar code in the branch.

ybznek:
process58. test this please

ybznek:
file probe: test this please

jan-cerny:
Makes sense to me. Just waiting for CI.

mpreisler:
ACK

PR #521, opened 2016-09-06
xccdf_policy_remediate.c: Disable run of ansible remediation

ybznek:


mpreisler:
Would it be easier to just not add it to the map for now?

ybznek:
@mpreisler heh.... shame on me...

ybznek:
I am not very happy with such change - it can break something what was already working.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -104,7 +104,7 @@ static const char *_get_supported_interpret(const char *sys, const struct _inter
 		{"urn:xccdf:fix:script:csh",		"/bin/csh"},
 		{"urn:xccdf:fix:script:tclsh",		"/usr/bin/tclsh"},
 		{"urn:xccdf:fix:script:javascript",	"/usr/bin/js"},
-		{"urn:xccdf:fix:script:ansible",	"/usr/bin/ansible-playbook"},
+		//{"urn:xccdf:fix:script:ansible",	"/usr/bin/ansible-playbook"}, // see commit message
```
mpreisler: proper comment please

jan-cerny:
@mpreisler @ybznek I don't like this solution. It removes a feature that is present in OpenSCAP for last few releases, since OpenSCAP 1.2.7. The commit message says: *We cannot run ansible snippet easily, now - due to missing header.* But that's a problem of SSG, in SSG the headers are currently missing. There can exist other SCAP content that has ansible remediations including headers.

Can we let the user select which remediations he prefers and make the bash default?

mpreisler:
The issue here is not that we can't run the ansible playbook. It's that doing that makes **no sense** in any sensible usage of ansible. If the user chooses ansible remediations they are using ansible to manage their machines. If those two are true the last thing the user wants is to run our openscap generated playbook on their machine as it is. In other words this is caused by us not thinking the whole use-case through in the past.

We have no reports of people using this feature. The XCCDF fix system attribute is our invention and AFAIK there is no content out there in the wild using this. Go ahead and search the interwebs for "urn:xccdf:fix:script:ansible". You will only find mentions in OpenSCAP codebase and SSG.

In my opinion your reservations are valid but the risk is only theoretical. Real users are very unlikely to be affected by this.

mpreisler:
Of course we could keep the current "urn:xccdf:fix:script:ansible" as it is and add "urn:xccdf:fix:script:ansible_not_retarded". I think shutting down a use-case that makes no sense is a good thing though.

jan-cerny:
@mpreisler OK, I understand.

jan-cerny:
@ybznek Lovely! ACK.

PR #523, opened 2016-09-12
Added support to not to export system characteristics

ybznek:
We can export full/thin results using OVAL directives, but we always export system characteristics, too.

Added support to not export system characteristics.

More info in commit message.

How it should work, you can see from attached test.

### DIFF
--- a/src/OVAL/oval_session.c
+++ b/src/OVAL/oval_session.c
@@ -74,6 +74,7 @@ struct oval_session {
 	} reporter;
 
 	bool validation;
+	bool export_sys_chars;
 	bool full_validation;
 };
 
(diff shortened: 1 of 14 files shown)


jan-cerny:
Hi, I have tested your pull request and it works for me as expected. The code looks good for me.

One nitpick though: I had a look into man page and there is always only "syschar", not "syschars". I suggest changing the parameter name to "--no-syschar". But this is not a blocker.

Please don't forget to describe the new feature in the man page.

Thank you very much that you have also included a test. :birthday: :gift: :smiley:

jan-cerny:
I confirm that results produced by oscap oval eval --results results.xml --without-syschar are valid.

ACK.

PR #527, opened 2016-09-15
Fix export of overrided oscap_text

yuumasato:
This change allows attributes marked as overriders by
oscap_text_set_override() to be exported.

Exportation of an overrider attribute is conditioned by override_given.
But override_given is only set when an imported attribute has override
capability set. So, when an oscap_text was marked as overrider by
oscap_text_set_set_override() the attribute was never exported.

Related to RHBZ#1320194.

### DIFF
--- a/src/common/text.c
+++ b/src/common/text.c
@@ -146,7 +146,7 @@ xmlNode *oscap_text_to_dom(struct oscap_text *text, xmlNode *parent, const char 
 
 	if (text->lang)
 		xmlNodeSetLang(text_node, BAD_CAST text->lang);
-	if (text->traits.can_override && text->traits.override_given)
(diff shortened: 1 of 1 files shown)


mpreisler:
Makes sense to me. The worst that can happen here is that we export the attribute when not necessary and that is not a major problem.

ACK but will wait for Jenkins.

mpreisler:
I think the test file in parser unit test needs to be changed, override=false needs to be added so that the exported is exactly the same as imported. Then it should pass.

yuumasato:
Restricted the export of override attribute to true values only. There is no need to export override="true".

mpreisler:
OK, ACK.

Issue #525, opened 2016-09-13
XCCDF rules with @role="unscored" should give a result of "informational"


jan-cerny:
XCCDF specification, in section "Check processing"  (page 59, table 39), describes how the results of a check should be processed. It says that:
> If this <xccdf:Rule> has a @role attribute whose value is “unscored”,
give this <xccdf:Rule> a result of “informational”. Otherwise, give this <xccdf:Rule> a result
appropriate to the result returned by the checking engine.

However, OpenSCAP does not follow the specification in this particular requirement. I have discovered that OpenSCAP propagates the results to the xccdf:rule-result/result element always, regardless the role  atrribute in Result element was set.

We should follow the specification when creating XCCDF results. If a XCCDF rule has role="unscored", OpenSCAP should return a result of "informational" for that rule.

jan-cerny:
Fixed in https://github.com/OpenSCAP/openscap/pull/533

PR #532, opened 2016-09-19
Fix results of XCCDF rules with @role="unscored".

yuumasato:
When a XCCDF rule has @role="unscored", perform all the checks but mark
the result as XCCDF_RESULT_INFORMATIONAL.

https://github.com/OpenSCAP/openscap/issues/525

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy.c
+++ b/src/XCCDF_POLICY/xccdf_policy.c
@@ -1079,6 +1079,10 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
 	}
 	if ((xccdf_test_result_type_t) ret == XCCDF_RESULT_NOT_CHECKED)
 		message = "None of the check-content-ref elements was resolvable.";
+
+	if (role == XCCDF_ROLE_UNSCORED)
+        ret = XCCDF_RESULT_INFORMATIONAL;
+
 	xccdf_check_content_ref_iterator_free(content_it);
 	oscap_list_free(bindings, (oscap_destruct_func) xccdf_value_binding_free);
 	/* Negate only once */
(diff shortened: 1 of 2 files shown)


src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -1079,6 +1079,10 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
 	}
 	if ((xccdf_test_result_type_t) ret == XCCDF_RESULT_NOT_CHECKED)
 		message = "None of the check-content-ref elements was resolvable.";
+
+	if (role == XCCDF_ROLE_UNSCORED)
+        ret = XCCDF_RESULT_INFORMATIONAL;
+
```
mpreisler: openscap uses tabs for indents.

mpreisler:
Shouldn't this go into `maint-1.0`? I will defer to @jan-cerny because he has more knowledge about this ticket.

jan-cerny:
@yuumasato 
I hope this comment will help you to do the fix correctly.
Feel free to ask if something is unclear :smile:

Martin is true, this patch should go to maint-1.0, because:
* XCCDF_ROLE_UNSCORED is present in maint-1.0
* the code that you change is same in maint-1.0
* this is a bug fix, not a new feature (in other words, the specification has been requiring this for a long time, we just haven't noticed that)

I have already tried and tested your patch applied on the top of maint-1.0 and it works well.

You can get the commit to the correct branch by doing something like:
```
git checkout maint-1.0
git cherry-pick maint-1.2-role-unscored-result-informational
```
Then replace spaces by tabs in your patch for xccdf_policy.c and modify the commit.
You can use `git commit --amend`.

Finally. please close this and open a new pull request against maint-1.0 branch.

yuumasato:
Thank you, @jan-cerny.

Issue #526, opened 2016-09-13
Remote resources mapping by catalog element in datastreams doesn't work


jan-cerny:
The remote resources support in datastreams is incomplete in OpenSCAP in my opinion.

I have run into this issue when examining test content from the SCAP validation test suite released by NIST. 
The following snippet is copy-pasted for NIST validation test suite, requirement R2910.
Explanation of the problem will follow below.

```
  <data-stream id="scap_gov.nist_datastream_r2910-datastream" scap-version="1.2" timestamp="2016-05-25T14:00:00" use-case="CONFIGURATION">
    <dictionaries>
      <component-ref id="scap_gov.nist_cref_r2910-cpe-dictionary.xml" xlink:href="#scap_gov.nist_comp_r2910-cpe-dictionary.xml">
        <cat:catalog>
          <cat:uri name="r2910-cpe-oval.xml" uri="#scap_gov.nist_cref_r2910-cpe-oval.xml"/>
        </cat:catalog>
      </component-ref>
    </dictionaries>
    <checklists>
      <component-ref id="scap_gov.nist_cref_r2910-xccdf.xml" xlink:href="#scap_gov.nist_comp_r2910-xccdf.xml">
        <cat:catalog>
          <cat:uri name="r2910-oval.xml" uri="#scap_gov.nist_cref_r2910-oval.xml"/>
        </cat:catalog>
      </component-ref>
    </checklists>
    <checks>
      <component-ref id="scap_gov.nist_cref_r2910-oval.xml" xlink:href="http://scap.nist.gov/validation/downloads/r2910-external-oval-component.xml"/>
      <component-ref id="scap_gov.nist_cref_r2910-cpe-oval.xml" xlink:href="#scap_gov.nist_comp_r2910-cpe-oval.xml"/>
    </checks>
  </data-stream>
```

XCCDF Rules in this datastream reference the OVAL check as eg. ```<xccdf:check-content-ref href="r2910-oval.xml" name="oval:nist.validation.variable:def:11"/>``` .  

However the "r2910-oval.xml" is not a local file, but it is a remote resource and it needs to be downloaded from the internet.

I will now explain how OpenSCAP should figure out that  "r2910-oval.xml" is a remote resource.

1. It should process the `<checklist>` element to find an XCCDF checklist.
2. It should find an XCCDF component there.
3. That component contains an catalog. The purpose of a catalog is to map URIs to datastream IDs. The @name attribute is the
[...]

PR #534, opened 2016-09-21
Add support for external (http) components

ybznek:
https://github.com/OpenSCAP/openscap/issues/526

### DIFF
--- a/src/DS/public/ds_sds_session.h
+++ b/src/DS/public/ds_sds_session.h
@@ -29,6 +29,7 @@
 #include "oscap.h"
 #include "oscap_source.h"
 #include "scap_ds.h"
+#include "xccdf_download_cb.h"
 
 /**
  * The ds_sds_session is structure tight closely to oscap_source.
@@ -171,6 +172,16 @@ int ds_sds_session_set_target_dir(struct ds_sds_session *session, const char *ta
  * @param session The Source DataStream session to reset
  */
 void ds_sds_session_reset(struct ds_sds_session *session);
+
+/**
+ * Set properties of remote content.
+ * @memberof ds_sds_session
+ * @param session The Source DataStream Session
+ * @param allowed Whether is download od remote resources allowed in this session (defaults to false)
+ * @param callback used to notify user about download proceeds. This might be safely set
+ * to NULL -- ignoring user notification.
+ */
+void ds_sds_session_set_remote_resources(struct ds_sds_session *session, bool allowed, download_progress_calllback_t callback);
 
 /**
  * Returns HTML representation of selected checklist in form of OpenSCAP guide.
--- a/src/DS/public/ds_sds_session.h
+++ b/src/DS/public/ds_sds_session.h
@@ -174,10 +174,10 @@ int ds_sds_session_set_target_dir(struct ds_sds_session *session, const char *ta
 void ds_sds_session_reset(struct ds_sds_session *session);
 
 /**
- * Set properties of remote content.
+ * Set property of remote content.
  * @memberof ds_sds_session
  * @param session The Source DataStream Session
- * @param allowed Whether is download od remote resources allowed in this session (defaults to false)
+ * @param allowed Whether is download of remote resources allowed in this session (defaults to false)
  * @param callback used to notify user about download proceeds. This might be safely set
  * to NULL -- ignoring user notification.
  */
--- a/src/DS/sds.c
+++ b/src/DS/sds.c
@@ -308,6 +308,30 @@ static int ds_sds_dump_component(const char* external_file, const char* componen
 		return ret;
 }
 
+static int ds_sds_dump_file_component(const char* external_file, const char* component_id, struct ds_sds_session *session, const char *target_filename_dirname, const char *relative_filepath)
+{
+	int ret = 0;
+
+	struct oscap_source *source_file = load_referenced_source(session, external_file);
+	xmlDoc *doc = oscap_source_get_xmlDoc(source_file);
+
+	if (doc == NULL) {
+		ret = -1;
+		goto cleanup;
+	}
+
+	xmlNodePtr inner_root = ds_sds_get_component_root_by_id(doc, component_id);
+
+	if (ds_sds_register_component(session, doc, inner_root, component_id, target_filename_dirname, relative_filepath) != 0) {
+		ret = -1;
+		goto cleanup;
+	}
+
+	cleanup:
+		oscap_source_free(source_file);
+		return ret;
+}
+
 int ds_sds_dump_component_ref_as(const xmlNodePtr component_ref, struct ds_sds_session *session, const char* sub_dir, const char* relative_filepath)
 {
 	char* cref_id = (char*)xmlGetProp(component_ref, BAD_CAST "id");
--- a/src/DS/sds.c
+++ b/src/DS/sds.c
@@ -229,6 +229,18 @@ static int ds_sds_register_sce(struct ds_sds_session *session, xmlNodePtr compon
 	return ret;
 }
 
+static int ds_sds_register_xmlDoc(struct ds_sds_session *session, xmlDoc* doc, xmlNodePtr component_inner_root, const char *relative_filepath)
+{
+	xmlDoc *new_doc = ds_doc_from_foreign_node(component_inner_root, doc);
+	if (new_doc == NULL) {
+		return -1;
+	}
+
+	struct oscap_source *component_source = oscap_source_new_from_xmlDoc(new_doc, relative_filepath);
+
+	return ds_sds_session_register_component_source(session, relative_filepath, component_source);
+}
+
 static int ds_sds_register_component(struct ds_sds_session *session, xmlDoc* doc, xmlNodePtr component_inner_root, const char* component_id, const char* target_filename_dirname, const char* relative_filepath)
 {
 	int ret = 0;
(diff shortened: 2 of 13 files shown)


ybznek:
After this PR it will crash without ```--fetch-remote-resources``` option if datastream reference remote file.

I don't know direct/clean way how to skip components which weren't successfully downloaded.
I suggest to fill bug and don't fix it now, because
- It doesn't break any old use case
- with --fetch-remote-resources it works
- we don't currently use content with remote components
- we need it only for certification where --fetch-remote-resources will be used

ybznek:
Planed related actions (but I will open another PR):
- exit oscap if component cannot be referenced
- check other actions (ds split?) where should be enabled --fetch-remote-resources option
- add --fetch-remote-resources option to man & bash_competition

mpreisler:
We also should look into how SCAP Workbench will handle these outputs to avoid yet another https://bugzilla.redhat.com/show_bug.cgi?id=1367923

src/DS/sds.c
```diff
@@ -246,108 +270,179 @@ static int ds_sds_dump_component(const char* external_file, const char* componen
...
+	ds_sds_session_remote_resources_progress(session)(false, "Downloading: %s ... ", url);
+
+	char* mem = oscap_acquire_url_download(url, &memory_size);
+	if (mem == NULL) {
+		ds_sds_session_remote_resources_progress(session)(false, "error", url);
 		return -1;
 	}
 
-	char* xlink_href = (char*)xmlGetNsProp(component_ref, BAD_CAST "href", BAD_CAST xlink_ns_uri);
+	ds_sds_session_remote_resources_progress(session)(false, "ok", url);
```
jan-cerny: I think there should be "ok\n", because when I run `oscap info`, I got messed output:
```
Checklists:
	Ref-Id: scap_gov.nist_cref_r2910-xccdf.xml
Downloading: http://scap.nist.gov/validation/downloads/r2910-external-oval-component.xml ... ok		Status: draft
```

jan-cerny: Also it makes weird thing with xccdf eval:
```
DownloadTest that the check_existence property is properly supported using the specified valuekTitle
Rule    xccdf_gov.nist_rule_validation.r2910_rule_1
Result  pass
```

There is a confilct with terminal manipulation kind of things, but adding a simple "\n" fixes it for me :smile:

src/DS/sds.c
```diff
@@ -246,108 +270,179 @@ static int ds_sds_dump_component(const char* external_file, const char* componen
...
+	return 0;
+}
 
-	char* filename_cpy = oscap_sprintf("./%s", relative_filepath);
-	char* file_reldir = dirname(filename_cpy);
+int ds_sds_dump_component_ref_as(const xmlNodePtr component_ref, struct ds_sds_session *session, const char* sub_dir, const char* relative_filepath)
+{
+	char* cref_id = (char*)xmlGetProp(component_ref, BAD_CAST "id");
+	if (!cref_id)
+	{
```
jan-cerny: Coding style

src/DS/public/ds_sds_session.h
```diff
@@ -173,6 +174,16 @@ int ds_sds_session_set_target_dir(struct ds_sds_session *session, const char *ta
 void ds_sds_session_reset(struct ds_sds_session *session);
 
 /**
+ * Set properties of remote content.
+ * @memberof ds_sds_session
+ * @param session The Source DataStream Session
+ * @param allowed Whether is download od remote resources allowed in this session (defaults to false)
```
jan-cerny: s/od/of

utils/oscap-info.c
```diff
@@ -57,11 +57,23 @@ struct oscap_module OSCAP_INFO_MODULE = {
...
+    "   --fetch-remote-resources \r\t\t\t\t - Download remote content referenced by DataStream.\n",
     .opt_parser = getopt_info,
     .func = app_info
 };
 
+static void _download_reporting_callback(bool warning, const char *format, ...)
+{
+	va_list argptr;
+	va_start(argptr, format);
+	vfprintf(stderr, format, argptr);
```
jan-cerny: Why does the download callback for `oscap info` write on stderr? The same messages in `xccdf eval` are written on stdoud. I think it should be consistent, because the text of message is same and has same meaning.

src/DS/sds.c
```diff
@@ -246,108 +270,179 @@ static int ds_sds_dump_component(const char* external_file, const char* componen
...
-	const char* target_filename_dirname = oscap_sprintf("%s/%s",sub_dir, file_reldir);
-	oscap_free(filename_cpy);
+	char* xlink_href = (char*)xmlGetNsProp(component_ref, BAD_CAST "href", BAD_CAST xlink_ns_uri);
 
-	ds_sds_dump_component(filename, component_id, session, target_filename_dirname, relative_filepath);
+	char* target_filename_dirname = compose_target_filename_dirname(relative_filepath, sub_dir);
 
+	char* component_id = NULL;
+
+	int ret = ds_sds_dump_component_by_href(session, xlink_href, target_filename_dirname, relative_filepath, cref_id, &component_id);
```
jan-cerny: This isn't much important, but I have noticed one thing while doing the review.

I think that `cref_id` and `xlink_href` are not used after this line anymore.
What if you put here this
```
xmlFree(cref_id);
xmlFree(xlink_href);
```
and remove those from the rest of the function? It's 6 times there.

utils/oscap-info.c
```diff
@@ -396,11 +411,29 @@ static int app_info(const struct oscap_action *action)
...
+	int c;
+	while ((c = getopt_long(argc, argv, "o:i:", long_options, NULL)) != -1) {
+		switch(c) {
+			case 0: break;
+			default: return oscap_module_usage(action->module, stderr, NULL);
+		}
+	}
+
+	if (optind >= argc) {
+		oscap_module_usage(action->module, stderr, "SCAP file need to be specified!\n");
```
jan-cerny: needs

utils/oscap-info.c
ybznek: Because these messages are printed to oscap info output and "downloading progress messages" could damage this output

src/DS/sds.c
yuumasato: I can confirm that SCAP Workbench needs the '\n' at the end. Without it scap-workbench misses the progress status of the first rule.

src/DS/sds.c
```diff
@@ -246,108 +270,179 @@ static int ds_sds_dump_component(const char* external_file, const char* componen
...
+			*sep = '\0';
+			*component_id = sep + 1;
+		}
+
+		if (!ds_sds_session_fetch_remote_resources(session)) {
+			static bool fetch_remote_resources_suggested = false;
+
+			if (!fetch_remote_resources_suggested) {
+				fetch_remote_resources_suggested = true;
+				ds_sds_session_remote_resources_progress(session)(false, "Datastream component '%s' points out to the remote '%s'. "
```
yuumasato: Would suggest to make it a warning so that scap-workbench does not parse it while reading rule id and results.
... (true, "WARNING: Datastream component ...

utils/oscap-info.c
jan-cerny: @ybznek But in `xccdf eval` it also "damages" the output.

src/DS/sds.c
ybznek: Great catch

src/DS/sds.c
```diff
@@ -218,26 +218,50 @@ static struct oscap_source *load_referenced_source(const struct ds_sds_session *
...
+		oscap_seterr(OSCAP_EFAMILY_XML, "Found component (id='%s') but it has no element contents, nothing to dump, skipping...", component_id);
+		return -1;
+	}
+
+	// If the inner root is script, we have to treat it in a special way
+	if (strcmp((const char*)component_inner_root->name, "script") == 0) {
+		return ds_sds_register_sce(session, component_inner_root, component_id, target_filename_dirname, relative_filepath);
+	// Otherwise we create a new XML doc we will dump the contents to.
+	// We can't just dump node "innerXML" because namespaces have to be
+	// handled.
```
mpreisler: indentation

utils/oscap-info.c
mpreisler: Good point. I think we should write warnings to stderr even in `xccdf eval` rather than writing these warnings to stdout just because that's what we do in `xccdf eval`. Thoughts?

src/DS/sds.c
```diff
@@ -246,108 +270,180 @@ static int ds_sds_dump_component(const char* external_file, const char* componen
...
+			*component_id = sep + 1;
+		}
+
+		if (!ds_sds_session_fetch_remote_resources(session)) {
+			static bool fetch_remote_resources_suggested = false;
+
+			if (!fetch_remote_resources_suggested) {
+				fetch_remote_resources_suggested = true;
+				ds_sds_session_remote_resources_progress(session)(true, "Datastream component '%s' points out to the remote '%s'. "
+									"Use '--fetch-remote-resources' option to download it.\n", cref_id, url);
```
yuumasato: Please, add the prefix "WARNING: " so that oscap users see this as a warning, and so that scap-workbench can treat this message as a warning too, otherwise it will be shown like an error in DiagnosticsDialog.

utils/oscap-info.c
jan-cerny: @mpreisler That would be a good solution.

jan-cerny:
@ybznek Please fix the conflict in `src/XCCDF/xccdf_session.c`. Please address also the minor issues that other reviewers have found. It should be an easy job. Then I will merge it.

ybznek:
@jan-cerny https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/DS/sds.c#L349 Old code doesn't check return value of this call, but my new code "accidentally" check it - it is why my changes cause the issue.

src/DS/sds.c
ybznek: Comment is for "else", not for return. It is like here https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/DS/sds.c#L275
I could move the comment after "else".

mpreisler: yeah, please move it, it's more obvious if it's inside the block

ybznek:
I have rebased due to conflict. New changes are after "	man: Add --fetch-remote-resources to info"

jan-cerny:
I have verified that CPE datastrems are working again.

Thank you very much for your awesome work!

Issue #529, opened 2016-09-19
Segfault with multiple datastream components with same href


ybznek:
```
$ oscap info  ./r2910-datastream.xml 
Document type: Source Data Stream
Imported: 2016-09-19T13:33:25
Segmentation fault (core dumped)
```
From backtrace:
**"There is already a mapping from component id '%s' to component-ref id '%s'. Having multiple mappings is legal in a datastream but it may prove problematic when selecting component-refs using XCCDF Ben"..., ap=<optimized out>) at util.c:181**

Full backtrace (maybe have inconsistent line numbers due to my change)
```
#0  0x00007ffff50bffb4 in vfprintf () from /lib64/libc.so.6
#1  0x00007ffff50e89b9 in vsnprintf () from /lib64/libc.so.6
#2  0x00007ffff7b27a75 in oscap_vsprintf (
    fmt=0x7ffff7b938a8 "There is already a mapping from component id '%s' to component-ref id '%s'. Having multiple mappings is legal in a datastream but it may prove problematic when selecting component-refs using XCCDF Ben"..., ap=<optimized out>) at util.c:181
#3  0x00007ffff7b23b88 in __oscap_seterr (file=file@entry=0x7ffff7b939b1 "sds_index.c", line=line@entry=200, 
    func=func@entry=0x7ffff7b93a50 <__PRETTY_FUNCTION__.9640> "ds_stream_index_parse", family=family@entry=2) at error.c:118
#4  0x00007ffff7b2d82a in ds_stream_index_parse (reader=0x61e6b0) at sds_index.c:197
#5  ds_sds_index_parse (reader=0x61e6b0) at sds_index.c:356
#6  0x00007ffff7b2a5f3 in ds_sds_session_get_sds_idx (session=session@entry=0x623210) at ds_sds_session.c:108
#7  0x000000000040cfc2 in app_info (action=0x7fffffffd7f0) at oscap-info.c:264
#8  0x0000000000407fee in oscap_module_call (action=0x7fffffffd7f0) at oscap-tool.c:261
#9  oscap_module_process (module=0x615ba0 <OSCAP_INFO_MODULE>, module@entry=0x614a40 <OSCAP_ROOT_MODULE>, argc=argc@entry=3, argv=argv@entry=0x7fffffffda88)
    at oscap-tool.c:346
#10 0x000000000040706f in main (argc=3, argv=0x7fffffffda88) at oscap.c:80
```

Content:
[...]

PR #537, opened 2016-09-22
Issue #529: Fix a segmentation fault

jan-cerny:
Strings expected in format strings were not supplied.
Compiler was OK with that because it is a function with
variable count of parameters.


jan-cerny:
Link to issue #529

PR #540, opened 2016-09-23
Consider suggestion to use "--fetch-remote-resources" a warning

yuumasato:
SCAP Workbench is parsing this message as progress status and
showing errors, while actually this message is a warning.

Related to RHBZ#1367923


jan-cerny:
ACK. Thanks!

PR #546, opened 2016-09-29
Add puppet to report generation

redhatrises:


mpreisler:
Just a note: I checked the codebase and we don't seem to be using this fix/@ system yet. It is used in SSG though so the @ system attribute wasn't invented in this PR.

Issue #541, opened 2016-09-27
Add 'complexity', 'disruption', 'reboot', and 'strategy' to --report if the attributes exist


redhatrises:
With https://github.com/OpenSCAP/scap-security-guide/issues/790 now merged in SSG, we should add 'complexity', 'disruption', 'reboot', and 'strategy', if the attributes exist, to the --report generation

jan-cerny:
Fixed in https://github.com/OpenSCAP/openscap/pull/548

PR #548, opened 2016-10-03
xccdf report: Add fix attributes

ybznek:
Do you have a better idea how to show it?

image

https://github.com/OpenSCAP/openscap/issues/541


mpreisler:
This looks great.

PR #549, opened 2016-10-04
Add support for tailoring of external datastreams

jan-cerny:
This pull request allows user to evaluate a tailoring within datastream that customizes benchmarks located in other datastream in on other datastream collection in an other separate file.

### DIFF
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -524,13 +524,43 @@ static int _acquire_xccdf_checklist_from_tailoring(struct xccdf_session* session
 		xmlFree(tailoring_xmlDoc);
 		return 1;
 	}
-	char *benchmark_id = strchr(benchmark_ref, '#') + 1;
 
-	struct oscap_source *xccdf_source = ds_sds_session_select_checklist(ds_sds_session,
-			NULL, benchmark_id, NULL);
+	struct oscap_source *xccdf_source = NULL;
+	if (oscap_str_startswith(benchmark_ref, "file:")) {
+		char* sep = strchr(benchmark_ref, '#');
+		const char *filename = benchmark_ref + strlen("file:");
+		const char *component_ref = NULL;
+		if (sep != NULL) {
+			component_ref = sep + 1;
+			*sep = '\0';
+		}
+
+		struct oscap_source *external_file = oscap_source_new_from_file(filename);
+		if (oscap_source_get_scap_type(external_file) == OSCAP_DOCUMENT_SDS) {
+			if (component_ref == NULL) {
+				oscap_seterr(OSCAP_EFAMILY_OSCAP,
+						"The referenced component is a datastream, but no datastream component was specified.");
+				oscap_source_free(external_file);
+				xmlFree(tailoring_xmlDoc);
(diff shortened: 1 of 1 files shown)


src/XCCDF/xccdf_session.c
```diff
@@ -524,13 +524,33 @@ static int _acquire_xccdf_checklist_from_tailoring(struct xccdf_session* session
...
+	struct oscap_source *xccdf_source = NULL;
+	if (oscap_str_startswith(benchmark_ref, "file:")) {
+		char* sep = strchr(benchmark_ref, '#');
+		const char *filename = benchmark_ref + strlen("file:");
+		const char *component_ref = sep + 1;
+		*sep = '\0';
+
+		struct oscap_source *external_file = oscap_source_new_from_file(filename);
+		if (oscap_source_get_scap_type(external_file) == OSCAP_DOCUMENT_SDS) {
+			struct ds_sds_session *ds_sds_session_tmp = ds_sds_session_new_from_source(external_file);
```
mpreisler: do you really need an extra variable here? why not just free ds_sds_session and reallocate it?

jan-cerny: yeah, great, thanks, that would be better and more readable

src/XCCDF/xccdf_session.c
```diff
@@ -524,13 +524,33 @@ static int _acquire_xccdf_checklist_from_tailoring(struct xccdf_session* session
 		xmlFree(tailoring_xmlDoc);
 		return 1;
 	}
-	char *benchmark_id = strchr(benchmark_ref, '#') + 1;
 
-	struct oscap_source *xccdf_source = ds_sds_session_select_checklist(ds_sds_session,
-			NULL, benchmark_id, NULL);
+	struct oscap_source *xccdf_source = NULL;
+	if (oscap_str_startswith(benchmark_ref, "file:")) {
+		char* sep = strchr(benchmark_ref, '#');
```
jan-cerny: Ok, I fix this below.

jan-cerny:
I have improved this. I have added checks for NULL and I have removed a temporary variable.

ybznek:
I think we should have test for this. Or do you think that we can use NIST testuite to be test for it?

src/XCCDF/xccdf_session.c
```diff
@@ -524,13 +524,43 @@ static int _acquire_xccdf_checklist_from_tailoring(struct xccdf_session* session
...
+		const char *filename = benchmark_ref + strlen("file:");
+		const char *component_ref = NULL;
+		if (sep != NULL) {
+			component_ref = sep + 1;
+			*sep = '\0';
+		}
+
+		struct oscap_source *external_file = oscap_source_new_from_file(filename);
+		if (oscap_source_get_scap_type(external_file) == OSCAP_DOCUMENT_SDS) {
```
ybznek: It is good to check, could you please add this condition to sds.c where can be found similar code?

jan-cerny: @ybznek yes, I can, but let's do that in other PR

ybznek:
LGTM, thank you!

PR #551, opened 2016-10-05
xccdf-share.xsl: HTML report - fix empty table

ybznek:
I forgot to check it in previous PR.
If these attributes are not provided we don't want to print empty table.

It should be part of https://github.com/OpenSCAP/openscap/pull/548


ybznek:
process58 fail...

PR #553, opened 2016-10-06
Fix a bug in Arithmetic complex function

rsprudencio:
Code was not considering that one of the operands could be a string.


mpreisler:

jan-cerny:
Great! ACK.

PR #554, opened 2016-10-08
Rerun failing tests

ybznek:
It is not very nice way how to deal with it, but it could save our time.

If some test fail, it will rerun it again. If we really break something, it will fail as should.

### DIFF
--- a/tests/probes/file/test_probes_file.sh
+++ b/tests/probes/file/test_probes_file.sh
@@ -118,8 +118,8 @@ function test_probes_file_invalid_utf8 {
 
 test_init "test_probes_file.log"
 
-test_run "test_probes_file" test_probes_file
-test_run "test_probes_file_filenames" test_probes_file_filenames
-test_run "test_probes_file_invalid_utf8" test_probes_file_invalid_utf8
+run_multiple_times 5 test_run "test_probes_file" test_probes_file
+run_multiple_times 5 test_run "test_probes_file_filenames" test_probes_file_filenames
+run_multiple_times 5 test_run "test_probes_file_invalid_utf8" test_probes_file_invalid_utf8
 
 test_exit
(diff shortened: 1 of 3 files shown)


mpreisler:
This is pretty awesome but I am concerned that it goes against unit testing principles. This makes our life easier in the short term but in the long term this is likely to hide issues.

If something fails 4 times and then passes, how is it reported in the results?

ybznek:
@mpreisler 
- (I am not sure, that our tests are "unit" tests. They don't check units, but whole functionality: parsing of parameters -> processing through bunch of functions -> output)

- we can only fill some issue comment. Output log is too long and "without" structure. If I print something, nobody will "see" it in such long output.

- We have issues with random fails of tests for too long term. If think that dirty "solution" (=automatic rerun of single test) is better than no solution (=manual rerun of whole test suite - it needs more of our time and more of jenkins time)

Do you have any better opinion?

jan-cerny:
I agree with @ybznek that these tests aren't unit tests, Therefore we have been going against unit testing principles for quite a long time :-)
But I can understand what @mpreisler is concerned about. 
There is a possibility that we introduce another non-deterministic behaviour in OpenSCAP, but if the probability of the bug will be less than 80 %, the test will keep passing and we will not notice that. 
I know that it sounds scary, but despite this fact, I would accept this pull request, because only other solution that I see is to spend a few months with valgrind and gdb. I think that would be a  clean solution, but we already tried that.

ybznek:
It reruns only a few specific tests - and we see what tests need it.

If we increase rate of random failures, we will not see anything either with current state of testing.

mpreisler:
We discussed this PR today with the team and decided against it. It was a tough decision because this change would make our lives easier. It is likely to hide issues and will make it less likely that we actually fix them.

PR #555, opened 2016-10-13
Separate stderr from stdout in SCE

mpreisler:
We used to put both into interleaved output. This makes reuse of SCE results difficult for some of our users. This patch separates them.

Keep in mind that while this doesn't break API or ABI it does alter behavior! "stdout" is no longer stdout and stderr interleaved, it just has stdout in it now. I have added stderr into the HTML report so I think for the vast majority of use-cases this will improve the experience and not break it. However I see this as a gray area so I'd appreciate discussion.

I know we suggested stdout_only, stdout and stderr to keep current behavior but I thought that that was to confusing.

image

### DIFF
--- a/src/SCE/sce_engine.c
+++ b/src/SCE/sce_engine.c
@@ -57,6 +57,7 @@ struct sce_check_result
 	char* href;
 	char* basename;
 	char* std_out;
+	char* std_err;
 	int exit_code;
 	struct oscap_stringlist* environment_variables;
 	xccdf_test_result_type_t xccdf_result;
@@ -68,6 +69,7 @@ struct sce_check_result* sce_check_result_new(void)
 	ret->href = NULL;
 	ret->basename = NULL;
 	ret->std_out = NULL;
+	ret->std_err = NULL;
 	ret->environment_variables = oscap_stringlist_new();
 	ret->xccdf_result = XCCDF_RESULT_UNKNOWN;
 
@@ -85,6 +87,8 @@ void sce_check_result_free(struct sce_check_result* v)
 		oscap_free(v->basename);
 	if (v->std_out)
 		oscap_free(v->std_out);
+	if (v->std_err)
+		oscap_free(v->std_err);
 
 	oscap_stringlist_free(v->environment_variables);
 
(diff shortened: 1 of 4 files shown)


ybznek:
file probe... I will collect verbose output and rerun test.

ybznek:
nothing new in output. Jenkins, test this please

src/SCE/sce_engine.c
```diff
@@ -444,8 +464,9 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
 	env_values[env_value_count] = NULL;
 
 	// We open a pipe for communication with the forked process
-	int pipefd[2];
-	if (pipe(pipefd) == -1)
+	int stdout_pipefd[2];
+	int stderr_pipefd[2];
+	if (pipe(stdout_pipefd) == -1 || pipe(stderr_pipefd))
```
ybznek: Please, check both pipes in the same way.

src/SCE/sce_engine.c
```diff
@@ -467,17 +488,19 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
...
-		    // we won't read from the pipe, so close the reading fd
-		    close(pipefd[0]);
+		    // we won't read from the pipes, so close the reading fd
+		    close(stdout_pipefd[0]);
+		    close(stderr_pipefd[0]);
 
 		    // forward stdout and stderr to the opened pipe
-			dup2(pipefd[1], fileno(stdout));
-			dup2(pipefd[1], fileno(stderr));
+			dup2(stdout_pipefd[1], fileno(stdout));
```
ybznek: Not consistent indentation with comment above.

radzy: Looks like the previous comment is no longer valid.  "forward stdout to the opened pipe"

mpreisler: Fixed

mpreisler:
I also added stderr element to the XSD schema for SCE results

mpreisler:
`1.2.11` was just released so this will be in `1.2.12` at the earliest.

mpreisler:
@ybznek Anything else here?

ybznek:
@mpreisler 
Sorry. LGTM.

ybznek:
CG, PR ```#555```

mpreisler:
> CG, PR #555

?? @ybznek

ybznek:
555 - triple 5... doesn't matter...

PR #556, opened 2016-10-13
Print error message if oscap loads bz2 files and doesn't support them

ybznek:
oscap is able to recognize bz2 format even if it was compiled without bzip2 support.

If user with oscap(without bzip2 support) wanted to load bzip2 file, he got an ugly xml parsing error message. Now he will get less ugly error message.

https://github.com/OpenSCAP/scap-security-guide/issues/1509

### DIFF
--- a/src/source/bz2.c
+++ b/src/source/bz2.c
@@ -25,25 +25,23 @@
 #include "config.h"
 #endif
 
-#ifdef HAVE_BZ2
-
-#include <bzlib.h>
 #include <libxml/tree.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
 
-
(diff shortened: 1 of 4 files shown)


src/source/bz2.c
```diff
@@ -25,7 +25,6 @@
 #include "config.h"
 #endif
 
-#ifdef HAVE_BZ2
 
 #include <bzlib.h>
```
mpreisler: we shouldn't be including `bzlib.h` is HAVE_BZ2 is undefined

ybznek: Thank, you. I overlooked this.

ybznek: I squashed last commit, it is fixed now.

PR #552, opened 2016-10-06
Update SUSE 11 CPE

GautamSatish:
Hello,

This PR is to update the SUSE 11 CPE to address issue 1469 raised in SSG based on the comment from @redhatrises . [https://github.com/OpenSCAP/scap-security-guide/issues/1469#issuecomment-248900441]

I have used the OVAL content from "installed_OS_is_sle11" in SSG's shared OVAL folder as reference.[https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/oval/installed_OS_is_sle11.xml]

If the changes look good, I will make them for SUSE 10 and 12 as well.

Thank you!

PR #558, opened 2016-10-18
opensuse leap is an opensuse product, use the opensuse assigned CPE f…

msmeissn:
…rom /etc/os-release

add opensuse leap 42.2

(do you also want a pull rq for maint-1.2?)

### DIFF
--- a/cpe/openscap-cpe-dict.xml
+++ b/cpe/openscap-cpe-dict.xml
@@ -121,9 +121,13 @@
             <title xml:lang="en-us">openSUSE 13.2</title>
             <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.opensuse:def:132</check>
       </cpe-item>
-      <cpe-item name="cpe:/o:novell:leap:42.1">
+      <cpe-item name="cpe:/o:opensuse:leap:42.1">
             <title xml:lang="en-us">openSUSE 42.1</title>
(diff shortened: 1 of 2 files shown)


jan-cerny:
@msmeissn Thank you for proposing a pull request.

The patch LGTM.

Usually such change should go only into master, but I think that this can go to maint-1.2, because this is very similar case to https://github.com/OpenSCAP/openscap/pull/552. There we also changed CPE IDs in a stable branch to comply with assigned IDs. @mpreisler Do you agree?

mpreisler:
@jan-cerny If these IDs are NIST assigned official CPE IDs I agree.

mpreisler:
Just checked and this patch goes against official CPE ID naming.

```
$ grep "leap" official-cpe-dictionary_v2.3.xml | grep novell
  <cpe-item name="cpe:/o:novell:leap:42.0">
    <cpe-23:cpe23-item name="cpe:2.3:o:novell:leap:42.0:*:*:*:*:*:*:*"/>
  <cpe-item name="cpe:/o:novell:leap:42.1">
    <cpe-23:cpe23-item name="cpe:2.3:o:novell:leap:42.1:*:*:*:*:*:*:*"/>
```

In other words we currently use the NIST CPE naming and this patch would change that to something custom.

msmeissn:
NIST currently uses o:novell:leap*

I need to work with them

ybznek:
jenkins, test this please again

mpreisler:
@msmeissn I am very reluctant to accept this if it goes against the NIST CPE dictionary.

msmeissn:
I will reopen it once I get the Mitre CPE dict changed

PR #559, opened 2016-10-19
Temporary fix to remove the need of environment variables

rsprudencio:
to feed system_info for chroot/vm scans, it currently relies
on the use of grub2.

Small change in system_info probe logic, it doesn't fail if
it was not possible to collect system_info, it reports as
"Unknown" OS name/version/arch and hostname.

### DIFF
--- a/src/OVAL/probes/independent/system_info.c
+++ b/src/OVAL/probes/independent/system_info.c
@@ -97,11 +97,25 @@
 #include <string.h>
 #include <net/if.h>
 #include <arpa/inet.h>
-#include <pcre.h>
 #include <sys/types.h>
 
-#define MAX_ENTRY_SIZE  256
-#define MAX_BUFFER_SIZE 4096
+#if defined USE_REGEX_PCRE
+#include <pcre.h>
+#define _REGEX_MENUENTRY       "(?<=menuentry ').*?(?=')"
+#define _REGEX_SAVED_ENTRY_NR  "(?<=saved_entry=)[0-9]+"
+#define _REGEX_SAVED_ENTRY     "(?<=saved_entry=).*"
+#define _REGEX_ARCH            "(?<=\\.)i386|i686|x86_64|ia64|alpha|amd64|arm|armeb|armel|hppa|m32r" \
+		               "|m68k|mips|mipsel|powerpc|ppc64|s390|s390x|sh3|sh3eb|sh4|sh4eb|sparc"
+#elif defined USE_REGEX_POSIX
+#include <regex.h>
+#define _REGEX_MENUENTRY       "^menuentry '.*'"
+#define _REGEX_SAVED_ENTRY     "^saved_entry=.*"
+#define _REGEX_ARCH            "\\.(i386|i686|x86_64|ia64|alpha|amd64|arm|armeb|armel|hppa|m32r" \
+		               "|m68k|mips|mipsel|powerpc|ppc64|s390|s390x|sh3|sh3eb|sh4|sh4eb|sparc)"
+#endif
+#define _REGEX_RES_VECSIZE     3
+#define MAX_ENTRY_SIZE         256
+#define MAX_BUFFER_SIZE        4096
 
 static int fd=-1;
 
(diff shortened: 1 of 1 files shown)


src/OVAL/probes/independent/system_info.c
```diff
@@ -65,6 +65,7 @@
 #include <string.h>
 #include <ctype.h>
 #include <errno.h>
+#include <limits.h>
```
mpreisler: Is this include used?

src/OVAL/probes/independent/system_info.c
```diff
@@ -334,6 +340,145 @@ static ssize_t __sysinfo_saneval(const char *s)
...
 
+static char * _offline_chroot_get_menuentry(int entry_num)
+{
+	FILE *fp;
+	pcre *re;
+	const char *error;
+	char *ptr, grubcfg[MAX_BUFFER_SIZE+1] = { '\0' };
+	int erroffset, rc, ovector[30] = { 0 }, len;
+
+	fp = fopen("/boot/grub2/grub.cfg", "r");
```
mpreisler: Seems to be leaking the file descriptor.

src/OVAL/probes/independent/system_info.c
```diff
@@ -334,6 +340,145 @@ static ssize_t __sysinfo_saneval(const char *s)
 	return (ssize_t)real_length;
 }
 
+static char * _offline_chroot_get_menuentry(int entry_num)
+{
+	FILE *fp;
+	pcre *re;
+	const char *error;
+	char *ptr, grubcfg[MAX_BUFFER_SIZE+1] = { '\0' };
+	int erroffset, rc, ovector[30] = { 0 }, len;
```
mpreisler: This is a bit of a personal preference but we don't require C89 so you don't have to declare variables at the beginning of scopes. Declaring variables where they are used and limiting their visibility is advised.

src/OVAL/probes/independent/system_info.c
rsprudencio: It is used for HOST_NAME_MAX in _offline_chroot_get_hname()

src/OVAL/probes/independent/system_info.c
```diff
@@ -96,6 +97,11 @@
 #include <string.h>
 #include <net/if.h>
 #include <arpa/inet.h>
+#include <pcre.h>
```
rsprudencio: Will work on that, second should be <regex.h>?

src/OVAL/probes/independent/system_info.c
rsprudencio: Fixed

src/OVAL/probes/independent/system_info.c
rsprudencio: I did some refactor to get variables closer to the used scope

src/OVAL/probes/independent/system_info.c
```diff
@@ -334,6 +340,164 @@ static ssize_t __sysinfo_saneval(const char *s)
 	return (ssize_t)real_length;
 }
 
+static char * _offline_chroot_get_menuentry(int entry_num)
+{
+	FILE *fp;
+	pcre *re;
+	char *ptr, *ret = NULL;
```
mpreisler: ptr is not used in the function

rsprudencio: Thanks, this one sneaked in commit :)

src/OVAL/probes/independent/system_info.c
```diff
@@ -334,6 +340,164 @@ static ssize_t __sysinfo_saneval(const char *s)
...
+	char grubcfg[MAX_BUFFER_SIZE+1] = { '\0' };
+	int rc, ovector[30] = { 0 }, len;
```
ybznek: What does the magic `30` means? Please use constant.

src/OVAL/probes/independent/system_info.c
```diff
@@ -334,6 +340,164 @@ static ssize_t __sysinfo_saneval(const char *s)
...
+	pcre *re;
+	const char *error;
+	int erroffset, rc, ovector[30] = { 0 };
+	char *ptr;
+
+	if (os == NULL)
+		return NULL;
+
+	/* Check if grubenv is a menuentry # or identifier */
+	re = pcre_compile("(?<=\.)i386|i686|x86_64|ia64|alpha|amd64|arm|armeb|armel|hppa|m32r"
```
rsprudencio: Thank you, I fixed that.

src/OVAL/probes/independent/system_info.c
```diff
@@ -96,6 +97,25 @@
...
+#define _REGEX_ARCH            "(?<=\\.)i386|i686|x86_64|ia64|alpha|amd64|arm|armeb|armel|hppa|m32r" \
+		               "|m68k|mips|mipsel|powerpc|ppc64|s390|s390x|sh3|sh3eb|sh4|sh4eb|sparc"
+#elif defined USE_REGEX_POSIX
+#include <regex.h>
+#define _REGEX_MENUENTRY       "^menuentry '.*'"
+#define _REGEX_SAVED_ENTRY     "^saved_entry=.*"
+#define _REGEX_ARCH            "\\.(i386|i686|x86_64|ia64|alpha|amd64|arm|armeb|armel|hppa|m32r" \
+		               "|m68k|mips|mipsel|powerpc|ppc64|s390|s390x|sh3|sh3eb|sh4|sh4eb|sparc)"
+#endif
+#define _REGEX_RES_VECSIZE     3
```
mpreisler: this is PCRE only, we could move it under the PCRE define

src/OVAL/probes/independent/system_info.c
```diff
@@ -343,11 +626,14 @@ void *probe_init(void)
 int probe_main(probe_ctx *ctx, void *arg)
 {
 	SEXP_t* item;
-	char* os_name, *os_version, *architecture, *hname;
+	char* os_name, *architecture, *hname;
+	const char *os_version = NULL;
 	struct utsname sname;
 	probe_offline_flags offline_mode = PROBE_OFFLINE_NONE;
 	(void)arg;
+	char unknown[] = "Unknown";
```
mpreisler: Shouldn't this be `const char *unknown = "Unknown";`? I would also just `strdup` it to avoid the logic of freeing or not freeing the entries.

mpreisler:
Found just 2 minor nitpicks. This is very close to merging now, great work, kudos!

src/OVAL/probes/independent/system_info.c
```diff
@@ -383,7 +678,15 @@ int probe_main(probe_ctx *ctx, void *arg)
...
 
-	if (!offline_mode) {
+	if (offline_mode) {
+		if (os_name != unknown)
+			free(os_name);
+		// Don't free os_version, it shares same memory from os_name!
+		if (architecture != unknown)
+			free(architecture);
+		if (hname != unknown)
+			free(hname);
```
yuumasato: It's a bit confusing here. 
At first glance it looked like a mistake, because it's comparing pointers, but it is in fact correct.
Getting rid of this free or not free logic by `strdup`ing the `unknown` string would eliminate this confusion.

PR #560, opened 2016-10-20
Update is_wrlinux to check /etc/os-release

radzy:
Instead of checking the presence of /etc/wrlinux-release, we open
/etc/os-release and check for the CPE_NAME entry.

Signed-off-by: T.O. Radzy Radzykewycz radzy@windriver.com

### DIFF
--- a/src/OVAL/probes/unix/runlevel.c
+++ b/src/OVAL/probes/unix/runlevel.c
@@ -53,6 +53,7 @@
 #include <stdio.h>
 #include <stdbool.h>
 #include <string.h>
+#include <strings.h>
 #include <errno.h>
 #include <assert.h>
 #include <unistd.h>
@@ -68,6 +69,9 @@
 #include <probe/option.h>
 #include <alloc.h>
 #include "common/debug_priv.h"
+
+#define RELEASENAME_MAX_SIZE	256
+#define RELEASENAME_PATTERN	"CPE_NAME=\"%255s\""
 
 struct runlevel_req {
         SEXP_t *service_name_ent;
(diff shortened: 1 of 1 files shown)


radzy:
I've been meaning to redo the is_wrlinux() function for a while.  For discussion:

1)  I used the CPE_NAME field of /etc/os-release in this sample.  Not sure if that's the best to use, but given the variability of the names in /etc/os-release, I don't think there's really a better one.

2)  Would it be better to specify the keyword as an argument, in addition to the value ?  That would avoid hard-coding CPE_NAME w.r.t. (1)

3)  If we do this, it might make sense to re-work the distro_tbl[] values: add a parameter (two?) to the table, and just cal the parse_os_release function directly.  Thoughts ?

mpreisler:
The logic is much more complex than just checking a presence of a file. Is there any value in parsing os-release instead of checking for the wrlinux-release file? Is WRLinux planning to drop that file from their distro?

src/OVAL/probes/unix/runlevel.c
```diff
@@ -281,6 +285,35 @@ static int get_runlevel_common (struct runlevel_req *req, struct runlevel_rep **
 
 #if !defined(LINUX_DISTRO)
 # define LINUX_DISTRO generic
+
+static int parse_os_release(char * cpe)
+{
+        struct stat st;
```
mpreisler: indent

src/OVAL/probes/unix/runlevel.c
```diff
@@ -281,6 +285,35 @@ static int get_runlevel_common (struct runlevel_req *req, struct runlevel_rep **
...
+	FILE * osrelease;
+	char buf[RELEASENAME_MAX_SIZE];
+	char *releasename = &buf[0];
+	char c;
+	char * cp;
+
+	got = stat ("/etc/os-release", &st);
+	if ( got ) return (got == 0);
+
+	osrelease = fopen("/etc/os-release", "r");
```
mpreisler: leaked file descriptor

mpreisler:
Ah, OK, makes sense.

radzy:
Thanks for the review.  Any comments about the original questions?  Is CPE_NAME optimal if this gets used by other distros ?  Should it be parameterized ?  Should I integrate this with other simplifications for other platforms ?

radzy:
@mpreisler : Well, since @MorrisR2 indicated that this is not needed for Amazon Linux, then the questions become irrelevant.  So I guess this doesn't need any other work, and it can be merged at a convenient time if no further changes are required.

src/OVAL/probes/unix/runlevel.c
```diff
@@ -281,6 +285,43 @@ static int get_runlevel_common (struct runlevel_req *req, struct runlevel_rep **
...
+static int parse_os_release(char * cpe)
+{
+        struct stat st;
+        int got;
+        int ret;
+        FILE * osrelease;
+        char buf[RELEASENAME_MAX_SIZE];
+        char *releasename = &buf[0];
+        char c;
+        char * cp;
```
rsprudencio: Looks like this variable is unused

radzy:
Thanks, @rsprudencio !

rsprudencio:
Looks good now @radzy, thank you!

PR #561, opened 2016-10-24
Issue#475: RHBZ#1387248: Fix oscap-docker reporting incompliance

jan-cerny:
Compliance scan of a Docker image/container using oscap-docker reported
incorrectly that there had been an error even if scan had been successful
but incompliance of the assessed system had been found.

### DIFF
--- a/utils/oscap_docker_python/oscap_docker_util.py
+++ b/utils/oscap_docker_python/oscap_docker_util.py
@@ -119,19 +119,19 @@ class OscapHelpers(object):
         os.environ["OSCAP_PROBE_"
                    "PRIMARY_HOST_NAME"] = "{0}-{1}".format(target, image)
         cmd = ['oscap'] + [x for x in oscap_args]
-        try:
-            run = subprocess.check_output(cmd)
-        except Exception as error:
-            print("\nCommand: {0} failed!\n".format(" ".join(cmd)))
-            print("Error was:\n")
-            print(error)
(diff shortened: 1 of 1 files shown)


utils/oscap_docker_python/oscap_docker_util.py
```diff
@@ -120,16 +120,19 @@ def oscap_chroot(self, target, image, chroot_path, *oscap_args):
...
-        except Exception as error:
-            print("\nCommand: {0} failed!\n".format(" ".join(cmd)))
-            print("Error was:\n")
-            print(error)
-
-            # Clean up
-            self._cleanup_by_path(chroot_path)
-
-            sys.exit(1)
+            run = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
```
mpreisler: This looks like a hack, should we be using `subprocess.call` and checking whether exit_code is in [0, 2]? It gives us one way to get stderr and one way to get stdout. I think it would be cleaner.

jan-cerny: `Subprocess.call` doesn't allow to check the output of command, which we need.

From https://pymotw.com/2/subprocess/:
> The standard input and output channels for the process started by call() are bound to the parent’s input and output. That means the calling programm cannot capture the output of the command. 

But we can do it using `subprocess.Popen()` and `Popen().communicate`

jan-cerny:
I have re-worked the fix using `Popen()`. It's more readable that the previous idea.

PR #565, opened 2016-10-27
Change @style from SCAP_1.1 to SCAP_1.2 when converting XCCDF from 1.…

mpreisler:
…1 to 1.2

A workaround for this bug is deployed in SSG but unfortunately this was never reported in OpenSCAP. Let's fix the issue at its source!

https://github.com/OpenSCAP/scap-security-guide/blob/master/RHEL/7/Makefile#L68


mpreisler:
We could also fix this in `maint-1.0`. I am torn on this one.

mpreisler:
Let's fix this in maint-1.0, it is a bug.

PR #566, opened 2016-10-27
Change @style from SCAP_1.1 to SCAP_1.2 when converting XCCDF from 1.…

mpreisler:
…1 to 1.2

A workaround for this bug is deployed in SSG but unfortunately this was never reported in OpenSCAP. Let's fix the issue at its source!

https://github.com/OpenSCAP/scap-security-guide/blob/master/RHEL/7/Makefile#L68

### DIFF
--- a/xsl/xccdf_1.1_to_1.2.xsl
+++ b/xsl/xccdf_1.1_to_1.2.xsl
@@ -124,6 +124,11 @@ Authors:
     </xsl:attribute>
   </xsl:template>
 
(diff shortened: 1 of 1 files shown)


jan-cerny:
Hi, you have opened this again on maint-1.2 :-), so I have merged it manually as https://github.com/OpenSCAP/openscap/commit/9933ce82f3fb2ca03958af9cadfdc4377ae217d7 into maint-1.0

Issue #383, opened 2016-04-08
oscap ds sds-compose: Allow @schematron-version attribute to be set to something else than just hardcoded "1.0" value


iankko:
OpenSCAP DataStream `sds-compose` functionality (when creating DataStream from XCCDF) doesn't allow to specify other DataStream `@schematron-version` than just `"1.0"` one, which is hardcoded when generating the datastream. See:
&nbsp; &nbsp; [1] https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/DS/sds.c#L972
&nbsp; &nbsp; [2] https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/DS/sds.c#L990

But the `SCAP Version 1.2 Requirements Matrix` (copy of which can be found for example at:
&nbsp; &nbsp; [3] http://people.redhat.com/swells/nist-scap-validation/scap-val-requirements-1.2.html )
clarifies:

```
If applicable, each component MUST validate against its associated Schematron stylesheet.
For the SCAP source data stream collection, it MUST validate against the version of the SCAP
Schematron rules as specified on the <ds:data-stream-collection> element's
@schematron-version attribute, and it SHOULD also validate against the latest Schematron
rules.
```

Yet from [SCAP 1.2 specification](http://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-126r2.pdf) (`
Table 2 - ds:data -stream-collection`):

```
shematron-version     literal– token     1

The version of the SCAP Requirements Schematron rule set to which 
the  data stream collection conforms.
```

Latest source datastream schematron version is `"1.2"` -- see http://scap.nist.gov/revision/1.2/scap-schematron-rules.zip .

Currently it's not possible to meet the aforementioned `@schematron-version == "1.2"` requirement when creating source datastream from oscap (e.g. via `oscap ds sds-compose`) directly.

Therefore support for a new CLI option `--schematron-version` should be added for `oscap ds sds-compose` command (and possibly also for other commands creating datastream by other way than just from XCCDF file). Specifying such a datastream schematron version after the
 `--schematron-version` parameter would set the `@schematron-version` attribute to the provided value.

mpreisler:
Why would the user want to specify the latest schematron version. Shouldn't OpenSCAP figure that out and just do the right thing?

Anybody opposed to us changing the literal to `1.2` and closing this? We can update it whenever a new schematron is released for source datastream.

mpreisler:
[...]

PR #567, opened 2016-10-27
Changed sds schematron-version attribute to the latest, 1.2

mpreisler:
See https://github.com/OpenSCAP/openscap/issues/383


yuumasato:
ACK.

PR #572, opened 2016-11-02
Add [ref] links to HTML guide

mpreisler:
That way users can right click the ref links, copy URL and use that URL to reference XCCDF groups and rules.

image


mpreisler:
This will be useful for http://static.open-scap.org/ssg-guides/ssg-rhel7-guide-stig-rhel7-server-upstream.html

jan-cerny:
I have generated a new guide from SSG RHEL7 content using your patches. The reference links in the guide work well for me. Thank you.

PR #573, opened 2016-11-04
Fixed a bug in SEXP_parse(): when a SEXP partial buffer ends

### LINKED ISSUE #310: File probe test failed
I have seen failed test for file probe (`tests/probes/file`) in Jenkins. https://jenkins.open-scap.org/job/openscap-maint/394/

May be similar to still unresolved https://fedorahosted.org/openscap/ticket/442

rsprudencio:
with "#d0:", it was not properly checking the correct buffer
size and was trying to parse the token after ":", which is in
the next SEXP partial buffer.

Probably this fix #310, which is the same bug in #447 comments, which was probably mistakenly related to the SIGSEGV.

### DIFF
--- a/src/OVAL/probes/SEAP/sexp-parser.c
+++ b/src/OVAL/probes/SEAP/sexp-parser.c
@@ -239,7 +239,7 @@ static inline bool isnextexp (int c)
 
 /**
  * The S-expression parser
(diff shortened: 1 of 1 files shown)


rsprudencio:
This was actually a wrong fix, it broke SEAP API tests.

PR #575, opened 2016-11-07
RHEL 6 ComputeNode CPE and RHEL 7 ComputeNode CPE

yuumasato:
Adding definitions to openscap-shipped CPE dictionary.
Related to rhbz#1311054.


mpreisler:

PR #577, opened 2016-11-07
Small refactor in spb_pick() fix when size == 0, just return

rsprudencio:
right away instead of getting inside of copy block.


rsprudencio:
Thx @dkopecek :smiley:

PR #582, opened 2016-11-10
Fix invalid ARF results

jan-cerny:
We don't pass validation performed by the SCAPVal tool.

```
ERROR: SCHEMATRON - [results.xml] If the <xccdf:result> value for a <xccdf:rule-result> is 'error', 'unknown', 'notapplicable', or 'notchecked', then the result of at least one OVAL definition referenced by that rule SHALL be 'error', 'unknown', 'not applicable', or 'not evaluated', respectively. If the <xccdf:result> value is 'fail' then the result of at least one of the OVAL definitions referenced SHALL match the fail category as defined in the SCAP table. If the <xccdf:result> value is 'pass' then the result of all of the OVAL definitions referenced SHALL match the pass category as defined in the SCAP table.
ERROR: SCHEMATRON - [results.xml] The @href attribute SHALL contain "#" + the @id of the <arf:report> containing the check result. This approach provides traceability between XCCDF and check results.
ERROR: SCHEMATRON - [results.xml] Depending on the checking engine used (OVAL or OCIL), the the <arf:report> element should contain the relevant (OVAL or OCIL) content.
```
etc.

Those errors are reported because in XCCDF results part of ARF file, in xccdf:check-content-ref@href attribute OpenSCAP provides references to the XCCDF. This is wrong because it should contain reference to arf:report where OVAL results can be found.

This pull request wants to fix those errors reported by SCAPVal. Basically, when publishing XCCDF results to ARF it will replace a reference to OVAL definitions by reference to OVAL results document.

To be honest, I don't like the way how is this done, but I wasn't able to find anything better now. Any suggestions are welcome.

### DIFF
--- a/src/DS/rds.c
+++ b/src/DS/rds.c
@@ -448,7 +448,29 @@ static int ds_rds_report_inject_ai_target_id_ref(xmlDocPtr doc, xmlNodePtr test_
 	return 0;
 }
 
-static void ds_rds_report_inject_rule_result_check_refs(xmlDocPtr doc, xmlNodePtr rule_result, char *desired_href)
+static void ds_rds_report_inject_check_content_ref(xmlNodePtr check_content_ref, struct oscap_htable *arf_report_mapping)
+{
+	if (check_content_ref->type == XML_ELEMENT_NODE) {
+		if (strcmp((const char*)check_content_ref->name, "check-content-ref") == 0) {
+			char *oval_filename = (char *) xmlGetProp(check_content_ref,
+					BAD_CAST "href");
+			if (oval_filename == NULL) {
+				return;
+			}
+			char *report_id = oscap_htable_get(arf_report_mapping, oval_filename);
+			if (report_id == NULL) {
+				oscap_free(oval_filename);
+				return;
+			}
+			char *desired_href = oscap_sprintf("#%s", report_id);
+			xmlSetProp(check_content_ref, BAD_CAST "href", BAD_CAST desired_href);
+			oscap_free(desired_href);
+			oscap_free(oval_filename);
+		}
+	}
+}
+
+static void ds_rds_report_inject_rule_result_check_refs(xmlDocPtr doc, xmlNodePtr rule_result, struct oscap_htable *arf_report_mapping)
 {
 	xmlNodePtr child = rule_result->children;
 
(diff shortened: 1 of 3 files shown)


src/DS/rds.c
```diff
@@ -780,14 +782,16 @@ int ds_rds_create(const char* sds_file, const char* xccdf_result_file, const cha
...
+		struct oscap_source *target_rds = ds_rds_create_source(sds_source, xccdf_result_source, oval_result_sources, oval_result_mapping, arf_report_mapping, target_file);
 		result = target_rds == NULL;
 		if (result == 0) {
 			result = oscap_source_save_as(target_rds, NULL);
 		}
 		oscap_source_free(target_rds);
 	}
 	oscap_htable_free(oval_result_sources, (oscap_destruct_func) oscap_source_free);
+	oscap_htable_free(oval_result_mapping, (oscap_destruct_func) oscap_free);
+	oscap_htable_free(arf_report_mapping, (oscap_destruct_func) arf_report_mapping);
```
jan-cerny: Ooops

src/DS/rds.c
```diff
@@ -481,25 +498,21 @@ static void ds_rds_report_inject_rule_result_check_refs(xmlDocPtr doc, xmlNodePt
  *
  * TODO: Consider dropping this functionality if 370-1 is changed / clarified.
```
mpreisler: Outdated comment

src/DS/rds.c
```diff
@@ -718,17 +729,18 @@ static int ds_rds_create_from_dom(xmlDocPtr* ret, xmlDocPtr sds_doc, xmlDocPtr x
...
 
-	unsigned int oval_report_suffix = 2;
-	struct oscap_htable_iterator *hit = oscap_htable_iterator_new(oval_result_sources);
+	struct oscap_htable_iterator *hit = oscap_htable_iterator_new(arf_report_mapping);
 	while (oscap_htable_iterator_has_more(hit)) {
-		struct oscap_source *oval_source = oscap_htable_iterator_next_value(hit);
+		const struct oscap_htable_item *report_mapping_item = oscap_htable_iterator_next(hit);
+		char *oval_filename = report_mapping_item->key;
+		char *report_id = report_mapping_item->value;
+		char *report_file = oscap_htable_get(oval_result_mapping, oval_filename);
```
mpreisler: const correctness, make it const if you can.

jan-cerny:
I have removed the outdated comment and added const modifier on some variables.

src/XCCDF/xccdf_session.c
```diff
@@ -1280,6 +1284,13 @@ static char *_xccdf_session_export_oval_result_file(struct xccdf_session *sessio
 		return NULL;
 	}
 
+	static int counter = 0;
+	char *report_id = oscap_sprintf("oval%d", counter++);
+	const char *original_name = oval_agent_get_filename(oval_session);
+	char *results_file_name = oscap_strdup(name);
```
rsprudencio: is this free'd anywhere?

jan-cerny: Yes, it is free'd when the hash table is destroyed on line 266.

PR #587, opened 2016-11-15
Maint 1.2 improve generate fix

rsprudencio:
Added --verbose support in generate fix.

Removed a filter in generate fix that was filtering all fixes by current machine CPE, which makes no sense.


mpreisler:
`--devel` support? Did you mean `--verbose`?

rsprudencio:
Where did you see --devel? It's written --verbose!!!

PR #588, opened 2016-11-15
Maint 1.2 improve generate fix

rsprudencio:
Added --verbose support in generate fix.

Removed a filter in generate fix that was filtering all fixes by current machine CPE, which makes no sense.


rsprudencio:
I changed 2 tests that were semantically wrong IMO. Please review

tests/API/XCCDF/applicability/test_report_anaconda_fixes.sh
```diff
@@ -16,18 +16,18 @@ line3='^\W*passwd --minlen=14$'
 $OSCAP xccdf generate fix --template urn:redhat:anaconda:pre \
 	--output $result $srcdir/${name}.xccdf.xml 2>&1 > $stderr
 [ -f $stderr ]; [ ! -s $stderr ]
-grep "$line1" $result && false
-grep "$line2" $result && false
+grep "$line1" $result && true
+grep "$line2" $result && true
 grep "$line3" $result && false
```
rsprudencio: Good question, I just flipped it during the tests and it really seems this is not needed at all.

jan-cerny:
Great! Thank you.

PR #593, opened 2016-11-23
Fixed a corner case when /etc/hostname is empty, now it sets

rsprudencio:
sysinfo hostname to 'Unknown' instead of an empty string.


mpreisler:
Makes sense, LGTM

PR #595, opened 2016-11-28
Fixed a potential null deref in ds_rds_lookup_container().

rsprudencio:
If xmlDocGetRootElement returns NULL to root, it will dereference a null pointers, leading to a crash or maybe something worse.


mpreisler:
This should probably be fixed in `maint-1.0` instead.

rsprudencio:
Thanks, closing this PR as I pushed a new one in maint-1.0 branch.

PR #597, opened 2016-11-28
Added result-oriented 'generate fix' whenever an user pass a

rsprudencio:
result-id, else it will fallback to profile-oriented fix
generation, which get fixes from all selected rules in
provided profile.

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -615,11 +615,11 @@ static int _write_script_header_to_fd(const char *sys, int output_fd)
 int xccdf_policy_generate_fix(struct xccdf_policy *policy, struct xccdf_result *result, const char *sys, int output_fd)
 {
 	__attribute__nonnull__(policy);
+	int ret = 0;
 
+	dI("Generating fixes for policy(profile/@id=%s)", xccdf_policy_get_id(policy));
 	if (result == NULL) {
 		// No TestResult is available. Generate fix from the stock profile.
-		dI("Generating fixes for policy(profile/@id=%s)", xccdf_policy_get_id(policy));
-		int ret = 0;
 		struct xccdf_benchmark *benchmark = xccdf_policy_get_benchmark(policy);
 		if (benchmark == NULL) {
 			oscap_seterr(OSCAP_EFAMILY_OSCAP, "Could not find benchmark model for policy id='%s' when generating fixes.", xccdf_policy_get_id(policy));
@@ -640,7 +640,18 @@ int xccdf_policy_generate_fix(struct xccdf_policy *policy, struct xccdf_result *
 		return ret;
 	}
 	else {
-		// TODO.
-		return 1;
+		struct xccdf_rule_result_iterator *rr_it = xccdf_result_get_rule_results(result);
+		while (xccdf_rule_result_iterator_has_more(rr_it)) {
+			struct xccdf_rule_result *rr = xccdf_rule_result_iterator_next(rr_it);
+			if (xccdf_rule_result_get_result(rr) != XCCDF_RESULT_FAIL)
+				continue;
+			struct xccdf_rule *rule = _lookup_rule_by_rule_result(policy, rr);
+			ret = _xccdf_policy_rule_generate_fix(policy, rule, sys, output_fd);
+			if (ret != 0)
+				break;
+		}
+		xccdf_rule_result_iterator_free(rr_it);
+
+		return ret;
 	}
 }
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -617,9 +617,9 @@ int xccdf_policy_generate_fix(struct xccdf_policy *policy, struct xccdf_result *
 	__attribute__nonnull__(policy);
 	int ret = 0;
 
-	dI("Generating fixes for policy(profile/@id=%s)", xccdf_policy_get_id(policy));
 	if (result == NULL) {
 		// No TestResult is available. Generate fix from the stock profile.
+		dI("Generating profile-oriented fixes for policy(profile/@id=%s)", xccdf_policy_get_id(policy));
 		struct xccdf_benchmark *benchmark = xccdf_policy_get_benchmark(policy);
 		if (benchmark == NULL) {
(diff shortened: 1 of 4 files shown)


src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -615,11 +615,11 @@ static int _write_script_header_to_fd(const char *sys, int output_fd)
 int xccdf_policy_generate_fix(struct xccdf_policy *policy, struct xccdf_result *result, const char *sys, int output_fd)
 {
 	__attribute__nonnull__(policy);
+	int ret = 0;
 
+	dI("Generating fixes for policy(profile/@id=%s)", xccdf_policy_get_id(policy));
```
rsprudencio: Makes sense, fixed

mpreisler:
Can we also remove the XSLT since we are not using it anymore?

utils/oscap-xccdf.c
```diff
@@ -757,76 +757,53 @@ int app_generate_fix(const struct oscap_action *action)
...
-		 * Thus, this mode is usefull to review which fixes has been executed as a part
-		 * of given scan (TestResult)
-		 *
-		 * TODO: Once this legacy is dropped, we need to document --cpe and --datastream
-		 * support in man pages.
-		 */
-		fprintf(stderr, "#Warning: OpenSCAP will now use fallback mode (XSLT) to generate fixes. "
-			"Some of the functionality might be missing (Text substitution, CPE processing, DataStream support, and tailoring). "
-			"Please ommit --result-id option to trigger advanced processing.\n");
-		return app_xccdf_xslt(action);
```
rsprudencio: It's gone already @mpreisler

mpreisler: That's not true, it's still there... See https://github.com/OpenSCAP/openscap/blob/maint-1.2/xsl/legacy-fix.xsl and related files.

rsprudencio: Oh that makes sense, I removed the usage of those functions but not the functions itself and related content, good call! Should I remove the related functions or just that content?

mpreisler: If they are public we can't remove them, if they are private we can and IMO we should. The smaller the codebase the better, as long as it does what's expected :-)

mpreisler: On a second thought we do install the `legacy-fix.xsl` file in `/usr/share/openscap/xsl` and people may be relying on it. At the very least I would add a comment that the file is not used by openscap and only installed for compatibility reasons. And when this gets merged I'd remove it in `master`.

rsprudencio: I'll add a comment for that.

utils/oscap-xccdf.c
```diff
@@ -757,76 +757,53 @@ int app_generate_fix(const struct oscap_action *action)
...
+		if (xccdf_session_build_policy_from_testresult(session, action->id) != 0)
+			goto cleanup2;
+
+		struct xccdf_policy *policy = xccdf_session_get_xccdf_policy(session);
+		struct xccdf_result *result = xccdf_policy_get_result_by_id(policy, action->id);
+		if (xccdf_policy_generate_fix(policy, result, action->tmpl, output_fd) == 0)
+			ret = OSCAP_OK;
+	} else { // Fallback to profile if result id is missing
+		/* Profile-oriented fixes */
+		if (!xccdf_session_set_profile_id(session, action->profile)) {
```
yuumasato: I noticed that when both `--profile` and `--result-id ` are not passed, it defaults to `default` profile. Because usually there are no selected rules in default profile nothing is generated, and no errors are ouput too.

I think this may cause to confusion to newcomers. Would printing a warning telling that we are selecting default profile as no `--profile` or `--result-id` were passed break any use case?

utils/oscap-xccdf.c
```diff
@@ -757,76 +757,53 @@ int app_generate_fix(const struct oscap_action *action)
...
 	if (action->f_results != NULL) {
 		if ((output_fd = open(action->f_results, O_CREAT|O_TRUNC|O_NOFOLLOW|O_WRONLY, 0700)) < 0) {
 			fprintf(stderr, "Could not open %s: %s", action->f_results, strerror(errno));
 			goto cleanup;
 		}
 	}
-	if (xccdf_policy_generate_fix(policy, NULL, action->tmpl, output_fd) == 0)
-		ret = OSCAP_OK;
 
+	if (action->id != NULL) {
```
yuumasato: `--result-id` has precedence over `--profile`. Does it make sense to print a warning if both parameters are set?
Is this precedence documented to the user?

rsprudencio: If we have a result-id, it means that a scan was ran and a profile was set already during the scan, so I think it makes sense to only consider results which are FAIL during generating fixes, because it doesn't make sense to generate fix for passes or for the whole profile blindly.

I think we should document it, should be enough, instead of printing a warning.

utils/oscap-xccdf.c
rsprudencio: I think this is harder to document, as not putting options leaves us no specific place for comments. Also printing is not a good idea because people rely on the output to do fix scripts.

rsprudencio:
I improved the documentation in the man page, it should make the precedence problem clear and also distinguish between the 3 possibilities for generating fix scripts.

utils/oscap.8
```diff
@@ -310,7 +310,7 @@ To use the ability to include additional information from SCE in XCCDF result fi
 .TP
 .B \fBfix\fR  [\fIoptions\fR] xccdf-file
 .RS
-Generate a script that shall bring the system to a state of compliance with given XCCDF Benchmark.
+Generate a script that shall bring the system to a state of compliance with given XCCDF Benchmark. There are 3 possibilities when generating fixes: Result-oriented fixes (--result-id), Profile-oriented fixes (--profile) or Default Profile-oriented fixes (no option provided to oscap). Result-oriented takes precedences over Profile-oriented, so if you provide --result-id oscap will ignore any profile provided.
```
mpreisler: SCAP terminology is `(default)` profile to denote the default profile. Furthermore, it's not really a third option, right? There are just 2 possibilities, either it's fixes from results or fixes from profiles.

mpreisler: I think this warrants proper explanation and formatting as well. Go into specifics of what result oriented and profile oriented means. That's something I am looking for when browsing the manpage.

rsprudencio: I added 2 lines defining what are Profile/Result-oriented fixes, not sure if they are clear, comments are welcome! :smile:

mpreisler:
Looks good now, great work!

ACK

PR #600, opened 2016-11-29
Added a better description to xccdf_session_set_validation() doc.

rsprudencio:


mpreisler:
The docs are really misleading now, I suggest to separate the 3 modes and add that as a paragraph. Write how it connects the function and the CLI tool.

mpreisler:
Oh, this was already merged.

rsprudencio:
I can patch it. You mean adding '--skip-valid' and FULL_VALIDATION environ to related mode?

mpreisler:
On a second read this is fine and I don't think we need to patch it. It's just that my expectation was slightly different.

PR #602, opened 2016-11-29
Remove the ARF schematron workaround when there are no applicable checks

mpreisler:
The workaround helps with old ARF schematron issues, see
http://making-security-measurable.1364806.n2.nabble.com/Xccdf-dev-xccdf-rule-result-element-properties-td7582721.html
However it also causes issues in SCAPVal which expects no check elements
in rule-result if none are applicable.

Resolves issues with SCAPVal requirements 370-1 and 370-2.

I considered committing this to `maint-1.0` but then decided against it. The old branch is used for OVAL older than 5.10 and this workaround was added for the old NIST certification. I don't want there to be any unforeseen side effects.


mpreisler:
The workaround is wired into a lot of XCCDF unit tests. I will need to change them before this passes Jenkins.

rsprudencio:
The patch looks good to me, but I'm struggling with the reasoning. Why we had this workaround? It was a flaw in the specification or some implementation specifics?

mpreisler:
There was a flawed schematron rule for ARF that checked that every rule-result has at least one check element. If no applicable checks can be found we used to not put any check elements in rule-result and that is valid from an XSD and specification standpoint but would trigger ARF schematron warnings so the ARFs wouldn't pass through SCAPVal.

This was however fixed in later versions of SCAPVal so we no longer need this workaround.

PR #604, opened 2016-11-30
Add first part of the NIST validation test suite to our upstream test suite

jan-cerny:
This pull request adds the tests that can esaily be moved upstream, without bigger complications.
This way we will regularly check whether we pass at least some of the validation requirements during CI, eg. on Jenkins.
It might help us to ensure we don't regress in certification requirements.


tests/nist/nist_test.sh
```diff
@@ -0,0 +1,37 @@
...
+# Author:
+#     Jan Černý <jcerny@redhat.com>
+
+. ${srcdir}/../test_common.sh
+
+set -e -o pipefail
+
+function test_nist {
+	test_dir="$1"
+	./test_worker.py --scanner "$(top_builddir)/run $OSCAP" $test_dir
```
jan-cerny: probably missing ${srcdir}

mpreisler:
This PR is quite weird, why is it split up like this if nothing worked before 2582d23b945d1a63b86bb0de4536bd79897fd69b has been committed?

jan-cerny:
@mpreisler That's because I wanted to make more smaller commits instead of one very big commit.

mpreisler:
> @mpreisler That's because I wanted to make more smaller commits instead of one very big commit.

If it's not atomic it might as well be one big commit... The value of small commits is that you can check them out, build them, test them, bisect them, etc... All of this you can't do if we do PRs like this.

To do this right I suggest to rebase and reorder the commits. Don't add them to Makefile.am in a separate commit, that makes no sense, you need to add a test folder and add it to Makefile.am in one commit so that the commit is atomic and testable.

mpreisler:
Considering the effort required to reorganize the commits I think it's fine if you keep it and just get the test suite running. But please keep this in mind when making new PRs.

mpreisler:
@jan-cerny is busy so I will take over this PR and get it merged.

mpreisler:
I think I got it working, will wait for Jenkins.

I wish test_worker.py wasn't a kitchen sink but I won't fight that too hard.

mpreisler:
I did a license check to figure out whether we can actually merge this into OpenSCAP and I am not 100% sure we can. There is no mention of any license. I assume it's public domain but that's never spelled out anywhere in the test-suite.

I guess it's public domain because it comes from NIST and is intended to be used by tool vendors and users but I wish they said that in the archive or on the site.

I checked:
- https://scap.nist.gov/validation/faq.html#validationtestsuitefaq_objective
- https://scap.nist.gov/validation/downloads/SCAP1.2ValidationTestContent_1-2.1.1.0.zip
- https://scap.nist.gov/validation/downloads/change_log_1-2.1.1.0.txt

**Assuming it's public domain and therefore we can merge it.**

mpreisler:
There are major exception safety issues in the code but I don't think that's a big deal in a test-suite. We can fix that later IMO.

mpreisler:
Ok, there is one more issue where the ARFs are output into locations that don't exist in builddir. I will get that fixed.

mpreisler:
Oh man, this is going to be a massive pain in the ***. The script assumes in-source builds and has no options for out of source builds...

mpreisler:
I think this will pass now but we'll see. I have to run.

@yuumasato @rsprudencio A review would be great!

rsprudencio:

mpreisler:
Ah crap, there is a typo in `Makefile.am`, this will be an easy fix, gimme a minute.

mpreisler:
There is a bunch of paths that are cleaned in tests/nist/Makefile.am that I believe are never there. Perhaps @jan-cerny can comment. Is this just a copy paste error or something else?

I left it as it is so that it doesn't block merging of this PR but it does look like a mistake to me.

rsprudencio:
This PR got huge, is should be functional already? With all those requirements automated?

I'll try to review this later.

mpreisler:
> This PR got huge, is should be functional already? With all those requirements automated?
> 
> I'll try to review this later.

It's a subset of the NIST test-suite that doesn't require any external conditions.

tests/nist/test_worker.py
```diff
@@ -79,7 +79,7 @@ def run_test(test, scanner):
         sys.stderr.write("Unable to run the test! Check the catalog.xml.\n")
         return None
     print(command)
-    with open(test_id + ".stdout", "w") as stdout_file:
+    with open(os.path.join(output_dir, test_id + ".stdout"), "w") as stdout_file:
```
yuumasato: When running `test_worker.py` by hand and specifying `output_dir`, `output_dir` is not created. I would expect it to be created if it doesn't exist already.

tests/nist/Makefile.am
```diff
@@ -1,7 +1,7 @@
 EXTRA_DIST = $(top_srcdir)/tests/test_common.sh
 
-DISTCLEANFILES = *.log *.results oscap_debug.log.* *.result.xml *.results_arf.xml */*.results_oval.xml */*.stdout
-CLEANFILES = *.log *.results oscap_debug.log.* *.result.xml *.results_arf.xml */*.results_oval.xml */*.stdout
+DISTCLEANFILES = *.log *.results oscap_debug.log.* *.result.xml */*.results_arf.xml */*.results_oval.xml */*.stdout
+CLEANFILES = *.log *.results oscap_debug.log.* *.result.xml */*.results_arf.xml */*.results_oval.xml */*.stdout
```
yuumasato: Nice side effect is that it cleans output put in specified `--outputdir` if it is inside `tests/nist` directory.
I guess some people might not expect it.

yuumasato: Not requesting changes. I'm ok with the way it is.

tests/nist/test_worker.py
yuumasato: As discussed with @mpreisler , it makes sense for the tool not to create the output directory.
Mistakes in the parameters don't end up creating a directory..

yuumasato:
Ack.

PR #607, opened 2016-12-06
Only check ID of profile if profile exists

yuumasato:
Freeing a default policy, which has no profile assigned, was causing a segfault.


mpreisler:
Makes sense to me, I just wish there was a comment explaining that this is done because of the default profile.

mpreisler:
Thx!

Issue #431, opened 2016-06-13
oscap-vm / oscap-ssh / oscap-chroot: Couple of issues reported by CoverityScan / ShellCheck


iankko:
There's couple of issues being reported while running ShellCheck on openscap:
- 1)
  
  ```
  /usr/bin/oscap-vm:99:27: error: Can't use > in [ ]. Escape it or use [[..]]. [SC2073]
  #    97|   if [ `which guestunmount` > /dev/null ]; then
  #    98|       UNMOUNT_COMMAND="guestunmount"
  #    99|-> elif [ `which fusermount` > /dev/null ]; then
  #   100|       echo "guestunmount command not present on the system, using simpler fusermount instead"
  #   101|       UNMOUNT_COMMAND="fusermount -u"
  ```
- 2)

```
1. /usr/bin/oscap-vm:99:8: warning: Quote this to prevent word splitting. [SC2046]
#    97|   if [ `which guestunmount` > /dev/null ]; then
#    98|       UNMOUNT_COMMAND="guestunmount"
#    99|-> elif [ `which fusermount` > /dev/null ]; then
#   100|       echo "guestunmount command not present on the system, using simpler fusermount instead"
#   101|       UNMOUNT_COMMAND="fusermount -u"
```
- 3)

```
1. /usr/bin/oscap-vm:97:27: error: Can't use > in [ ]. Escape it or use [[..]]. [SC2073]
#    95|   which guestmount > /dev/null || die "Cannot find guestmount, please install libguestfs utilities."
#    96|   
#    97|-> if [ `which guestunmount` > /dev/null ]; then
#    98|       UNMOUNT_COMMAND="guestunmount"
#    99|   elif [ `which fusermount` > /dev/null ]; then
```
- 4)

```
1. /usr/bin/oscap-vm:97:6: warning: Quote this to prevent word splitting. [SC2046]
#    95|   which guestmount > /dev/null || die "Cannot find guestmount, please install libguestfs utilities."
#    96|   
#    97|-> if [ `which guestunmount` > /dev/null ]; then
#    98|       UNMOUNT_COMMAND="guestunmount"
#    99|   elif [ `which fusermount` > /dev/null ]; then
```
- 5)

```
1. /usr/bin/oscap-ssh:149:18: warning: Quote this to prevent word splitting. [SC2046]
#   147|   
#   148|   # We have to rewrite various paths to a remote temp dir
#   149|-> for i in $(seq 0 `expr $# - 1`); do
[...]

PR #608, opened 2016-12-06
ShellCheck warnings mostly fixed in our utils scripts

mpreisler:
Minor improvements to our util scripts to get rid of some `ShellCheck` warnings. There are more to fix but this PR fixes everything in #431 except false positives listed in the ticket.


rsprudencio:
Looks good to me

PR #609, opened 2016-12-06
Runlevel probe fixes

### LINKED ISSUE #591: runlevel related coverity issues
```
1. Defect type: CHAR_IO 
1. openscap-1.2.12/src/OVAL/probes/unix/runlevel.c:309: char_io: Assigning the return value of "fgetc" to char "c" truncates its value.
Expand
2. Defect type: NULL_RETURNS 
7. openscap-1.2.12/src/OVAL/probes/unix/runlevel.c:307: dereference: Dereferencing a pointer that might be null "osrelease" when calling "fscanf".
```

mpreisler:
Fixed coverity warnings and other issues found in `parse_os_release` in `runlevel.c`. I found some logic errors in the code and many things were over-complicated.

Would be awesome if @radzy could review this.

I tested it using gdb.

```
./run cgdb src/OVAL/probes/.libs/probe_runlevel
```

Inside gdb do:
```
break main
run
print parse_os_release("whatever_you_wish");
```

Fixes #591

### DIFF
--- a/src/OVAL/probes/unix/runlevel.c
+++ b/src/OVAL/probes/unix/runlevel.c
@@ -296,8 +296,7 @@ static int parse_os_release(const char * cpe)
 	if (osrelease < 0)
 		return !errno;
 
-	char buf[RELEASENAME_MAX_SIZE];
-	char *releasename = &buf[0];
+	char releasename[RELEASENAME_MAX_SIZE];
 	memset(releasename, 0, RELEASENAME_MAX_SIZE);
 
 	got = fscanf(osrelease, RELEASENAME_PATTERN, releasename);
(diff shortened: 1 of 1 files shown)


rsprudencio:
```
+ * - parse_os_release("cpe") returns 1 (!!!)
 + * - parse_os_release("cpe:/o:fedoraproject:fedora:*") returns 0 (!!!)
```

Do we want this behavior?

mpreisler:
@rsprudencio The behavior is surprising but it does work with the use-case. As long as it's a private function where the surprising results are documented I think it's fine.

radzy:
Looks good to me.

rsprudencio:
Looks good to me too.

PR #610, opened 2016-12-07
Added support for --without-syschar in XCCDF modules, now it is

rsprudencio:
possible to generate OVAL and/or XCCDF results with or without
OVAL System Characteristics.

### DIFF
--- a/utils/oscap.8
+++ b/utils/oscap.8
@@ -100,6 +100,11 @@ Write XCCDF results into FILE.
 Writes results to a given FILE in Asset Reporting Format. It is recommended to use this option instead of --results when dealing with datastreams.
 .RE
 .TP
+\fB\-\-without-syschar\fR
+.RS
+Don't provide system characteristics in result file.
+.RE
+.TP
 \fB\-\-report FILE\fR
 .RS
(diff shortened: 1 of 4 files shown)


mpreisler:
Could you also add this to the manpage?

rsprudencio:
Sorry, forgot about that!

utils/oscap.8
```diff
@@ -100,6 +100,11 @@ Write XCCDF results into FILE.
 Writes results to a given FILE in Asset Reporting Format. It is recommended to use this option instead of --results when dealing with datastreams.
 .RE
 .TP
+\fB\-\-without-syschar\fR
+.RS
+Don't provide system characteristics in result file.
```
mpreisler: in OVAL result file, right? Does this have any effect if I don't request OVAL results or ARF but just request XCCDF results?

rsprudencio: Oops, indeed it was kinda misleading, added a more descriptive text for this option.

rsprudencio:
Another unrelated failure in process58 probe, @openscap-jenkins please retest this

rsprudencio:

PR #611, opened 2016-12-07
[DO NOT MERGE] Implemented xccdf_session_new_from_source() which creates

rsprudencio:
XCCDF session from loaded sources instead of a file.

This is not finished, it doesn't load the result for some reason. Will be used for future implementations of 'generate fix'.

### DIFF
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -119,6 +119,48 @@ static void _xccdf_session_free_oval_result_sources(struct xccdf_session *sessio
 static const char *oscap_productname = "cpe:/a:open-scap:oscap";
 static const char *oval_sysname = "http://oval.mitre.org/XMLSchema/oval-definitions-5";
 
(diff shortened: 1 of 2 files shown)


src/XCCDF/xccdf_session.c
```diff
@@ -119,6 +119,48 @@ static void _xccdf_session_free_oval_result_sources(struct xccdf_session *sessio
...
 
+struct xccdf_session *xccdf_session_new_from_source(struct oscap_source *source, struct oscap_source *result_source)
+{
+	if (source == NULL)
+		return NULL;
+
+	struct xccdf_session *session = (struct xccdf_session *) oscap_calloc(1, sizeof(struct xccdf_session));
+
+	session->source = source;
+	session->xccdf.result_source = result_source;
```
mpreisler: It looks like result_source can be NULL but it isn't documented in the function docs.

mpreisler:
@rsprudencio We should probably look at finishing this or closing the PR.

rsprudencio:
@mpreisler will close this PR as I won't touch generate fix for now

PR #613, opened 2016-12-13
Added --thin option to the XCCDF module, this enables thin results in OVAL and ARF files.

rsprudencio:
Full Results is still the default option.

If the user needs both no syschars and thin results they can use `--without-syschar --thin --results-arf /tmp/arf.xml`.

### DIFF
--- a/src/OVAL/results/oval_resultSystem.c
+++ b/src/OVAL/results/oval_resultSystem.c
@@ -407,7 +407,7 @@ static void _oval_result_definition_to_dom_based_on_directives(struct oval_resul
 			if (criteria)
 				_oval_result_system_scan_criteria_for_references(criteria, tstmap);
 		} else if (content == OVAL_DIRECTIVE_CONTENT_THIN) {
-			// TODO
+			// NOOP
 		}
 	}
 }
--- a/src/OVAL/results/oval_resultSystem.c
+++ b/src/OVAL/results/oval_resultSystem.c
@@ -406,6 +406,8 @@ static void _oval_result_definition_to_dom_based_on_directives(struct oval_resul
 			/* collect the tests that are referenced from reported definitions */
 			if (criteria)
 				_oval_result_system_scan_criteria_for_references(criteria, tstmap);
+		} else if (content == OVAL_DIRECTIVE_CONTENT_THIN) {
+			// TODO
 		}
 	}
 }
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -903,22 +903,16 @@ int xccdf_session_load_oval(struct xccdf_session *session)
 		}
 
 		if (session->export.thin_results) {
-			struct oval_result_directives *dir = oval_result_directives_new();
-			if (dir == NULL) {
-				oscap_seterr(OSCAP_EFAMILY_OSCAP, "Failed to create new directives model for: '%s'.", contents[idx]->href);
-				return 1;
-			}
+			struct oval_results_model *res_model = oval_agent_get_results_model(tmp_sess);
+			struct oval_directives_model *dir_model = oval_results_model_get_directives_model(res_model);
+			// This is the worst function name in existence, despite its name,
+			// it's getting the oval_result_directives of the oval_directives_model.
+			// You would expect oval_directives_model_getresdirs at least, but no..
+			struct oval_result_directives *dir = oval_directives_model_get_defdirs(dir_model);
 			oval_result_directives_set_content(dir,  OVAL_RESULT_TRUE | OVAL_RESULT_FALSE |
 							OVAL_RESULT_UNKNOWN | OVAL_RESULT_NOT_EVALUATED |
 							OVAL_RESULT_NOT_APPLICABLE | OVAL_RESULT_ERROR,
 							OVAL_DIRECTIVE_CONTENT_THIN);
-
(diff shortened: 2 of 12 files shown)


mpreisler:

mpreisler:
Causes linking issues with the python bindings. Caught by Jenkins.

src/XCCDF/xccdf_session.c
```diff
@@ -890,6 +896,25 @@ int xccdf_session_load_oval(struct xccdf_session *session)
 			return 2;
 		}
 
+		if (session->export.thin_results) {
+			struct oval_results_model *res_model = oval_agent_get_results_model(tmp_sess);
+			struct oval_result_directives *dir = oval_directives_model_new();
```
mpreisler: Note to self: The pointer types are not compatible! This is undefined behavior and causes crashes...

mpreisler:
I got it into a working state, now let's propagate the thin results stuff to CPE OVAL.

mpreisler:
Fixed up the CPE session thin results. Do not merge yet, I want to do some refactoring because one part of the code is duplicated in CPE session and XCCDF session.

mpreisler:
I think `oval_directives_model_set_defdirs` leaks. Any thoughts? From the code it looks like oval_directives owns oval_result_directives. CC @rsprudencio

mpreisler:
CC @rsprudencio A review would be great. I can't assign you via github because you are the author of this PR :-)

rsprudencio:
@mpreisler I will review this today :smile:

src/XCCDF/xccdf_session.c
```diff
@@ -890,6 +902,19 @@ int xccdf_session_load_oval(struct xccdf_session *session)
...
 		}
 
+		if (session->export.thin_results) {
+			struct oval_results_model *res_model = oval_agent_get_results_model(tmp_sess);
+			struct oval_directives_model *dir_model = oval_results_model_get_directives_model(res_model);
+			// This is the worst function name in existence, despite its name,
+			// it's getting the oval_result_directives of the oval_directives_model.
+			// You would expect oval_directives_model_getresdirs at least, but no..
+			struct oval_result_directives *dir = oval_directives_model_get_defdirs(dir_model);
+			oval_result_directives_set_content(dir,  OVAL_RESULT_TRUE | OVAL_RESULT_FALSE |
```
rsprudencio: Shouldn't we set oval_result_directives_set_reported() too?

src/OVAL/results/oval_resultSystem.c
```diff
@@ -406,6 +406,8 @@ static void _oval_result_definition_to_dom_based_on_directives(struct oval_resul
 			/* collect the tests that are referenced from reported definitions */
 			if (criteria)
 				_oval_result_system_scan_criteria_for_references(criteria, tstmap);
+		} else if (content == OVAL_DIRECTIVE_CONTENT_THIN) {
+			// NOOP
```
rsprudencio: I remember tracing the flow and reaching this if, but I never actually knew what this was for, have you came to any conclusion and are we ok with a NOP in here?

src/XCCDF/xccdf_session.c
mpreisler: I removed the call because I think it's not necessary. Why do you think it's necessary?

src/OVAL/results/oval_resultSystem.c
mpreisler: Yeah, the code was mostly correct all along, I just made it explicit with the comment. We are OK with a noop because thin results do not have results for criteria. Only the higher level results for definitions.

utils/oscap.8
```diff
@@ -95,6 +95,11 @@ Use given CPE dictionary or language (auto-detected) for applicability checks. (
 Write XCCDF results into FILE.
 .RE
 .TP
+\fB\-\-thin-results\fR
```
mpreisler: @rsprudencio Is it `--thin` or `--thin-results`?

src/XCCDF/xccdf_session.c
rsprudencio: Actually when directive is created it's automatically set to true, so I think it's not really need, my bad

mpreisler:
Shouldn't we have the same thing for `oval eval`?

rsprudencio:
@mpreisler I think we do, but makes more sense to be in a separate PR, I'll create a ticket for that

rsprudencio:
The patches looks good to me, maybe we need another reviewer as we both made those patches hehehe

utils/oscap.8
mpreisler: Yeah, you are right. I was using `--thin` all the time which also works so I assumed it's `--thin`.

mpreisler:
@rsprudencio Yeah, I requested @yuumasato to review this as well.

utils/oscap.8
```diff
@@ -95,6 +95,11 @@ Use given CPE dictionary or language (auto-detected) for applicability checks. (
...
+.RS
+Set OVAL/ARF results to Thin Results.
```
yuumasato: We can provide better description of what are results with "Thin Results", similar to `--without-syschar`.
Like "Provide only the minimal amount of information in OVAL/ARF result files."

mpreisler:
@yuumasato I think this is now ready for your review.

rsprudencio:
@yuumasato Good one, didn't rearrange in man page, it's done now :smile:

PR #616, opened 2016-12-21
Use oscap_enum_to_string instead of just index lookup for OVAL affected family conversion

mpreisler:
This fixes a segfault when trying to convert OVAL_AFCFML_UNKNOWN to a string.

Also fixed a typo in the ODAFAMILY enum string mapping. Although keep in mind that that was not necessary to fix the segfault.


yuumasato:
Good fix. Ack.

PR #618, opened 2016-12-22
We always build the system_info probe in OpenSCAP, mark it as such

mpreisler:
Previously we were testing for headers and disabling the tests if they were missing. Not disabling the probe. That makes no sense, let's always build and always test system_info.

system_info deps are hard to check for because the probe has different implementations for BSD, Linux and Solaris.


mpreisler:
I thought about patching this in `maint-1.0` but I think it's not a bugfix but rather an enhancement / change of reporting. Let's keep all the autoconf variables in `maint-1.0` even if they are not being used.

yuumasato:
Ack, too.

PR #619, opened 2016-12-22
Stop checking pcre.h AND regex.h in all the probes

mpreisler:
We check for PCRE OR POSIX regex.h globally, makes no sense to check again for all the probes.

We need pcre.h OR POSIX regex.h whereas in the probes we check that both are present. In rare circumstances this may disable probes even though all the necessary deps are present.


mpreisler:
Rebased to fix conflicts.

yuumasato:
Ack.

Issue #617, opened 2016-12-22
XCCDF score re-calculation after remediation.


GautamSatish:
Hello,

I had brought this up on the mailing list sometime ago. This is a report of my observations with some more experiments.

For calculating the compliance score, oscap calls "xccdf_policy_get_score" in the function "xccdf_session_evaluate". 

When running "oscap xccdf evaluate" with the "--remediate" option, the report file mentions all fixed rules as passed but the compliance score is from before running remediation, i.e. based on the initial evaluation. So even if the number of rules that pass increases, compliance score does not change. When an evaluation is run again, the number of rules that passes remains the same as with the previous report but compliance score is now updated.

This behavior is the same when running "oscap xccdf remediate". The report of this operation still the mentions the same score from evaluation result passed as input to it.

From what I understand, the problem is that oscap doesn't update the score in "xccdf_session_remediate".

The functions which calculates score like xccdf_item_get_default_score() are capable for reporting the score even for "fixed" rules:

`
		if ((xccdf_rule_result_get_result(rule_result) == XCCDF_RESULT_PASS) ||
                    (xccdf_rule_result_get_result(rule_result) == XCCDF_RESULT_FIXED)) {
`

Its just that it doesn't get invoked after remediation. 

I notice that there is a convenient function called **"xccdf_result_recalculate_scores"** which currently doesn't seem to be used anywhere. By invoking this function after the remediation, I am able to get the correct compliance score in the reports. I have tested this with both "oscap xccdf eval --remediate" and "oscap xccdf remediate".

I am not sure if this was one of the intended uses of "xccdf_result_recalculate_scores" but it appears to be doing the job. If this seems acceptable, I can submit a patch for review.

Thank you!

-Gautam.

shawndwells:
FYI - Red Hat is on a corporate holiday shutdown through the new year, so @mpreisler and team may have delayed responses.

.... but for what it's worth, I encounter this frequently during demos and labs. I'd submit the patch.

mpreisler:
@GautamSatish sounds good, we'd appreciate a patch for this.

GautamSatish:
@shawndwells and @mpreisler, 
I was running my patch through "make check" and I see that the unit tests in API/XCCDF have started to fail.

The problem is that after remediation, the assert checks that the score is still zero. Like here for example: https://github.com/OpenSCAP/openscap/blob/maint-1.2/tests/API/XCCDF/unittests/test_remediation_simple.sh#L43

[...]

PR #624, opened 2017-01-02
Update compliance score after remediation.

GautamSatish:
This PR addresses the issue discussed in #617. 

I have introduced a new wrapper function `xccdf_policy_update_score` similar to `xccdf_policy_get_score`. This function calls the `xccdf_result_recalculate_scores` routine.

In `xccdf_session_remediate`, rather than returning `xccdf_policy_remediate`, on success, it returns `xccdf_policy_update_score`. Both `xccdf_policy_update_score` and `xccdf_policy_remediate` have status as the return value.

I have also updated some of the unit tests to check for the corrected compliance score after remediation.

Thank you!

### DIFF
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -1558,7 +1558,7 @@ int xccdf_session_remediate(struct xccdf_session *session)
 	if ((res = xccdf_policy_remediate(xccdf_session_get_xccdf_policy(session), session->xccdf.result)) != 0)
 		return res;
 
-	return xccdf_policy_update_score(xccdf_session_get_xccdf_policy(session), session->xccdf.result);
+	return xccdf_policy_recalculate_score(xccdf_session_get_xccdf_policy(session), session->xccdf.result);
 }
 
 int xccdf_session_build_policy_from_testresult(struct xccdf_session *session, const char *testresult_id)
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -1554,7 +1554,11 @@ int xccdf_session_remediate(struct xccdf_session *session)
 	xccdf_result_set_version(session->xccdf.result,
 			benchmark != NULL ? xccdf_benchmark_get_version(benchmark) : NULL);
 	xccdf_result_fill_sysinfo(session->xccdf.result);
-	return xccdf_policy_remediate(xccdf_session_get_xccdf_policy(session), session->xccdf.result);
+
+	if ((res = xccdf_policy_remediate(xccdf_session_get_xccdf_policy(session), session->xccdf.result)) != 0)
+		return res;
+
+	return xccdf_policy_update_score(xccdf_session_get_xccdf_policy(session), session->xccdf.result);
 }
 
 int xccdf_session_build_policy_from_testresult(struct xccdf_session *session, const char *testresult_id)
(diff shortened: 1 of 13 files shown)


rsprudencio:

rsprudencio:

src/XCCDF/xccdf_session.c
```diff
@@ -1554,7 +1554,11 @@ int xccdf_session_remediate(struct xccdf_session *session)
 	xccdf_result_set_version(session->xccdf.result,
 			benchmark != NULL ? xccdf_benchmark_get_version(benchmark) : NULL);
 	xccdf_result_fill_sysinfo(session->xccdf.result);
-	return xccdf_policy_remediate(xccdf_session_get_xccdf_policy(session), session->xccdf.result);
+
+	if ((res = xccdf_policy_remediate(xccdf_session_get_xccdf_policy(session), session->xccdf.result)) != 0)
```
rsprudencio: Why do we even have this function as int return if all it does it run remediate functions and then return 0? @mpreisler shouldn't we make this void return? As this *if* will never be true, I don't think it's really needed.

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -2060,6 +2060,11 @@ struct xccdf_score * xccdf_policy_get_score(struct xccdf_policy * policy, struct
 	return xccdf_result_calculate_score(test_result, (struct xccdf_item *) benchmark, scsystem);
 }
 
+int xccdf_policy_update_score(struct xccdf_policy * policy, struct xccdf_result * test_result)
```
rsprudencio: IMO this could be named xccdf_policy_recalculate_scores to be consistent?

src/XCCDF/xccdf_session.c
mpreisler: We can't change that in a maintenance branch anyway.

GautamSatish: I did not want to change the signature of any of the functions. I added the `if` statement to ensure that in case `xccdf_policy_remediate` gets updated in the future to return anything else, `xccdf_session_remediate` would also return it as before. If the plan is to make it return void, I will remove the `if` statement.

src/XCCDF_POLICY/xccdf_policy.c
GautamSatish: Since the wrapper around `xccdf_result_calculate_score` was called `xccdf_policy_get_score`, I was going for something that follows a similar logic. :-)

Nevertheless, I will make this change.

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -2060,6 +2060,11 @@ struct xccdf_score * xccdf_policy_get_score(struct xccdf_policy * policy, struct
 	return xccdf_result_calculate_score(test_result, (struct xccdf_item *) benchmark, scsystem);
 }
 
+int xccdf_policy_recalculate_score(struct xccdf_policy * policy, struct xccdf_result * test_result)
+{
+    struct xccdf_benchmark * benchmark = xccdf_policy_model_get_benchmark(xccdf_policy_get_model(policy));
```
rsprudencio: Wrong indentation but whole file is messed up already.

mpreisler: Good catch, I will fix this myself after merging.

mpreisler:
ACK, great work, thanks a lot for your contribution!

src/XCCDF_POLICY/xccdf_policy.c
mpreisler: Fixed in c5222b5d27f15ea6ad733caa7d8c5b64b7bfe5de

PR #626, opened 2017-01-03
Test results directive --without-syschar and --thin-results

yuumasato:


mpreisler:

mpreisler:
I'd expect `Makefile.am` changes but I don't see them in the diff.

tests/nist/nist_test.sh
```diff
@@ -21,6 +21,60 @@ function test_nist {
...
+function test_results_directive {
+	test_dir="$1"
+	run_script="$(cd "${builddir}"; pwd)/run"
+	# make sure the output dir in builddir exists
+	mkdir -p "${builddir}/tests/nist/$test_dir/"
+	local datastream="ind_family_test-datastream.xml"
+
+	local result="${builddir}/tests/nist/${test_dir}/results_directive.results_arf.xml"
+	# the tests of this evaluation dont need to pass, return 0
+	${run_script} $OSCAP xccdf eval --profile "xccdf_gov.nist_profile_family-validation" --results-arf ${result} "${srcdir}/$test_dir/${datastream}" > /dev/null && 0
```
dahaic: Not sure what the construct `&& 0` means - should be `|| true`?

mpreisler: Good catch, this does look fishy.

yuumasato: Yeah, I was a little bit confused by exit statuses.
But I still don't know why it is working with `&&` if the evaluation of this data stream returns 2.

tests/nist/nist_test.sh
```diff
@@ -21,6 +21,65 @@ function test_nist {
 	fi
 }
 
+function test_results_directive {
+	test_dir="$1"
+	run_script="$(cd "${builddir}"; pwd)/run"
```
dahaic: As we have test_common.sh sourced in this file, you can use $OSCAP, instead of defining run_script by yourself.

yuumasato: Yes, I noticed that, but wanted to keep it consistent with `test_nist()`, which also uses `${run_script}`.

mpreisler: Looks like a mistake in the original PR to me. Let's use `$OSCAP` everywhere.

mpreisler:
Looks good to me now, ACK.

PR #629, opened 2017-01-04
Warn when injecting XCCDF 1.1 results into an ARF

mpreisler:
OpenSCAP can do it but the result won't be a valid ARF, only XCCDF 1.2
is allowed in ARFs.

Example:
```
oscap xccdf eval --verbose WARNING --results-arf /tmp/arf.xml ~/d/scap-security-guide/RHEL/7/output/ssg-rhel7-xccdf-1.2.xml 
WARNING: This content points out to the remote resources. Use `--fetch-remote-resources' option to download them.
WARNING: Skipping https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2 file which is referenced from XCCDF content
```

vs.

```
$ oscap xccdf eval --verbose WARNING --results-arf /tmp/arf.xml ~/d/scap-security-guide/RHEL/7/output/ssg-rhel7-xccdf.xml 
WARNING: This content points out to the remote resources. Use `--fetch-remote-resources' option to download them.
WARNING: Skipping https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2 file which is referenced from XCCDF content
W: oscap: Exporting ARF from XCCDF 1.1 is not allowed by SCAP specification. The resulting ARF will not validate. Convert the input to XCCDF 1.2 to get valid ARF results.
```

### DIFF
--- a/src/DS/rds.c
+++ b/src/DS/rds.c
@@ -32,6 +32,7 @@
 #include "common/_error.h"
 #include "common/util.h"
 #include "common/list.h"
+#include "common/debug_priv.h"
 
 #include "ds_common.h"
 #include "ds_rds_session.h"
@@ -600,6 +601,13 @@ static void ds_rds_add_xccdf_test_results(xmlDocPtr doc, xmlNodePtr reports,
 		const char* report_request_id, struct oscap_htable *arf_report_mapping)
 {
 	xmlNodePtr root_element = xmlDocGetRootElement(xccdf_result_file_doc);
+
+	if (root_element->ns && root_element->ns->href &&
+			oscap_str_endswith((const char*)root_element->ns->href, "xccdf/1.1")) {
+		dW("Exporting ARF from XCCDF 1.1 is not allowed by SCAP specification. "
+		   "The resulting ARF will not validate. Convert the input to XCCDF 1.2 "
+		   "to get valid ARF results.");
+	}
 
 	// There are 2 possible scenarios here:
 
--- a/src/DS/rds.c
+++ b/src/DS/rds.c
@@ -606,7 +606,8 @@ static void ds_rds_add_xccdf_test_results(xmlDocPtr doc, xmlNodePtr reports,
 			oscap_str_endswith((const char*)root_element->ns->href, "xccdf/1.1")) {
 		dW("Exporting ARF from XCCDF 1.1 is not allowed by SCAP specification. "
 		   "The resulting ARF will not validate. Convert the input to XCCDF 1.2 "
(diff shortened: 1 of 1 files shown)


src/DS/rds.c
```diff
@@ -601,6 +602,13 @@ static void ds_rds_add_xccdf_test_results(xmlDocPtr doc, xmlNodePtr reports,
 {
 	xmlNodePtr root_element = xmlDocGetRootElement(xccdf_result_file_doc);
 
+	if (root_element->ns && root_element->ns->href &&
```
rsprudencio: root_element is not getting checked to be non-NULL, if it's NULL this will lead to a NULL deref and crash in the best scenario.

mpreisler: Unless you pass an invalid doc it won't be NULL. The rest of the code of the function also assumes it's non-NULL.

What do you suggest I change?

rsprudencio: I know, but if anyone ever uses this passing NULL arg or maybe malformed xmlDoc it might lead to NULL deref. I know it's a corner case, but I'd put some null checking before dereferencing that pointer just to be sure.

mpreisler: I'd have to change the entire function for that to have any effect. The rest of the function is unrelated to this change. Furthermore, the function is only used internally.

Do you feel strongly about adding the check?

rsprudencio: @mpreisler actually I saw several functions not checking that, I think we can skip it here and do a security refactoring later on

src/DS/rds.c
```diff
@@ -601,6 +602,13 @@ static void ds_rds_add_xccdf_test_results(xmlDocPtr doc, xmlNodePtr reports,
...
+			oscap_str_endswith((const char*)root_element->ns->href, "xccdf/1.1")) {
+		dW("Exporting ARF from XCCDF 1.1 is not allowed by SCAP specification. "
+		   "The resulting ARF will not validate. Convert the input to XCCDF 1.2 "
```
yuumasato: Convertion of XCCDF 1.1 to XCCDF 1.2 is done through `xccdf_1.1_to_1.2.xsl`, right? Should we point users to this?

mpreisler: Good point, I will change that.

mpreisler: Fixed

mpreisler:
Will wait for jenkins to finish and then self-merge, @yuumasato reviewed this and I made a fix based on his feedback.

PR #630, opened 2017-01-04
Make WARNING the default verbosity level

mpreisler:
~~DO NOT MERGE, up for discussion.~~

@jan-cerny I have noticed that your oscap_set_verbose code defaults to unknown verbosity and no output. Do you recall what was the reasoning behind that? I expected WARNING level by default and stderr output, which is what this PR does.

Does the UNKNOWN verbose level have any meaning? It seems it's just used to flag that `oscap_set_verbose` was never run with non-NULL verbosity level.

Perhaps I am missing something.

### DIFF
--- a/tests/API/OVAL/unittests/test_external_variable.sh
+++ b/tests/API/OVAL/unittests/test_external_variable.sh
@@ -9,7 +9,10 @@ stderr=$(mktemp ${name}.err.XXXXXX)
 echo "stderr file: $stderr"
 
 $OSCAP oval eval --results $result --variables $srcdir/external_variables.xml $srcdir/$name.oval.xml 2> $stderr
-[ ! -s $stderr ] && rm $stderr
+# filter out the expected warnings in stderr
+sed -i -E "/^W: oscap:     Referenced variable has no values \(oval:x:var:[13689]\)/d" "$stderr"
+[ -f $stderr ]; [ ! -s $stderr ]; rm $stderr
+
 [ -s $result ]
 
 assert_exists 10 '/oval_results/oval_definitions/variables/external_variable'
--- a/tests/API/OVAL/unittests/test_external_variable.sh
+++ b/tests/API/OVAL/unittests/test_external_variable.sh
@@ -10,7 +10,8 @@ echo "stderr file: $stderr"
 
 $OSCAP oval eval --results $result --variables $srcdir/external_variables.xml $srcdir/$name.oval.xml 2> $stderr
 # filter out the expected warnings in stderr
-sed -i -E "/^W: oscap:     Referenced variable has no values \(oval:x:var:[13689]\)/d" "$stderr"
+
+sed -i -E "/^W: oscap:[ ]+Referenced variable has no values \(oval:x:var:[13689]\)/d" "$stderr"
 [ -f $stderr ]; [ ! -s $stderr ]; rm $stderr
 
 [ -s $result ]
--- a/src/XCCDF/public/xccdf_session.h
+++ b/src/XCCDF/public/xccdf_session.h
@@ -358,8 +358,11 @@ int xccdf_session_load_oval(struct xccdf_session *session);
  *
  * @memberof xccdf_session
  * @param session XCCDF Session
+ * @param plugin_name Name of the plugin to load
+ * @param quiet If true we will not output errors if loading fails
  * @returns zero on success
  */
+int xccdf_session_load_check_engine_plugin2(struct xccdf_session *session, const char* plugin_name, bool quiet);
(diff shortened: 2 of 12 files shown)


mpreisler:
(The Jenkins failure is expected)

jan-cerny:
@mpreisler 

I think that UNKNOWN level doesn't have any meaning, probably it is just a fallback value for oscap_string_to_enum.

However, now, in OpenSCAP you must directly specify verbosity level. The '--verbose' has a required argument.

If you omit the level, you will get an error:

```
oscap oval eval --verbose  /usr/share/xml/scap/ssg/content/ssg-fedora-oval.xml 
ERROR:    Invalid verbosity level! Verbosity level must be one of: DEVEL, INFO, WARNING, ERROR.
```

What's the reason behind this pull request? Do you want to reuse the function somewhere else? Or did you want to make the code more readable?

mpreisler:
I want to make "warnings and up" visible by default, even without `--verbose`

jan-cerny:
@mpreisler OK, That makes sense. Great idea! I think we can ignore the UNKNOWN value and default to WARNING.

mpreisler:
@dahaic How do you feel about changing this in a maintenance branch? `maint-1.2`. If everything goes well there will be no change in output, if there are any warnings they will be shown in `stderr`.

dahaic:
Personally, I see it as a bugfix, because proposed behaviour is something I would expect from any utility. So I have no issues with inclusion into the maintenance branch.

mpreisler:
Ok, so we have a consensus here. I will change the tests accordingly and finish this up.

jan-cerny:
@mpreisler The Jenkins fail here looks legit, there are multiple failing test cases in XCCDF directory.

mpreisler:
> @mpreisler The Jenkins fail here looks legit, there are multiple failing test cases in XCCDF directory.

Yeah, it's expected. Check out comment https://github.com/OpenSCAP/openscap/pull/630#issuecomment-270515296

I need to change the tests one by one.

mpreisler:
TODO:
- tests/API/XCCDF/applicability
- tests/API/XCCDF/unittests
- tests/API/XCCDF/variable_instance
- tests/bz2
- tests/DS
- tests/probes/sysctl

mpreisler:
```
$ cat /tmp/test_remediation_subs_unresolved.out.RClcIQ
W: oscap: The xccdf:rule-result/xccdf:instance element was not found.
```

This looks like a serious bug found by this change. I will investigate further after I fix up the rest.

mpreisler:
Does anybody understand the code at https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/XCCDF_POLICY/xccdf_policy_substitute.c#L168 ? I think it might be mistaking fix/instance with rule-result/instance but I don't know enough about this part of OpenSCAP to know for sure.

@jan-cerny @ybznek

rsprudencio:
The only instance I can recall in the specs is the one related to textfilecontent result and according to the code, that's the correct relation: `xccdf_rule_result_get_instances(data->rule_result);`

I don't know the relation between those instances and this text substitution though, sorry :disappointed:

mpreisler:
I took another look and the instance it's looking for is really not in the rule-results. So it could be that the code is fine. I am leaning towards backlisting this warning in the test. So we will keep warning about this situation but the tests will pass when this warning is in that particular test. I will make the commit shortly.

mpreisler:

rsprudencio:

mpreisler:
> @openscap-jenkins test this please

This was passed before AFAIK.

rsprudencio:
> This was passed before AFAIK.

Yes I know, I was testing Jenkins, it's only failing on SSG PR's :disappointed:

tests/API/OVAL/unittests/test_external_variable.sh
```diff
@@ -9,7 +9,10 @@ stderr=$(mktemp ${name}.err.XXXXXX)
 echo "stderr file: $stderr"
 
 $OSCAP oval eval --results $result --variables $srcdir/external_variables.xml $srcdir/$name.oval.xml 2> $stderr
-[ ! -s $stderr ] && rm $stderr
+# filter out the expected warnings in stderr
+sed -i -E "/^W: oscap:     Referenced variable has no values \(oval:x:var:[13689]\)/d" "$stderr"
```
jan-cerny: What if the indentation level will change in future? I would put here instead of 6 spaces a regular expression matching at least one space.

mpreisler: Fixed

jan-cerny:
@mpreisler Great job! Thank you.

PR #641, opened 2017-01-09
Make sure OSCAP_FULL_VALIDATION=1 is in TESTS_ENVIRONMENT of all the test Makefile.am files

mpreisler:
For some reason it was omitted in some cases. Probably copy paste errors. I have added it everywhere.

I had to blacklist 2 test cases in tests/nist, I commented why.

### DIFF
--- a/tests/nist/nist_test.sh
+++ b/tests/nist/nist_test.sh
@@ -15,7 +15,7 @@ function test_nist {
 	mkdir -p "${builddir}/tests/nist/$test_dir/"
 	(
 		# workaround for OVAL 5.4 issue where var_ref regex pattern is too
-		# strict and doesn't allow numbers or underscores
+		# strict and doesnt allow numbers or underscores
 		if [ "x$test_dir" == "xR1100" ]; then
 			unset OSCAP_FULL_VALIDATION
 		fi
--- a/tests/nist/nist_test.sh
+++ b/tests/nist/nist_test.sh
@@ -13,7 +13,19 @@ function test_nist {
 	test_dir="$1"
 	# make sure the output dir in builddir exists
 	mkdir -p "${builddir}/tests/nist/$test_dir/"
-	"${srcdir}/test_worker.py" --scanner "$OSCAP" --outputdir "${builddir}/tests/nist/$test_dir/" "${srcdir}/$test_dir/"
+	(
+		# workaround for OVAL 5.4 issue where var_ref regex pattern is too
+		# strict and doesn't allow numbers or underscores
+		if [ "x$test_dir" == "xR1100" ]; then
+			unset OSCAP_FULL_VALIDATION
+		fi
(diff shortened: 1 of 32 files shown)


mpreisler:
```
$ find . -name Makefile.am -print0 | xargs -r0 grep -L 'OSCAP_FULL_VALIDATION'
./API/CPE/Makefile.am
./API/XCCDF/Makefile.am
./API/Makefile.am
./probes/Makefile.am
./Makefile.am
```

All of these are Makefiles that don't do any testing themselves, they just reference their subdirs.

tests/nist/nist_test.sh
```diff
@@ -13,7 +13,19 @@ function test_nist {
...
+			unset OSCAP_FULL_VALIDATION
+		fi
+		# workaround for missing idref in <platform> in an internally used
+		# XCCDF. this XCCDF is never exported and only used internally.
+		if [ "x$test_dir" == "xR1200" ]; then
+			unset OSCAP_FULL_VALIDATION
+		fi
+		"${srcdir}/test_worker.py" --scanner "$OSCAP" --outputdir "${builddir}/tests/nist/$test_dir/" "${srcdir}/$test_dir/"
+	)
 	ret_val=$?
```
dahaic: I would vote for changing the subshell monster to remembering the variable before and resetting it back afterwards :). It currently breaks default vim syntax.. (and is ugly, anyway)

mpreisler: That does not work, check the source for details. We have to unset and we have to do it in a subshell. Unfortuantely, oscap checks whether OSCAP_FULL_VALIDATION is non-NULL, setting it to 0 or "" still makes it non-NULL.

mpreisler: Read https://github.com/OpenSCAP/openscap/blob/maint-1.2/utils/oscap-xccdf.c#L463 and weep :D

mpreisler: I guess we could unset and then reset in the same shell but the logic would be more complex than the subshell.

dahaic: Ok, this is quite ugly. But it does not break syntax highlight :)

`env | grep 'OSCAP_FULL_VALIDATION=' && local reset_command="OSCAP_FULL_VALIDATION=1"
...
eval $reset_command`

mpreisler: image

This is gvim. Could it be that you have some custom broken syntax highlighting plugins? I just use vim inbuilt for bash.

```
$ gvim --version
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Dec 15 2016 16:47:14)
Included patches: 1-133
Modified by <bugzilla@redhat.com>
Compiled by <bugzilla@redhat.com>
```

dahaic: I have checked it on various systems, and the issue is on RHEL7 machines, not present on Fedora.
highlight

PR #646, opened 2017-01-11
Check for errors when mounting image or guestfs domain in oscap-vm

mpreisler:
Otherwise we would still run oscap on an empty dir and potentially provide misleading results.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1391754


rsprudencio:
Looks great to me 😄

Issue #512, opened 2016-08-24
Hard-coded colors in terminal output


ybznek:
If somebody use white terminal, he will not see some output texts, because they have same color.

https://github.com/OpenSCAP/openscap/blob/maint-1.2/utils/oscap-xccdf.c#L297

I wanted to fix it, but we use all - grey/white/bold formats.

If I replace white/gray by "default", all these texts will be visually same.

We could use [italic]not applicable[/italic], but e.g. my terminal cannot display italic.
If we use [bold]not applicable[/bold] it will have same style as rule title.

Any tips how to don't use hard-coded colors and don't break much current styles?

shawndwells:
On 8/24/16 8:01 AM, Zbyněk Moravec wrote:
>
> If somebody use white terminal, he will not see some output texts,
> because they have same color.
>
> https://github.com/OpenSCAP/openscap/blob/maint-1.2/utils/oscap-xccdf.c#L297
>
> I wanted to fix it, but we use all - grey/white/bold formats.
>
> If I replace white/gray by "default", all these texts will be visually
> same.
>
> We could use [italic]not applicable[/italic], but e.g. my terminal
> cannot display italic.
> If we use [bold]not applicable[/bold] it will have same style as rule
> title.
>
> Any tips how to don't use hard-coded colors and don't break much
> current styles?
>

Pass:        Green
Fail:          Red
notapplicable:    Black (no action or concern needed)
notchecked:        Yellow. Maybe orange.

A good thought process on color selection:
http://www.neurotriggersagency.com/wp-content/uploads/2015/04/image2-northamericanculture.jpg

(while it says "North American Culture," I wager this is more international)

PR #647, opened 2017-01-11
Don't use white colors in oscap xccdf eval output

mpreisler:
A lot of users use white background terminals and white on white is not visible. Instead let's use default color and bold.

I thought about applying this to `maint-1.0` but wasn't sure either way. I think it's an enhancement and not a bugfix so I decided to go with `maint-1.2`.


mpreisler:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1365911

mpreisler:
Also fixes #512

yuumasato:
Ack.

PR #648, opened 2017-01-11
Word-wrap in pre elements in HTML guide and report

### LINKED ISSUE #504: HTML guide for lower screen resolution
When we open guide with lower resolution than full HD, horizontal scrollbar occurs, because content isn't resized properly.

We should reduce `<table>` size if user has lower resolution.

Please ensure, that inner content like remediation scripts is readable.

image

mpreisler:
This prevents the pre elements increasing width of the other HTML elements because of extremely long lines suggested for configuration files.

Fixes #504 

Requires CSS 2.1 but I think that's a fair requirement and is widely supported.


mpreisler:
image

yuumasato:
Ack

PR #649, opened 2017-01-12
Show severity in HTML guide

mpreisler:
Ran into this today when demoing HTML guide in a meeting. No idea why we didn't show severity in HTML guide.

I also have to say that the HTML guide looks much worse than our HTML report. And the filtering, searching and grouping features are not there. Wish we could bring some of those features to the HTML guide in the future.

I also sneaked in a bugfix. In HTML report we just assume severity attribute is there. This is not guaranteed by the specification! If severity is missing we should assume it's "unknown". I have made changes so that that's what gets displayed in HTML report and HTML guide.

image

### DIFF
--- a/xsl/xccdf-guide-impl.xsl
+++ b/xsl/xccdf-guide-impl.xsl
@@ -213,6 +213,11 @@ Authors:
                     </p>
                 </xsl:if>
 
+                <div class="severity">
+                    <span class="label label-warning">severity:</span>&#160;
+                    <xsl:call-template name="item-severity"><xsl:with-param name="item" select="$item" /></xsl:call-template>
+                </div>
+
                 <div class="identifiers">
                     <xsl:call-template name="item-idents-refs">
                         <xsl:with-param name="item" select="$item"/>
--- a/xsl/xccdf-guide-impl.xsl
+++ b/xsl/xccdf-guide-impl.xsl
@@ -214,8 +214,10 @@ Authors:
                 </xsl:if>
 
                 <div class="severity">
-                    <span class="label label-warning">severity:</span>&#160;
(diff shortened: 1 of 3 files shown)


mpreisler:
> I have one nitpick to consider though - "identifier" and "references" has some vertical space around. But newly added severity is glued to identifiers field. Can we keep the vertical pause there for consistency reasons?

I have also noticed that but I have no clue what causes this. In the code it's all `div` elements with no special styling that could cause this space. Any suggestions how to fix this appreciated.

yuumasato:
Adding a paragraph this fixes the vertical spacing.
screenshot from 2017-01-16 11-05-41

mpreisler:
@yuumasato Could you please just push that to my branch so I don't have to replicate your changes?

mpreisler:
I have wrapped severity in `<p>` element. Anything else?

yuumasato:
@mpreisler Actually I just edited the HTML in place in the browser.
And I guess that's just it, a `<p>`, element.

jan-cerny:
@mpreisler I have reviewed and tested this pull request and it looks good to me.

One nitpick though. The labels "Rationale:" and "Remediation Shell script:" start with a capital letter, whereas "references", "severity", "warning" and "identifiers" start with a small letter. I think it should be consistent.

mpreisler:
@jan-cerny You are not the first to notice that, it's a known issue that goes back to the HTML report and guide rewrite. The code that writes that out is shared between report and guide and in report we do capitalize.

It is unrelated to this PR, perhaps somebody else can fix that in a separate PR.

jan-cerny:
@mpreisler OK, ACK Then,

I have created an issue for the nitpick problem https://github.com/OpenSCAP/openscap/issues/655

PR #652, opened 2017-01-13
Add missing oval xsd imports

mpreisler:
OVAL 5.11.1 and 5.11 have some additional OVAL definition namespaces we are not importing. Found and fixed by @GaryGapinski


mpreisler:
We may also want to change https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/OVAL/public/oval_types.h#L41 and https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/OVAL/oval_enumerations.c#L368 but I will leave that for another PR.

jan-cerny:
ACK

Issue #634, opened 2017-01-05
Coverity issues in runlevel probe


mpreisler:
```
3. openscap-1.2.13/src/OVAL/probes/unix/runlevel.c:312: check_return: Calling "fgetc(osrelease)" without checking return value. This library function may fail and return an error code.
```

yuumasato:
A  possible fix is commented out in the source itself.
`/*const char c = */fgetc(osrelease);`

Did you have intentions of using the char?

mpreisler:
@yuumasato I added that there because of the compiler warning but I am not the author of this code. I was just fixing some issues and doing refactoring.

PR #654, opened 2017-01-17
Issue #634: Coverity: Check return value of fgetc

jan-cerny:
Adressing:
Calling "fgetc(osrelease)" without checking return value. This library
function may fail and return an error code.

Fixes https://github.com/OpenSCAP/openscap/issues/634


yuumasato:
Thank you, this looks good to me.

PR #661, opened 2017-01-21
manual: Mentioned fetch-remote-resources in more detail in Advanced o…

mpreisler:
…scap usage

Suggested for `maint-1.2` only because the messages are slightly different in `maint-1.0` and we don't support bzip2 there.


yuumasato:
The jenkins failure is due to not enough disk space.

yuumasato:
I've executed and checked the command and outputs and it is consistent.

PR #662, opened 2017-01-21
manual: Mention how to split source datastreams and how to view OVAL …

mpreisler:
…referenced

Only `maint-1.2`, the outputs are slightly different in `maint-1.0` and we don't support bzip2 there.


dahaic:
In other places,  we try to capitalize Source DataStream and Result DataStream, can you update it to be consistent?

PR #665, opened 2017-01-22
Manual more detailed idents

mpreisler:
For `maint-1.2` mainly because of group-by and identifier/reference lookup. Both of these are `maint-1.2` only.


rsprudencio:
@openscap-jenkins retest this

rsprudencio:

PR #667, opened 2017-01-22
Manual: More about vulnerability assessment

mpreisler:
All of this is applicable on `maint-1.0` and `maint-1.2`.


dahaic:
For posterity: my addition to the PR was based purely on maint-1.2  branch. This is why I have proposed to make it maint-1.2 PR for now.

yuumasato:
For posterity, we found out changes are actually applicable for both branches, but we agreed upon merging it only on maint-1.2.

PR #671, opened 2017-02-04
Terminate HTML guide generation if the profile ID doesn't match any p…

### LINKED ISSUE #656: Generating HTML guide with a nonexisting profile should fail
When I try to generate a HTML guide and I specify a profile that does not exist in the given datastream, the HTML guide is generated. 
[guide.tar.gz](https://github.com/OpenSCAP/openscap/files/714065/guide.tar.gz)

```
[jcerny@thinkpad content]$ oscap xccdf generate guide --profile ijklmnop --output /tmp/guide.html ssg-fedora-ds.xml
[jcerny@thinkpad content]$ echo $?
0
```
The resulting HTML guide contains no rules. I think that an error message should be displayed instead, and no guide should be generated.

I think my situation is similar to a situation when an user wants to scan his system and he specifies wrong profile.  But in that case he will get error:

```
[jcerny@thinkpad content]$  oscap xccdf eval --profile ijklmnop ssg-fedora-ds.xml 
Profile "ijklmnop" was not found. Get available profiles using:
$ oscap info "ssg-fedora-ds.xml"
```

I think that fixing this issue would help to avoid mistakes when generating guides.

mpreisler:
…rofile

Improves but doesn't fully fix #656, to fix it entirely (exit code = 1) we need to change to xsltApplyStylesheetUser and check contents of the context.

```
$ ./run ./utils/.libs/oscap xccdf generate guide --profile asdfasdfvzxcv ~/d/scap-security-guide/RHEL/7/output/ssg-rhel7-xccdf.xml
Profile "asdfasdfvzxcv" was not found. Get available profiles using "oscap info".
$ echo $?
0
```

### DIFF
--- a/xsl/xccdf-guide-impl.xsl
+++ b/xsl/xccdf-guide-impl.xsl
@@ -484,6 +484,9 @@ Authors:
     </xsl:call-template>
 
     <xsl:variable name="profile" select="$benchmark/cdf:Profile[@id = $profile_id]"/>
+    <xsl:if test="$profile_id != '' and not($profile)">
+        <xsl:message terminate="yes">Profile "<xsl:value-of select="$profile_id"/>" was not found. Get available profiles using "oscap info".</xsl:message>
+    </xsl:if>
 
     <xsl:text disable-output-escaping='yes'>&lt;!DOCTYPE html></xsl:text>
     <html lang="en">


rsprudencio:
Lgtm

PR #674, opened 2017-02-07
OVAL to XCCDF xsl

mpreisler:
Resurrecting an old useful XSLT written by @isimluk 

https://www.redhat.com/archives/open-scap-list/2013-May/msg00003.html

It is very useful when you need to evaluate a plain OVAL file using Satellite 5 or 6 or any other integration or scanning tool that doesn't support direct OVAL evaluation.

Usage:
```
$ xsltproc --stringparam ovalfile $OVAL oval-to-xccdf.xslt $OVAL
```


yuumasato:
Cool, but I think it would be nice if generated xccdf would also be consumable by oscap  and validate in NIST scapval tool.
IDs of generated Benchmark and Rule don't `_benchmark_`, `_rule_` or have prefix `xccdf_`.

mpreisler:
> Cool, but I think it would be nice if generated xccdf would also be consumable by oscap and validate in NIST scapval tool.
> IDs of generated Benchmark and Rule don't _benchmark_, _rule_ or have prefix xccdf_.

lol

I am just testing you, of course.

yuumasato:
Thank you, ack.

PR #680, opened 2017-02-14
Fix SCE stderr stalling

mpreisler:
We have been reading stdout and only then stderr. This caused stalls when stderr was filled up and stdout was still waiting for input. This PR changes the SCE stdout and stderr pipe reading to be interleaved. We do non-blocking reads from both pipes.

Thx goes to @dahaic for the test
Thx goes to @rsprudencio for the fcntl nonblock impl

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1420811

### DIFF
--- a/src/SCE/sce_engine.c
+++ b/src/SCE/sce_engine.c
@@ -34,6 +34,7 @@
 #include "common/util.h"
 #include "common/list.h"
 #include "common/oscap_acquire.h"
+#include "common/oscap_string.h"
 #include "sce_engine_api.h"
 
 #include <stdlib.h>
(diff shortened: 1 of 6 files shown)


src/SCE/sce_engine.c
```diff
@@ -327,6 +328,44 @@ void sce_parameters_allocate_session(struct sce_parameters* v)
...
+		const int read_status = read(fd, &readbuf, 1);
+		if (read_status == 1) {  // successful read
+			if (readbuf == '&') {
+				// & is a special case, we have to "escape" it manually
+				// (all else will eventually get handled by libxml)
+				oscap_string_append_string(string, "&amp;");
+			} else {
+				oscap_string_append_char(string, readbuf);
+			}
+			//printf("Read %i from fd=%i\n", readbuf, fd);
```
ybznek: I think, it should be removed or in one of debug macros

mpreisler: Can't use debug macros, SCE is outside libopenscap. I can remove the printfs if you want, I figured they'd be useful to whoever was testing/reviewing this.

ybznek: OK, makes sense. Dont need to remove it.

mpreisler: Crap, I have removed them in the meantime :D Doesn't matter though.

mpreisler:
Please squash to one commit when merging.

PR #684, opened 2017-02-20
Recategorize some warning messages

jan-cerny:
After merging https://github.com/OpenSCAP/openscap/pull/630 some new messages have appeared on stderr. When scanning SSG Fedora common profile using the latest OpenSCAP from git I got stderr completely convoluted with warnings. Most of them were ```Getting extended ACL for file ```... and ``lstat() failed``. However these messages don't signalise anything harmful. I propose moving them to a lower verbosity level.


mpreisler:
Good stuff, we may have to do more of these as more warnings show up.

PR #522, opened 2016-09-06
ansible generate fix - add header & indentation 

ybznek:
Example of usage:
```
$ myoscap xccdf generate fix --template urn:xccdf:fix:script:ansible /tmp/res.xml | tee  /tmp/ansible.yml
---
 - hosts: localhost # set required host
   tasks:

    - name: Ensure aide is installed
      package:
        name="{{item}}"
        state=present
      with_items:
        - aide

    - name: Ensure vsftpd is installed
      package:
        name="{{item}}"
        state=present
      with_items:
        - vsftpd

```
```
$ sudo ansible-playbook /tmp/ansible.yml 
[sudo] password for zmoravec: 
 [WARNING]: provided hosts list is empty, only localhost is available


PLAY [localhost] ***************************************************************

TASK [setup] *******************************************************************
ok: [localhost]

TASK [Ensure aide is installed] ************************************************
changed: [localhost] => (item=[u'aide'])

TASK [Ensure vsftpd is installed] **********************************************
ok: [localhost] => (item=[u'vsftpd'])

PLAY RECAP *********************************************************************
localhost                  : ok=3    changed=1    unreachable=0    failed=0   


```

PR #686, opened 2017-02-24
Fix missing header in result-oriented Ansible remediations

jan-cerny:
When we generate Ansible playbook from an entire profile
we include a header at the beginning of the generated playbook.
However, when we generate fixes only for failed rules based on
OpenSCAP scan result (so-called result-oriented fixes), we don't
include any header. This commit will cause including a header
in result-oriented fixes as well.


jan-cerny:
See how it is done for profile-oriented fixes in https://github.com/OpenSCAP/openscap/pull/522

mpreisler:
ACK

Issue #681, opened 2017-02-15
Check for fcntl() return value


jan-cerny:
In `src/SCE/sce_engine.c` on lines 564 and 565 we should check for return value of `fcntl(stdout_pipefd[0], F_SETFL, flag_stdout | O_NONBLOCK);` to ensure flag was set correctly on the file descriptor.

mpreisler:
http://man7.org/linux/man-pages/man2/fcntl.2.html

We need to check whether the return value is different from 0. If it is we should abort and print an error.

rsprudencio:
Fixed in #688

PR #688, opened 2017-02-27
Issue #681: Check for return value of fcntl

jan-cerny:
The fcntl function manipulates file descriptors. It may fail.
On error, it returns -1. We should check for the return value,
and if it fails, cancel the operation and display an error
message.


rsprudencio:
Great work :smile:

PR #690, opened 2017-02-28
Introduce a test for XCCDF 1.1 to XCCDF 1.2 transformation

jan-cerny:
We ship an XSLT template that allows us to convert older
XCCDF files in XCCDF 1.1 to XCCDF 1.2. The template can be found
in xsl/xccdf_1.1_to_1.2.xsl. However we don't test the transformation.
Since the transforamtion is mentioned in user manual, we should
have a test that ensures it is working. This commit introduces
a simple test based on XCCDF content from NIST test suite.

### DIFF
--- a/tests/API/XCCDF/unittests/Makefile.am
+++ b/tests/API/XCCDF/unittests/Makefile.am
@@ -193,6 +193,7 @@ EXTRA_DIST += \
 	test_xccdf_sub_title.xccdf.xml \
 	test_xccdf_test_system.sh \
 	test_xccdf_test_system.xccdf.xml \
+	test_xccdf_transformation.sh \
 	test_xccdf_xml_escaping_value.sh \
 	test_xccdf_xml_escaping_value.xccdf.xml \
 	unit_helper.c \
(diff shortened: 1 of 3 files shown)


mpreisler:
NACK.

This uses xsltproc which we never check for during configure or while running the check. We should at least check that it's in path (see `hash` command). If it's not we should skip the test.

mpreisler:
The test fails for an unrelated reason though.

```
$ /usr/bin/xpath /tmp/test_xccdf_transformation.xccdf12.MHDuQn 'count(/Benchmark[namespace-uri()="http://checklists.nist.gov/xccdf/1.2"])'
namespace-uri: Function not supported
```

dahaic:
@mpreisler when this task was discussed, you also mentioned regressions happened in the past - is simple validation enough, or should we add target XCCDF, and compare those two to be the same? My point is there is actually no check the content is correct. It might create some empty template, and it will pass this test.

jan-cerny:
@mpreisler Thanks for review. It seems to me that this function is not available on RHEL6. I will try to find a better solution. I will also add a check to ensure `xsltproc` is available (I believed that it always comes together with libxslt).

mpreisler:
@jan-cerny The fixes look great. ACK

@dahaic That is true but I appreciate that it's really hard to write a good test for this kind of stuff. We could check verbatim against a good XCCDF 1.2 but that we are basically checking that the behavior never changes. I think this improves the situation so we can merge it and improve the test later.

Issue #664, opened 2017-01-22
Show more info about ARF files in `oscap info`


mpreisler:
This is pathetic and we have all the API we need to improve it:

```
$ oscap info arf.xml
Document type: Result Data Stream

Asset: asset0
 - collection1 -> xccdf1
```

Show evaluation start and end dates at least. And possibly CPEs of the asset.

mpreisler:
Similar situation with XCCDF test results:

```
$ oscap info xccdf-results.xml 
Document type: XCCDF Checklist
Checklist version: 1.1
Imported: 2017-01-21T19:17:35
Status: draft
Generated: 2017-01-19
Resolved: true
Profiles:
        standard
        pci-dss
        C2S
        rht-ccp
        common
        stig-rhel7-workstation-upstream
        stig-rhel7-server-gui-upstream
        stig-rhel7-server-upstream
        stig-rhevh-upstream
        ospp-rhel7-server
        nist-cl-il-al
        cjis-rhel7-server
        docker-host
        nist-800-171-cui
Referenced check files:
        ssg-rhel7-oval.xml
                system: http://oval.mitre.org/XMLSchema/oval-definitions-5
        ssg-rhel7-ocil.xml
                system: http://scap.nist.gov/schema/ocil/2
        https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2
                system: http://oval.mitre.org/XMLSchema/oval-definitions-5
Test Results:
        xccdf_org.open-scap_testresult_common
```

It should list some more info for the Test Results. I am putting it here because the changes will most likely be closely related.

jan-cerny:
@ybznek Do you still work on this?

mpreisler:
Fixed in https://github.com/OpenSCAP/openscap/pull/693

PR #693, opened 2017-03-02
Issue #664: Show more information about ARF files in "oscap info"

jan-cerny:
This commit will display additional information about an ARF file,
such as evaluation date, source benchmark and profile, platform CPEs,
result ID.

See the improvement:

Before:
```
[jcerny@thinkpad arf_info]$ oscap info arf.xml
Document type: Result Data Stream

Asset: asset0
 - collection1 -> xccdf1
```

After:
```
[jcerny@thinkpad arf_info]$ oscap info arf.xml
Document type: Result Data Stream
Imported: 2017-03-02T12:05:07

Asset: asset0
	ARF report: xccdf1
		Result ID: xccdf_org.open-scap_testresult_xccdf_org.ssgproject.content_profile_common
		Source benchmark: /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml
		Source profile: xccdf_org.ssgproject.content_profile_common
		Report request: collection1
		Evaluation started: 2017-03-02T12:02:02
		Evaluation finished: 2017-03-02T12:05:07
		Platform CPEs:
			cpe:/o:fedoraproject:fedora:24
```

### DIFF
--- a/tests/API/XCCDF/unittests/test_xccdf_results_arf_no_oval.sh
+++ b/tests/API/XCCDF/unittests/test_xccdf_results_arf_no_oval.sh
@@ -6,7 +6,8 @@ set -o pipefail
 name=$(basename $0 .sh)
 result=$(mktemp -t ${name}.out.XXXXXX)
 resultArf=$(mktemp -t ${name}.arf.out.XXXXXX)
-stderr=$(mktemp -t ${name}.out.XXXXXX)
+stdout=$(mktemp -t ${name}.out.XXXXXX)
+stderr=$(mktemp -t ${name}.err.XXXXXX)
 
 $OSCAP xccdf eval --results $result --results-arf $resultArf $srcdir/${name}.xccdf.xml 2> $stderr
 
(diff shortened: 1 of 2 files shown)


mpreisler:
Strange failure in jenkins, could be related to the changes:
`Exported arf results from xccdf without reference to oval   FAIL: all.sh`

@openscap-jenkins test this again

mpreisler:

mpreisler:
@jan-cerny Do you think we could do something similar for XCCDF test result? Right know we assume it's a benchmark and we never list the TestResult elements. The code would most likely be closely related to the ARF info code.

jan-cerny:
@mpreisler  The failure in Jenkins is related. The test https://github.com/OpenSCAP/openscap/blob/maint-1.2/tests/API/XCCDF/unittests/test_xccdf_results_arf_no_oval.sh#L20 greps for
 `- collection1 -> xccdf1` which I have removed and replaced by something more meaningful.

Does something depend on keeping that string in stdout? If not, I would change the test and also improve it to test the new feature.

I think that we could do something similar for XCCDF results, because here I also use functions from xccdf_result module. I will look into that and let you know. Thanks for suggestion 👍

mpreisler:
> Does something depend on keeping that string in stdout? If not, I would change the test and also improve it to test the new feature.

I do not think we need to keep that. IMO feel free to change the test.

jan-cerny:
OK, I will change the test.

jan-cerny:
I think it's ready for review.

mpreisler:
Greak work, @jan-cerny ! ACK

PR #695, opened 2017-03-03
Updated test coverage regarding CPE in generate fix scenario.

dahaic:
Test covers generation from profile or from results, when CPE
is defined. Expected outcome is that CPE is ignored for profile
generate fix.

### DIFF
--- a/tests/DS/test_ds.sh
+++ b/tests/DS/test_ds.sh
@@ -141,9 +141,55 @@ function test_eval {
     diff /dev/null $stderr; rm $stderr
 }
 
-function test_generate_fix {
+function test_eval_cpe {
+    local stdout=$(mktemp -t ${name}.out.XXXXXX)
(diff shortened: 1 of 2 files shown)


yuumasato:
Error in process58 probe.

tests/DS/test_ds.sh
```diff
@@ -141,9 +141,55 @@ function test_eval {
     diff /dev/null $stderr; rm $stderr
 }
 
-function test_generate_fix {
+function test_eval_cpe {
+    local stdout=$(mktemp -t ${name}.out.XXXXXX)
+    local stderr=$(mktemp -t ${name}.err.XXXXXX)
+    local ret=0
+
+    $OSCAP xccdf eval --progress "${srcdir}/$1" 1> $stdout 2> $stderr || ret=$?
```
yuumasato: Very clever to use `--progress` to check individual result of rules.

PR #697, opened 2017-03-03
Show Anaconda snippets labeled properly in HTML report & guide

mpreisler:
I noticed that we don't show Anaconda snippets labeled properly and we use them quite a bit in SSG. This PR fixes that by introducing proper labels in HTML report and guide.

image


yuumasato:
Ack.

PR #699, opened 2017-03-07
RHBZ #1414303, #1414312: Fix a segfault in rpm probes

jan-cerny:
Function probe_init() can possibly fail and return NULL.
We should check if initialization has been done properly
before we start the function.
Now we don't check the return value, which can possibly lead
to synchronization errors between threads and causes
a segmentation fault in probe_fini().

### DIFF
--- a/src/OVAL/probes/unix/linux/dpkginfo.c
+++ b/src/OVAL/probes/unix/linux/dpkginfo.c
@@ -91,6 +91,9 @@ int probe_main (probe_ctx *ctx, void *arg)
         struct dpkginfo_reply_t *dpkginfo_reply = NULL;
         int errflag;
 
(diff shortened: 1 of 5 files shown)


jan-cerny:
This goes to maint-1.0 because it's a small bugfix in old code.

mpreisler:
@jan-cerny Shouldn't we start checking for this in all the probes? Not just rpm probes?

jan-cerny:
@mpreisler Yes, we can. I recall in some of probes we already do that, but in most of them we don't. Sometimes it would not be necessary, because sometimes probe_init() functions "never fail" or simply don't initialise anything. But it will definitely make the code less error-prone.

mpreisler:
ACK

PR #702, opened 2017-03-14
Changed from Error message when selector id doesn't exist in Benchmar…

### LINKED ISSUE #614: Exit status is 0 when scanner fails to find selector ID in Benchmark
When running a scan against profile from a tailoring file and the tailoring file references a benchmark that is not in the XCCDF file, the scanner outputs the rules that could not be found:

`OpenSCAP Error: Selector ID(xccdf_org.ssgproject.content_rule_firefox_preferences-on-secure_page_warning) does not exist in Benchmark.`

The resulting report has all rules from XCCDF file marked as 'notselected':
[arf_report.xml.gz](https://github.com/OpenSCAP/openscap/files/651547/arf_report.xml.gz), which makes sense, but the exit status is 0.

Steps to reproduce:
1) Use SCAP Workbench to create a tailoring file (based on Upstream Firefog STIG), unselect at least one rule.
2) run a scan with tailoring file created in the previous step. Supply a XCCDF file different from what was used to create tailoring file (ssg-jave-ds.xml): 

`oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_stig-firefox-upstream_customized --tailoring-file ssg-firefox-ds-tailored.xml --results-arf arf.xml ssg-java-ds.xml`

3) Notice there will be errors (for each rule that was deselected in tailoring).
4) run `echo $?`, it will be 0

Is this intended bahavior? Why would exit status be 0, if there were clearly errors?

rsprudencio:
…k to a Warning message

Fixes #614

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy.c
+++ b/src/XCCDF_POLICY/xccdf_policy.c
@@ -1666,7 +1666,7 @@ _xccdf_policy_add_selector_internal(struct xccdf_policy *policy, struct xccdf_be
 	/** If the selector doesn't point to an item it can still point to the cluster. */
 	struct oscap_htable_iterator *hit = xccdf_benchmark_get_cluster_items(benchmark, xccdf_select_get_item(sel));
 	if (hit == NULL) {
-		oscap_seterr(OSCAP_EFAMILY_XCCDF, "Selector ID(%s) does not exist in Benchmark.", xccdf_select_get_item(sel));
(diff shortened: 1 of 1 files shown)


jan-cerny:
@rsprudencio We don't usually start our warnings with `"Warning: "` because there is a `W:` automatically printed out on the beginning of the line when using `dW()` macro.

rsprudencio:
@jan-cerny true, updated :smile:

jan-cerny:
Thanks!

Issue #513, opened 2016-08-30
Manual.adoc - change links to updated guides/results/...


ybznek:
https://github.com/OpenSCAP/openscap/blob/maint-1.2/docs/manual/manual.adoc#34-content-transformation

I think we could use some newer examples in the documentation.

jan-cerny:
Fixed in https://github.com/jan-cerny/openscap/commit/d4fb56e40cbc7877e121c2b585233064bbf3a516
New examples stored on static.open-scap.org.

PR #703, opened 2017-03-14
Issue #513: Provide better examples in user manual

jan-cerny:
In our examples, we referenced to old-style guides and reports
as generated by OpenSCAP 1.0.x. Instead we will show our new
nice reports and guides.

### DIFF
--- a/docs/manual/manual.adoc
+++ b/docs/manual/manual.adoc
@@ -833,32 +833,32 @@ human-readable form. The following commands represent the most common
 cases:
 
 * Creating a guide (see an
-http://mpreisle.fedorapeople.org/openscap/guide.html[example]):
+https://static.open-scap.org/examples/guide.html[example]):
 --------------------------------------------------------
 $ oscap xccdf generate guide scap-xccdf.xml > guide.html
 --------------------------------------------------------
 
 * Creating a guide with profile checklist (see an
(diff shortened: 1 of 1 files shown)


docs/manual/manual.adoc
```diff
@@ -833,32 +833,32 @@ human-readable form. The following commands represent the most common
...
 * Generating the XCCDF scan report (see an
-http://mpreisle.fedorapeople.org/openscap/report-xccdf.html[example]):
+https://static.open-scap.org/examples/report-xccdf.html[example]):
 -------------------------------------------------------------------
 $ oscap xccdf generate report xccdf-results.xml > report-xccdf.html
 -------------------------------------------------------------------
 
 * Generating the OVAL scan report (see an
-http://mpreisle.fedorapeople.org/openscap/report-oval.html[example]):
+https://static.open-scap.org/examples/report-oval.html[example]):
```
rsprudencio: This report is the same as the old one hosted at fedorapeople.org, was that intentional?

jan-cerny: @rsprudencio Yeah, because OVAL reports still look the same as before.

rsprudencio: Ok, great work :smile:

PR #705, opened 2017-03-14
Added verbose support for `oscap xccdf remediate` module

### LINKED ISSUE #520: verbose mode with oscap xccdf remediate doesn't work
Version:
latest oscap version in Fedora24 & latest upstream maint-1.2

`oscap xccdf remediate  --verbose INFO --verbose-log-file /tmp/verbose.xml --report /tmp/rep.xml  /tmp/res.xml`

`--verbose` parameters are correctly validated, but verbose mode doesn't print anything.

/tmp/verbose.xml doesn't exist after command run, so it seems that verbose mode doesn't work completely for this use case.

rsprudencio:
Fixes #520


jan-cerny:
Thank you!

Issue #249, opened 2015-12-04
Docs: User manual should include section about oscap-ssh/oscap-vm/oscap-docker


isimluk:
Thanks!

mpreisler:
This ticket is up for grabs. I think we can still deliver it for 1.2.14.

rsprudencio:
Fixed in #706

PR #706, opened 2017-03-14
Issue #249: Mention oscap-ssh, oscap-docker in User's manual

jan-cerny:
This adds a few paragraphs into OpenSCAP User Manual, however I didn't want to copy everything from man pages.


rsprudencio:
Great, thanks!

PR #708, opened 2017-03-17
RHBZ #1431186: Fix crashing systemd probes

jan-cerny:
Systemd probes are crashing during scanning Docker
containers. This probably happens because containers
do not have systemd running inside them by default.
Instead, we will report an error.


mpreisler:

jan-cerny:
I think there is a fail in OVAL unit tests. I will investigate it.

jan-cerny:
The fail should be unrelated to this, but is related to patch https://github.com/OpenSCAP/openscap/commit/8585318d02649ed5f30721394a4dfd7f558fd41f that I pushed directly. Unfortunately it had a side effect that I wasn't aware of.  A fix is proposed https://github.com/OpenSCAP/openscap/pull/710

mpreisler:

mpreisler:
ACK

Issue #707, opened 2017-03-14
Error messages are duplicated


jan-cerny:
Errors that are reported using oscap_seterr are displayed twice.

For example:

```
$ oscap xccdf eval dsadasd
E: oscap: (oscap_source.c:284:oscap_source_get_xmlDoc()) Unable to open file: 'dsadasd'
OpenSCAP Error: Unable to open file: 'dsadasd' [oscap_source.c:284]
```

This is probably caused by https://github.com/OpenSCAP/openscap/pull/630

jan-cerny:
See https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/common/error.c#L57

PR #710, opened 2017-03-20
Do not rely on number of lines on stderr

jan-cerny:
We fixed duplicate error messages (issue #707). But in this test,
we assumed that there are 6 lines on stderr. Actually two of them
were duplicated. We should not count number of lines on stderr
because we may add a new warning somewhere in our code and it will
break the test again. This commit stops testing stderr line count.

### DIFF
--- a/tests/API/OVAL/unittests/test_object_component_type.sh
+++ b/tests/API/OVAL/unittests/test_object_component_type.sh
@@ -8,8 +8,6 @@ set -o pipefail
 $OSCAP oval eval $srcdir/test_object_component_type.oval.xml 2> $stderr || ret=$?
 [ $ret -eq 1 ]
 
-stderr_line_count=`cat $stderr | wc -l`
-[ $stderr_line_count -eq 6 ]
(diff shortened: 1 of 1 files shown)


mpreisler:
Why not check for 3 lines? The idea of the test is to make sure we do output errors in stderr and we keep doing that.

jan-cerny:
@mpreisler The test still greps for some messages on stderr.
What if we add a new warning?

mpreisler:
> @mpreisler The test still greps for some messages on stderr.
> What if we add a new warning?

Then we change the test to grep for different warnings. It's not ideal, it kind of sucks but I don't know how to test for this any better.

jan-cerny:
@mpreisler I'm not sure if we are on the same page. What I think is wrong about the test is that we have harcoded number of lines here. I propose to remove the check with "wc -l", but at the same time I keep checking that a specific message is on stderr. Am I missing something? How can I do this better?

mpreisler:
Ah, OK, since we are grepping for the warnings I don't think we need to check number of lines.

PR #717, opened 2017-04-03
Improve test of Ansible fixes

jan-cerny:
We shouldn't change the format of the output to ensure
the fix generated by oscap is consumable by Ansible.


yuumasato:
Ack.

PR #719, opened 2017-04-04
Change category of a message, emit warning only if necessary

jan-cerny:
It is pretty common to have variables with no values, but
the warnings flood the stderr after we have enabled them by default.
Still, this will keep informing user if there was a problem.


mpreisler:

rsprudencio:
LGTM, Thanks :smile:

PR #721, opened 2017-04-13
Updates for Python 3 bindings

Hexadorsimal:
Added a Vagrantfile with ansible provisioning for easy creation of a build environment.
Updated the swig files so all the public functions get exported to python
Changed enable python3 to default to yes
Added some extra accessor and mutator functions so we can export functioning XCCDF and OVAL files made from scratch in python.

### DIFF
--- a/swig/python2/Makefile.am
+++ b/swig/python2/Makefile.am
@@ -11,6 +11,7 @@ AM_CPPFLAGS =	@xml2_CFLAGS@ \
 		-I${top_srcdir}/src/common/public \
 		-I${top_srcdir}/src/source/public \
 		-I$(top_srcdir)/src/CVE/public \
+		-I$(top_srcdir)/src/DS/public \
 		-I/usr/include -include "sys/types.h" \
 		-D_LARGEFILE64_SOURCE \
 		-D_LARGEFILE_SOURCE=1
--- a/swig/python3/Makefile.am
+++ b/swig/python3/Makefile.am
@@ -10,6 +10,7 @@ AM_CPPFLAGS =	@xml2_CFLAGS@ \
 		-I${top_srcdir}/src/common/public \
 		-I${top_srcdir}/src/source/public \
 		-I$(top_srcdir)/src/CVE/public \
+		-I$(top_srcdir)/src/DS/public \
 		-I/usr/include -include "sys/types.h" \
 		-D_LARGEFILE64_SOURCE \
 		-D_LARGEFILE_SOURCE=1
--- a/swig/src/openscap.i
+++ b/swig/src/openscap.i
@@ -215,6 +215,17 @@
 
 %module openscap
 %{
+ #include "../../src/DS/public/ds_rds_session.h"
+ #include "../../src/DS/public/ds_sds_session.h"
+ #include "../../src/DS/public/scap_ds.h"
+%}
+%include "../../src/DS/public/ds_rds_session.h"
+%include "../../src/DS/public/ds_sds_session.h"
+%include "../../src/DS/public/scap_ds.h"
+
+
+%module openscap
+%{
  #include "../../src/XCCDF/public/xccdf_benchmark.h"
  #include "../../src/XCCDF/public/xccdf_session.h"
  #include "../../src/XCCDF_POLICY/public/xccdf_policy.h"
@@ -226,22 +237,32 @@
 
 %module openscap
 %{
+ #include "../../src/OVAL/public/oval_adt.h"
  #include "../../src/OVAL/public/oval_agent_api.h"
  #include "../../src/OVAL/public/oval_definitions.h"
+ #include "../../src/OVAL/public/oval_directives.h"
  #include "../../src/OVAL/public/oval_system_characteristics.h"
  #include "../../src/OVAL/public/oval_results.h"
+ #include "../../src/OVAL/public/oval_schema_version.h"
+ #include "../../src/OVAL/public/oval_session.h"
  #include "../../src/OVAL/public/oval_types.h"
  #include "../../src/OVAL/public/oval_variables.h"
+ #include "../../src/OVAL/public/oval_version.h"
  #include "../../src/OVAL/public/oval_probe.h"
  #include "../../src/OVAL/public/oval_probe_handler.h"
  #include "../../src/OVAL/public/oval_probe_session.h"
 %}
+%include "../../src/OVAL/public/oval_adt.h"
 %include "../../src/OVAL/public/oval_agent_api.h"
 %include "../../src/OVAL/public/oval_definitions.h"
+%include "../../src/OVAL/public/oval_directives.h"
 %include "../../src/OVAL/public/oval_system_characteristics.h"
 %include "../../src/OVAL/public/oval_results.h"
+%include "../../src/OVAL/public/oval_schema_version.h"
+%include "../../src/OVAL/public/oval_session.h"
 %include "../../src/OVAL/public/oval_types.h"
 %include "../../src/OVAL/public/oval_variables.h"
+%include "../../src/OVAL/public/oval_version.h"
 %include "../../src/OVAL/public/oval_probe.h"
 %include "../../src/OVAL/public/oval_probe_handler.h"
 %include "../../src/OVAL/public/oval_probe_session.h"
@@ -251,6 +272,12 @@
  #include "../../src/OVAL/public/oval_agent_xccdf_api.h"
 %}
 %include "../../src/OVAL/public/oval_agent_xccdf_api.h"
+
+%module openscap
+%{
+ #include "../../src/source/public/oscap_source.h"
+%}
+%include "../../src/source/public/oscap_source.h"
 
 /*%typemap(in) value[ANY] {
 


mpreisler:
Note: Breaks API and ABI, do not merge yet.

mpreisler:

configure.ac
```diff
@@ -10,7 +10,7 @@ AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_AUX_DIR([config])
 AC_CONFIG_MACRO_DIR([m4])
 
-AM_INIT_AUTOMAKE([foreign tar-pax])
+AM_INIT_AUTOMAKE([foreign tar-pax subdir-objects])
```
mpreisler: We can't do this because it breaks compatibility with old automake. there is no subdir-objects in old automake.

configure.ac
```diff
@@ -1510,10 +1510,10 @@ AC_CONFIG_FILES([Makefile
...
-		tests/codestyle/Makefile
-		tests/oval_details/Makefile
-		tests/nist/Makefile
+                 tests/bz2/Makefile
+                 tests/codestyle/Makefile
+                 tests/oval_details/Makefile
+                 tests/nist/Makefile
 
                  src/SCE/Makefile
                  tests/sce/Makefile])
```
mpreisler: Please don't change indents, this just creates merge conflicts for no real gain. Plus we use tabs, not spaces.

configure.ac
Hexadorsimal: Ok.  It was just throwing a ton of warnings at me.  I've never submitted a merge request before.  What happens now?  Should I change it?

configure.ac
Hexadorsimal: Ok sorry.  I just thought I was fixing it because it was a mix of tabs and spaces.

mpreisler: No problem. I would do the same thing to make it look prettier but the constant merge conflicts just aren't worth it :-) As a general rule of thumb most people will change indents to the right indents if they are also changing that particular line.

configure.ac
mpreisler: Yeah, you have a couple options. You could use `git rebase -i maint-1.2` and choose to edit the commit to remove this change.

mpreisler:
This contains some great work but right now we can't merge it all. I will take pieces that can be merged immediately and then I suggest we rebase and take it from there.

2574a349ca1c28ac8919b639811f17c6f68c9866 merged - benchmark supported version_info

mpreisler:
https://github.com/OpenSCAP/openscap/pull/721/commits/cc317a2c36326878bf8f0f7c9592deb4761ffe19 is impossible for me to merge, too many things happening at once, the subdir option breaks build on old distros, whitespace changes, etc... I will skip that for now.

mpreisler:
We will pass on vagrant and ansible for now. They are useful but let's not keep that in the upstream git. OpenSCAP is not specific to F25 and there are countless ways to build it.

swig/python2/Makefile.am
```diff
@@ -3,14 +3,15 @@ AM_CFLAGS = -fPIC
...
+		-I$(top_srcdir)/src/CVE/public \
 		-I$(top_srcdir)/src/CVSS/public \
+		-I$(top_srcdir)/src/DS/public \
 		-I$(top_srcdir)/src/OVAL/public \
+		-I${top_srcdir}/src/source/public \
 		-I$(top_srcdir)/src/XCCDF/public \
 		-I$(top_srcdir)/src/XCCDF_POLICY/public \
-		-I${top_srcdir}/src/common/public \
-		-I${top_srcdir}/src/source/public \
-		-I$(top_srcdir)/src/CVE/public \
```
mpreisler: why the reorg here? The real change is addition of DS/public but it's getting lost because you moved a bunch of other includes around.

Hexadorsimal: I was alphabetizing it to make it easier to see what was missing.  It's not really important, I was just making myself too comfortable and pretending it was my own project, because that's what I would do.

src/OVAL/public/oval_definitions.h
```diff
@@ -1828,15 +1901,15 @@ struct oval_component *oval_variable_get_component(struct oval_variable *);	//ty
  * It should be freed after use by the calling application.
  * @memberof oval_variable
  */
-struct oval_iterator *oval_variable_get_possible_values(struct oval_variable *variable);
+struct oval_variable_possible_value_iterator *oval_variable_get_possible_values(struct oval_variable *variable);
```
mpreisler: We can't do this in a maintenance branch. This breaks API.

mpreisler:
Merged aaf774d1336781d7f23346299ecf67d4af638fa0 - oval_object_content_set_filter and oval_object_content_get_filter

src/OVAL/public/oval_definitions.h
Hexadorsimal: Breaks it how?  As far as I could tell it was already broken, at least from the python perspective.  I had access to oval_variable_get_possible_values() which returned an oval_iterator but there are zero exposed functions that can work with a plain oval_iterator.  I considered exposing the plain functions, but decided it would be better to make special iterator functions like you do for every other iterator.

Hexadorsimal: Oh, because people using the C version are expecting the plain iterators.  Gotcha.

swig/python2/Makefile.am
mpreisler: Ah, understood. The reason why I frown upon changes like this is because this is in multiple branches and will cause conflicts as we forward-port fixes.

src/OVAL/public/oval_definitions.h
mpreisler: Changing signatures of public functions breaks API and possibly ABI. We simply can't do that in maintenance branches, no exceptions. Distributions wouldn't be able to easily update to lates openscap upstream if we broke API/ABI. Even if the function itself was broken the best we can do is create a function with a different name that does it right and deprecate the broken function in maintenance branch, to be removed in the future.

Hexadorsimal: I guess if we wanted to maintain API compatibility AND fix python we'd have to expose the plain iterator functions.

mpreisler:
Want to merge https://github.com/OpenSCAP/openscap/pull/721/commits/9026368c2d795c422cf39525326e3af16dd25125 with some minor formatting changes but I get conflicts. @Hexadorsimal could you please rebase against fresh `maint-1.2`?

swig/python2/Makefile.am
shawndwells: Total fan of alphabetically sorting things, if we can keep consistency between branches. Just did alphabetical sorting of rule selectors in an SSG profile (https://github.com/OpenSCAP/scap-security-guide/pull/1963)

/me ducks and runs

mpreisler: As long as it's done separately and in `maint-1.0` and merged all the way up it's doable. Mixing it with actual changes backfires badly when doing `git blame` or `git bisect`.

src/OVAL/public/oval_definitions.h
mpreisler: @Hexadorsimal or introduce `oval_variable_get_possible_values2` and deprecate `oval_variable_get_possible_values`. It's sad but that's what we have to do to avoid problems for existing users of the API.

Hexadorsimal:
How do I rebase against fresh main-1.2?

mpreisler:
> How do I rebase against fresh main-1.2?

@Hexadorsimal depends on your set-up but here is what I do.

```
$ git remote -v
fork    git@github.com:mpreisler/openscap.git (fetch)
fork    git@github.com:mpreisler/openscap.git (push)
upstream  git@github.com:OpenSCAP/openscap.git (fetch)
upstream  git@github.com:OpenSCAP/openscap.git (push)

# make sure you have both your fork and the upstream repo there.
# You may have to use the https upstream repo because you don't have rights to push into it.

$ git checkout maint-1.2
# pull upstream changes into your local maint-1.2 branch
$ git pull upstream maint-1.2
# go back to your feature branch that is based on old version of maint-1.2
$ git checkout my_feature_branch
# rebase the feature branch so that it is based on the new HEAD of maint-1.2
$ git rebase -i maint-1.2
# the --force below is necessary because this will be a non-fw push
$ git push --force fork my_feature_branch
```

HTH!

Hexadorsimal:
Did I get it?

mpreisler:
> Did I get it?

Nope, what you did is merge upstream maint-1.2. Probably by doing `git pull upstream maint-1.2` while having the feature branch checked out. This is inferior to actual rebasing because it will make the git history very complex in the long run.

Hexadorsimal:
Okay, I'm going to go read about rebasing.  This is embarrassing.

Hexadorsimal:
So when I fork the project, should I also be creating a feature branch to work on and not just working on the maint-1.2 branch in my fork?

mpreisler:
> So when I fork the project, should I also be creating a feature branch to work on and not just working on the maint-1.2 branch in my fork?

It's 100% up to you what you do in your fork. However I strongly recommend always doing `git checkout -b new_feature_123` simply because you can work on multiple PRs that way and it makes everything more manageable.

Hexadorsimal:
I'm pretty sure it's rebased now.

mpreisler:
> I'm pretty sure it's rebased now.

It's rebased but it's all based on a commit from April 7. Could you please pull latest changes from upstream and rebase on top of that?

Hexadorsimal:
How does this look now?

Hexadorsimal:
Probably pretty good.  It's just a single commit now.

mpreisler:

mpreisler:
Looking good, if this passes jenkins and my testing I will merge it.

mpreisler:
Good stuff, thanks for your contribution @Hexadorsimal !

Hexadorsimal:
Thanks for your patience!  That was a very good learning experience for me.  I'd never rebased anything before.

PR #725, opened 2017-04-24
Fixing sysctl test on s390 architecture

dahaic:


jan-cerny:
@dahaic This should go to maint-1.0 branch.

PR #728, opened 2017-04-28
Easier language and more colors in OVAL details

mpreisler:
I am always confused when reading the OVAL details and have to read the lines several times to understand it. This PR aims to improve that. I use colors to highlight which condition passed and failed. I use simpler language - passed or failed - instead of satisfying and violating.

Before:
image

image

After:
image

image

I feel like this makes the report way easier to read, especially for users not familiar with OVAL and XCCDF. Let me know what you think.

### DIFF
--- a/xsl/xccdf-report-oval-details.xsl
+++ b/xsl/xccdf-report-oval-details.xsl
@@ -71,16 +71,16 @@ Authors:
         <!-- if there are items to display, go ahead -->
         <xsl:when test='$items'>
             <h4>
-                <xsl:choose>
-                    <xsl:when test='$result="pass"'>Items found satisfying </xsl:when>
-                    <xsl:otherwise>Items found violating </xsl:otherwise>
-                </xsl:choose>
                 <span class="label label-primary">
                     <xsl:choose>
                         <xsl:when test='$title'><xsl:value-of select='$title'/></xsl:when>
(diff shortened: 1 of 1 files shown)


mpreisler:
One more change:

image

shawndwells:
This will be a forcing function to cleanup OVAL check language... e.g. reading the gpgcheck=0, I kept wondering why the rule failed. Not such a bad thing.

I like this, a lot.

rsprudencio:
Looks really good, I'd just change the size of those boxes to match the font size, looks smaller than the font. If it's not possible or too much of a hassle, we can merge this and add it to TODO list.

mpreisler:
> If it's not possible or too much of a hassle, we can merge this and add it to TODO list.

Not sure how to do that, I just use the class label similar to the OVAL condition box.

mpreisler:
@rsprudencio @shawndwells Anything else that's blocking before we merge this?

shawndwells:
LGTM.

shawndwells:
I would merge this myself, but don't feel comfortable mucking around in core OpenSCAP code + branches. Looking forward to seeing this feature merged!

mpreisler:
> I would merge this myself, but don't feel comfortable mucking around in core OpenSCAP code + branches.

Only one way to fix that :-P

rsprudencio:
Nope, LGTM :)

Issue #716, opened 2017-03-27
OpenSCAP emits many warnings while scanning Common profile for Fedora


jan-cerny:
When I run a scan using SSG common profile for Fedora, I can see some warning messages at the end of the output:

I run this command:
```oscap xccdf eval --results-arf arf.xml --profile xccdf_org.ssgproject.content_profile_common --report report.html /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml```

And the output is:
```
[snip]
...
Title   Enable the Audit Daemon
Rule    xccdf_org.ssgproject.content_rule_service_auditd_enabled
Result  pass

W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
W: oscap: rbt_str_add: non-zero return code
...
[snip]
```

I'm using OpenSCAP 1.2.14 on Fedora 24 Workstation:
```
$ rpm -q openscap
openscap-1.2.14-1.fc24.x86_64
```

jan-cerny:
Experienced also on F25.

jan-cerny:
Fixed by #729

PR #729, opened 2017-05-02
Change the category of messages

jan-cerny:
These messages are disturibing, moreover there is no way
how a user can mitigate them.

### DIFF
--- a/src/OVAL/probes/unix/linux/rpminfo.c
+++ b/src/OVAL/probes/unix/linux/rpminfo.c
@@ -143,7 +143,7 @@ static void pkgh2rep (Header h, struct rpminfo_rep *r)
 
 	if (regexec(&g_keyid_regex, str, 1, keyid_match, 0) != 0) {
 		sid = NULL;
-		dW("Failed to extract the Key ID value: regex=\"%s\", string=\"%s\"",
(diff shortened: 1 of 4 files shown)


src/OVAL/probes/unix/linux/rpminfo.c
```diff
@@ -143,7 +143,7 @@ static void pkgh2rep (Header h, struct rpminfo_rep *r)
 
 	if (regexec(&g_keyid_regex, str, 1, keyid_match, 0) != 0) {
 		sid = NULL;
-		dW("Failed to extract the Key ID value: regex=\"%s\", string=\"%s\"",
+		dI("Failed to extract the Key ID value: regex=\"%s\", string=\"%s\"",
```
yuumasato: This looks like it should be a developer level warning, it has regular expression and a string.
It is not clear to me what is criteria.

jan-cerny: @yuumasato yes, you're true

jan-cerny: changed

jan-cerny:
Also can fix #716

yuumasato:
Ack, thank you.

PR #730, opened 2017-05-04
 Fix for dubious assert when renaming benchmarks

Hexadorsimal:
Changed xccdf_find_target_htable to work with benchmarks too because benchmarks are definitely going through this code when they get renamed.

### DIFF
--- a/src/XCCDF/benchmark.c
+++ b/src/XCCDF/benchmark.c
@@ -849,7 +849,7 @@ bool xccdf_benchmark_rename_item(struct xccdf_item *item, const char *newid)
 struct oscap_htable *
 xccdf_benchmark_find_target_htable(const struct xccdf_benchmark *benchmark, xccdf_type_t type)
 {
-	assert(type & (XCCDF_ITEM | XCCDF_PROFILE | XCCDF_RESULT));
+	assert(type & XCCDF_OBJECT);
 	if (type == XCCDF_PROFILE)
 		return XITEM(benchmark)->sub.benchmark.profiles_dict;
 	if (type == XCCDF_RESULT)
--- a/src/XCCDF/public/xccdf_benchmark.h
+++ b/src/XCCDF/public/xccdf_benchmark.h
@@ -49,7 +49,7 @@
  * Type of an XCCDF object.
  *
  * When checking the type, you can use either operator == for type equivalence,
- * or operator &amp; to take a type inheriritance hierarchy into account.
+ * or operator &amp; to take a type inheritance hierarchy into account.
  * For example, XCCDF_ITEM & XCCDF_RULE evaluates to true,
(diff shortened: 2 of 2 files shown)


jan-cerny:
@openscap-jenkins ok to test

mpreisler:
So it will return a htable of just one benchmark? We might consider changing the code to avoid all of this for benchmarks.

Hexadorsimal:
Well, the benchmark is the structure that's holding the hash tables for all of the items in the benchmark.  When any element is being parsed from the XML, it gets added to the hash table of the containing benchmark so it can be looked up.  This happens even for benchmarks, they just get added to their own hash table.
https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/XCCDF/benchmark.c#L147

Then whenever you want to rename an item, it renames it by looking it up in its containing benchmark (or itself).  
https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/XCCDF/item.c#L911-L912

When you rename it, it goes through this code just like any other item.
https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/XCCDF/benchmark.c#L833

Then you hit the assert because the hash table find function doesn't want to allow benchmarks.

Hexadorsimal:
Yeah, you're right, a benchmark doesn't really need to store a reference to itself in its own hash table.  That's probably the right fix.

Hexadorsimal:
I mean, I'm still pretty new to this so maybe I'm completely off, but I think this is what's happening.

mpreisler:
> I mean, I'm still pretty new to this so maybe I'm completely off, but I think this is what's happening.

From my perspective I think you are right. IMO let's handle this special case in the renaming function and not by having a htable of 1 element.

jan-cerny:
@Hexadorsimal @mpreisler What is the status of this?

Hexadorsimal:
So, I haven't actually been doing any more work on this particular issue, but I have given it a few spare brain cycles over the weeks, and I actually think this is a fine fix how it is.  It's not actually holding an htable for one benchmark, it's just adding a reference to the benchmark into the htable for all the generic items that are contained in the benchmark.  It makes sense if you consider the benchmark to be an item, and it actually IS, because the benchmark macro includes the item macro inside of it.  It's only weird if you think of the htable as being a repository for the names of the items that are contained within it, so it's containing itself, but apart from that I don't think it's a big deal.

I think this simple change of the object type in the assert is the right fix.

Hexadorsimal:
https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/XCCDF/item.h#L221-L232
When I said "macro" I obviously meant this struct, and not a macro.  This is how a benchmark is an item.  There's too many macros in this codebase.

jan-cerny:
OK, thanks for explanation. ACK.

PR #732, opened 2017-05-04
Expose cpe_dict_model_add_item()

Hexadorsimal:
Changed cpe_dict_model_add_item() to be a publicly accessible function.

### DIFF
--- a/src/CPE/cpedict_priv.c
+++ b/src/CPE/cpedict_priv.c
@@ -155,8 +155,7 @@ OSCAP_ACCESSOR_STRING(cpe_generator, product_name)
 
 OSCAP_GETTER(struct cpe_generator *, cpe_dict_model, generator)
 OSCAP_ACCESSOR_SIMPLE(int, cpe_dict_model, base_version)
-OSCAP_IGETTER_GEN(cpe_item, cpe_dict_model, items)
-OSCAP_ITERATOR_REMOVE_F(cpe_item)
+OSCAP_IGETINS_GEN(cpe_item, cpe_dict_model, items, item) OSCAP_ITERATOR_REMOVE_F(cpe_item)
 OSCAP_IGETINS_GEN(cpe_vendor, cpe_dict_model, vendors, vendor) OSCAP_ITERATOR_REMOVE_F(cpe_vendor)
 
 /* ****************************************
@@ -274,8 +273,6 @@ OSCAP_ACCESSOR_STRING(cpe_language, value)
  * These function shoud not be called from outside. For exporting these elements
  * has to call parent element's 
  */
-static bool cpe_dict_model_add_item(struct cpe_dict_model *dict, struct cpe_item *item);
-
 static struct cpe_reference *cpe_reference_parse(xmlTextReaderPtr reader);
 static struct cpe_check *cpe_check_parse(xmlTextReaderPtr reader);
 static struct cpe_notes *cpe_notes_parse(xmlTextReaderPtr reader);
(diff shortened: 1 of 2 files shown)


rsprudencio:

rsprudencio:

src/CPE/cpedict_priv.c
```diff
@@ -155,8 +155,7 @@ OSCAP_ACCESSOR_STRING(cpe_generator, product_name)
 
 OSCAP_GETTER(struct cpe_generator *, cpe_dict_model, generator)
 OSCAP_ACCESSOR_SIMPLE(int, cpe_dict_model, base_version)
-OSCAP_IGETTER_GEN(cpe_item, cpe_dict_model, items)
-OSCAP_ITERATOR_REMOVE_F(cpe_item)
+OSCAP_IGETINS_GEN(cpe_item, cpe_dict_model, items, item) OSCAP_ITERATOR_REMOVE_F(cpe_item)
```
rsprudencio: Any specific reason for this change?

Hexadorsimal:
The function cpe_dict_model_add_item() was static/private to that file it
was in, but I wanted to be able to add an item to a cpe_dict_model, so I
deleted the static function and changed the macro from OSCAP_IGETTER_GEN to
OSCAP_IGETINS_GEN, which will create the function for us publicly, so that
we can add cpe_items to cpe_dict_models.

src/CPE/cpedict_priv.c
```diff
@@ -291,25 +288,9 @@ static void cpe_notes_export(const struct cpe_notes *notes, xmlTextWriterPtr wri
...
-
-	if (dict == NULL || item == NULL)
-		return false;
-
-	oscap_list_add(dict->items, item);
-	return true;
-}
 
 /***************************************************************************/
+
```
mpreisler: ?

src/CPE/public/cpe_dict.h
```diff
@@ -522,6 +522,9 @@ bool cpe_item_add_title(struct cpe_item *item, struct oscap_text *new_title);
 OSCAP_DEPRECATED(bool cpe_item_add_note(struct cpe_item *item, struct oscap_text *new_title));
 
 /// @memberof cpe_dict_model
+bool cpe_dict_model_add_item(struct cpe_dict_model *dict, struct cpe_item *new_item);
```
mpreisler: Where is this function implemented?

Hexadorsimal: The difference between OSCAP_IGETTER_GEN and OSCAP_IGETINS_GEN is that IGETINS also implements an insertion function (I'm assuming the INS stands for insert).  cpe_dict_model_add_item() is that insertion function.  It's just buried deep in a macro, but the prototype in the public header doesn't use the macro so it's easier for to read.  I didn't come up with this, I'm just following the convention that all the other functions were using in the module.

src/CPE/cpedict_priv.c
Hexadorsimal: I didn't really do it on purpose, but it was how the code ended up after I deleted the function.  I was just balancing the blank lines above and below because I thought it looked nice.  We can change it if it bothers you.

src/CPE/public/cpe_dict.h
Hexadorsimal: https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/common/util.h#L291-L300

mpreisler: Alright

mpreisler:
ACK. Thanks!

PR #733, opened 2017-05-05
Change RPM probes to return "not applicable" when run on non-RPM environment

yuumasato:
These changes makes RPM probes return "not applicable" when RPM config files are not found.
They also fix possible undefined behaviours when destroying an uninitialised mutex.

I have changed order rpmFreeXXXX calls in probe_fini() functions, but I'm not sure if I can do it.

I will still look into writing checks for init and fini functions.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1447629

### DIFF
--- a/src/OVAL/probes/unix/linux/rpminfo.c
+++ b/src/OVAL/probes/unix/linux/rpminfo.c
@@ -275,18 +275,19 @@ void *probe_init (void)
 #ifdef HAVE_RPM46
 	rpmlogSetCallback(rpmErrorCb, NULL);
 #endif
-        if (rpmReadConfigFiles ((const char *)NULL, (const char *)NULL) != 0) {
-                dI("rpmReadConfigFiles failed: %u, %s.", errno, strerror (errno));
-                return (NULL);
-        }
-
-        g_rpm.rpmts = rpmtsCreate();
-        pthread_mutex_init (&(g_rpm.mutex), NULL);
-
 	if (regcomp(&g_keyid_regex, g_keyid_regex_string, REG_EXTENDED) != 0) {
 		dE("regcomp(%s) failed.");
 		return NULL;
 	}
+
+	if (rpmReadConfigFiles ((const char *)NULL, (const char *)NULL) != 0) {
+		dI("rpmReadConfigFiles failed: %u, %s.", errno, strerror (errno));
+		g_rpm.rpmts = NULL;
+		return ((void *)&g_rpm);
+        }
+
+        g_rpm.rpmts = rpmtsCreate();
+        pthread_mutex_init (&(g_rpm.mutex), NULL);
 
         return ((void *)&g_rpm);
 }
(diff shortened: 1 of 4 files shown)


src/OVAL/probes/unix/linux/rpminfo.c
```diff
@@ -378,9 +387,16 @@ int probe_main (probe_ctx *ctx, void *arg)
         struct rpminfo_req request_st;
         struct rpminfo_rep *reply_st;
 
+	/*
+	 * arg is NULL if regex compilation failed
+	 */
```
mpreisler: Please use `//`.

mpreisler:
Looks good. I added a small nitpick about using single line comments via `//` instead of the three line parchment style comments.

I wonder whether this doesn't break behavior. Consider you are checking that package `abc` is not installed. If you don't have RPM on the system is the right answer notapplicable or not installed?

shawndwells:
On 5/8/17 1:14 PM, Martin Preisler wrote:
> I wonder whether this doesn't break behavior. Consider you are
> checking that package |abc| is not installed. If you don't have RPM on
> the system is the right answer notapplicable or not installed?

Applicability should be determined by the OVAL rule itself, not the
probe... e.g. CPEs / <platform> tags.

Would expect the rule to be structured similar to:
`````
if os=rhel
    test_rpm_foo_installed
elif os=debian
    test_dpkg_foo_installed
`````

yuumasato:
> I wonder whether this doesn't break behavior. 

Behavior in 1.2.13 was to result in "not applicable", in 1.2.14 it changed to "error".

> Consider you are checking that package abc is not installed. If you don't have RPM on the system is the right answer notapplicable or not installed?

That is a good question. Not installed is a misleading answer, IMO. And not applicable may imply that the system is not RPM based.

An answer for the object could be "not collected".
`not collected  |  This value MUST be used when no attempt was made to collect the OVAL Object.`
And this will imply that OVAL test will result in `unknown`, and this kinda makes sense, because we cannot say if it is installed or not.
This flag also allows inclusion of a message explaining the flag, that could be used to clarify that RPM was not found in the system.

mpreisler:
@yuumasato would reporting `unknown` break openscap-daemon CPE and atomic scan?

mpreisler:
Thinking about this some more I think we should just revert to previous behavior.

@yuumasato Is there anything missing in this PR or is it marked WIP because of the discussion?

yuumasato:
@mpreisler it is marked as WIP progress because I would like to add checks for the case when RPM is not available, which is reproducible by chroot'ing into a tree without RPM properly configured.
But as that will require privileges, I'll skip it.

yuumasato:
Changed the comments style, I'm really confused about which style should be used.

mpreisler:
> Changed the comments style, I'm really confused about which style should be used.

Use the modern style. Parts of OpenSCAP were written in C89 but we no longer enforce that. Similarly everybody prefers when you declare variables in the smallest possible scope and as close to their use as possible. We no longer have to declare at the beginning of the block.

mpreisler:
ACK

PR #737, opened 2017-05-09
Do not remove consecutive newlines from ansible fixes when generating

mpreisler:
Use thread-safe strchr instead of strtok that uses global state.


rsprudencio:

yuumasato:
ACK.

PR #740, opened 2017-05-12
Fixes #739

### LINKED ISSUE #739: <Rule> with <refine-rule> loses <Rule>-specific severity attribute during evaluation
Using oscap 1.2.10 on RHEL 7.3.

When a `<refine-rule>` which cites only the idref and role attributes is present in a `<Profile>` for evaluation, the severity attribute of the target `<Rule>` is not propagated to the corresponding evaluation `<rule-result>` element. The neglected attribute appears as lacking severity in the evaluation report.

WesleyCeraso:
Hi Team, this pull request fixes a unit test, creates more asserts and fixes the issue #739.
Each commit has its own description.


rsprudencio:
LGTM, thx :smile:

PR #741, opened 2017-05-12
CPE Notes bugfix

Hexadorsimal:
Some of the code was treating the notes in a cpe_item as a list of oscap_texts, and some of the code was treating it as a cpe_notes object.  As a result cpe_item_get_notes() was returning garbage because it was trying to cast the memory for the cpe_notes struct as the first string in the list of note texts.
I fixed it, but it breaks the API a little bit because the return value of cpe_item_get_notes() is different now.  I don't feel bad because it was returning garbage before, so anyone who was using it was just getting garbage.  I bet nobody was using it.

### DIFF
--- a/src/CPE/cpedict_priv.c
+++ b/src/CPE/cpedict_priv.c
@@ -79,7 +79,7 @@ struct cpe_item {		// the node <cpe-item>
 
 	struct oscap_list *references;	// list of references
 	struct oscap_list *checks;	// list of checks
-	struct oscap_list *notes;	// list of notes - it's the same structure as titles
+	struct oscap_list *notes;	// list of cpe_notes (which contain lists of oscap_text)
 	struct cpe_item_metadata *metadata;	// element <meta:item-metadata>
 	struct cpe23_item *cpe23_item;		///< element <cpe23-item>
 	struct {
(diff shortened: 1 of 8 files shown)


mpreisler:
Breaks API.

Hexadorsimal:
Should I try to put this into master instead of the maint-1.2 build since it breaks the API?  Or what?

mpreisler:
We have two options:

1) If you want this in a future 1.2.x release you have to add these fixes as new functions, just suffix them - cpe_item_get_notes2. Make sure the broken old ones are marked as deprecated. Then we remove deprecated functions in master and rename it back.

2) We can merge it straight to master. But then it will take a lot longer before this gets into a release. We are looking at more than a year from now.

Hexadorsimal:
I'm not sure I really care if it gets put into 1.2.  I mainly just don't like my fork diverging too far from the main project.

shawndwells:
On 5/12/17 3:30 PM, John Whipple wrote:
> I'm not sure I really care if it gets put into 1.2. I mainly just
> don't like my fork diverging too far from the main project.
Given the ecosystem and customer around this work, strongly suggest
Option #1 for supportability and merging into mainline RHEL/CentOS releases.

mpreisler:
Hmm, haven't figured out what does it but some hotkey in github makes a review request to the person github suggests. Wish I could turn that off :-/

Hexadorsimal:
Yeah, I can see why having a working version of the function in the maint-1.2 branch would be nice, and under normal circumstances Martin is right that we shouldn't break the API, but in this case the original API is returning garbage.  Using lists with void pointers and casting the memory to whatever you think it should be only works if you get the datatype right, and whoever wrote this originally made a mistake and picked the wrong datatype.

I actually started to do option #1 and make cpe_item_get_notes2(), but I really feel like it's just confusing the issue for no real benefit and it was hurting my programming sensibilities.  That function is already broken and there is really no reason to even try to maintain that tiny corner of the API.  Ultimately, it's your project though so you should do what you think is best.

mpreisler:
> Yeah, I can see why having a working version of the function in the maint-1.2 branch would be nice, and under normal circumstances Martin is right that we shouldn't break the API, but in this case the original API is returning garbage. Using lists with void pointers and casting the memory to whatever you think it should be only works if you get the datatype right, and whoever wrote this originally made a mistake and picked the wrong datatype.

It's as simple as not breaking `abi-compliance-check`. No exceptions.

It will cause linker errors in bindings even though the user is not using the broken API.

Hexadorsimal:
Alright, hopefully this works.

PR #743, opened 2017-05-17
Type-specific oval variable iterators

Hexadorsimal:
These functions were previously returning the generic oscap_iterators, which are not available publicly, and there are no python swig interfaces for them.


yuumasato:
LGTM.
Thank you for your contribution.

Issue #744, opened 2017-05-19
get_html_guide returns NULL since 1.2.11


evgeni:
Ohai,

given the following code, all releases until 1.2.10 would print me some HTML guide:

```c
#include <stdio.h>
#include <oscap.h>
#include <oscap_source.h>
#include <ds_sds_session.h>
#include <oscap_debug.h>
#include <oscap_error.h>

int main() {
	oscap_init();
	oscap_set_verbose("DEBUG", NULL, false);
	struct oscap_source * source = oscap_source_new_from_file("/tmp/smart_proxy_openscap/test/data/ssg-rhel7-ds.xml");
	struct ds_sds_session * sds = ds_sds_session_new_from_source(source);
	struct oscap_source * source2 = ds_sds_session_select_checklist(sds, NULL, NULL, NULL);
	char * html = ds_sds_session_get_html_guide(sds, "xccdf_org.ssgproject.content_profile_rht-ccp");
	printf("%s", html);
	if (html == NULL) {
		char * error = oscap_err_get_full_error();
		printf("%s", error);
	}
	return 0;
}
```

However, starting with 1.2.11, I just get NULL.
The file I am using is this one: https://github.com/theforeman/smart_proxy_openscap/blob/master/test/data/ssg-rhel7-ds.xml

Now I wonder, is the code wrong, or is that a bug in OpenSCAP?

mpreisler:
Could you please include `<oscap_debug.h>` and use:
```
oscap_set_verbose("DEBUG", NULL, false);
```
?

Let's see if openscap tells us what's wrong.

evgeni:
There seems to be no additional output after calling that.

evgeni:
OK, for added fun: I wanted to try to bisect that, and fetched OpenSCAP from git. But even with 1.2.9 and 1.2.10 from git I get NULL back.
The previous observation about the breakage between 1.2.10 and 1.2.11 was based on the Fedora 25 packages.

evgeni:
So with the error code in place, I get the following output on bad versions:
```
Internal error: Could not acquire handle to xccdf.xml source. [ds_sds_session.c:313]
```

and now really bisecting, says that the following is the first bad commit:
```
commit b8defed4fdea58a1ae6b88ef11b20d56c499cd80
Author: Jan Černý <jcerny@redhat.com>
Date:   Thu Sep 22 12:42:40 2016 +0200

    Support tailoring of a checklist located in a different datstream
    
[...]

PR #745, opened 2017-05-19
pass session->checklist_id to oscap_htable_get in get_html_guide

evgeni:
In b8defed we moved to using `s->checklist_id` instead of hard coding `xccdf.xml`. However, this change was forgotten in `ds_sds_session_get_html_guide` which leads to the following error when trying to get the HTML guide:

    Internal error: Could not acquire handle to xccdf.xml source.

This change migrates `ds_sds_session_get_html_guide` to also use `s->checklist_id` and thus fixes the issue.

Tests missing and are left to the more knowledgeable maintainer ;)

Fixes: #744


mpreisler:

mpreisler:
Awesome! Thanks for your contribution!

PR #746, opened 2017-06-05
Suggest users to install bzip2-devel instead of bzip2

isimluk:
We use this ('devel support') and similar language as a pattern across the build pipeline.

🐟


isimluk:
Wow this is awesome project!

It takes weeks before I get PR merged into manageiq.

What awesome community!!!

PR #748, opened 2017-06-06
Added functionality for single rule evaluation

matusmarhefka:
* added '--rule' cmd option for xccdf eval
* usage:
    oscap xccdf eval [--remediate] --profile PROFILE --rule RULE INPUT_FILE

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy.c
+++ b/src/XCCDF_POLICY/xccdf_policy.c
@@ -690,6 +690,13 @@ static int _xccdf_policy_report_rule_result(struct xccdf_policy *policy,
 	} else
 		xccdf_check_free(check);
 
+	/* If policy selects only one rule, skip reporting for the other
+	 * unselected rules - only the selected rule will be reported. */
+	if (policy->rule != NULL) {
+		const char* rule_id = xccdf_rule_get_id(rule);
+		if (strcmp(policy->rule, rule_id) != 0)
+			return ret;
+	}
 	ret = xccdf_policy_report_cb(policy, XCCDF_POLICY_OUTCB_END, (void *) rule_result);
 	return ret;
 }
@@ -950,14 +957,25 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
 	const char* rule_id = xccdf_rule_get_id(rule);
 	const bool is_selected = xccdf_policy_is_item_selected(policy, rule_id);
 	const char *message = NULL;
+	int report = 0;
 
-	int report = xccdf_policy_report_cb(policy, XCCDF_POLICY_OUTCB_START, (void *) rule);
+	/* If policy selects only one rule and the rule currently being
+	 * evaluated is not equal to the selected rule, do not evaluate it and
+	 * mark it as notselected. */
+	if (policy->rule != NULL) {
+		if (strcmp(policy->rule, rule_id) != 0) {
+			return _xccdf_policy_report_rule_result(policy, result, rule, NULL, XCCDF_RESULT_NOT_SELECTED, NULL);
+		}
+		policy->rule_found = 1;
+	}
+	/* Otherwise start reporting */
+	report = xccdf_policy_report_cb(policy, XCCDF_POLICY_OUTCB_START, (void *) rule);
 	if (report)
 		return report;
 
 	struct xccdf_refine_rule_internal* r_rule = oscap_htable_get(policy->refine_rules_internal, rule_id);
-
 	xccdf_role_t role = xccdf_get_final_role(rule, r_rule);
+
 	if (!is_selected) {
 		dI("Rule '%s' is not selected.", rule_id);
 		return _xccdf_policy_report_rule_result(policy, result, rule, NULL, XCCDF_RESULT_NOT_SELECTED, NULL);
@@ -2037,6 +2055,12 @@ struct xccdf_result * xccdf_policy_evaluate(struct xccdf_policy * policy)
 			break;
 	}
 	xccdf_item_iterator_free(item_it);
+
+	if (policy->rule != NULL && !policy->rule_found) {
+		oscap_seterr(OSCAP_EFAMILY_XCCDF,
+			"Rule '%s' not found in selected profile.", policy->rule);
+		return NULL;
+	}
 
 	xccdf_policy_add_final_setvalues(policy, xccdf_benchmark_to_item(benchmark), result);
 
--- a/tests/API/XCCDF/unittests/Makefile.am
+++ b/tests/API/XCCDF/unittests/Makefile.am
@@ -131,6 +131,11 @@ EXTRA_DIST += \
 	test_report_without_oval_poses_no_errors.sh \
 	test_report_without_oval_poses_no_errors.xccdf.xml.result.xml \
 	test_report_without_xsl_fails_gracefully.sh \
+	test_single_rule.sh \
+	test_single_rule.ds.xml \
+	test_single_rule.oval.xml \
+	test_single_rule-tailoring.xml \
+	test_single_rule.xccdf.xml \
 	test_xccdf_fix_attr_export.sh \
 	test_xccdf_fix_attr_export.xccdf.xml \
 	test_xccdf_complex_check_and_notchecked.sh \
--- a/utils/oscap-tool.h
+++ b/utils/oscap-tool.h
@@ -118,6 +118,7 @@ struct oscap_action {
 	char *f_verbose_log;
 	/* others */
         char *profile;
+	const char *rule;
         char *show;
         char *format;
         const char *tmpl;
--- a/utils/oscap-xccdf.c
+++ b/utils/oscap-xccdf.c
@@ -139,6 +139,7 @@ static struct oscap_module XCCDF_EVAL = {
 		"INPUT_FILE - XCCDF file or a source data stream file\n\n"
         "Options:\n"
         "   --profile <name>\r\t\t\t\t - The name of Profile to be evaluated.\n"
+	"   --rule <name>\r\t\t\t\t - The name of a single rule to be evaluated.\n"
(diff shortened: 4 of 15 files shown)


mpreisler:

utils/oscap-xccdf.c
```diff
@@ -1041,6 +1046,9 @@ bool getopt_xccdf(int argc, char **argv, struct oscap_action *action)
                 } else {
                     action->f_ovals = NULL;
                 }
+
+		if (action->rule != NULL && action->profile == NULL)
+			return oscap_module_usage(action->module, stderr, "Profile needs to be specified when evaluating a single rule!");
```
mpreisler: Does it really have to? if profile is NULL we just use the default profile so it should all still work, right?

utils/oscap-tool.h
```diff
@@ -118,6 +118,7 @@ struct oscap_action {
 	char *f_verbose_log;
 	/* others */
         char *profile;
+        const char *rule;
```
mpreisler: indents

utils/oscap-xccdf.c
matusmarhefka: Well, I've tried the following on Fedora/RHEL7 (without specifying profile) and it returned zero without evaluating anything:
`$ oscap xccdf eval /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml`
`$ oscap xccdf eval /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml`
How exactly can I trigger this "default" profile evaluation?

utils/oscap-tool.h
matusmarhefka: You mean that indents should be tabs? Because this whole structure uses a mix of tabs and spaces for indentation, I was not sure which one to use and AFAIK there is no coding guidelines for openscap project, right?

jan-cerny: @matusmarhefka Yes, it should be tabs. We follow the Linux Kernel coding style. https://www.kernel.org/doc/html/v4.10/process/coding-style.html

Unfortunately, some old code doesn't follow the coding style and uses spaces. In new pull request we use tabs, so that's why there is a mix. We haven't replaced spaces by tabs to avoid merge conflicts and possible problems with downstream patches.

utils/oscap-xccdf.c
yuumasato: It doesn't evaluate anything because `default` profile has no rules selected.
I guess the question is whether the one rule cannot be evaluated as if part of the `default` profile, when no profile is provided.

jan-cerny: @matusmarhefka "Default" profile is a virtual profile that is used when no profile is specified. You can trigger it if you run `oscap xccdf eval` without `--profile`.

If no profile is specified, OpenSCAP evaluates all  rules (`<Rule>`) with attribute `selected="true"`, regardless of which profile they belong.

The reason why it does nothing in your case is that in SSG content there is no Rule with attribute `selected="true"`, in other words, the  virtual "Default profile" contains 0 rules.

But in other content, there can be some of there rules. You can have also content with no profiles. For example this file has no profile, but it's a valid XCCDF: https://github.com/OpenSCAP/openscap/blob/maint-1.2/tests/API/XCCDF/unittests/test_xccdf_oval_relative_path.xccdf.xml

utils/oscap-tool.h
matusmarhefka: Fixed the indent.

utils/oscap-xccdf.c
matusmarhefka: @jan-cerny Thanks for explanation, I removed this "if" chunk of code and I agree that printing error does not make sense if profile is NULL.

utils/oscap.8
```diff
@@ -75,6 +75,11 @@ You may specify OVAL Definition files as the last parameter, XCCDF evaluation wi
 Select a particular profile from XCCDF document.
 .RE
 .TP
+\fB\-\-rule RULE\fR
+.RS
+Select a particular rule from XCCDF document (also requires profile to be specified). Only this rule will be evaluated.
+.RE
+.TP
```
shawndwells: thank you for putting a note in the man page! This --rule concept has been asked for a lot, and documentation really helps get the word out.

matusmarhefka: @shawndwells Now that you mentioned it, I should probably remove the part which says that profile is required to be specified (based on the discussion above). Thanks.

matusmarhefka:
Commit 8734c0e should have all set based on all review comments.

utils/oscap-xccdf.c
```diff
@@ -139,6 +139,7 @@ static struct oscap_module XCCDF_EVAL = {
 		"INPUT_FILE - XCCDF file or a source data stream file\n\n"
         "Options:\n"
         "   --profile <name>\r\t\t\t\t - The name of Profile to be evaluated.\n"
+	"   --rule <name>\r\t\t\t\t - The name of a single rule to be evaluated.\n"
```
dahaic: Indentation should be spaces.

Also can you specify that profile is necessary to specify for this rule, and what are the reasons?

utils/oscap.8
```diff
@@ -75,6 +75,11 @@ You may specify OVAL Definition files as the last parameter, XCCDF evaluation wi
 Select a particular profile from XCCDF document.
 .RE
 .TP
+\fB\-\-rule RULE\fR
+.RS
+Select a particular rule from XCCDF document. Only this rule will be evaluated.
```
dahaic: Same as in the --help text:
Can you specify that profile is necessary to be specified for this option and what are the reasons?

mpreisler: `--profile` is not necessary. That's why Matus removed that sentence.

utils/oscap-xccdf.c
jan-cerny: @dahaic No, we use tabs for indentation. That's a general rule for all our code written in C. As I mentioned in one of previous comments:

>We follow the Linux Kernel coding style. https://www.kernel.org/doc/html/v4.10/process/coding-style.html Unfortunately, some old code doesn't follow the coding style and uses spaces. In new pull request we use tabs, so that's why there is a mix. We haven't replaced spaces by tabs to avoid merge conflicts and possible problems with downstream patches.

Also, profile is not necessary to specify.

dahaic: @jan-cerny please note that lines surrounding this addition are indented by spaces, thus using tab in this particular context is wrong, imho. Either fix it in whole block, or use what is used, and fix it later in separate commit. Sorry for not justifying my objection.

utils/oscap.8
dahaic: @mpreisler wasn't aware {{--profile}} is no longer necessary - still by specifying profile, we get values as specified in that particular profile, right? I am missing this (in my opinion quite useful) behaviour being pointed out.

mpreisler:
That is not what we do. We have a convention that we fix indentation on the
lines we are changing. We never reindent blocks of code or entire files
because that causes merge conflicts. In this case I don't really care but
imo it's faster to just follow the rule than to bikeshed :-)

utils/oscap.8
mpreisler: `(default)` profile is a profile and also can specify Values. As far as the API is concerned it is a proper profile.

I agree that mentioning that --profile affects values that are used with the rules makes sense.

matusmarhefka:
I added unit tests (make check) and updated the commit message.

Just a note (I also added it into the commit msg): when both profile and rule are specified and the rule is notselected in that profile, it will not be evaluated and 0 will be returned. Do you agree with this behaviour? I based this on the original functionality of 'oscap xccdf eval' without single-rule feature - when some profile is specified and it does not contain any selected rules, oscap returns 0 without evaluating anything.

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -950,14 +957,26 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
...
-
-	int report = xccdf_policy_report_cb(policy, XCCDF_POLICY_OUTCB_START, (void *) rule);
+	int report = 0;
+
+	/* If policy selects only one rule and the rule currently being
+	 * evaluated is not equal to the selected rule, do not evaluate it and
+	 * mark it as notselected. */
+	if (policy->rule != NULL) {
+		if (strcmp(policy->rule, rule_id) != 0) {
+			dI("Rule '%s' is not selected.", rule_id);
```
mpreisler: This should be dD at best IMO.

mpreisler:
I just have one minor nitpick regarding the logging level. Other than that I think this is ready to be merged once jenkins passes.

Thanks for this great contribution!

src/XCCDF_POLICY/xccdf_policy.c
matusmarhefka: @mpreisler Good point, it would cause a huge spam. I removed the whole dI line.

matusmarhefka:
Hopefully the last change, commit eb4e967 changes:
* removed logging about notselected rules
* added a note about values into the man page (thanks to @dahaic)

mpreisler:
Fails on Jenkins:
```
+ Test single-rule evaluation                                 FAIL: all.sh
==========================================
1 of 1 test failed
Please report to open-scap-list@redhat.com
==========================================
```

tests/API/XCCDF/unittests/Makefile.am
```diff
@@ -131,6 +131,7 @@ EXTRA_DIST += \
 	test_report_without_oval_poses_no_errors.sh \
 	test_report_without_oval_poses_no_errors.xccdf.xml.result.xml \
 	test_report_without_xsl_fails_gracefully.sh \
+	test_single_rule.sh \
```
mpreisler: Need to add the other resources too, otherwise distcheck is going to fail.

matusmarhefka: Fixed in 4637b13.

Issue #722, opened 2017-04-20
oscap generate fix doesn't support multi benchmark datastreams


mpreisler:
```
Profile "xccdf_org.ssgproject.content_profile_pci-dss_centric" was not found. Get available profiles using:
$ oscap info "/home/mpreisle/d/scap-security-guide/build/ssg-rhel7-ds.xml"
Fatal error encountered when generating role '/tmp/ssg-rhel7-PCIDSS-RHEL-7-role-pci-dss_centric.sh'. Error details:
Command '['oscap', 'xccdf', 'generate', 'fix', '--benchmark-id', 'xccdf_org.ssgproject.content_benchmark_PCIDSS-RHEL-7', '--profile', 'xccdf_org.ssgproject.content_profile_pci-dss_centric', '--template', 'asdf', '/home/mpreisle/d/scap-security-guide/build/ssg-rhel7-ds.xml']' returned non-zero exit status 1
```

jan-cerny:
Fixed in https://github.com/jan-cerny/openscap/commit/0346f3c6b360c27d7dc982f04cad6f16b4905d84

mpreisler:
Ref to the upstream repo: https://github.com/OpenSCAP/openscap/commit/0346f3c6b360c27d7dc982f04cad6f16b4905d84

PR #751, opened 2017-06-09
Issue #722: Support multiple benchmark datastreams in "xccdf generate fix"

jan-cerny:
This commit will allow to select a benchmark from which the fix
will be generated if there are multiple benchmarks in a single
datastream.


mpreisler:
ACK

PR #752, opened 2017-06-09
Change to clarify tools needed for autogen

kjankov:
Since you need the tools for running autogen, they should be mentioned first.


mpreisler:

jan-cerny:
Thanks! And welcome!

PR #753, opened 2017-06-13
Added and implemented the xccdf_session_set_profile_id_by_suffix meth…

kjankov:
…od to xccdf_session.c to improve command line interface for the user: instead of requiring the full extended profile id, the user only needs to give the profile suffix as an argument. Later commits will make sure that only one profile matches the suffix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ Makefile.in
 *.lo
 *.a
 *.la
+*.cproject
+*.project
+.settings/language.settings.xml
 
 oscap_debug.log.*
 
--- a/.gitignore
+++ b/.gitignore
@@ -13,8 +13,8 @@ Makefile.in
 *.lo
 *.a
 *.la
-*.cproject
-*.project
+.cproject
+.project
 .settings/language.settings.xml
 
 oscap_debug.log.*
--- a/src/XCCDF/public/xccdf_session.h
+++ b/src/XCCDF/public/xccdf_session.h
@@ -302,6 +302,15 @@ bool xccdf_session_set_report_export(struct xccdf_session *session, const char *
 bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *profile_id);
 
 /**
+ *Select XCCDF Profile for evaluation with only profile suffix as input
+ *@memberof xccdf_session
+ *@param session XCCDF Session
+ *@param profile_suffix unique profile ID or suffix of the ID of the profile to set
+ *@returns true on success
+ */
+bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix);
+
+/**
  * Retrieves ID of the profile that we will evaluate with, or NULL.
  * @memberof xccdf_session
  * @param session XCCDF Session
--- a/src/XCCDF/public/xccdf_session.h
+++ b/src/XCCDF/public/xccdf_session.h
@@ -302,13 +302,14 @@ bool xccdf_session_set_report_export(struct xccdf_session *session, const char *
 bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *profile_id);
 
 /**
- *Select XCCDF Profile for evaluation with only profile suffix as input
+ *Select XCCDF Profile for evaluation with only profile suffix as input. Reports error
+ *if multiple profiles match the suffix.
  *@memberof xccdf_session
  *@param session XCCDF Session
  *@param profile_suffix unique profile ID or suffix of the ID of the profile to set
- *@returns true on success
+ *@returns 0 on success, 1 if profile is not found, and 2 if multiple matches are found.
  */
-bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix);
+int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix);
 
 /**
  * Retrieves ID of the profile that we will evaluate with, or NULL.
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -486,7 +486,8 @@ int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const 
 	} else if (multiple) {
 		return 2;
 	} else {
-		assert(xccdf_session_set_profile_id(session, full_profile_id)==0);
+		const bool search_result = xccdf_session_set_profile_id(session, full_profile_id);
+		assert(search_result);
 		return 0;
 	}
 }
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -491,7 +491,7 @@ int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const 
 		return 2;
 	}
 	else {
-		xccdf_session_set_profile_id(session, full_profile_id);
+		assert(xccdf_session_set_profile_id(session, full_profile_id)==0);
 		return 0;
 	}
 }
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -469,7 +469,10 @@ bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const
 		xccdf_profile_iterator_free(profit_bench);
 	}
 
-	return xccdf_session_set_profile_id(session, full_profile_id);
+	if (full_profile_id == NULL)
+		return false;
+	else
+		return xccdf_session_set_profile_id(session, full_profile_id);
 }
 
 const char *xccdf_session_get_profile_id(struct xccdf_session *session)
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -452,12 +452,10 @@ int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const 
 						full_profile_id, tailoring_profile_id);
 					multiple = true;
 					break;
-				}
-				else {
+				} else {
 					full_profile_id = tailoring_profile_id;
 				}
 			}
-
 		}
 		xccdf_profile_iterator_free(profit_tailoring);
 	}
@@ -475,8 +473,7 @@ int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const 
 						full_profile_id, bench_profile_id);
 					multiple = true;
 					break;
-				}
-				else {
+				} else {
 					full_profile_id = bench_profile_id;
 				}
 			}
@@ -486,11 +483,9 @@ int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const 
 
 	if (full_profile_id == NULL) {
 		return 1;
-	}
-	else if (multiple) {
+	} else if (multiple) {
 		return 2;
-	}
-	else {
+	} else {
 		assert(xccdf_session_set_profile_id(session, full_profile_id)==0);
 		return 0;
 	}
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -431,6 +431,47 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
 	return true;
 }
 
+bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix)
+{
+	const char *full_profile_id = NULL;
+	struct xccdf_benchmark *bench = xccdf_policy_model_get_benchmark(session->xccdf.policy_model);
+
+	// Tailoring Profiles
+	struct xccdf_tailoring *tailoring = xccdf_policy_model_get_tailoring(session->xccdf.policy_model);
+	if (tailoring != NULL)   {
+		struct xccdf_profile_iterator *profit_tailoring = xccdf_tailoring_get_profiles(tailoring);
+		while (xccdf_profile_iterator_has_more(profit_tailoring)) {
+			struct xccdf_profile *tailoring_profile = xccdf_profile_iterator_next(profit_tailoring);
+			const char *tailoring_profile_id = xccdf_profile_get_id(tailoring_profile);
+
+			if(oscap_str_endswith(tailoring_profile_id, profile_suffix)) {
+				full_profile_id = tailoring_profile_id;
+				break;
+			}
+
+		}
+		xccdf_profile_iterator_free(profit_tailoring);
+	}
+
+	// Benchmark Profiles
+	if (full_profile_id == NULL) {
+		struct xccdf_profile_iterator *profit_bench = xccdf_benchmark_get_profiles(bench);
+		while (xccdf_profile_iterator_has_more(profit_bench)) {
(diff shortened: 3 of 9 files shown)


src/XCCDF/public/xccdf_session.h
```diff
@@ -292,6 +292,10 @@ bool xccdf_session_set_report_export(struct xccdf_session *session, const char *
  */
 bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *profile_id);
 
+
+bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix);
```
mpreisler: Needs documentation and some minor whitespace fixes.

src/XCCDF/xccdf_session.c
```diff
@@ -425,6 +425,47 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
 	return true;
 }
 
+bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix)
+{
+	const char *full_profile_id = NULL;
+    struct xccdf_benchmark *bench = xccdf_policy_model_get_benchmark(session->xccdf.policy_model);
+
+    // Tailoring Profiles
```
mpreisler: indented by spaces

src/XCCDF/xccdf_session.c
```diff
@@ -425,6 +425,47 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
 	return true;
 }
 
+bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix)
+{
+	const char *full_profile_id = NULL;
+    struct xccdf_benchmark *bench = xccdf_policy_model_get_benchmark(session->xccdf.policy_model);
```
mpreisler: indented by spaces

mpreisler:
Besides documentation and some minor formatting issues this is looking good.

shawndwells:
On 6/13/17 1:17 PM, Katarina Jankov wrote:
>
>  *
>
>
>         File Changes
>
>   * *A* .cproject
>     <https://github.com/OpenSCAP/openscap/pull/753/files#diff-0> (46)
>   * *A* .project
>     <https://github.com/OpenSCAP/openscap/pull/753/files#diff-1> (27)
>   * *A* .settings/language.settings.xml
>     <https://github.com/OpenSCAP/openscap/pull/753/files#diff-2> (15)
>

Should these files be committed to the repo?

src/XCCDF/xccdf_session.c
```diff
@@ -425,6 +425,47 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
...
+
+			if(oscap_str_endswith(bench_profile_id, profile_suffix)) {
+				full_profile_id = bench_profile_id;
+				break;
+			}
+		}
+
+		xccdf_profile_iterator_free(profit_bench);
+	}
+
```
mpreisler: `if full_profile_id == NULL` here we need to report a failure to match. passing NULL into set_profile_id means we want to select the `(default)` profile which will always be there. So we would fail silently.

mpreisler:
The code looks OK but this doesn't work for me. I am not sure why. I am testing with:
```
./run utils/.libs/oscap xccdf eval --profile common /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml
```

And it is not evaluating any rules. The full ID still works:
```
./run utils/.libs/oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_common /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml 
Title   gpgcheck Enabled In Main Dnf Configuration
Rule    xccdf_org.ssgproject.content_rule_ensure_gpgcheck_globally_activated
^C
```

src/XCCDF/xccdf_session.c
```diff
@@ -425,6 +425,47 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
...
+				full_profile_id = tailoring_profile_id;
+				break;
+			}
+
+		}
+		xccdf_profile_iterator_free(profit_tailoring);
+	}
+
+	// Benchmark Profiles
+	if (full_profile_id != NULL) {
```
mpreisler: This needs to be `full_profile_id == NULL` if I am understanding the code correctly.

mpreisler: After this change it works!

```
$ ./run utils/.libs/oscap xccdf eval --profile common /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml 
Title   gpgcheck Enabled In Main Dnf Configuration
Rule    xccdf_org.ssgproject.content_rule_ensure_gpgcheck_globally_activated
Result  pass

Title   gpgcheck Enabled For All Dnf Package Repositories
Rule    xccdf_org.ssgproject.content_rule_ensure_gpgcheck_never_disabled
Result  fail

```

.gitignore
```diff
@@ -13,6 +13,9 @@ Makefile.in
 *.lo
 *.a
 *.la
+*.cproject
```
mpreisler: should the star be there at the beginning? Isn't the file just `.cproject`?

.gitignore
```diff
@@ -13,6 +13,9 @@ Makefile.in
...
+*.project
```
mpreisler: same as above

src/XCCDF/xccdf_session.c
```diff
@@ -425,6 +425,52 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
...
+				full_profile_id = bench_profile_id;
+				break;
+			}
+		}
+
+		xccdf_profile_iterator_free(profit_bench);
+	}
+
+	if (full_profile_id == NULL) {
+		printf("Error: No matches were found for this profile suffix.");
```
mpreisler: We use `oscap_seterr` to report errors in the API. Check out https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/XCCDF/xccdf_session.c#L204 for an example.

Please include the suffix in the error message. I suggest something like:
"Can't find any profiles matching suffix '%s'."

You can look into how `*_set_profile_id` and the functions it calls report errors and match the style of the messages.

mpreisler: Thinking about this again we shouldn't report it like this. We should let the user of the function report the error. So just returning `false` and changing the message in `report_missing_profile` should be enough. If we find multiple matches we will report an error using oscap_seterr. But all that later.

utils/oscap-xccdf.c
```diff
@@ -446,6 +446,13 @@ static void report_missing_profile(const struct oscap_action *action)
 		"$ oscap info \"%s\"\n", action->profile, action->f_xccdf);
 }
 
+static void report_multiple_profile_matches(const struct oscap_action *action)
+{
+	fprintf(stderr,
+			"At least two profiles matched the given suffix. Use a more specific suffix "
+			"to get an exact match.\n");
```
mpreisler: Hint at oscap info here, similar to what's done in report_missing_profile.
Also please fix the indents to be same as report_missing_profile, not sure why it's indented so much.

utils/oscap-xccdf.c
```diff
@@ -493,10 +500,15 @@ int app_evaluate_xccdf(const struct oscap_action *action)
 	/* Select profile */
 	if (!xccdf_session_set_profile_id(session, action->profile)) {
 		if (action->profile != NULL) {
-			if (!xccdf_session_set_profile_id_by_suffix(session, action->profile)) {
+			int suffix_match_result = xccdf_session_set_profile_id_by_suffix(session, action->profile);
```
mpreisler: this can be `const`

src/XCCDF/xccdf_session.c
```diff
@@ -455,18 +471,30 @@ bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const
...
-	else
-		return xccdf_session_set_profile_id(session, full_profile_id);
+	if (full_profile_id == NULL) {
+		return 1;
+	}
+	else if (multiple) {
+		return 2;
+	}
+	else {
+		xccdf_session_set_profile_id(session, full_profile_id);
```
mpreisler: why force `return 0` and throw away the result of `xccdf_session_set_profile_id`?

src/XCCDF/xccdf_session.c
```diff
@@ -455,18 +471,30 @@ bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const
 			const char *bench_profile_id = xccdf_profile_get_id(bench_profile);
 
 			if(oscap_str_endswith(bench_profile_id, profile_suffix)) {
-				full_profile_id = bench_profile_id;
-				break;
+				if (full_profile_id != NULL) {
+					printf("Matches: \n%s\n", full_profile_id);
+					printf("%s\n", bench_profile_id);
```
mpreisler: You need to use `oscap_seterr`, this will print into stdout!

src/XCCDF/xccdf_session.c
```diff
@@ -438,9 +447,16 @@ bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const
 			struct xccdf_profile *tailoring_profile = xccdf_profile_iterator_next(profit_tailoring);
 			const char *tailoring_profile_id = xccdf_profile_get_id(tailoring_profile);
 
-			if(oscap_str_endswith(tailoring_profile_id, profile_suffix)) {
-				full_profile_id = tailoring_profile_id;
-				break;
+			if (oscap_str_endswith(tailoring_profile_id, profile_suffix)) {
+				if (full_profile_id != NULL) {
+					printf("Matches: \n %s\n", full_profile_id);
+					printf("%s\n", tailoring_profile_id);
```
mpreisler: oscap_seterr

src/XCCDF/xccdf_session.c
```diff
@@ -425,10 +425,19 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
...
 	const char *full_profile_id = NULL;
 	struct xccdf_benchmark *bench = xccdf_policy_model_get_benchmark(session->xccdf.policy_model);
+	bool multiple = false;
+
+	const char full_profile_id_prefix[] = "xccdf_org.ssgproject.content_profile_";
+	const char *suffix_full_id_form = strcat(full_profile_id_prefix, profile_suffix);
+	if (xccdf_session_set_profile_id(session, suffix_full_id_form)) {
+		full_profile_id = suffix_full_id_form;
+		return 0;
+	}
```
mpreisler: why is this block here? aren't you already doing this in `oscap-xccdf.c`?

src/XCCDF/xccdf_session.c
```diff
@@ -425,10 +425,19 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
...
 
-bool xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix)
+// 0 = no error, 1 = no matches, 2 = multiple matches
+int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix)
 {
 	const char *full_profile_id = NULL;
 	struct xccdf_benchmark *bench = xccdf_policy_model_get_benchmark(session->xccdf.policy_model);
+	bool multiple = false;
+
+	const char full_profile_id_prefix[] = "xccdf_org.ssgproject.content_profile_";
```
mpreisler: why the `[]` syntax here?

kjankov: I'm getting segmentation fault when I try it with pointers, I think because I am appending another string to it. Not sure

src/XCCDF/xccdf_session.c
mpreisler: on second thought I think we should assert that it returns 0 and then return 0. because if the code is correct this must always return 0.

Please change the line to:
`assert(xccdf_session_set_profile_id(session, full_profile_id) == 0);`

kjankov:
@dahaic I added a grep search to make sure the full profile id is used in the results file. Martin didn't think we necessarily needed to check the results-arf and report. I'm currently working on adding a tailoring file to the situation. If both tailoring and benchmark have the same profile or profiles with matching suffix, this should be a valid use case as long as tailoring only has one match.

.gitignore
```diff
@@ -13,6 +13,9 @@ Makefile.in
 *.lo
 *.a
 *.la
+.cproject
+.project
+.settings/language.settings.xml
```
jan-cerny: @kjankov Why is this here? I mean I'm OK with this, but I wonder why is this needed? Do you use some special cool editor? Maybe you use something better than I'm using :-) Thanks for reply.

src/XCCDF/xccdf_session.c
```diff
@@ -431,6 +431,71 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
...
+		xccdf_profile_iterator_free(profit_bench);
+	}
+
+	if (full_profile_id == NULL) {
+		return 1;
+	}
+	else if (multiple) {
+		return 2;
+	}
+	else {
```
jan-cerny: A small nitpick: We usually write it on one line:, like this: `} else {` That's because we follow the Linux Kernel coding style. If you have some time, you can read about it: https://www.kernel.org/doc/html/v4.10/process/coding-style.html

tests/API/XCCDF/unittests/Makefile.am
```diff
@@ -82,6 +82,8 @@ EXTRA_DIST += \
 	test_oval_without_definition.oval.xml \
 	test_oval_without_definition.sh \
 	test_oval_without_definition.xccdf.xml \
+	test_profile_selection_by_suffix.sh \
+	test_profile_selection_by_suffix.xccdf.xml \
 	test_remediate_perl.sh \
```
jan-cerny: Please, put `test_profile_selection_by_suffix_tailoring.xccdf.xml` here as well. Now it's causing `make distcheck` to fail.

jan-cerny:
I'm afraid that it doesn't work.

E.g.:
`oscap xccdf eval --profile common /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml`
Does nothing.

It should evaluate the Common profile for Fedora systems. The result should be the same as  this command:
`oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_common /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml`

jan-cerny:
@dahaic I think that main use-case for this feature are DataStreams. Shouldn't the test test the feature on DataStreams as well?

.gitignore
kjankov: I'm using Eclipse and it adds these files. Not sure if it's better, I'm just more accustomed to it :)

src/XCCDF/xccdf_session.c
kjankov: Fixed, thank you. Will reference that in the future

tests/API/XCCDF/unittests/Makefile.am
kjankov: Fixed

kjankov:
@jan-cerny Hello, made the changes you recommended and fixed the statement that was causing profile evaluation to not work.

mpreisler:

jan-cerny:
Great work! Thank you!

PR #754, opened 2017-06-14
RHBZ#1446451: Allow to identify problems with external variables

jan-cerny:
RHBZ#1446451 has discovered that we are unable to find reasons of errors caused by missing external variables. There was no chance that user could magically figure out that he needs to provide some external variables.

This PR wants to add a way to inform user about missing external variables by introducing some error messages.

### DIFF
--- a/src/OVAL/oval_component.c
+++ b/src/OVAL/oval_component.c
@@ -1462,12 +1462,21 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_VARREF(oval_argu_
 		return flag;
 	}
 
+	dIndent(1);
+	dI("Variable component references %s '%s'",
+		oval_variable_type_get_text(oval_variable_get_type(variable)),
+		oval_variable_get_id(variable));
+
 	if (argu->mode == OVAL_MODE_QUERY) {
-		if (oval_probe_query_variable(argu->u.sess, variable) != 0)
+		if (oval_probe_query_variable(argu->u.sess, variable) != 0) {
+			dIndent(-1);
 			return flag;
+		}
 	} else {
-		if (oval_syschar_model_compute_variable(argu->u.sysmod, variable) != 0)
+		if (oval_syschar_model_compute_variable(argu->u.sysmod, variable) != 0) {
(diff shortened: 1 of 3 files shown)


src/OVAL/oval_component.c
```diff
@@ -1462,12 +1462,21 @@ static oval_syschar_collection_flag_t _oval_component_evaluate_VARREF(oval_argu_
 		return flag;
 	}
 
+	dIndent(1);
+	dI("Variable component references %s '%s'",
```
mpreisler: shouldn't this be `dD`?

src/OVAL/results/oval_resultTest.c
```diff
@@ -564,6 +564,12 @@ static oval_result_t eval_item(struct oval_syschar_model *syschar_model, struct
 						oval_sysent_get_value(item_entity),
 						oval_sysitem_get_id(cur_sysitem), oval_state_get_id(state));
 			}
+			if (ent_val_res == OVAL_RESULT_ERROR) {
+				dI("Comparing entity '%s'='%s' of item '%s' to corresponding entity in state '%s' was not successful.",
```
mpreisler: why `dI`. this is an error, right?

mpreisler: Hmm, actually you are right here I think. Because this is not an error in the tool but rather in the content. It's a gray area, `dI` is fine if you really meant it that way.

jan-cerny: Yes, I did meant it this way, actually the scan process can continue here

src/OVAL/oval_component.c
jan-cerny: It could be, because it's a very detailed information, but on the other hand, it's related to SCAP content, and not to the scanner implementation, if you are trying to figure out some problems with variables in your content, it might be beneficial the some of your OVAL_variables are referencing something else, so the evaluation flow would be more obvious.

mpreisler: Ok

mpreisler:
ACK

PR #757, opened 2017-06-22
Bash script header generation for results of generate fix command

kjankov:


kjankov:
The header will have different info depending on whether it's based on a profile from a benchmark, or from results of an evaluation of a profile. Currently most of the info for the former option is there, but it seems that the result is always considered null (even in the generate_fix function). Not sure why

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -661,17 +661,71 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
 	return ret;
 }
 
-static int _write_script_header_to_fd(const char *sys, int output_fd)
+static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_result *result, const char *sys, int output_fd, struct xccdf_profile *profile)
```
mpreisler: you can get the profile from policy, so no need to pass both

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -688,12 +742,14 @@ int xccdf_policy_generate_fix(struct xccdf_policy *policy, struct xccdf_result *
 		// No TestResult is available. Generate fix from the stock profile.
 		dI("Generating profile-oriented fixes for policy(profile/@id=%s)", xccdf_policy_get_id(policy));
 		struct xccdf_benchmark *benchmark = xccdf_policy_get_benchmark(policy);
+		struct xccdf_policy *profile = xccdf_policy_get_profile(policy);
```
mpreisler: wrong types

kjankov: Oops.. fixed

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -711,10 +712,10 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
...
-			const char *result_start_time = xccdf_result_get_start_time(result);
-			const char *result_end_time = xccdf_result_get_end_time(result);
+			const char *start_time = xccdf_result_get_start_time(result);
+			const char *end_time = xccdf_result_get_end_time(result);
 
-			const char *fix_header = oscap_sprintf(
+			char *fix_header = oscap_sprintf(
 				"###############################################################################\n#\n"
 				"# Bash remediation role for the results of evaluation of profile %s \n#\n"
 				"# Evaluation Result Start Time:  %s\n"
```
mpreisler: Could we just say Evaluation Start Time? The result is redundant here I think.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -685,13 +676,23 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
...
 		if (result == NULL) {
 			// Profile-based remediation fix
+			struct xccdf_profile *profile = xccdf_policy_get_profile(policy);
+			const char *profile_id = xccdf_profile_get_id(profile);
+			struct oscap_text_iterator *profile_title_iterator = xccdf_profile_get_title(profile);
+			const char *profile_title = oscap_textlist_get_preferred_plaintext(profile_title_iterator, NULL);
+			struct oscap_text_iterator *profile_description_iterator = xccdf_profile_get_description(profile);
+			const char *profile_description = oscap_textlist_get_preferred_plaintext(profile_description_iterator, NULL);
+
+			oscap_text_iterator_free(profile_title_iterator);
```
mpreisler: Could you please rearrange this so that we get the iterator, get the string and then free the iterator. then we do the same for the other iterator. It decreases the scope for variables and makes it easier to read and spot mistakes.

src/XCCDF_POLICY/xccdf_policy_remediate.c
kjankov: Fixed

src/XCCDF_POLICY/xccdf_policy_remediate.c
kjankov: Fixed

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -661,17 +661,78 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+			const char *start_time = xccdf_result_get_start_time(result);
+			const char *end_time = xccdf_result_get_end_time(result);
+
+			char *fix_header = oscap_sprintf(
+				"###############################################################################\n#\n"
+				"# Bash remediation role for the results of evaluation of profile %s \n#\n"
+				"# Evaluation Start Time:  %s\n"
+				"# Evaluation End Time:  %s\n#\n"
+				"# This file was generated by OpenSCAP %s using one of the following:\n"
+				"# 	$ oscap xccdf generate fix --result-id xccdf-results.xml \n"
```
mpreisler: this won't work because --result-id has to have some value to it

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -661,17 +661,84 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+
+			char *fix_header = oscap_sprintf(
+				"###############################################################################\n#\n"
+				"# Bash remediation role for the results of evaluation of profile %s \n"
+				"# XCCDF Version:  %s\n#\n"
+				"# Evaluation Start Time:  %s\n"
+				"# Evaluation End Time:  %s\n#\n"
+				"# This file was generated by OpenSCAP %s using one of the following:\n"
+				"# 	$ oscap xccdf generate fix --result-id %s xccdf-results.xml \n"
+				"# 	$ oscap xccdf generate fix --result-id %s arf.xml \n#\n"
```
mpreisler: Could you please add `--template %s` where `sys` is substituted for %s to all the command line examples? It can happen that an XCCDF file has multiple fixes and they are all "bash-like" but you get a different set depending on which `--template` value you use. `--template` is just a very shitty name for fix-type.

If you do that you will also be able to recycle all this code for ansible as well.

kjankov:
When using normal xccdf-results.xml files (instead of arf files), you will get a warning similar to:
WARNING: Skipping ./ssg-fedora-oval.xml file which is referenced from XCCDF content

which occurs in the maint-1.2 branch without these modifications as well

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -661,20 +661,91 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+
+		fix_header = oscap_sprintf(
+			"###############################################################################\n#\n"
+			"# Bash remediation role for the results of evaluation of profile %s \n"
+			"# XCCDF Version:  %s\n#\n"
+			"# Evaluation Start Time:  %s\n"
+			"# Evaluation End Time:  %s\n#\n"
+			"# This file was generated by OpenSCAP %s using one of the following:\n"
+			"# 	$ oscap xccdf generate fix --result-id %s --template %s xccdf-results.xml \n"
+			"# 	$ oscap xccdf generate fix --result-id %s --template %s arf.xml \n#\n"
```
mpreisler: if sys is "" this will be invalid.

kjankov:
@mpreisler Ok, I will fix that. Valgrind was reporting it as a memory leak so I wasn't sure

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -661,20 +661,91 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
-			"   tasks:\n";
+	if (!(oscap_streq(sys, "") || oscap_streq(sys, "urn:xccdf:fix:script:sh") || oscap_streq(sys, "urn:xccdf:fix:commands") ||
+		oscap_streq(sys, "urn:xccdf:fix:script:ansible")))
+		return 0; // no header required
+
+	const char *oscap_version = oscap_get_version();
+	char *fix_header;
 
-		return _write_text_to_fd(output_fd, ansible_header);
+	// Bash Script
```
mpreisler: what is this comment for? seems outdated

src/XCCDF_POLICY/xccdf_policy_remediate.c
kjankov: This is fixed. I think I amended the commit, but you have to refresh it to see it

tests/API/XCCDF/unittests/test_fix_arf.sh
```diff
@@ -33,9 +34,10 @@ grep -q "$bash_line1" $script
 grep -q "$bash_line2" $script
 
 # Generate an Ansible playbook from a profile in ARF file
-$OSCAP xccdf generate fix --profile $profile --template $ansible_template --output $playbook $results_arf >$stdout 2>$stderr
-diff $playbook $srcdir/$name.playbook1.yml
-[ -f $stdout ]; [ ! -s $stdout ]; rm $stdout
+$OSCAP xccdf generate fix --profile $profile --template $ansible_template $results_arf | grep -Ev $regex >$playbook 2>$stderr
+echo "$(cat $playbook)"
```
mpreisler: this should probably be removed.

tests/API/XCCDF/unittests/test_fix_arf.sh
```diff
@@ -50,9 +52,10 @@ grep -q -v "$bash_line1" $script
 grep -q "$bash_line2" $script
 
 # Generate  an Ansible playbook based on scan results stored in ARF file
-$OSCAP xccdf generate fix --result-id $result_id --template $ansible_template --output $playbook $results_arf >$stdout 2>$stderr
-diff $playbook $srcdir/$name.playbook2.yml
-[ -f $stdout ]; [ ! -s $stdout ]; rm $stdout
+$OSCAP xccdf generate fix --result-id $result_id --template $ansible_template $results_arf | grep -Ev $regex >$playbook 2>$stderr
+echo "$(cat $playbook)"
```
mpreisler: debugging left overs

tests/API/XCCDF/unittests/test_fix_arf.sh
```diff
@@ -50,9 +52,10 @@ grep -q -v "$bash_line1" $script
...
+diff -B $playbook $srcdir/$name.playbook2.yml
+#[ -f $stdout ]; [ ! -s $stdout ]; rm $stdout
```
mpreisler: remove this, $stdout is not used in the test that precedes this

tests/API/XCCDF/unittests/test_fix_bash_header.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
```
mpreisler: this file suggests we are testing bash headers but we are also testing ansible headers, I suggest we rename it or split it into 2 files

tests/API/XCCDF/unittests/test_fix_arf.sh
```diff
@@ -12,6 +12,7 @@ ansible_task1a="\- name: ensure everything passes"
 ansible_task1b="shell: /bin/true"
 ansible_task2a="\- name: correct the failing case"
 ansible_task2b="shell: /bin/false"
+regex="^$|[\#]$|^([\#][[:blank:]][^\-])"
```
mpreisler: I highly recommend commenting what regexes do, regexes are hard to read even for people with a ton of experience with them

src/XCCDF_POLICY/xccdf_policy_remediate.c
mpreisler: Could we avoid having --template in the example in case sys is "". That is the most correct solution and will work even if the assumption that "" is bash remediations is broken in the future.

tests/API/XCCDF/unittests/test_fix_bash_header.sh
kjankov: Is test_fix_script_header ok?

mpreisler: Yeah, rename the related XCCDF as well please.

tests/API/XCCDF/unittests/test_fix_arf.sh
kjankov: Done

tests/API/XCCDF/unittests/test_fix_bash_header.sh
kjankov: Done

tests/API/XCCDF/unittests/test_fix_arf.sh
kjankov: Fixed

tests/API/XCCDF/unittests/test_fix_arf.sh
kjankov: Fixed

tests/API/XCCDF/unittests/test_fix_arf.sh
kjankov: Fixed

utils/oscap-xccdf.c
```diff
@@ -858,8 +858,19 @@ int app_generate_fix(const struct oscap_action *action)
 	} else { // Fallback to profile if result id is missing
 		/* Profile-oriented fixes */
 		if (!xccdf_session_set_profile_id(session, action->profile)) {
-			report_missing_profile(action);
-			goto cleanup2;
```
mpreisler: there used to be `cleanup2` but now we just to `cleanup`. Is that intentional?

mpreisler: this was fixed

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -661,20 +661,93 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+		struct xccdf_profile *profile = xccdf_policy_get_profile(policy);
+		const char *profile_id = xccdf_profile_get_id(profile);
+		// Title
+		struct oscap_text_iterator *title_iterator = xccdf_profile_get_title(profile);
+		const char *profile_title = oscap_textlist_get_preferred_plaintext(title_iterator, NULL);
+		oscap_text_iterator_free(title_iterator);
+		// Description
+		struct oscap_text_iterator *description_iterator = xccdf_profile_get_description(profile);
+		const char *profile_description = description_iterator != NULL ?
+				oscap_textlist_get_preferred_plaintext(description_iterator, NULL) : "Not available";
```
jan-cerny: I think that this is a memory leak. Profile_title and profile_description need to be freed.

```
==726== 43 bytes in 1 blocks are definitely lost in loss record 2 of 4
==726==    at 0x4C2BBAD: malloc (vg_replace_malloc.c:299)
==726==    by 0x793F079: strdup (in /usr/lib64/libc-2.23.so)
==726==    by 0x4E8E5D8: oscap_strdup (util.c:65)
==726==    by 0x4E8DD99: oscap_text_get_plaintext (text.c:203)
==726==    by 0x4E8DF97: oscap_textlist_get_preferred_plaintext (text.c:255)
==726==    by 0x4F3E4B8: _write_script_header_to_fd (xccdf_policy_remediate.c:682)
==726==    by 0x4F3E832: xccdf_policy_generate_fix (xccdf_policy_remediate.c:770)
==726==    by 0x40D9D8: app_generate_fix (oscap-xccdf.c:876)
==726==    by 0x4089C9: oscap_module_call (oscap-tool.c:261)
==726==    by 0x408E4E: oscap_module_process (oscap-tool.c:346)
==726==    by 0x40781C: main (oscap.c:80)
```

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -661,20 +661,93 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+
+		struct xccdf_benchmark *benchmark = xccdf_policy_get_benchmark(policy);
+		const char *benchmark_version_info = xccdf_benchmark_get_version(benchmark);
+		const char *benchmark_id = xccdf_benchmark_get_id(benchmark);
+		const struct xccdf_version_info *xccdf_version = xccdf_benchmark_get_schema_version(benchmark);
+		const char *xccdf_version_name = xccdf_version_info_get_version(xccdf_version);
+
+		fix_header = oscap_sprintf(
+			"###############################################################################\n#\n"
+			"# Bash remediation role for profile %s\n"
```
kjankov: Fixed

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -661,20 +661,93 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+		fix_header = oscap_sprintf(
+			"###############################################################################\n#\n"
+			"# Bash remediation role for the results of evaluation of profile %s \n"
+			"# XCCDF Version:  %s\n#\n"
+			"# Evaluation Start Time:  %s\n"
+			"# Evaluation End Time:  %s\n#\n"
+			"# This file was generated by OpenSCAP %s using one of the following:\n"
+			"# 	$ oscap xccdf generate fix --result-id %s%s%s xccdf-results.xml \n"
+			"# 	$ oscap xccdf generate fix --result-id %s%s%s arf.xml \n#\n"
+			"# This script is generated from the results of a profile evaluation.\n"
```
kjankov: So I've been looking into how to do this, and i either would have to pass the filename into the xccdf_policy_generate_fix as a parameter (which causes all sorts of issues with the python wrapper) or add some kind of getter/setter into one of the structures (policy, benchmark, etc)

src/XCCDF_POLICY/xccdf_policy_remediate.c
kjankov: Fixed

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -661,20 +661,95 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+			"# This file was generated by OpenSCAP %s using:\n"
+			"# 	$ oscap xccdf generate fix --profile %s%s%s sds.xml \n#\n"
+			"# This script is generated from an OpenSCAP profile without preliminary evaluation.\n"
+			"# It attempts to fix every selected rule, even if the system is already compliant.\n#\n"
+			"###############################################################################\n\n",
+				script ? "Ansible" : "Bash", profile_id, profile_title, profile_description, benchmark_id,
+				benchmark_version_info, xccdf_version_name, oscap_version, profile_id, template, format);
+
+		oscap_free(profile_title);
+		oscap_free(profile_description);
```
mpreisler: this gives compiler warnings

kjankov:
@jan-cerny  I fixed the memory leak, and the filename input into the header isn't really feasible (it breaks the API/Python wrapper).

mpreisler:

src/XCCDF_POLICY/xccdf_policy_remediate.c
kjankov: Fixed, made profile_title and profile_description not const

src/XCCDF_POLICY/xccdf_policy_remediate.c
kjankov: Fixed

src/XCCDF_POLICY/xccdf_policy_remediate.c
kjankov: Fixed

src/XCCDF_POLICY/xccdf_policy_remediate.c
kjankov: Fixed

jan-cerny:
Awesome! Thank you.

ACK.

PR #758, opened 2017-06-23
[tests] Add clang analyzer script to check false positive

xiangzhai:
Hi openscap developers,

clang analyzer might falsely flag bugs https://bugzilla.gnome.org/show_bug.cgi?id=777077 @pwithnall it wasted developers' valuable time to double check the False Positive issues, so I simply added clang_analyzer.sh, it is easy to use:

```
$ ./clang_analyzer.sh ../lib/vasnprintf.c
.
.
.
../lib/vasnprintf.c:5191:31: note: Use of memory after it is freed
                           : (dp->conversion == 'c' || dp->conversion == 's'
                              ^~~~~~~~~~~~~~
1 warning generated.
```

if you considered it might be a false positive, please report the bug https://bugs.llvm.org//enter_bug.cgi?product=clang and Cc to me lesliezhai@llvm.org.cn thanks a lot! 

Regards,
Leslie Zhai

### DIFF
--- a/tests/clang_analyzer.sh
+++ b/tests/clang_analyzer.sh
@@ -36,11 +36,14 @@ cmd="clang -cc1 -analyze -disable-free -disable-llvm-verifier \
     -mthread-model posix -fmath-errno -masm-verbose -mconstructor-aliases \
     -munwind-tables -fuse-init-array -target-cpu x86-64 \
     -momit-leaf-frame-pointer -dwarf-column-info -debugger-tuning=gdb \
-    -resource-dir /usr/lib/clang/$clang_version -D HAVE_CONFIG_H -I .. -I ../.. \
-    -I /usr/include/libxml2 -I ./public -I ../../src -I ../../src/common/public \
-    -I ../../src/source/public -I ../../src/XCCDF_POLICY/public \
-    -I ../../src/XCCDF/public -I ../../src/CPE/public -D _GNU_SOURCE \
-    -D OSCAP_THREAD_SAFE -D _POSIX_C_SOURCE=200112L -D PIC \
+    -resource-dir /usr/lib/clang/$clang_version -D HAVE_CONFIG_H \
+    -I.. -I/usr/include/libxml2 -I../src -I../src/common/public \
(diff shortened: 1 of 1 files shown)


jan-cerny:
This is interesting. Do you think that we can include the static analysis into the upstream test suite? Or perhaps create a Jenkins job that would run the analysis nightly?

xiangzhai:
@jan-cerny CI https://travis-ci.org/ is helpful! and KDE use Jenkins https://build.kde.org/job/k3b%20master%20kf5-qt5/ I am the maintainer of K3B, but I am puzzed how to distinguish whether it is False Positive or Potential bugs, my script is help people to recognize them manually :)

matejak:
Hello, thank you for your contribution! And sorry to keep you waiting for so long.
While testing the script to proceed towards its merge, we have hit the `'stddef.h' file not found` [issue](https://clang.llvm.org/docs/FAQ.html). The amount of arguments you supply to `clang` is huge, some of it is probably platform-specific and it may cause headers not to be found.
Could you please comment those arguments somewhere in the script? As a side note, arguments are best passed as an array, so you can use the pattern
```
args=()
...
# Now we do this to achieve something
args+=(-option -another-option=value)
...
# Now we execute clang with args
clang "${args[@]}"
```
and explain the trickiest arguments to `clang` at the moment you add them.

matejak:
I took a look again at this PR @xiangzhai  are you interested in working on it?

- The problem with `stddef.h` was caused by some options passed to `clang`. Namely, the problematic args were those containing `-internal-isystem ...`, and `-internal-externc-isystem ...` and `-resource-dir /usr/lib/clang/$clang_version`.
- Then, plenty of arguments were not recognized by `clang5.0` and I had to take them out in order to avoid errors (e.g. `-main-file-name`, `-mrelocation-model`, ...).
- Finally, after cleaning up the script, I got rid of errors, but it started to complain about missing symbols, which indicated that clang thought more about compiling an executable than analysing the source.

Leslie, we need your help to get this merged, the static analysis capabilities of clang are not documented well enough and the script is not working for us at all with `clang5.0`.

matejak:
We are closing this due to inactivity.

Issue #723, opened 2017-04-20
oscap generate fix complains about remote resources even though it doesn't need them


mpreisler:
No need to load OVAL to generate remediations.

```
Scanning dependencies of target generate-ssg-centos5-role.yml
[ 40%] [centos5-guides] generating urn:xccdf:fix:script:ansible remediation roles for all profiles in ssg-centos5-ds.xml
WARNING: This content points out to the remote resources. Use `--fetch-remote-resources' option to download them.
WARNING: Skipping http://www.redhat.com/security/data/oval/Red_Hat_Enterprise_Linux_5.xml file which is referenced from XCCDF content
WARNING: This content points out to the remote resources. Use `--fetch-remote-resources' option to download them.
WARNING: Skipping http://www.redhat.com/security/data/oval/Red_Hat_Enterprise_Linux_5.xml file which is referenced from XCCDF content
Generated '/home/mpreisle/d/scap-security-guide/build/roles/ssg-centos5-role-default.yml' for profile ID '' in benchmark 'xccdf_org.ssgproject.content_benchmark_RHEL-5', template=urn:xccdf:fix:script:ansible.Generated '/home/mpreisle/d/scap-security-guide/build/roles/ssg-centos5-role-stig-rhel5-server-upstream.yml' for profile ID 'xccdf_org.ssgproject.content_profile_stig-rhel5-server-upstream' in benchmark 'xccdf_org.ssgproject.content_benchmark_RHEL-5', template=urn:xccdf:fix:script:ansible.[ 40%] Built target generate-ssg-centos5-role.yml
Scanning dependencies of target generate-ssg-centos5-role.sh
[ 40%] [centos5-guides] generating urn:xccdf:fix:script:sh 
```

jan-cerny:
I think we can skip OVAL completely in "oscap xccdf generate fix". I will verify.

mpreisler:
Fixed https://github.com/OpenSCAP/openscap/pull/764

PR #764, opened 2017-06-27
Issue #723: Don't require other resources in "generate fix"

jan-cerny:
Checks are not necessary to load OVAL because the remediation
scripts are stored in the XCCDF. This will remove the warning
for remote resources that are not needed.

### DIFF
--- a/src/XCCDF/public/xccdf_session.h
+++ b/src/XCCDF/public/xccdf_session.h
@@ -42,6 +42,20 @@
 struct xccdf_session;
 
 /**
+ * Loading flags for XCCDF session
+ * @memberof xccdf_session
+ * The flags set which components will be loaded in XCCDF session.
+ */
+typedef enum {
+	XCCDF_SESSION_LOAD_NONE = 0,
+	XCCDF_SESSION_LOAD_XCCDF = 1 << 0,
+	XCCDF_SESSION_LOAD_CPE = 1 << 1,
+	XCCDF_SESSION_LOAD_OVAL = 1 << 2,
+	XCCDF_SESSION_LOAD_CHECK_ENGINE_PLUGINS = 1 << 3,
+	XCCDF_SESSION_LOAD_ALL = XCCDF_SESSION_LOAD_XCCDF | XCCDF_SESSION_LOAD_CPE | XCCDF_SESSION_LOAD_OVAL | XCCDF_SESSION_LOAD_CHECK_ENGINE_PLUGINS
+} xccdf_session_loading_flags_t;
+
+/**
  * Costructor of xccdf_session. It attempts to recognize type of the filename.
  * @memberof xccdf_session
  * @param filename path to XCCDF or DS file.
@@ -202,6 +216,14 @@ void xccdf_session_set_user_tailoring_cid(struct xccdf_session *session, const c
  * to NULL -- ignoring user notification.
  */
 void xccdf_session_set_remote_resources(struct xccdf_session *session, bool allowed, download_progress_calllback_t callback);
+
+/**
+ * Disable or allow loading of depending content (OVAL, SCE, CPE)
+ * @memberof xccdf_session
+ * @param session XCCDF Session
+ * @param flags Bit mask that sets loading of other content in the session.
+ */
+void xccdf_session_set_loading_flags(struct xccdf_session *session, xccdf_session_loading_flags_t flags);
 
 /**
  * Set custom oval files for this session
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -112,6 +112,7 @@ struct xccdf_session {
 	bool full_validation;				///< True value indicates that every possible step will be validated by XSD.
 
 	struct oscap_list *check_engine_plugins; ///< Extra non-OVAL check engines that may or may not have been loaded
+	xccdf_session_loading_flags_t loading_flags; ///< Load referenced files while loading XCCDF
 };
 
 static int _xccdf_session_autonegotiate_tailoring_file(struct xccdf_session *session, const char *original_path);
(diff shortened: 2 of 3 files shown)


mpreisler:
I know this is not necessary here but thinking further ahead I think we should go with a bitfield for this.

Have something like:
```
XCCDF_SESSION_LOAD_CPE = 1 << 0
XCCDF_SESSION_LOAD_OVAL = 1 << 1
XCCDF_SESSION_LOAD_CHECK_ENGINE_PLUGINS = 1 << 2

XCCDF_SESSION_LOAD_ALL = XCCDF_SESSION_LOAD_CPE | XCCDF_SESSION_LOAD_OVAL | XCCDF_SESSION_LOAD_CHECK_ENGINE_PLUGINS

xccdf_session_set_load_flags(session, XCCDF_SESSION_LOAD_CPE | XCCDF_SESSION_LOAD_OVAL)
...
```

The advantage is that we can extend this without breaking the API.

What do you think?

mpreisler:

jan-cerny:
That's a nice idea.

I can imagine a use-case that we would like to test something on eg. SSG RHEL7, but we're using Fedora, and we don't want to spend time with VM, so we could evaluate a benchmark without the CPE check, by adding something like `--no-cpe`.

I don't think that adding another variable and functions would break the API, but flags are more flexible.

jan-cerny:
I have improved the commit. The commit now introduces new type, xccdf_session_flags_t, which specifies flags for XCCDF session. These flags drive loading content in XCCDF session.

src/XCCDF/public/xccdf_session.h
```diff
@@ -198,6 +211,14 @@ void xccdf_session_set_user_tailoring_cid(struct xccdf_session *session, const c
 void xccdf_session_set_remote_resources(struct xccdf_session *session, bool allowed, download_progress_calllback_t callback);
 
 /**
+ * Disable or allow loading of depending content (OVAL, SCE, CPE)
+ * @memberof xccdf_session
+ * @param session XCCDF Session
+ * @param flags Bit mask that sets loading of other content in the session.
+ */
+void xccdf_session_set_loading_flags(struct xccdf_session *session, xccdf_session_loading_flags_t flags);
```
mpreisler: shouldn't this take an int or a short instead of enum?

src/XCCDF/xccdf_session.c
```diff
@@ -549,12 +556,23 @@ int xccdf_session_load(struct xccdf_session *session)
...
 	if ((ret = xccdf_session_load_xccdf(session)) != 0)
 		return ret;
-	if ((ret = xccdf_session_load_cpe(session)) != 0)
-		return ret;
-	if ((ret = xccdf_session_load_oval(session)) != 0)
-		return ret;
-	if ((ret = xccdf_session_load_check_engine_plugins(session)) != 0)
-		return ret;
+
+	xccdf_session_loading_flags_t flags = session->loading_flags;
```
mpreisler: const

src/XCCDF/public/xccdf_session.h
```diff
@@ -42,6 +42,19 @@
 struct xccdf_session;
 
 /**
+ * Loading flags for XCCDF session
+ * @memberof xccdf_session
+ * The flags set which components will be loaded in XCCDF session.
+ */
+typedef enum {
+	XCCDF_SESSION_LOAD_NONE = 0,
```
mpreisler: this is misleading, it will still load the XCCDF

src/XCCDF/public/xccdf_session.h
jan-cerny: Why? The `xccdf_session_loading_flags_t` was defined using `typedef`.

src/XCCDF/xccdf_session.c
jan-cerny: OK

src/XCCDF/public/xccdf_session.h
jan-cerny: Thank you. What wouldn't be misleading? XCCDF_SESSION_LOAD_ONLY_XCCDF or XCCDF_SESSION_LOAD_NO_COMPONENT or something else?

yuumasato: Maybe just XCCDF_SESSION_LOAD_XCCDF?

mpreisler: The problem is that since it's zero there is no way to "remove" it from the bitfield. Maybe it would be enough to have _NONE there but have it documented that it does load the XCCDF and will load parse the DS to get the XCCDF out of it.

jan-cerny: I'm sorry, but now I'm confused. Should it be non-zero value? If yes, should it be called XCCDF_SESSION_LOAD_XCCDF? Or do you just suggest keeping this and writing a comment there?

Tbh I can't find a meaningful example of creating an XCCDF session without loading an XCCDF document 😬

jan-cerny: @mpreisler @yuumasato Can you help me with my confusion?

yuumasato: > The problem is that since it's zero there is no way to "remove" it from the bitfield. 

Oh, now I see it.

> Tbh I can't find a meaningful example of creating an XCCDF session without loading an XCCDF document 😬

I understand you, but who knows what use cases might come up in the future? 😬
And I see XCCDF_SESSION_LOAD_XCCDF to be non zero and removable from bitfield as  future proof implementation.

So, I'd suggest to have
XCCDF_SESSION_LOAD_NONE = 0 and 
XCCDF_SESSION_LOAD_XCCDF = 1 << 0
and shift the other values.

jan-cerny: OK

jan-cerny:
I have improved the flags.

yuumasato:
Ack.

src/XCCDF/public/xccdf_session.h
mpreisler: If we pass in a bitflag combination is it still the enum?

yuumasato: Oh, yeah, It will not be any of the defined enum then.

PR #766, opened 2017-06-28
Remove unrelated text in the middle of other text

jan-cerny:


mpreisler:
hehe, probably copy pasta...

PR #767, opened 2017-06-28
[WIP] Suffix/short id option for --result-id arguments 

kjankov:
Allows for user to input suffix for the result id; i.e., common instead of xccdf_org.open-scap_testresult_xccdf_org.ssgproject.content_profile_common 
Works for xccdf generate fix --result-id and xccdf generate report --result-id

### DIFF
--- a/src/XCCDF/benchmark.c
+++ b/src/XCCDF/benchmark.c
@@ -518,6 +518,10 @@ struct xccdf_result *xccdf_benchmark_get_result_by_id(struct xccdf_benchmark *be
 
 struct xccdf_result *xccdf_benchmark_get_result_by_id_suffix(struct xccdf_benchmark *benchmark, const char *testresult_suffix)
 {
+	struct xccdf_result *init_result = xccdf_benchmark_get_result_by_id(benchmark, testresult_suffix);
+	if (init_result != NULL)
+		return init_result;
+
 	const char *final_result_id = NULL;
 	struct xccdf_result_iterator *result_iterator = xccdf_benchmark_get_results(benchmark);
 
@@ -537,12 +541,7 @@ struct xccdf_result *xccdf_benchmark_get_result_by_id_suffix(struct xccdf_benchm
 	}
 	xccdf_result_iterator_free(result_iterator);
 
-	if (final_result_id == NULL) {
-		return NULL;
-	} else {
-		struct xccdf_result *final_result =  xccdf_benchmark_get_result_by_id(benchmark, final_result_id);
-		return final_result;
-	}
+	return xccdf_benchmark_get_result_by_id(benchmark, final_result_id);
 }
 
 bool xccdf_benchmark_add_content(struct xccdf_benchmark *bench, struct xccdf_item *item)
--- a/src/XCCDF/benchmark.c
+++ b/src/XCCDF/benchmark.c
@@ -516,6 +516,35 @@ struct xccdf_result *xccdf_benchmark_get_result_by_id(struct xccdf_benchmark *be
 	return result;
 }
 
+struct xccdf_result *xccdf_benchmark_get_result_by_id_suffix(struct xccdf_benchmark *benchmark, const char *testresult_suffix)
+{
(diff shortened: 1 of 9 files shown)


mpreisler:
was this branched off maint-1.2? it seems to have some commits from the script header PR #757.

could you please rebase off fresh maint-1.2 to get rid of the merge commit? drop the commits that are in #757 to keep things tidy. thx!

src/XCCDF/benchmark.c
```diff
@@ -516,6 +516,35 @@ struct xccdf_result *xccdf_benchmark_get_result_by_id(struct xccdf_benchmark *be
...
+				final_result_id = result_full_id;
+			}
+		}
+	}
+	xccdf_result_iterator_free(result_iterator);
+
+	if (final_result_id == NULL) {
+		return NULL;
+	} else {
+		struct xccdf_result *final_result =  xccdf_benchmark_get_result_by_id(benchmark, final_result_id);
```
mpreisler: don't use a separate variable for this, we can just return it as it is

mpreisler:
I don't like that `xccdf_session_build_policy_from_testresult` gets these super powers of selecting testresult by suffix. It's not mentioned in docs and IMO it's not the right place to implement it. I'd prefer to have a function that will get you the full result ID from a suffix and then you can use it all over the API. It's more readable and transparent too that way.

src/XCCDF/benchmark.c
```diff
@@ -516,6 +516,34 @@ struct xccdf_result *xccdf_benchmark_get_result_by_id(struct xccdf_benchmark *be
...
+					final_result_id, result_full_id);
+				break;
+			} else {
+				final_result_id = result_full_id;
+			}
+		}
+	}
+	xccdf_result_iterator_free(result_iterator);
+
+	return xccdf_benchmark_get_result_by_id(benchmark, final_result_id);
```
mpreisler: Can't we just remember the result from the loop and return that? Any reason why we do this lookup again?

kjankov: Fixed

src/XCCDF/benchmark.c
kjankov: Fixed

jan-cerny:
@kjankov Can we have also a test for ARF (result datastreams)?

Although ARF is more complicated format than XCCDF results, we recommend users using ARF instead of XCCDF results, because it contains all the information in a single file, including OVAL system characteristics. Therefore I think users will use the short IDs with ARF more often than with XCCDF results, so it might be worth testing.

jan-cerny:
@kjankov Pull requests should never contain `Merge branch ..` commits. 

You can get rid of it using interactive rebase `git rebase -i origin/maint-1.2` and then `git push --force [your branch]`.

kjankov:
@jan-cerny Merge commits were fixed.

The suffix matching for --result-id works with arf files, but implementing a test for it would be fairly difficult- I would want to have a reasonably sized xml file for the test in the unittests directory, but using a "fake" benchmark doesn't really work with arf. It has a lot of other dependencies and sources needed for a valid result file. Multiple files would probably be necessary. It might be possible but I don't really think it's worth it.

jan-cerny:
@kjankov 
If you have some time, you can read about improving the commit messages:
https://chris.beams.io/posts/git-commit/

mpreisler:
@jan-cerny I am confused, this was a WIP PR, right?

plus:

> I don't like that xccdf_session_build_policy_from_testresult gets these super powers of selecting testresult by suffix. It's not mentioned in docs and IMO it's not the right place to implement it. I'd prefer to have a function that will get you the full result ID from a suffix and then you can use it all over the API. It's more readable and transparent too that way.

jan-cerny:
@mpreisler It was done, there was just the WIP forgotten.

The function that will get you the full result ID from a suffix and then you can use it all over the API was implemented in https://github.com/OpenSCAP/openscap/pull/767/commits/386627c19f78dda13a00b6f5dc38c5c33506ea6c

PR #769, opened 2017-06-29
Implement xccdf requires

yuumasato:
Parse XCCDF `requires` element and check if their `idrefs` are selected.
Rule results in 'notselected' if any `requires` element is not met.

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy.c
+++ b/src/XCCDF_POLICY/xccdf_policy.c
@@ -1001,11 +1001,11 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
 		// One requires element has no idref met
 		if (!any_require_selected) {
 			all_requires_met = false;
-			dI("A requires element from Rule '%s' requires that any of the following XCCDF Items is also selected.", rule_id);
+			dW("A requires element from Rule '%s' requires that any of the following XCCDF Items is also selected.", rule_id);
 			requires_ids = oscap_stringlist_get_strings(requires_element);
 			while (oscap_string_iterator_has_more(requires_ids)) {
 				const char *requires_id = oscap_string_iterator_next(requires_ids);
-				dI("%s", requires_id);
+				dW("%s", requires_id);
 			}
 			oscap_string_iterator_free(requires_ids);
 		}
@@ -1013,7 +1013,7 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
 	oscap_stringlist_iterator_free(requires_elements);
 
 	if (!all_requires_met) {
-		dI("Rule '%s' will be unselected because one of its requires element was not met.", rule_id);
+		dW("Rule '%s' will be unselected because one of its requires element was not met.", rule_id);
 
 		// We need to report result before unselecting rule
 		// This is necessary for the callback to know if a rule result is notselected
--- a/src/XCCDF_POLICY/xccdf_policy.c
+++ b/src/XCCDF_POLICY/xccdf_policy.c
@@ -981,6 +981,47 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
 		return _xccdf_policy_report_rule_result(policy, result, rule, NULL, XCCDF_RESULT_NOT_SELECTED, NULL);
 	}
 
+	bool all_requires_met = true;
+	// Check if all requires elements are met
+	struct oscap_stringlist_iterator *requires_elements = xccdf_item_get_requires((struct xccdf_item*)rule);
+	while (oscap_stringlist_iterator_has_more(requires_elements) && all_requires_met) {
+
+		struct oscap_stringlist *requires_element = oscap_stringlist_iterator_next(requires_elements);
+		struct oscap_string_iterator *requires_ids = oscap_stringlist_get_strings(requires_element);
+		bool any_require_selected = false;
+		while (oscap_string_iterator_has_more(requires_ids) && !any_require_selected) {
+			const char *requires_id = oscap_string_iterator_next(requires_ids);
+			if (xccdf_policy_is_item_selected(policy, requires_id)){
+				dD("Rule '%s' requires %s, which is selected.", rule_id, requires_id);
+				any_require_selected = true;
+			} else {
+				dD("Rule '%s' requires %s, which is not selected.", rule_id, requires_id);
+			}
+		}
+		oscap_string_iterator_free(requires_ids);
+
+		// One requires element has no idref met
+		if (!any_require_selected) {
+			all_requires_met = false;
+			dI("A requires element from Rule '%s' requires that any of the following XCCDF Items is also selected.", rule_id);
+			requires_ids = oscap_stringlist_get_strings(requires_element);
+			while (oscap_string_iterator_has_more(requires_ids)) {
+				const char *requires_id = oscap_string_iterator_next(requires_ids);
+				dI("%s", requires_id);
+			}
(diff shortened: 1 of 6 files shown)


yuumasato:
Sample evaluation output from test case:
```
$ oscap xccdf eval --profile xccdf_com.example.www_profile_test_requires_10 --results results.xml --report report.html test_requires.xccdf.xml 
Title   Rule number 3
Rule    xccdf_com.example.www_rule_3
Result  pass

Title   Rule number 4
Rule    xccdf_com.example.www_rule_4
Result  pass

Title   Rule number 10
Rule    xccdf_com.example.www_rule_10
W: oscap: A requires element from Rule 'xccdf_com.example.www_rule_10' requires that any of the following XCCDF Items is also selected.
W: oscap: xccdf_com.example.www_rule_5
W: oscap: xccdf_com.example.www_rule_6
W: oscap: Rule 'xccdf_com.example.www_rule_10' will be unselected because one of its requires element was not met.
Result  notselected
```

And snippet from `results.xml`
```
<rule-result idref="xccdf_com.example.www_rule_10" time="2017-06-29T15:43:22" weight="1.000000">
      <result>notselected</result>
      <message severity="info">At least one requires element was not met.</message>
    </rule-result>
```

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -981,6 +979,52 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
 		return _xccdf_policy_report_rule_result(policy, result, rule, NULL, XCCDF_RESULT_NOT_SELECTED, NULL);
 	}
 
+	bool all_requires_met = true;
```
mpreisler: the impl for this shouldn't be here, this is slow and unnecessary. we need to do this at resolve stage.

mpreisler:
What about Group requires?

I think this all needs to be done at resolve stage. The recursive code (O(n) in a pathological case) can't be run for every item, that makes the whole thing O(n^2)!

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -981,6 +979,52 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
...
+		oscap_string_iterator_free(requires_ids);
+
+		// One requires element has no idref met
+		if (!any_require_selected) {
+			all_requires_met = false;
+			dW("A requires element from Rule '%s' requires that any of the following XCCDF Items is also selected.", rule_id);
+			requires_ids = oscap_stringlist_get_strings(requires_element);
+			while (oscap_string_iterator_has_more(requires_ids)) {
+				const char *requires_id = oscap_string_iterator_next(requires_ids);
+				dW("%s", requires_id);
```
jan-cerny: There can be some whole sentence.

src/XCCDF_POLICY/xccdf_policy.c
yuumasato: I don't understand how this should be done at resolve stage. 
I didn't manage to reach `xccdf_resolve_rule` or `xccdf_resolve_group` during evaluation.

If I understand correctly, by looking at XCCDF specification, resolving is related to loading the XCCDF document.
And the `requires` and `conflicts` are handled during traversal, and described in section "Item Processing Algorithm".

mpreisler: @yuumasato each profile is resolved before it can be evaluated. We have all the information we need to select or unselect things at resolve stage. Can you elaborate why this can't be done at resolve stage? You probably didn't reach xccdf_resolve_* because your benchmark was already resolved.

yuumasato: >  I didn't manage to reach xccdf_resolve_rule or xccdf_resolve_group during evaluation.

They were not reached because Rule and Group elements in my Benchmark don't extend any other Rule or Group. 

From specification:
> An XCCDF document that has no @extends or @abstract attributes is called a resolved document.

The Resolve stage is about resolving item inheritance defined in `extends` attribute.

@mpreisler It is possible to do it there, but do we want to  process `requires` and `conflicts` elements at this stage?

yuumasato:
> The recursive code (O(n) in a pathological case) can't be run for every item, that makes the whole thing O(n^2)!

Could you elaborate more on the code being O(n^2)?
The nested while is there because`requires_elements` is list of list of strings.

src/XCCDF_POLICY/xccdf_policy.c
yuumasato: Dwelling on possibilities, at the resolve stage there may be Items not resolved whose selected state may change according to inheritance of its parent item.

mpreisler:
@yuumasato to make it not O(n^2) we need to cache all our full require resolves. For example:

```
A requires B
B requires C
C requires D
...
Y requires Z
```

Your algo:
```
for every require of A look for every require of B look for every require of C, ...
for every require of B look for every require of C ...
```

Suggested algo:
```
For every require of A look for every require of B, once we figure out requires of B cache them. requires_satisfied[B] = true.
...
```

Your algo is O(n^2), suggested algo is O(n).

I claim we can do this at resolve stage easily. Resolve should (probably) fatal error when some requires can't be satisfied in some profiles. This even makes content builds safer. And we do no work when users are evaluating. Users will resolve just once and then evaluate on a lot of machines. Saving even more work.

src/XCCDF_POLICY/xccdf_policy.c
mpreisler: I would assume we do requires as the last step of the resolve. If the spec goes against us there is nothing we can do but it's a shame.

mpreisler: I just reviewed the spec and it's not mentioned in the resolving workflow. Maybe we have to do it for every evaluated even if the benchmark has been resolved. IMO that's an issue in the spec, looks like the bolted on "requires" after they had the resolve workflow set-up.

yuumasato:
@mpreisler Sorry, that is not how the algorithm is working. And suggested algo is not how it is supposed to work too.

The algo should not follow down the links of requires until there is not more requires. It should just look at the referenced item and verify if is selected.

The algo is working like this:
```
for every requires of A, look if at least one of its idref is selected. If any requires doesn't have at least one of its idref selected, unselect A.
for every requires of B, look ...
```

`Example #2 – Ordering of requires/conflicts` in section 7.2.3.3.2 illustrates in more detail the algorithm.

mpreisler:
@yuumasato I don't understand the difference. Is it just that it's looking for anything in requires that is satisfied and that's enough? Instead of looking through everything? How does that make a difference to complexity?

What if A requires B but B requires C and C is unselected. That should unselect A and B as well, right?

yuumasato:
>I don't understand the difference. Is it just that it's looking for anything in requires that is satisfied and that's enough? Instead of looking through everything? 

Yes, a requires element is a list of idrefs, if any of the listed id is selected, this requires element is good. Now a Rule can have multiple requires elements, and for each requires element, at least one idref should be selected.
The tricky part is that, only the moment when the Rule is evaluated that its required items selection state matter, if one of its requires go unselected after, it will be inconssistent, but thats how the spec says it should be.

> How does that make a difference to complexity?

Every Rule is evaluated once, every `requires` element is evaluated only once. It should be linear to the quantity and size of the `requires` elements.

> What if A requires B but B requires C and C is unselected. That should unselect A and B as well, right?

It will depend on the order of the rules in the Benchmark.

If the rules are listed as A, B and C
- A will be left selected,
- B will be deselected, 
- C will be left selected.

If rules in the Benchmark are ordered like C, B, A, all rules will be unselected.

yuumasato:
Closing this for now.

Issue #773, opened 2017-07-04
Remediation scripts are a mix of Bash and Ansible code


jan-cerny:
The Bash remediation script, generated from SSG Datastream, contains both Ansible and Bash code.

That means the script cannot be run!

I use this command:
```
oscap xccdf generate fix --profile xccdf_org.ssgproject.content_profile_common ../scap-security-guide-0.1.34/ssg-fedora-ds.xml > fix.sh
```
See the attached file
[fix.sh.zip](https://github.com/OpenSCAP/openscap/files/1121671/fix.sh.zip)

I used OpenSCAP build from current upstream head, and SSG 0.1.34 from tarball downloaded from GitHub. The issue is reproducible with openscap-1.2.14-1.fc24.x86_64 as well.

```
bash fix.sh
Remediating rule 1/74: 'xccdf_org.ssgproject.content_rule_ensure_gpgcheck_globally_activated'
fix.sh: line 24: -: command not found
fix.sh: line 25: stat:: command not found
fix.sh: line 26: path:: command not found
fix.sh: line 27: register:: command not found
fix.sh: line 28: check_mode:: command not found

......
.....

```

jan-cerny:
If `--template` is specified, the issue doesn't happen. But if no `--template` is provided, it simply takes the last `<xccdf:fix>`, no matter what language is it. That's wrong. It should take bash as default.

kjankov:
@jan-cerny See PR https://github.com/OpenSCAP/openscap/pull/777

mpreisler:
@kjankov @jan-cerny offered a fix in #776 , it looks like your fix is a duplicate.

jan-cerny:
I'm sorry

jan-cerny:
Fixed in https://github.com/OpenSCAP/openscap/commit/a6c35af3f8cbb19f799323a912d816d6e13b6a31

PR #776, opened 2017-07-04
Issue #773: Ensure that bash fixes are always selected

jan-cerny:
Bash is a default language for remediation scripts. If we don't
provide a template explicitely, the result script shouldn't
contain Ansible or other snippets.

### DIFF
--- a/utils/oscap-xccdf.c
+++ b/utils/oscap-xccdf.c
@@ -255,6 +255,7 @@ static struct oscap_module XCCDF_GEN_FIX = {
     .usage = "[options] xccdf-file.xml",
     .help = GEN_OPTS
         "\nFix Options:\n"
+		"   --fix-type <type>\r\t\t\t\t - Fix type. Should be one of: bash, ansible, puppet, anaconda (default: bash).\n"
         "   --output <file>\r\t\t\t\t - Write the script into file.\n"
         "   --result-id <id>\r\t\t\t\t - Fixes will be generated for failed rule-results of the specified TestResult.\n"
 		"   --template <id|filename>\r\t\t\t\t - Fix template. (default: bash)\n"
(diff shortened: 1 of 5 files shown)


mpreisler:
I am in favor of this. Have you checked why we don't do this already? Maybe there is a reason. Selecting a default explicitly is an obvious solution so maybe there is a catch.

Could we also get some shorthand for --fix_type bash? Instead of having to type the huge urn style template?

redhatrises:
Having the `--fix-type` or `--script-type` would be a handy feature. You could make it so that the default of the option is set to bash.

jan-cerny:
@mpreisler I don't think there is, I think we only haven't noticed before, because there used to be only bash in SSG. There is a code that selects the best fix according to various criteria - reboot, disruption, etc, but those criteria should be considered AFTER we select a language.

I think bash should be default if `--template` isn't specified, because I have found this in the man page:
```
                     --template ID|FILE
                            Template to be used to generate the script. If  it
                            contains a dot '.' it is interpreted as a location
                            of a file with the template definition.  Otherwise
                            it  identifies  a template from standard set which
                            currently includes: bash (default if no --template
                            switch  present). Brief explanation of the process
                            of writing your own templates is in the  XSL  file
                            xsl/legacy-fix.xsl in the openscap data directory.
                            You can also take a look at the  default  template
                            xsl/legacy-fixtpl-bash.xml.
```
and --help says:
```
   --template <id|filename>      - Fix template. (default: bash)
```

@redhatrises Yes, that would be nice. I can't remember `urn:xccdf:fix:script:sh`. I usually write it wrong. I suppose our users don't remember it as well. :) Moreover it's badly documented

jan-cerny:
Btw, `template` sounds weird for this.

jan-cerny:
I have added `--fix-type` option into the "oscap xccdf generate fix" module.

utils/oscap-xccdf.c
```diff
@@ -791,11 +792,41 @@ int app_generate_fix(const struct oscap_action *action)
...
+				"Option '--fix-type' is mutually exclusive with '--template'.\n"
+				"Please provide only one of them.\n");
+		return OSCAP_ERROR;
+	} else if (action->fix_type != NULL) {
+		if (strcmp(action->fix_type, "bash") == 0) {
+			template = "urn:xccdf:fix:script:sh";
+		} else if (strcmp(action->fix_type, "ansible") == 0) {
+			template = "urn:xccdf:fix:script:ansible";
+		} else if (strcmp(action->fix_type, "puppet") == 0) {
+			template = "urn:xccdf:fix:script:puppet";
```
mpreisler: anaconda would also be nice so that we support everything we use in SSG.

utils/oscap.8
```diff
@@ -331,6 +331,9 @@ Result-oriented fixes are generated using result-id provided to select only the
 .TP
 Profile-oriented fixes are generated using all rules within the provided profile. If no result-id/profile are provided, (default) profile will be used to generate fixes.
 .TP
+\fB\-\-fix-type TYPE\fR
+Specify fix type. There are multiple programming languages in which the fix script can be generated. TYPE should be one of: bash, ansible, puppet. Default is bash.
```
mpreisler: Explain that it is mutually exclusive with --template and why. Add anaconda.

mpreisler:
Fails the test suite. The failure looks legitimate.

utils/oscap-xccdf.c
jan-cerny: OK

utils/oscap.8
jan-cerny: OK

jan-cerny:
There is a problem: attribute `system` of element `fix` isn't required.

yuumasato:
LGTM. Ack.

PR #777, opened 2017-07-05
Added line that assigns default bash template when template argument …

kjankov:
…is missing from generate fix. Ensures that both bash and ansible scripts don't appear in same file

In function _find_fix_for_template in xccdf_policy_remediate.c, fixes for rules are only filtered by template if template is not null. Thus if template is null, no filtering occurs, and both bash and ansible scripts will be printed, but a missing template argument should default to bash script only. This fix assigns a null template to a bash template.


mpreisler:
Duplicate of #776 
I talked to @jan-cerny and he will work on #776 so don't worry about this issue :-)

Issue #772, opened 2017-07-04
Timestamp in result-oriented remediation script header is wrong


jan-cerny:
I generate a bash script from ARF results using this command:

```
oscap xccdf generate fix --result-id xccdf_org.open-scap_testresult_xccdf_org.ssgproject.content_profile_common arf.xml > fix.sh
```

Evaluation start time in the script header is wrong. It should report the time when the evaluation started. Instead, it shows the time when the fix was generated from an ARF.

See:

```
# Evaluation Start Time:  2017-07-04T11:34:14
# Evaluation End Time:  2017-07-04T11:26:07
```

It says that evaluation was started after it was finished 😬

mpreisler:
Fixed by #804

jan-cerny:
:1st_place_medal:

PR #779, opened 2017-07-11
Removed line that overwrites the original result start time that is s…

kjankov:
…et at the beginning of evaluation


kjankov:
Fix for issue https://github.com/OpenSCAP/openscap/issues/772

jan-cerny:
I think that this isn't the right fix. It breaks test `tests/API/XCCDF/unittests/test_remediate_simple.sh`.

mpreisler:
@kjankov @jan-cerny any updates here?

jan-cerny:
@mpreisler I don't have any update. I still think it isn't right fix, because it breaks test tests/API/XCCDF/unittests/test_remediate_simple.sh. Therefore we can't merge it now.

jan-cerny:
I think I have a right fix, see https://github.com/OpenSCAP/openscap/pull/804

mpreisler:
Replaced by #804

PR #780, opened 2017-07-13
Html report group by improvements (CJIS, 800-171)

mpreisler:
Added support for shortnames of 800-171 and FBI CJIS.

### DIFF
--- a/xsl/xccdf-report-impl.xsl
+++ b/xsl/xccdf-report-impl.xsl
@@ -456,10 +456,10 @@ Authors:
     <xsl:param name="href"/>
     <xsl:choose>
         <xsl:when test="starts-with($href, 'http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-53')">
-            <xsl:text>NIST SP 800-53 ID</xsl:text>
+            <xsl:text>NIST SP 800-53</xsl:text>
         </xsl:when>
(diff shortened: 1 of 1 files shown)


xsl/xccdf-report-impl.xsl
```diff
@@ -452,12 +452,15 @@ Authors:
...
 
-<xsl:template name="convert-url-to-name">
+<xsl:template name="convert-reference-url-to-name">
     <xsl:param name="href"/>
     <xsl:choose>
         <xsl:when test="starts-with($href, 'http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-53')">
             <xsl:text>NIST SP 800-53 ID</xsl:text>
         </xsl:when>
+        <xsl:when test="starts-with($href, 'http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-171')">
+            <xsl:text>NIST SP 800-171 ID</xsl:text>
```
shawndwells: We're inconsistent with naming. Some are appended with ``ID``. aka ``DISA CCI`` vs ``NIST SP 800-171 ID``.

Suggest we drop the ``ID``

jan-cerny: +1

mpreisler: done

PR #787, opened 2017-07-31
tests/mitre/test_mitre.sh: do not run SELinux test on systems without SELinux

matusmarhefka:
* added check if SELinux is enabled and if it is not the test
  linux-def_selinuxsecuritycontext_test.xml is skipped

### DIFF
--- a/tests/mitre/test_mitre.sh
+++ b/tests/mitre/test_mitre.sh
@@ -15,6 +15,8 @@ EXTVARFILE="${MITRE_FILES}/var/linux-external-variables.xml"
 DISTRO="$(cat /etc/*-release | head -n1)"
 DISTRO_NAME="$(cat /etc/*-release | awk '{print $1}' | head -n1)"
 DISTRO_RELEASE="$(cat /etc/*-release | sed -n 's|^[^0-9]*\([0-9]*\).*$|\1|p' | head -n1)"
+SELINUX_ENABLED=$(grep -iqE "(SELINUX=enforcing|SELINUX=permissive)" \
+	/etc/selinux/config 2>/dev/null; echo $?)
 
 function test_mitre {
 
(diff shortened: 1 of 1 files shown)


tests/mitre/test_mitre.sh
```diff
@@ -15,6 +15,8 @@ EXTVARFILE="${MITRE_FILES}/var/linux-external-variables.xml"
 DISTRO="$(cat /etc/*-release | head -n1)"
 DISTRO_NAME="$(cat /etc/*-release | awk '{print $1}' | head -n1)"
 DISTRO_RELEASE="$(cat /etc/*-release | sed -n 's|^[^0-9]*\([0-9]*\).*$|\1|p' | head -n1)"
+SELINUX_ENABLED=$(grep -iqE "(SELINUX=enforcing|SELINUX=permissive)" \
+	/etc/selinux/config 2>/dev/null)
```
matusmarhefka: Fixed in commit 60df8d1. The error was that instead of return value of that grep command I saved its output into SELINUX_ENABLED variable which was always empty because of '-q' cmd option passed to grep.

yuumasato:
Ack.

mpreisler:
Couldn't get my feedback here in time. Grepping a config file is an odd choice, can't we run `getenforce` and be safer? If `which getenforce` returns 1 we are sure SELinux is not on the machine.

matusmarhefka:
@mpreisler I was considering that, but what if your system runs SELinux and you don't have `libselinux-utils` package installed (which contains getenforce)? In this case `which getenforce` will return error and we would not detect that SELinux is actually running. Another reason is that the file `/etc/selinux/config` is part of `selinux-policy` package:
```
$ dnf repoquery --requires --resolve selinux-policy
bash-0:4.3.42-7.fc24.x86_64
selinux-policy-0:3.13.1-191.24.fc24.noarch
```
and `selinux-policy` does not have `libselinux-utils` package as a dependency. I think that in this case it is safer to check the configuration file rather than to rely on the binary which might not be installed on a system.

By the way, `getenforce` also looks into `/etc/selinux/config`:
```
$ strace getenforce 2>&1 | grep "/etc/selinux"
access("/etc/selinux/config", F_OK)     = 0
```

PR #789, opened 2017-08-01
Skip empty sysctl values

jan-cerny:
We will skip empty values to be consistent with "sysctl" tool.
If this behavior will be considered a bug in "sysctl" tool,
we will need to revert this commit. See RHBZ #1473207.

This should bring our broken tests working back.

### DIFF
--- a/src/OVAL/probes/unix/sysctl.c
+++ b/src/OVAL/probes/unix/sysctl.c
@@ -186,6 +186,16 @@ int probe_main(probe_ctx *ctx, void *probe_arg)
 
                         fclose(fp);
 
+			/* Skip empty values as sysctl tool does.
+			 * If this behavior of sysctl will be indentified as a bug,
+			 * we will have to remove this block.
+			 * See https://bugzilla.redhat.com/show_bug.cgi?id=1473207
+			 */
+			if (l == 0) {
+				dI("Skipping file '%s' because it has no value.", mibpath);
+				continue;
+			}
+
                         /*
                          * sanitize the value
                          *  - only printable and whitespace chars allowed


dahaic:
Jenkins test this please.

src/OVAL/probes/unix/sysctl.c
```diff
@@ -186,6 +186,12 @@ int probe_main(probe_ctx *ctx, void *probe_arg)
 
                         fclose(fp);
 
+						/* Skip empty values as sysctl tool does */
+						if (l == 0) {
```
matusmarhefka: Besides of 6 tabs indent (instead of 3 tabs) looks okay to me.

dahaic:
Can we reference the bugzilla even in the code comment? To help future selves with explanation what is going on?

jan-cerny:
@dahaic Yes, we can!

src/OVAL/probes/unix/sysctl.c
jan-cerny: Hmmm 🥇 Looked OK in my vim

jan-cerny:
I have fixed the indents, actually 4 tabs looks correct. I have improved the comment and added a link to Bugzilla.

src/OVAL/probes/unix/sysctl.c
```diff
@@ -186,6 +186,16 @@ int probe_main(probe_ctx *ctx, void *probe_arg)
 
                         fclose(fp);
 
+			/* Skip empty values as sysctl tool does.
+			 * If this behavior of sysctl will be indentified as a bug,
+			 * we will have to remove this block.
+			 * See https://bugzilla.redhat.com/show_bug.cgi?id=1473207
+			 */
+			if (l == 0) {
+				dI("Skippping file '%s' because it has no value.", mibpath);
```
dahaic: Sorry, one more nitpick - ppp

jan-cerny: Ooopsp. Fixed.

dahaic:
Thanks!

PR #790, opened 2017-08-01
[WIP] Fast iterator

mpreisler:
In another episode of "wtf is this in our codebase", I found that we are doing a ton of jumps and use the filtering iterator for everything even though the filters are only used on 2 lines in the entire codebase.

```
fast-iterator make check: 13m48.488s
maint-1.2 make check: 15m22.458s
```

And I haven't gotten rid of the delegation jumps yet. For every "typed" iterator (the type is actually fake, just for readability) we do a jump that is not inlined. If I finish getting rid of the delegation jumps and fix up the iterators we save a bit more.

Example:
```
#define OSCAP_ITERATOR_HAS_MORE(n) bool n##_iterator_has_more(OSCAP_ITERATOR(n) it) { return oscap_fast_iterator_has_more(ITERATOR_CAST(it)); }
#define OSCAP_ITERATOR_NEXT(t,n) t n##_iterator_next(OSCAP_ITERATOR(n) it) { return oscap_fast_iterator_next(ITERATOR_CAST(it)); }
```

so for oscap_string_iterator_has_more we jump into the function which immediately loads the arguments and jumps into oscap_iterator_has_more which then loads the filter (if it's not used it's not NULL as you would expect, it's a "always true" function instead but because it's a function pointer we have to jump into it and can't inline it) and jumps at least twice into the filter.

Opening a PR for discussion. I think a part of this can be done in maint-1.2 but if we want to really make iteration blazing fast we can only do it in master. What do you think?

(by blazing fast I mean iterators on the stack, 128bits per iterator, everything inlined, no extra features like filtering, no delegation jumps, no safety - if you use it wrong we segfault)

### DIFF
--- a/src/common/list.c
+++ b/src/common/list.c
@@ -337,6 +337,59 @@ void oscap_iterator_reset(struct oscap_iterator * it)
     it->cur = NULL;
 }
 
+void *oscap_fast_iterator_new(struct oscap_list *list)
+{
+	struct oscap_fast_iterator *it = oscap_calloc(1, sizeof(struct oscap_fast_iterator));
+	it->cur = NULL;
+	it->list = list;
+	return it;
+}
+
+void oscap_fast_iterator_free(struct oscap_fast_iterator *it)
+{
+	oscap_free(it);
+}
+
+void *oscap_fast_iterator_detach(struct oscap_fast_iterator *it)
+{
+	__attribute__nonnull__(it);
+
+	if (!it->cur)
+		return NULL;
+
+	struct oscap_list_item *item = it->cur;
+	void *value = item->data;
+
+	assert(it->list->first != NULL && it->list->last != NULL);
+
+	if (it->list->first == it->list->last) {
+		assert(it->list->first == it->list->last);
+		assert(it->list->first == item);
+		it->list->first = it->list->last = NULL;
+		it->cur = NULL;
+	} else if (item == it->list->first) {
+		assert(it->list->first != it->list->last);
+		assert(item->next != NULL);
+		it->list->first = item->next;
+		it->cur = NULL;
+	} else {
+		struct oscap_list_item *cur = it->list->first;
+		while (cur->next != item) {
+			assert(cur->next != NULL);
+			cur = cur->next;
+		}
+		assert(cur->next == item);
+		cur->next = item->next;
+		if (item == it->list->last)
+			it->list->last = cur;
+		it->cur = cur;
+	}
+
+	free(item);
+	--it->list->itemcount;
+	return value;
+}
+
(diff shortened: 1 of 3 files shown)


jan-cerny:
This sounds great! The code looks OK.

BTW, what is `oscap_iterator_detach` good for?

mpreisler:
> BTW, what is oscap_iterator_detach good for?

Looking at the code it's to remove the currently iterator over item from the list.

jan-cerny:
The currently iterator over item? What's that? I thought that iterator can walk only by a list.

mpreisler:
Correction: It removes the currently iterated over item.

So you iterate over a list, when you are at item 3 you can detach it and continue iterating. Detaching it from the list removes it from the list without deallocating it. We use it a bunch of times in the codebase.

jan-cerny:
@mpreisler Aha, thanks :D 

The changes that you propose look great.

Regarding your question, I would go for writing in maint-1.2 as much as possible. After merging this into maint-1.2 we can merge maint-1.2 into the master, then you could do the blazing iterators in a separate pull request on master.

jan-cerny:
@mpreisler If I understand it well, this PR is not WIP anymore, because the blazing iterators can't go to maint-1.2?

mpreisler:
> @mpreisler If I understand it well, this PR is not WIP anymore, because the blazing iterators can't go to maint-1.2?

This is WIP, this PR in this state breaks the test suite and uses undefined behavior :-) Currently it's just a test.

jan-cerny:
@mpreisler And the blazing iterators?

mpreisler:
> @mpreisler And the blazing iterators?

I have no clue what your question is. This PR needs to be finished before it can get merged into `maint-1.2` but if we do this carefully this can go into `maint-1.2`. Currently it breaks tests because it was just a quick hack to see if it would make a difference.

We can make iterators even better and faster in `master` by keeping them on the stack. That fixes memory leaks and makes the code way simpler in a lot of places. Plus it makes iterators even quicker. This can never go into `maint-1.2` though so that's why I am not doing that in this PR.

jan-cerny:
So the only thing remaining here is the breaking test, and the blazing iterators will not be here?

mpreisler:
> So the only thing remaining here is the breaking test

Way more things remain... For one all the iterator getters still return using the old constructor. So we allocate 4x64 bits but only use the first 2x64 bits... As I described again and again, this is a WIP prototype to see if this change has any impact. It needs to be finished and can't be merged at this point.

> and the blazing iterators will not be here?

That is correct if by blazing iterator you mean iterators allocated on the stack.

jan-cerny:
OK, thanks for clarification.

mpreisler:
..... nobody cares enough :-)

If I am wrong you are free to re-use these patches and resubmit.

Issue #792, opened 2017-08-07
OVAL textfilecontent54 multiline behavior incorrectly defaults to false under specific conditions


balleman:
Textfilecontent54's has a 'multiline' behavior that should default to true [1].  Under the following conditions, it appears that openscap defaults the multiline behavior to false:
* filepath element is used
* behaviors element exists that sets a behavior other than multiline, e.g., ignore_case="true"

The desired behavior can be forced by explicitly setting multiline="true", but this should be unnecessary, as it is the prescribed default.

[1] https://oval.mitre.org/language/version5.10.1/ovaldefinition/documentation/independent-definitions-schema.html#Textfilecontent54Behaviors

balleman:
I'm working on a pull request to add a test case and potential solution.

PR #793, opened 2017-08-07
OVAL Textfilecontent54 multiline behavior default bugfix

balleman:
This PR attempts to fix issue #792 by changing the "reset filebehavior attributes if 'filepath' entity is used" block to only set attributes if they exist, allowing the defaults in the probe_tfc54behaviors_canonicalize() function to be effective.

Test cases are also added to check some permutations of behaviors.  The checks for oval:filepath:tst:5 fail without the changes to the textfilecontent54 probe.

I'm more confident in the bug than the fix; feel free to point out problems in the fix or better approaches.


mpreisler:
This is awesome, thank you for your contribution! I will take a closer look at this later today but massive kudos for creating test cases.

mpreisler:
@openscap-ci test this please

jan-cerny:
@balleman Thank you for your contribution!

PR #796, opened 2017-08-09
No need to require libpcre when you depend on libopenscap

isimluk:
https://bugzilla.redhat.com/show_bug.cgi?id=1414777

Merge at your own risk. 😜 

🥞 🐼 🚽 🤙


mpreisler:
This change looks fine but we need to look into the semantics of the pkg-config file. Maybe we can drop all the Requires from it because we don't bleed them into the public API.

mpreisler:
Thanks for your contribution @isimluk !

:carousel_horse: :speedboat: :dolphin: :shipit:

PR #797, opened 2017-08-09
Deprecate oscap allocators

mpreisler:
We talked about this at today's C workshop, oscap_free, oscap_alloc, oscap_calloc, etc are all just thin wrappers around the real thing. There are some debugging additional capabilities but it's not really very useful. We have much much better tools for debugging memory issues.

This is not a big deal but it's confusing to new contributors because they have to learn distinctions between oscap_free vs free and when to use which and so on. It's confusing and frankly unnecessary.

As a hidden bonus we gain a little bit of speed because the compiler can use builtins. Plus we get better static analysis because the tools understand malloc, free, etc...

TODO:
- ~oscap_free~ DONE
- ~oscap_alloc~ DONE
- ~oscap_calloc~ DONE
- ~the rarer allocator functions~ DONE

EDIT: We were breaking const correctness in quite a few cases. This is undefined behavior according to the standard but because gcc isn't taking advantage of it it worked fine for years... It was hidden because oscap_free would always cast to `void*` silently. Moving to `free` uncovered these issues so I fixed them all. There was one public function that returns `const char*` even though you should free the buffer, I can't fix that in public API in maint-1.2 but I have commented that and we will fix it in `master` - `const char *oval_schema_version_to_cstr(oval_schema_version_t version)`


mpreisler:
```
make check
new code: 10m54.976s
old code: 11m16.147s
```

This is about what I would expect with a change like this. We avoid a few instructions, avoid a bunch of jumps and in rare cases the optimizer will even rip out malloc and free altogether because of builtins. Nothing to write home about...

I would also replace a few calloc calls with malloc. In cases where we immediately fill the entire structure in the ctor it makes no sense really to use calloc. I suspect the gains will be minimal though so I won't bother.

src/OVAL/oval_component.c
```diff
@@ -868,29 +868,29 @@ void oval_component_free(struct oval_component *component)
 	case OVAL_COMPONENT_OBJECTREF:{
 			oval_component_OBJECTREF_t *objectref = (oval_component_OBJECTREF_t *) component;
 			if (objectref->item_field != NULL)
```
jan-cerny: I think this condition can be removed. But there will be more of them I guess. It isn't necessary to fix them.

mpreisler: Fixed

src/OVAL/oval_behavior.c
```diff
@@ -115,12 +115,12 @@ void oval_behavior_free(struct oval_behavior *behavior)
 	__attribute__nonnull__(behavior);
 
 	if (behavior->value)
```
mpreisler: Fixed

src/OVAL/oval_test.c
```diff
@@ -260,10 +260,10 @@ void oval_test_free(struct oval_test *test)
 	__attribute__nonnull__(test);
 
 	if (test->comment != NULL)
-		oscap_free(test->comment);
+		free(test->comment);
 	if (test->id != NULL)
-		oscap_free(test->id);
-	oval_collection_free_items(test->notes, &oscap_free);
+		free(test->id);
+	oval_collection_free_items(test->notes, &free);
```
mpreisler: Fixed

src/XCCDF/value.c
```diff
@@ -397,12 +397,12 @@ struct xccdf_value_instance *xccdf_value_instance_new(xccdf_value_type_t type)
...
-		oscap_list_free(inst->choices, oscap_free);
-		oscap_free(inst->selector);
-        oscap_free(inst->match);
-        oscap_free(inst->value);
-        oscap_free(inst->defval);
-		oscap_free(inst);
+		oscap_list_free(inst->choices, free);
+		free(inst->selector);
+        free(inst->match);
+        free(inst->value);
```
mpreisler: Fixed

mpreisler:
Fixed the pointed out issues, there are countless of such issues though. I think it's not realistic to fix them all in this PR. Would be great to fix them gradually in other PRs.

jan-cerny:
@mpreisler Looks OK to me. So what is your opinion on merging it now and causing conflicts with other PRs, namely CVRF? Thank you.

mpreisler:
> @mpreisler Looks OK to me. So what is your opinion on merging it now and causing conflicts with other PRs, namely CVRF? Thank you.

IMO let's merge it and fix conflicts. CVRF is not ready yet.

PR #798, opened 2017-08-10
Don't reinvent builtins in our codebase

mpreisler:
I wanted to get rid of more of our reinvented builtins but in the end the best I could do was inline some of them. I was able to get rid of oscap_strtol though. Plus I discovered that oscap_find_file doesn't exist even though it's declared.

This is all just minor spring cleaning for fun, really. The performance gains are quite minimal.

### DIFF
--- a/src/XCCDF/value.c
+++ b/src/XCCDF/value.c
@@ -418,7 +418,7 @@ struct xccdf_value_instance *xccdf_value_new_instance(struct xccdf_value *val)
 	bool xccdf_value_instance_set_##WHAT##_string(struct xccdf_value_instance *inst, const char *newval) { \
 		oscap_free(inst->WHAT); inst->WHAT = oscap_strdup(newval); inst->flags.WHAT##_given = true; return true; }
 #define XCCDF_VALUE_INSTANCE_VALUE_ACCESSOR_NUM(WHAT) \
-	xccdf_numeric xccdf_value_instance_get_##WHAT##_number(const struct xccdf_value_instance *inst) { return oscap_strtol(inst->WHAT, NULL, 10); } \
+	xccdf_numeric xccdf_value_instance_get_##WHAT##_number(const struct xccdf_value_instance *inst) { return inst->WHAT == NULL ? NAN : strtol(inst->WHAT, NULL, 10); } \
 	bool xccdf_value_instance_set_##WHAT##_number(struct xccdf_value_instance *inst, xccdf_numeric newval) { \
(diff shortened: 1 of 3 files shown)


mpreisler:
Crap, atof depends on locale so this would break on Czech locale for example where a decimal point is `,` instead of `.`. What is the locale independent atof in C standard? My google-fu is failing me.

WesleyCeraso:
I could not find as well. Setting the locale to C is not an option?

mpreisler:
This is a can of worms, I will just use the old semi-broken behavior with `(float)strtol..` and we can fix this at some point later.

src/XCCDF/value.c
```diff
@@ -417,8 +417,9 @@ struct xccdf_value_instance *xccdf_value_new_instance(struct xccdf_value *val)
 	const char *xccdf_value_instance_get_##WHAT##_string(const struct xccdf_value_instance *inst) { return inst->WHAT; } \
 	bool xccdf_value_instance_set_##WHAT##_string(struct xccdf_value_instance *inst, const char *newval) { \
 		oscap_free(inst->WHAT); inst->WHAT = oscap_strdup(newval); inst->flags.WHAT##_given = true; return true; }
+// TODO: Parse floats correctly instead of converting long to float
```
mpreisler: or even better use fixed point arith but that would break API and ABI

src/common/util.h
```diff
@@ -403,15 +403,6 @@ void oscap_strtoupper(char *str);
 bool oscap_ptr_cmp(void *node1, void *node2);
 
 /**
- * find file with given name and mode in given paths
```
jan-cerny: It was removed in 71da0a6e96a913dcf66ece70c1b78da9d7fcf26d 5 years ago, so I guess we don't need it :D

jan-cerny:
When I clicked to resolve conflicts and I confirmed, it added automatically a commit "Merge branch" which is forbidden to be in pull request, so I am completely confused.

mpreisler:
> When I clicked to resolve conflicts and I confirmed, it added automatically a commit "Merge branch" which is forbidden to be in pull request, so I am completely confused.

It's not forbidden. It's actually a pretty common workflow to merge from main branch back to feature branch to fix conflicts. It's just that for the trivial cases we use rebase because it leaves simpler history. It's not a big deal here. Plus I like that the merge commit shows the conflict you had to fix.

mpreisler:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
Jenkins tested it, test passed, but the result is not reported here, because integration with Jenkins is broken at the moment, so I add a link to the build: https://jenkins.open-scap.org/job/openscap-pull-requests/117/

PR #799, opened 2017-08-11
Verify the container's filesystem location

### LINKED ISSUE #794: "oscap-docker" doesn't work with default container storage on Fedora 26
Fedora 26 has started to use overlay2 as a storage. Overlay has replaced devicemapper, which was used in older Fedoras. See https://fedoraproject.org/wiki/Changes/DockerOverlay2. 

Containers on devicemapper are mounted by `atomic.mount.DockerMount` into `[tmpdir]/rootfs`, but containers on overlay2 are mounted by `atomic.mount.DockerMount` into `[tmpdir]`, which isn't expected by oscap-docker.  It expects  `[tmpdir]/rootfs` always. See https://github.com/OpenSCAP/openscap/blob/maint-1.2/utils/oscap_docker_python/oscap_docker_util.py#L262 

As a result, OpenSCAP then tries to chroot to a non-existing directory, which leads to error of `oscap` utility, and subsequent tracebacks.

jan-cerny:
If the container or image doesn't mount in 'rootfs' subdirectory, we will fallback to the original mount directory.

Partially fixes #794 .

### DIFF
--- a/utils/oscap_docker_python/oscap_docker_util.py
+++ b/utils/oscap_docker_python/oscap_docker_util.py
@@ -218,9 +218,12 @@ class OscapScan(object):
             sys.stderr.write(str(e) + "\n")
             return None
 
-        # Remeber actual mounted fs in 'rootfs'
-        chroot = os.path.join(_tmp_mnt_dir, 'rootfs')
-
+        # Remember actual mounted fs in 'rootfs' for devicemapper
+        rootfs_path = os.path.join(_tmp_mnt_dir, 'rootfs')
+        if os.path.exists(rootfs_path):
+            chroot = rootfs_path
+        else:
+            chroot = _tmp_mnt_dir
 
         try:
             # Figure out which RHEL dist is in the chroot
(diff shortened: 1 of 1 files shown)


jan-cerny:
Should I fix these PEP8 issues in this pull request? The issues aren't in my code and aren't related to this pull request.

utils/oscap_docker_python/oscap_docker_util.py
```diff
@@ -203,6 +203,17 @@ def _remove_mnt_dir(self, mnt_dir):
...
+    def _find_chroot_path(self, mnt_dir):
+        '''
+        Remember actual mounted fs in 'rootfs' for devicemapper
+        '''
+        rootfs_path = os.path.join(mnt_dir, 'rootfs')
+        if os.path.exists(rootfs_path):
+            chroot = rootfs_path
+        else:
+            chroot = mnt_dir
+        return chroot
```
mpreisler: you can write all this in just one line: `return rootfs_path if os.path.exists(rootfs_path) else mnt_dir`

mpreisler:
> Should I fix these PEP8 issues in this pull request? The issues aren't in my code and aren't related to this pull request.

Nah, don't worry about it. I think we should avoid introducing more PEP8 errors but forcing everyone to fix all files whenever they are changed would be too much IMO.

utils/oscap_docker_python/oscap_docker_util.py
jan-cerny: Yes, I know, but that isn't readable.

yuumasato:
LGTM.

utils/oscap_docker_python/oscap_docker_util.py
mpreisler: IMO it is, this is a C-ism that is unnecessary in python. Just return it, don't use a variable. Shorter is better IMO. You don't need to deallocate anything before you return which is why this pattern is even used in C.

PR #800, opened 2017-08-14
Remove notselected rules and groups from HTML report

jan-cerny:
Notselected rules are never evaluated, so there is no need to include them in the Evaluation report.

This will remove the unwanted data from the HTML report, which will make the report smaller and possibly will make the searching and grouping functions faster.

See how size of the report is reduced:
```
[jcerny@thinkpad ~]$ ls -lh /tmp/report*.html        
-rw-rw-r--. 1 jcerny jcerny 723K Aug 14 12:50 /tmp/report-new.html                                        
-rw-rw-r--. 1 jcerny jcerny 1.4M Aug 14 12:50 /tmp/report-old.html
```

I also removed the `notselected` checkbox, because using it will have no effect.

### DIFF
--- a/xsl/xccdf-report-impl.xsl
+++ b/xsl/xccdf-report-impl.xsl
@@ -337,6 +337,7 @@ Authors:
     <xsl:variable name="ruleresult" select="key('testresult_ruleresults', concat($testresult/@id, '|', $item/@id))"/>
     <xsl:variable name="result" select="$ruleresult/cdf:result/text()"/>
 
+    <xsl:if test="$result != 'notselected'">
     <tr data-tt-id="{$item/@id}" class="rule-overview-leaf rule-overview-leaf-{$result} rule-overview-leaf-id-{$item/@id}" id="rule-overview-leaf-{generate-id($ruleresult)}">
         <xsl:attribute name="data-tt-parent-id">
             <xsl:value-of select="$item/parent::cdf:*/@id"/>
@@ -373,6 +374,7 @@ Authors:
             </div>
         </td>
     </tr>
+    </xsl:if>
 </xsl:template>
 
 <xsl:template name="rule-overview-inner-node">
@@ -385,6 +387,7 @@ Authors:
     <xsl:param name="error_rule_results"/>
     <xsl:param name="unknown_rule_results"/>
     <xsl:param name="notchecked_rule_results"/>
+    <xsl:param name="notselected_rule_results"/>
 
     <xsl:variable name="descendant_rules" select="$item/descendant::cdf:Rule"/>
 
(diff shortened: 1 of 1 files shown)


mpreisler:
If I remember correctly this is exactly what the report was doing when I first reimplemented it. But for some reason I had to add showing notselected as well. We need to look into why to make sure we don't break specification reqs or something like that.

jan-cerny:
@mpreisler I tried to look into why, but unfortunately I can't find anything in the commit history. Could you help me? Does specification say something about HTML reports?

mpreisler:
@jan-cerny The best I could find is https://martin.preisler.me/2014/07/openscap-html-report-redesign-part-2/ which is not great.

Also found these but no info about notselected.

```
commit f82d29ecf283ee1d8b28fc87dc8e2c7307833483
Author: Martin Preisler <mpreisle@redhat.com>
Date:   Tue Jul 22 18:09:20 2014 +0200

    "notselected" rules are hidden by default in HTML report
    
    Had to rewrite the rule hiding logic a bit to accomplish this.

commit 8562b378c829cdb59ae70ec7af420473e789af06
Author: Martin Preisler <mpreisle@redhat.com>
Date:   Thu Jul 17 17:11:53 2014 +0200

    Don't count notchecked and notselected into total rule count in HTML report
```

https://lists.fedorahosted.org/archives/list/scap-security-guide@lists.fedorahosted.org/thread/EVQMZQTLVQXLAQII5YQZKNJ3BAXFTYZX/

Personally I don't see any reason why we couldn't omit these from the HTML report. Does anybody else see any issue with not including notselected rules?

mpreisler:
(also wanted to say that this is a great PR title and description, gets to the point immediately and has all the info a reviewer needs)

shawndwells:
On 8/15/17 4:24 PM, Martin Preisler wrote:
>
> Personally I don't see any reason why we couldn't omit these from the
> HTML report. Does anybody else see any issue with not including
> notselected rules?
>
dropping them would be great. Frequently confuses users anyway.

jan-cerny:
@mpreisler @shawndwells Thank you.

xsl/xccdf-report-impl.xsl
```diff
@@ -588,6 +592,7 @@ Authors:
                 <xsl:variable name="error_rule_results" select="$testresult/cdf:rule-result[cdf:result/text() = 'error']"/>
                 <xsl:variable name="unknown_rule_results" select="$testresult/cdf:rule-result[cdf:result/text() = 'unknown']"/>
                 <xsl:variable name="notchecked_rule_results" select="$testresult/cdf:rule-result[cdf:result/text() = 'notchecked']"/>
+                <xsl:variable name="notselected_rule_results" select="$testresult/cdf:rule-result[cdf:result/text() = 'notselected']"/>
```
mpreisler: is this ever used for anything?

mpreisler:
ACK

mpreisler:
For the record jenkins https://jenkins.open-scap.org/job/openscap-pull-requests/118/ passed.

PR #803, opened 2017-08-15
Remove requires from pkgconfig file

mpreisler:
See #796
See https://bugzilla.redhat.com/show_bug.cgi?id=1414777

All the 3 "Requires" are not actually required when using openscap public API.


isimluk:
🏖 🥃

jan-cerny:
:swimming_man: :cocktail: :cake: :ice_cream: :tropical_drink: :trophy:

isimluk:
😋

PR #806, opened 2017-08-22
The enum value in oscap_string_map elements can be const

mpreisler:
And should be to avoid coding errors.

This is a pretty minor thing I noticed while working on the CVRF code. The assumption is that the elements of the oscap_string_map are const. We can make that clear by the const keyword on the int value.

Note fore reviewer: oscap_string_map is a bad name for what this is, it's actually one line/element of that map.


jan-cerny:
ACK

PR #807, opened 2017-08-23
Fix some warnings

jan-cerny:
This PR provides fixes for some of the warnings, but there is much more to be fixed.


src/OVAL/probes/SEAP/sexp-manip.c
```diff
@@ -2036,6 +2036,8 @@ static int __SEXP_sizeof_lmemb (const SEXP_t *s_exp, size_t *sz)
                 }
 
                 ret = SEXP_rawval_lblk_cb ((uintptr_t)SEXP_LCASTP(v_dsc.mem)->b_addr, (int(*)(SEXP_t *, void *))__SEXP_sizeof_lmemb, sz, 1);
+                (*sz) += sizeof (SEXP_valhdr_t) + v_dsc.hdr->size;
```
mpreisler: this is not explained in the commit message, I am not saying it's not a legit change but it should be explained.

mpreisler: oh actually it is sort of explained. nvm

mpreisler:
ACK

PR #808, opened 2017-08-23
Enable OFFLINE_RPMDB mode in rpminfo probe

### LINKED ISSUE #778: Scanning mounted filesystems without CHROOT
Use case:

Currently https://github.com/openshift/image-inspector/ uses OpenSCAP to scan Docker images. Root FS of a Docker image is pulled and extracted to a temp directory. We then run `oscap` commands with env vars set to enable `chroot` to the directory containing the image rootfs (in particular `OSCAP_PROBE_ROOT`).

The problem with this is that it requires `image-inspector` to have permission to run `chroot`. By default, pods (containers) running inside OpenShift are not able to `chroot` and must specifically request this permission.

We'd like to know if it's possible to scan mounted filesystems *without* chroot?

jan-cerny:
The custom path to RPM database was set before config files
were loaded, so it was overwritten by value from config files,
therefore it usually scanned the host RPMs. This should be
fixed by calling the addMacro() function after calling
rpmReadConfigFiles().

Helps to fix #778.

### DIFF
--- a/src/OVAL/probes/unix/linux/rpminfo.c
+++ b/src/OVAL/probes/unix/linux/rpminfo.c
@@ -270,7 +270,6 @@ void probe_preload ()
 void *probe_init (void)
 {
 	probe_setoption(PROBEOPT_OFFLINE_MODE_SUPPORTED, PROBE_OFFLINE_CHROOT|PROBE_OFFLINE_RPMDB);
-	addMacro(NULL, "_dbpath", NULL, getenv("OSCAP_PROBE_RPMDB_PATH"), 0);
 
 #ifdef HAVE_RPM46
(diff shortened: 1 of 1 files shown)


jan-cerny:
Example usage https://da.gd/BO5Y

jan-cerny:
We have discussed this today with the team and we think that for other probes it will be better to use OFFLINE_OWN mode, and replace usage of chroot. I will remove these last two commits. I will keep here only rpminfo probe. The other probes will be enabled in separate pull requests in future.

jan-cerny:
I have removed the useless commits.

yuumasato:
LGTM.
Something I noticed, `addMacro()` was deprecated in latest release http://rpm.org/wiki/Releases/4.14.0

PR #809, opened 2017-08-23
Include guard fixes

mpreisler:
Found a typo in one of the include guards that was causing issues when I included `sce_engine_api.h`. Fixed that. Then I ran `checkguard` to find issues in all the headers, fixed all of them.

Pretty minor stuff.


jan-cerny:
ACK

PR #810, opened 2017-08-25
Add python3 openscap_api.py

redhatrises:
- Moves openscap_api.py to src folder
- Updates openscap_api.py to handle python 3 as well as python 2.6+
- Python2 and Python3 now use the same openscap_api.py rather than having separate files.


jan-cerny:
@redhatrises This is a great idea.

However I don't like having two identical files in the repository. It's more difficult to maintain. If somebody changes one of the files, he will forget about the second file. Do you think that it's possible to share the code?

redhatrises:
@jan-cerny agreed as long as we don't have to support python2.5. What do you think about the proposed changes? (Can't tell why build is failing though.)

mpreisler:
@redhatrises Could you please rebase to make the history clean? This is adding an identical file, then removing the old file. This breaks rename tracking. Just move the file to the src path and don't add any file.

jan-cerny:
The tests are failing, because they can't find openscap_api now. I have found that there is a variable `PYTHONPATH` in `run.in` where the path is specified. Try to add the `swig/src` to `PYTHONPATH`. But I don't know if it will be the right fix, autotools are also confusing for me.

mpreisler:
@redhatrises I won't insist on it here but for future reference, it's not a good idea to do rename and change in one commit. It makes `merge` and `blame` confusing. Just change and then rename or vice versa.

mpreisler:
ACK

PR #812, opened 2017-08-28
RHBZ#1485876: Fix oscap running out of memory

jan-cerny:
https://bugzilla.redhat.com/show_bug.cgi?id=1485876

### DIFF
--- a/src/DS/sds.c
+++ b/src/DS/sds.c
@@ -624,7 +624,7 @@ static inline int ds_sds_compose_component_add_script_content(xmlNode *component
 	}
 }
 
-static int ds_sds_compose_add_component_internal(xmlDocPtr doc, xmlNodePtr datastream, const char* filepath, const char* comp_id, bool extended)
+static int ds_sds_compose_add_component_internal(xmlDocPtr doc, xmlNodePtr datastream, struct oscap_source *component_source, const char* comp_id, bool extended)
 {
 	xmlNsPtr ds_ns = xmlSearchNsByHref(doc, datastream, BAD_CAST datastream_ns_uri);
 	if (!ds_ns)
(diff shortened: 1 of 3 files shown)


jan-cerny:
Now I'll look into how to avoid loading evertyhing twice.

jan-cerny:
So far I have reduced memory usage on my test command `oscap xccdf eval --oval-results --results results.xml --report report.html /usr/share/xml/scap/ssg/content/ssg-rhel7-xccdf.xml` from 923 MB to 826 MB. Although I reduced it about 10 %, I think it's still too much.

mpreisler:
Good job!

But yeah, it's still too much. Try using valgrind's massif to figure out what takes this much memory. Check out http://valgrind.org/docs/manual/ms-manual.html

jan-cerny:
I'm doing massif right now.

jan-cerny:
Most of the memory is used by XML parsing. Most of the massif report starts somewhere in `libxml2` and goes sooner or later to `oscap_source_get_xmlDoc()`.

I have played with debugger and I have noticed that we parse `ssg-rhel7-oval.xml` twice when I run my test command.

mpreisler:
See http://www.xmlsoft.org/xmlmem.html#General4

```
If you are generating the DOM tree then memory requirements will grow nearly linear with the size of the data. In general for a balanced textual document the internal memory requirement is about 4 times the size of the UTF8 serialization of this document (example the XML-1.0 recommendation is a bit more of 150KBytes and takes 650KBytes of main memory when parsed). Validation will add a amount of memory required for maintaining the external Dtd state which should be linear with the complexity of the content model defined by the Dtd
```

How big is the final ARF? We can't do better than 4x its size in memory consumption.

mpreisler:
One easy way to reduce memory consumption would be to free `session->xccdf.source` after it's parsed into OpenSCAP models. It's a tradeoff though. If we need to reload the session we will need to load it from disk again. This is used in SCAP Workbench for example. For most usages in the `oscap` tool freeing it after it's parsed will have no effect and will return memory to libc.

jan-cerny:
I have discovered that to avoid loading OVAL twice we would have to have access to xccdf session when we build source datastram when we build ARF when we create report with OVAL details.

jan-cerny:
> I have discovered that to avoid loading OVAL twice we would have to have access to xccdf session when we build source datastram when we build ARF when we create report with OVAL details.

And from what I have seen in the code, that could be really hard and dangerous, we would have to merge whole source datastream and xccdf session logic. I think that should be done in a separate pull request. 
@mpreisler I will remove the WIP label from this.

jan-cerny:
To clarify, this PR doesn't fix the RHBZ#1485876 completely, it just helps a little to fix it.

mpreisler:
ACK

PR #813, opened 2017-08-28
Test generate guide update

dahaic:
Noticed we lack check that `oscap xccdf generate guide --profile` puts only rules enabled by the profile into the guide. When I was at it, I split that awful `all.sh` script to separate scenarios.


dahaic:
I have also noticed #814 but we cannot merge test for that until it's fixed :)

mpreisler:
ACK

PR #815, opened 2017-08-28
Test to cover filecontent test of non-utf files

dahaic:
Relates to commit https://github.com/OpenSCAP/openscap/commit/abc52f8ba07fb48efcc34228ed394234157b9c9c


dahaic:
FYI, forced push to move it from `API/OVAL/unittests` directory into `probes/textfilecontent54`

jan-cerny:
@dahaic Great! Thank you! :1st_place_medal:

PR #816, opened 2017-08-29
Enables make check to use a different oscap binary

WesleyCeraso:
This commit allows us to use any oscap tool in the system with the make check by setting the variable OSCAP.

However, the following tests link against libopenscap_testing.la and don't use the oscap tool:
./tests/probes/xinetd
./tests/probes/sysinfo
./tests/oscap_string
./tests/bz2
./tests/API/CVSS
./tests/API/XCCDF/parser
./tests/API/XCCDF/unittests
./tests/API/CCE
./tests/API/SEAP
./tests/API/crypt
./tests/API/probes
./tests/API/CPE/lang
./tests/API/CPE/name
./tests/API/CPE/dict
./tests/API/CVE
./tests/API/OVAL/schema_version
./tests/API/OVAL/glob_to_regex
./tests/API/OVAL
./tests/DS/ds_sds_index

I could turn these tests off in case the OSCAP variable is defined, does it make sense or is there a better alternative?

### DIFF
--- a/tests/API/XCCDF/parser/test_api_xccdf.sh
+++ b/tests/API/XCCDF/parser/test_api_xccdf.sh
@@ -49,10 +49,12 @@ function test_api_xccdf_validate {
 
 test_init "test_api_xccdf.log"
 
-test_run "export xccdf 1.1" test_api_xccdf_export xccdf11.xml
-test_run "validate xccdf 1.1" test_api_xccdf_validate xccdf11.xml "1.1"
-test_run "export xccdf 1.2" test_api_xccdf_export xccdf12.xml
-test_run "validate xccdf 1.2" test_api_xccdf_validate xccdf12.xml "1.2"
-test_run "export xccdf results 1.1" test_api_xccdf_export xccdf11-results.xml
+if [ -z ${CUSTOM_OSCAP+x} ] ; then
+    test_run "export xccdf 1.1" test_api_xccdf_export xccdf11.xml
+    test_run "validate xccdf 1.1" test_api_xccdf_validate xccdf11.xml "1.1"
+    test_run "export xccdf 1.2" test_api_xccdf_export xccdf12.xml
+    test_run "validate xccdf 1.2" test_api_xccdf_validate xccdf12.xml "1.2"
+    test_run "export xccdf results 1.1" test_api_xccdf_export xccdf11-results.xml
+fi
 
 test_exit
--- a/tests/API/XCCDF/unittests/all.sh
+++ b/tests/API/XCCDF/unittests/all.sh
@@ -8,15 +8,16 @@ test_init test_api_xccdf_unittests.log
 #
 # API C Tests
 #
-test_run "xccdf:complex-check -- NAND is working properly" ./test_xccdf_shall_pass $srcdir/test_xccdf_complex_check_nand.xccdf.xml
-test_run "xccdf:complex-check -- single negation" ./test_xccdf_shall_pass $srcdir/test_xccdf_complex_check_single_negate.xccdf.xml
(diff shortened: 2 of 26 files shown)


mpreisler:
We should document this somewhere but I am not sure how. Would it make sense to start `README.md` inside `tests/`?

> I could turn these tests off in case the OSCAP variable is defined, does it make sense or is there a better alternative?

I think this is our best alternative.

WesleyCeraso:
I tried to disable the checks in the Makefile.in but I couldn't, so I had to disable them in the shell script files. Additionaly I created the variable CUSTOM_OSCAP.

The checks that are bypassed when CUSTOM_OSCAP is set are the ones that tests other things not related to the oscap tool, therefore it makes no difference which oscap tool is being used.

tests/API/XCCDF/parser/test_api_xccdf.sh
```diff
@@ -49,10 +49,12 @@ function test_api_xccdf_validate {
...
-test_run "validate xccdf 1.1" test_api_xccdf_validate xccdf11.xml "1.1"
-test_run "export xccdf 1.2" test_api_xccdf_export xccdf12.xml
-test_run "validate xccdf 1.2" test_api_xccdf_validate xccdf12.xml "1.2"
-test_run "export xccdf results 1.1" test_api_xccdf_export xccdf11-results.xml
+if [ -z ${CUSTOM_OSCAP+x} ] ; then
+  test_run "export xccdf 1.1" test_api_xccdf_export xccdf11.xml
+  test_run "validate xccdf 1.1" test_api_xccdf_validate xccdf11.xml "1.1"
+  test_run "export xccdf 1.2" test_api_xccdf_export xccdf12.xml
+  test_run "validate xccdf 1.2" test_api_xccdf_validate xccdf12.xml "1.2"
+  test_run "export xccdf results 1.1" test_api_xccdf_export xccdf11-results.xml
```
mpreisler: indents...

tests/API/XCCDF/unittests/all.sh
```diff
@@ -8,15 +8,16 @@ test_init test_api_xccdf_unittests.log
...
-test_run "Assert for environment better" $OSCAP oval eval --id oval:moc.elpmaxe.www:def:1 $srcdir/test_xccdf_check_content_ref_without_name_attr.oval.xml
+if [ -z ${CUSTOM_OSCAP+x} ] ; then
+  test_run "xccdf:complex-check -- NAND is working properly" ./test_xccdf_shall_pass $srcdir/test_xccdf_complex_check_nand.xccdf.xml
+  test_run "xccdf:complex-check -- single negation" ./test_xccdf_shall_pass $srcdir/test_xccdf_complex_check_single_negate.xccdf.xml
+  test_run "Certain id's of xccdf_items may overlap" ./test_xccdf_shall_pass $srcdir/test_xccdf_overlaping_IDs.xccdf.xml
+  test_run "Test Abstract data types." ./test_oscap_common
+  test_run "xccdf_rule_result_override" $srcdir/test_xccdf_overrides.sh
 
+  test_run "Assert for environment" [ ! -x $srcdir/not_executable ]
+  test_run "Assert for environment better" $OSCAP oval eval --id oval:moc.elpmaxe.www:def:1 $srcdir/test_xccdf_check_content_ref_without_name_attr.oval.xml
```
mpreisler: indents

I will not comment other occurrences of wrong indents, they are in the whole patch

WesleyCeraso:
Indentation updated

jan-cerny:
I have tried this and it works as expected. The patch looks OK. Thank you. ACK.

Issue #817, opened 2017-08-30
./src/OVAL/probes/fsdev.c solaris specific code for __fsdev_init function issues


jlcharton:
Hi,

I think that the code is broken in several places.

line 246:
`struct mnttab *ment;`

But ment is never allocated before being used later on lines 271 and 284.
So maybe line 246 could be fixed like this:
`struct mnttab mentbuf, *ment = &mentbuf;`

lines 271 and 284:
`while ((getmntent(fp, ment)) != 0) {`

But according to solaris man pages for getmntent:

> RETURN VALUES
  getmntent() and getmntany()
     If the next entry is successfully read by getmntent()  or  a
     match  is  found with getmntany(), 0 is returned.  If an EOF
     is encountered on reading, these functions return -1. If  an
     error  is  encountered,  a value greater than 0 is returned.
     The following error values are defined in <sys/mnttab.h>:
     MNT_TOOLONG     A line in the  file  exceeded  the  internal
                     buffer size of MNT_LINE_MAX.
     MNT_TOOMANY     A line in the file contains too many fields.
     MNT_TOOFEW      A line in the file contains too few fields.

So getmntent() is supposed to return 0 on success and the test should probably be:
`while ((getmntent(fp, ment)) == 0) {`

Best Regards,

Jean-Louis Charton

mpreisler:
Hi @jlcharton

could you please submit these changes as a pull request?

matusmarhefka:
@mpreisler @jlcharton Can we close this issue? It looks like it was resolved by commit https://github.com/OpenSCAP/openscap/commit/78291468df36ea0e29755608e6bbd8dcd30f1537

jan-cerny:
Fixed in https://github.com/OpenSCAP/openscap/commit/78291468df36ea0e29755608e6bbd8dcd30f1537

PR #819, opened 2017-08-30
__fsdev_init() function fix suggestion for Solaris

jlcharton:
See issue #817.
BTW, the code sometimes properly checks malloc returned value and sometimes don't do it for realloc.
Any reason for this?

Regards,

Jean-Louis Charton

### DIFF
--- a/src/OVAL/probes/fsdev.c
+++ b/src/OVAL/probes/fsdev.c
@@ -243,7 +243,7 @@ static fsdev_t *__fsdev_init(fsdev_t * lfs, const char **fs, size_t fs_cnt)
 	FILE *fp;
 	size_t i;
 
-	struct mnttab *ment;
+	struct mnttab mentbuf, *ment = &mentbuf;
 	struct stat st;
 
 	fp = fopen(MNTTAB, "r");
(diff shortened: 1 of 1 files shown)


mpreisler:
@openscap-ci test this please

mpreisler:
> BTW, the code sometimes properly checks malloc returned value and sometimes don't do it for realloc.
> Any reason for this?

sounds like a bug

src/OVAL/probes/fsdev.c
```diff
@@ -243,7 +243,7 @@ static fsdev_t *__fsdev_init(fsdev_t * lfs, const char **fs, size_t fs_cnt)
 	FILE *fp;
 	size_t i;
 
-	struct mnttab *ment;
+	struct mnttab mentbuf, *ment = &mentbuf;
```
mpreisler: Can't we just allocate this on the stack and then use `&mentbuf` for the methods? this aliasing seems unnecessary.

jlcharton: Yes for sure.
I just wrote it this way to minimize the code that had to be changed.

mpreisler:
I will manually apply this to `maint-1.2` because it's a bugfix. I will also make it use a single variable instead of the alias.

mpreisler:
Applied as 78291468df36ea0e29755608e6bbd8dcd30f1537

PR #823, opened 2017-08-30
Update oval_fts.c

jlcharton:
path argument of valid_local_zone() function should be const char* since this function is called by OVAL_FTS_localp() which declared path as const char*.


jan-cerny:
@openscap-ci ok to test

jan-cerny:
Thanks.

PR #824, opened 2017-08-30
Update cpe_name.h

jlcharton:


mpreisler:
I am afraid this breaks API and ABI so we can't merge it to `maint-1.2`.

jlcharton:
However the prototypes are quite weird: why do they return a pointer to bool and not just a bool.

Moreover, this makes Oracle C compiler complain when compiliing ./src/CPE/cpelang.c since (for example):
> CC       libcpe_la-cpelang.lo
> "cpelang.c", line 174: warning: improper pointer/integer combination: op "="
> "cpelang.c", line 177: warning: improper pointer/integer combination: op "="


 This is because in function cpe_testexpr_evaluate one has:
`        case CPE_LANG_OPER_MATCH:`
`                ret = dict_cb(cpe_testexpr_get_meta_cpe(expr), usr);`
`                break;`
`        case CPE_LANG_OPER_CHECK:`
`                ret = check_cb(cpe_testexpr_get_meta_check_system(expr), cpe_testexpr_get_meta_check_href(expr), cpe_testexpr_get_meta_check_id(expr), usr);`
`                break;`

and ret is of type bool and dict_cb and check_cb from their prototypes are supposed to return a pointer to a bool.

mpreisler:
@openscap-ci add to whitelist

mpreisler:
@jlcharton yeah, the prototypes look wrong and this looks like the proper fix. We just can't remove the typedef from the API. We can't break ABI no matter what, it would make it hard for distributions to update to latest OpenSCAP.

We don't want to spoil https://abi-laboratory.pro/tracker/timeline/openscap/ :-)

mpreisler:
Are the typedefs used anywhere else in public API, can we just create 2 new typedefs and use them in the codebase? That would make it API compatible and also fix the issue.

jan-cerny:
@mpreisler Could we merge this to master branch instead?

jan-cerny:
What do you think about including this patch into master branch?

jan-cerny:
@mpreisler @WesleyCeraso @yuumasato @matejak @dahaic @matusmarhefka Any opinion will be strongly welcome.

matusmarhefka:
@jan-cerny IMO this would be okay to merge into the master branch instead of maint-1.2 branch.

jan-cerny:
@matusmarhefka Thank you for investigation, I agree with you that these functions should return bool. I suspect the code doesn't work properly now.

mpreisler:
We have consensus that this is a good patch but we can't do it in maint-1.2. Let's continue in #951 which has the same patch except for master branch.

PR #825, opened 2017-09-01
Fixed void* pointer arithmetic in spb.c

jlcharton:


jan-cerny:
This is awesome! :smiley:

PR #831, opened 2017-09-05
Update oscap-oval.c

jlcharton:
Fix the following warning raised by Oracle Developer Studio 12.5:
  CC       oscap-oscap-oval.o
"oscap-oval.c", line 361: warning: improper pointer/integer combination: arg #3.


mpreisler:
ACK.

Thanks!

PR #832, opened 2017-09-05
Update test_api_seap_spb.c

jlcharton:
Fix non standard void pointer arithmetic (GCC specific extension).


WesleyCeraso:
Awesome, thanks

PR #835, opened 2017-09-08
Fix for systems that lack POSIX.1-2008 realpath

jlcharton:
Also fix a couple of other minor issues (strdup returned value and error message).


jan-cerny:
ACK

PR #836, opened 2017-09-11
HTML report to report target as container or image, not as host

### LINKED ISSUE #305: HTML report says the target was host even if I scan VM or container
Seems to be related to offline scanning.

I will add more details later.

jan-cerny:
Fixes #305 
screenshot from 2017-09-11 13-46-38

### DIFF
--- a/src/XCCDF/result.c
+++ b/src/XCCDF/result.c
@@ -180,7 +180,7 @@ void xccdf_result_fill_sysinfo(struct xccdf_result *result)
 	_xccdf_result_clear_metadata(XITEM(result));
 
 	/* override target name by environment variable */
-	const char *target_hostname = getenv("OSCAP_PROBE_PRIMARY_HOST_NAME");
+	const char *target_hostname = getenv("OSCAP_EVALUATION_TARGET");
 	if (target_hostname == NULL) {
 		target_hostname = sname.nodename;
 	}
--- a/src/XCCDF/result.c
+++ b/src/XCCDF/result.c
@@ -178,8 +178,14 @@ void xccdf_result_fill_sysinfo(struct xccdf_result *result)
 		return;
 
 	_xccdf_result_clear_metadata(XITEM(result));
+
+	/* override target name by environment variable */
+	const char *target_hostname = getenv("OSCAP_PROBE_PRIMARY_HOST_NAME");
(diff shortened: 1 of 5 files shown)


jan-cerny:
A small idea:

The XSLT template could parse the target element contents. Based on that, it could produce 2 cells in the table:
1. Target type, which would be `machine` or `Docker image` or `Docker container` or `Virtual machine` or `Arbitrary mounted mount` depending on the target
2. Target name which would be machine's hostname or image name or image ID or container name or container ID or VM's name or the mounted filepath.
The different types could have different colors for example.

What do you think?

src/XCCDF/result.c
```diff
@@ -178,8 +178,14 @@ void xccdf_result_fill_sysinfo(struct xccdf_result *result)
 		return;
 
 	_xccdf_result_clear_metadata(XITEM(result));
+
+	/* override target name by environment variable */
+	const char *target_hostname = getenv("OSCAP_PROBE_PRIMARY_HOST_NAME");
```
mpreisler: The name of the environment variable is not very self-documenting. It's not related to OVAL probes and it is not primary.

jan-cerny: @mpreisler Yes, I agree, but the variable is already set in oscap-chroot in oscap-vm. Fortunately it's not read anywhere. Do you think I can rename it and rename all ocurences? Isn't it used somewhere outside of OpenSCAP?

mpreisler: I don't think it's used outside of OpenSCAP. I think we should rename it before people start using it.

jan-cerny:
Unfortunately this requires pythonX-docker package - @matusmarhefka  checked it and there is no such packages in RHEL7 extras repo (which provides docker and atomic packages). I need to find a way how to handle this problem. Please don't merge yet.

jan-cerny:
We could check if the class is available and 
 if not we could implement a fallback solution. The fallback solution could be similar to https://github.com/projectatomic/atomic/blob/master/Atomic/mount.py#L390 except creating a temporary container.

jan-cerny:
Using low-level API works for me on my RHEL7 virtual machine. This is ready to review.

mpreisler:
ACK

PR #837, opened 2017-09-11
Fix freeing non-heap memory

WesleyCeraso:
In the exceptional case when the function oscap_textlist_get_preferred_plaintext returns NULL, we'd try to free a non-heap memory.

Signed-off-by: Wesley Ceraso Prudencio <wcerasop@redhat.com>


mpreisler:
ACK

PR #838, opened 2017-09-11
Fix profile descriptions in generated roles

WesleyCeraso:
When there are newlines in profile description it now includes commentary marks in all lines

Signed-off-by: Wesley Ceraso Prudencio <wcerasop@redhat.com>


mpreisler:
Profile description will almost always be multi-line. Make a linebreak after the profile description label, then the fillter is just "# ".

```
...
# Profile description:
# bla
# bla
# bla
#
...
```

Makes it way easier to read.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -694,6 +694,31 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
 		const struct xccdf_version_info *xccdf_version = xccdf_benchmark_get_schema_version(benchmark);
 		const char *xccdf_version_name = xccdf_version_info_get_version(xccdf_version);
 
+		if (NULL != profile_description) {
+			unsigned int newLines = 0;
```
mpreisler: code style

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -694,6 +694,31 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
...
+			const char *c = profile_description;			
+			while (*c != '\0') {
```
mpreisler: for loop please

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -694,6 +694,31 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
...
+				if (*c == '\n')
+					++newLines;
+				++c;
+			}
+			
+			if (newLines > 0) {
+				unsigned int description_length = c - profile_description;
```
mpreisler: Isn't this the same as strlen(profile_description)? I know this is slightly faster but it's also unreadable. It can also be const and `size_t`.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -694,6 +694,31 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
...
+				const char filler[] = "#                       ";
```
mpreisler: why `filler[]` and not `filler*`?

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -694,6 +694,31 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
...
+				char *commented_description = malloc(description_length + newLines * (sizeof filler - 1));
+				for (unsigned int i = 0, j = 0; j < description_length; ++i, ++j) {
+					commented_description[i] = profile_description[j];
+					if (profile_description[j] == '\n') {
+						memcpy(&commented_description[i + 1], filler, (sizeof filler - 1));
```
mpreisler: This is undefined behavior. It sucks but that's what it is. I recommend looking into `memcpy` vs `memmove`.

mpreisler: you also probably want `strcpy` here

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -694,6 +694,31 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
...
+						i += (sizeof filler - 1);
```
mpreisler: brackets are unnecessary here

mpreisler:
I think we should guard against newlines in other fields. Titles for example can have newlines in them according to the spec. We almost never do it unless it's an accident. See https://github.com/OpenSCAP/scap-security-guide/pull/2215

src/XCCDF_POLICY/xccdf_policy_remediate.c
WesleyCeraso: So I can use sizeof
  const char *str1 = "Hello World";
  const char str2[] = "Hello World";
  std::cout << sizeof str1 << std::endl;
  std::cout << sizeof str2 << std::endl;
---------------
8
12

src/XCCDF_POLICY/xccdf_policy_remediate.c
WesleyCeraso: You mean the NULL coming first? I can change it, but it's a technique to avoid typing **var = NULL** instead **var == NULL**.

src/XCCDF_POLICY/xccdf_policy_remediate.c
WesleyCeraso: They are actually necessary, but I, hopefully, simplified the code and got rid of it.

src/XCCDF_POLICY/xccdf_policy_remediate.c
WesleyCeraso: I think memcpy is better than strncpy in this case as we know **exactly** how many bytes to copy. IMO, strncpy, is more suitable when you know how many bytes to copy **at most**.
Now, however, as I shortened the filler to only two characters "# ", I preferred to use a simpler for-loop that could be easily unrolled by the compiler.

src/XCCDF_POLICY/xccdf_policy_remediate.c
WesleyCeraso: strlen implies one more pass through the string and I tried to avoid that, however, I simplified the code and got rid of it.

src/XCCDF_POLICY/xccdf_policy_remediate.c
WesleyCeraso: Sure

WesleyCeraso:

mpreisler:
Camelcase newLines

WesleyCeraso:
@mpreisler code style fixed

mpreisler:
```
+ generate fix: generate header for bash script               FAIL: all.sh
```

WesleyCeraso:
@mpreisler The tests also don't support multiline titles or descriptions.

mpreisler:
@WesleyCeraso could you please commit fixes as separate commits next time? Would make it way easier for me to figure out what you changed.

mpreisler:
ACK

PR #839, opened 2017-09-12
process58: Do not collect SELinux labels when SELinux is not enabled

bachradsusi:
getpidcon() function returns a value from /proc/<pid>/attr/current but
this interface is used by other LSM's like AppArmor too so the value
could be SELinux unrelated. Therefore we should check if SELinux is
enabled before we try to collect a process label.


mpreisler:
ACK

PR #840, opened 2017-09-12
WCOREDUMP usage improvement in sch_pipe.c

jlcharton:
Fix for AIX and for other platforms where WCOREDUMP is not defined.
Follow Linux Programmer's Manual (i.e. man waitpid) for WCOREDUMP: "This macro should be employed only if WIFSIGNALED returned true".


jan-cerny:
@jlcharton Nice catch. You're right. Thank you for the patch.

PR #841, opened 2017-09-13
Added support for AIX in fsdev.c

jlcharton:
Also fix a couple of minor issues in __fsdev_init for Linux: potential FILE handle leak in case of malloc error - use endmntent instead of fclose.

Questions:
- Linux docs say that LSB deprecates the xyzmntent family of functions. So what should be used as a replacement?
- In match_fs function, why not just call bsearch?

### DIFF
--- a/src/OVAL/probes/fsdev.c
+++ b/src/OVAL/probes/fsdev.c
@@ -52,6 +52,13 @@
 # include <sys/param.h>
 # include <sys/ucred.h>
 # include <sys/mount.h>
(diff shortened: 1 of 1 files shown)


jan-cerny:
1. I don't know. I think that would be question to author of the docs. Anyway those functions are still present on latest Linux systems. I doubt they'll be removed.
2. I don't see any reason why not just call bsearch.

Your patch looks good to me. Thanks.

PR #842, opened 2017-09-13
Make sure formatting of profile descriptions is preserved when generating HTML reports/guides

WesleyCeraso:
When the '''white-space: pre-wrap''' option is used whitespace is preserved by the browser. "Text will wrap when necessary, and on line breaks.


mpreisler:
Could you provide a screenshot? You can just paste images into PR comments. IIRC we had profile descriptions as these small notes next to profile titles. I am wondering if this changes that. As descriptions get larger and formatted I think we should change how to display them completely. Have a special field for them and make it all work with Benchmark description.

WesleyCeraso:
@mpreisler Sure,

Profile without html, before:
image

After:
image

Profile with HTML, before:
image

After:
image

mpreisler:
This is definitely an upgrade but I think going forward we need a bigger overhaul of how we show descriptions of benchmarks and profiles.

ACK

PR #846, opened 2017-09-15
Add support for AIX and other legacy Unices

jlcharton:
Also fix a 32/64 bit issue in get_size function and clean the code for BSD.

### DIFF
--- a/src/OVAL/probes/unix/file.c
+++ b/src/OVAL/probes/unix/file.c
@@ -67,6 +67,7 @@
 #undef OS_SOLARIS
 #undef OS_SUNOS
 #undef OS_WINDOWS
+#undef OS_AIX
 
 #if defined(__FreeBSD__)
 # define OS_FREEBSD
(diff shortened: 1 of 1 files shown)


src/OVAL/probes/unix/file.c
```diff
@@ -80,6 +81,8 @@
 # endif
 #elif defined(_WIN32)
 # define OS_WINDOWS
+#elif defined(_AIX)
+# define OS_AIX
```
mpreisler: this define is never used, what is it for?

jlcharton: Yes I know and this is intended for at least the 2 following reasons:
I had to do it that way else the pre processor would fall there:
# error "Sorry, your OS isn't supported."
Moreover, please leave this #define OS_AIX because I have to add in a future patch some code specific to AIX to implement acl_trival function (or equivalent) for the has_extended_acl function.

jan-cerny: Aha, OS_WINDOWS is never used. I though it is used in master branch, but it isn't used there as well.

jan-cerny:
@jlcharton Thanks for your patch. I think we can resolve the problem of OS_WINDOWS variable later. In fact the OS_WINDOWS was there for a long time and AFAIK no issue was reported.

ACK

PR #852, opened 2017-09-15
Minor fix for AIX

jlcharton:
BTW, are all the includes really necessary?

### DIFF
--- a/src/OVAL/probes/SEAP/sch_pipe.c
+++ b/src/OVAL/probes/SEAP/sch_pipe.c
@@ -31,9 +31,6 @@
 #include <limits.h>
 #include <sys/stat.h>
 #include <sys/uio.h>
-#if !defined(_AIX)
-#include <sys/unistd.h>
-#endif
 #include <sys/wait.h>
 #include <sys/socket.h>
 #include <sys/select.h>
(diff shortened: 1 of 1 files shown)


src/OVAL/probes/SEAP/sch_pipe.c
```diff
@@ -31,7 +31,9 @@
 #include <limits.h>
 #include <sys/stat.h>
 #include <sys/uio.h>
+#if !defined(_AIX)
 #include <sys/unistd.h>
```
mpreisler: are you sure this is not available on AIX?

jlcharton: On my AIX, I've got this compiler error: 1506-296 (S) #include file <sys/unistd.h> not found.
But beyond AIX, are you sure that sys/uio.h, sys/unistd.h and even fcntl.h are really necessary?
On my CentOS, Solaris and AIX boxes, I can compile sch_pipe.c without these 3 includes.

mpreisler: In that case just remove the include, it could be redundant/forgotten.

jan-cerny:
@jlcharton I would prefer if those 2 commits are squashed into one commit and I would be happy if the commit message says what is removed and why.

jan-cerny:
@jlcharton I know that working with commits caused you problems in past, so I have done that manually. See https://github.com/OpenSCAP/openscap/commit/fdcc3324d02c96e3b1c6f1a3bb7d29d6a0355602
I recommmend https://git-scm.com/book/en/v2/, it's a good read.

jlcharton:
@jan-cerny: thank you :-)

PR #860, opened 2017-09-18
Update seap.h

jlcharton:
Avoid redefining EOPNOTSUPP and ECANCELED if they are already defined in <errno.h>.

### DIFF
--- a/src/OVAL/probes/SEAP/public/seap.h
+++ b/src/OVAL/probes/SEAP/public/seap.h
@@ -33,6 +33,7 @@ extern "C" {
 #endif
 
 #include <stdint.h>
+#include <errno.h>
 #include <sexp.h>
 #include <seap-types.h>
 #include <seap-message.h>


jan-cerny:
@jlcharton I think I wasn't necessary to create a separate pull request for each patch. Those patches are small, it's OK to open just a single pull request next time.

jan-cerny:
I'm not sure. I have found that those macros are defined at multiple places each time with different values. If we override it, will OpenSCAP still work?

jlcharton:
@jan-cerny for me there are only 2 files with code that define EOPNOTSUPP and ECANCELED. They are lib/errno.in.h and src/OVAL/probes/SEAP/public/seap.h.

In seap.h EOPNOTSUPP is set to 1001 and ECANCELED to 1002.

In lib.errno.in.h EOPNOTSUPP is set to 130 but only for WIN32 platforms.
ECANCELED is set to 105 for WIN32 platforms and to 2008 for other platforms that don't define it.
But in fact is this lib/errno.h file used elsewhere than in the lib directory? No code in the lib directory actually uses ECANCELED or EOPNOTSUPP.

So I think it should be safe to include <errno.h> at the beginnig of <seap.h>. After all, seap.h is supposed to define ECANCELED and EOPNOTSUPP only on systems that don't define them. Including <errno.h> just make sure the system official values (if they exist) are set before seap.h eventually try to set them if they are not defined.

BTW, here are the official values for AIX, Solaris and Linux
AIX:  ECANCELED = 117, EOPNOTSUPP = 64
Solaris: ECANCELED = 47, EOPNOTSUPP = 122
Linux: ECANCELED = 125, EOPNOTSUPP = 95.

jan-cerny:
Thanks for detailed explanation. Sounds reasonable.

PR #861, opened 2017-09-18
Fix AIX broken NAN issue

jlcharton:


jan-cerny:
Could we only redefine the wrong definition on AIX? Introducing new macro instead of the standard macro makes the code more difficult to understand.

jlcharton:
@jan-cerny : sorry please don't take my last commit (984ca14) into account. Actually, it doesn't work on AIX because then the compiler produces this error message "1506-232 (I) Divisor for modulus or division operator cannot be zero" at several location in the code such as for example:
```if (!cvss_metrics_is_valid(impact->base_metrics)) return NAN;```
or
```metrics->score = NAN;```

So my problem is that I can (and have to) use (0.0f/0.0f) as NAN in array initialization but nowhere else in the code where I need to use the "official" NAN definition from math.h (that is (*((float *)(&_SQNAN)))).
This is why I had to define the NAN_INIT macro for use in array initialization in my first patch and thus preserve the official NAN macro from math.h for use in the rest of the code.

I don't see any other simple way to fix the code. Any suggestion is welcome.

**Comment by jan-cerny** [comment]
@jlcharton OK, in this situation it looks like your first solution was the best. I'm sorry I wasn't aware of problem the redefinition may cause. Please remove 984ca14 from this pull request.

**Comment by jlcharton** [comment]
@jan-cerny sorry but I can't find anywhere in the github web interface how to remove 984ca14. Do you know how to remove this patch from the pull request and revert to 80269ab?

**Comment by jan-cerny** [comment]
@jlcharton I usually do that using interactive rebase and then push-force.
```
git rebase -i upstream/maint-1.2
<editor where you can remove the commit will be displayed>
git push -f <your_branch>
```

jan-cerny:
Ooops. I think you have  used the wrong branch. I'll merge your commit bfb255b75ce6f78984a5a4e5570563d4ad3ac71e manually

Issue #848, opened 2017-09-15
html report not nice on my test data


isimluk:
Please bear (🍺) with me, I didn't look too much in depth into this but

```
wget https://raw.githubusercontent.com/OpenSCAP/openscap/maint-1.2/tests/sce/test_sce_in_ds.xml
oscap xccdf eval --profile xccdf_org.open-scap.sce-community-content_profile_pci-dss --report /tmp/report2.html  test_sce_in_ds.xml
firefox /tmp/report2.html
```

the frontpage html looks like this:

![scrot](https://user-images.githubusercontent.com/6666052/30485569-1e1d64a4-9a2e-11e7-8136-8df67a7221ff.jpg)

Do we have a nicer example of SCE report?

Many thanks for the great tool 🥂  and take care! 🤶

jan-cerny:
The header of HTML report just displays what the input XCCDF file contains in `Benchmark/title`, `Benchmark/description` and `Benchmark/notice` elements. The file `test_sce_in_ds.xml` has those elements empty, that's why there is nothing in the HTML report.

The bug that I see is that it shouldn't show the label and the blue field if these elements are present but empty. I think that will be easy to fix.

@isimluk I remember you were asking to embed HTML output of an SCE script. What did you mean by that?

Right now we have only an option to include the SCE script stdout and stderr in the report. The option is `--check-engine-results`. They will be shown in rule details.

Nice example where the SCE script lists contents of root directory on stdout:
```
cd tests/sce
oscap xccdf eval --check-engine-results --report /tmp/report.html test_sce_in_report.xml
```
![screenshot from 2017-09-18 15-15-46](https://user-images.githubusercontent.com/9050916/30543740-7719b60e-9c84-11e7-8c14-effea1748b1d.png)

evgenyz:
@jan-cerny So, this particular bug was fixed in #864, or I'm missing some subtle details that weren't addressed?

isimluk:
Thx, looks great now.

PR #864, opened 2017-09-18
Don't reserve space for empty elements in report

jan-cerny:
If some elements are empty, we shouldn't try to populate them in the report,
because it looks very ugly.
Helps to fix issue #848.


mpreisler:
Could you please you attach a screenshot of #848 with this fix?

jan-cerny:
Yes, I could.

Before:
screenshot from 2017-09-18 16-47-03

After:
screenshot from 2017-09-18 16-45-33

mpreisler:
ACK, jenkins is temporarily dead.

PR #867, opened 2017-09-18
Fix CVSS_SUBMODULES array declaration

jlcharton:
IBM XLC/C++: 1506-131 (W) Explicit dimension specification or initializer required for an auto or static array.
GCC -pedantic: error: array size missing in CVSS_SUBMODULES

### DIFF
--- a/utils/oscap-cvss.c
+++ b/utils/oscap-cvss.c
@@ -40,7 +40,7 @@ static bool getopt_cvss(int argc, char **argv, struct oscap_action *action);
 static int app_cvss_score(const struct oscap_action *action);
 static int app_cvss_describe(const struct oscap_action *action);
 
-static struct oscap_module* CVSS_SUBMODULES[];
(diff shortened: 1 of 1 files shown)


jan-cerny:
@openscap-ci test this please

jan-cerny:
Nice catch. However `3` is a magical constant. It might not be obvious why the size of the array is 3. Could you define a named constant?

jlcharton:
@jan-cerny do you mean something like this:
```
#define CVSS_SUBMODULES_NUM 3
static struct oscap_module* CVSS_SUBMODULES[CVSS_SUBMODULES_NUM];

...

static struct oscap_module* CVSS_SUBMODULES[CVSS_SUBMODULES_NUM] = {
    &CVSS_SCORE_MODULE,
    &CVSS_DESCRIBE_MODULE,
    NULL
};
```

jan-cerny:
@jlcharton Yes, exactly.

jan-cerny:
ACK.

PR #868, opened 2017-09-18
Fix CPE_SUBMODULES array declaration

jlcharton:


jan-cerny:
This is again introducing a magic constant.

jan-cerny:
@jlcharton Great. Thanks.

PR #870, opened 2017-09-19
Extend 'oscap info'

matejak:
This PR aims to address #685 by parametrizing the `info` module of the `oscap` binary.

Issue #685, opened 2017-02-21
[RFE] add option to `oscap info` to print only list of profiles, also provide names of the profiles


dahaic:
This is probably request specific for testing, as common use case is to use some specific profile you know about.

It would be great to have option to print only easily parsable profile list, basically what is in "Profiles:".

I guess it would be also helpful to have way to print human readable names of the profiles in CLI. I would expect `oscap info` to do that.

dahaic:
+ I would appreciate to have option to select profiles of which xccdf are to be printed. And to print all xccdfs in the file.

matejak:
More than 80% of the issue has been implemented in #874 , so please check if it is OK. Further expansion of `oscap info` should be discussed from the big-picture point of view.

jan-cerny:
I think that they would be OK. But we usually close the Github issues after the patches are merged.

PR #872, opened 2017-09-20
Extend 'oscap info'

matejak:
This PR aims to address #685 by parametrizing the info module of the `oscap` binary.
This is follow-up of the #870 which was not created according to the contribution guideline.


yuumasato:
Sorry, for late heads up, I believe this should go to maint-1.2 branch.

mpreisler:
It looks like this patch has been based on `master` so it needs to be rebased.

matejak:
Rebase didn't work and we have decided with @yuumasato that charry-picking was more appropriate, so let's use the follow-up PR #874

PR #873, opened 2017-09-20
Add path specification to wildcard function arguments.

### LINKED ISSUE #871: build instructions to install schemas
When I build and install `openscap` according to build instructions, no files end up being installed in the `<prefix>/share/openscap/schemas/xccdf/...`. There is just an `sce` subdirectory of `schemas`. This is a problem as `scap-security-guide` requires more schemas.
The `openscap` [Fedora RPM package](http://pkgs.fedoraproject.org/cgit/rpms/openscap.git/tree/openscap.spec) contains those files. As I am a `spec` file newbie, so I don't see the reason right away, but ending up with those files installed seems to be a manner of documentation or build setup adjustment.

matejak:
Automake doesn't understand the wildcard function which is evaluated
at install-time. Therefore, one has to make manually sure that the wildcard
argument corresponds to the intended filesystem location even when the build
happens outside the source tree.

This fixes #871 . The 100% clean solution would be to [remove the wildcards](https://www.gnu.org/software/automake/manual/html_node/Wildcards.html), but I can imagine we don't want to go for it.

### DIFF
--- a/schemas/Makefile.am
+++ b/schemas/Makefile.am
@@ -24,37 +24,37 @@ cpe23dir = $(pkgdatadir)/schemas/cpe/2.3/
 cvedir = $(pkgdatadir)/schemas/cve/
 commondir = $(pkgdatadir)/schemas/common/
 
-oval53_DATA = $(wildcard oval/5.3/*.xsd oval/5.3/*.xsl)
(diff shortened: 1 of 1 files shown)


mpreisler:
ACK

PR #874, opened 2017-09-20
Enabled customization of the 'oscap info' datastream output format.

matejak:
This PR aims to address #685 by parametrizing the info module of the `oscap` binary.
This is follow-up of the #870 which was not created according to the contribution guideline and #872, which was created against `master`.

### DIFF
--- a/src/XCCDF/public/xccdf_session.h
+++ b/src/XCCDF/public/xccdf_session.h
@@ -337,7 +337,7 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
  *@param profile_suffix unique profile ID or suffix of the ID of the profile to set
  *@returns 0 on success, 1 if profile is not found, and 2 if multiple matches are found.
  */
-int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix);
+oscap_profile_match_t xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix);
 
 /**
  * Retrieves ID of the profile that we will evaluate with, or NULL.
--- a/src/XCCDF/public/xccdf_session.h
+++ b/src/XCCDF/public/xccdf_session.h
@@ -337,7 +337,7 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
  *@param profile_suffix unique profile ID or suffix of the ID of the profile to set
  *@returns 0 on success, 1 if profile is not found, and 2 if multiple matches are found.
  */
-oscap_profile_match_t xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix);
+int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix);
 
 /**
  * Retrieves ID of the profile that we will evaluate with, or NULL.
--- a/utils/oscap-info.c
+++ b/utils/oscap-info.c
@@ -60,7 +60,9 @@ struct oscap_module OSCAP_INFO_MODULE = {
     .help = "Print information about a file\n"
     "\n"
     "Options:\n"
-    "   --fetch-remote-resources \r\t\t\t\t - Download remote content referenced by DataStream.\n",
+    "   --fetch-remote-resources \r\t\t\t\t - Download remote content referenced by DataStream.\n"
+    "   --profile <id>\r\t\t\t\t - Show profile info when displaying a DataStream.\n"
+    "   --profiles\r\t\t\t\t - Only show profiles when displaying a DataStream. Outputs them in a <id>:<title> format, one profile per line.\n",
     .opt_parser = getopt_info,
     .func = app_info
 };
@@ -89,11 +91,8 @@ static inline void _print_xccdf_status(struct xccdf_status *status, const char *
 	}
 }
 
-static inline void _print_xccdf_profiles(struct xccdf_profile_iterator *prof_it, const char *prefix)
+static void _print_xccdf_profile_default(const struct xccdf_profile *prof, const char *prefix)
 {
-	printf("%sProfiles:\n", prefix);
-	while (xccdf_profile_iterator_has_more(prof_it)) {
-		struct xccdf_profile * prof = xccdf_profile_iterator_next(prof_it);
 		struct oscap_text_iterator *title_it = xccdf_profile_get_title(prof);
 		char *profile_title = oscap_textlist_get_preferred_plaintext(title_it, NULL);
 		oscap_text_iterator_free(title_it);
@@ -102,6 +101,28 @@ static inline void _print_xccdf_profiles(struct xccdf_profile_iterator *prof_it,
 		printf("%s\t\tId: %s%s\n", prefix,
 			xccdf_profile_get_id(prof),
 			xccdf_profile_get_abstract(prof) ? " (abstract)" : "");
+}
+
+static void _print_xccdf_profile_terse(const struct xccdf_profile *prof, const char *prefix)
+{
+		struct oscap_text_iterator *title_it = xccdf_profile_get_title(prof);
+		char *profile_title = oscap_textlist_get_preferred_plaintext(title_it, NULL);
+		oscap_text_iterator_free(title_it);
+		printf("%s:%s\n",
+			xccdf_profile_get_id(prof),
+			profile_title);
+		free(profile_title);
+}
+
+static inline void _print_xccdf_profiles(struct xccdf_profile_iterator *prof_it, const char *prefix, void (*print_one_profile)(const struct xccdf_profile *, const char *))
+{
+	if (prefix)
+		printf("%sProfiles:\n", prefix);
+	if (print_one_profile == 0)
+		print_one_profile = & _print_xccdf_profile_default;
+	while (xccdf_profile_iterator_has_more(prof_it)) {
+		struct xccdf_profile * prof = xccdf_profile_iterator_next(prof_it);
+		print_one_profile(prof, prefix);
 	}
 	xccdf_profile_iterator_free(prof_it);
 }
@@ -160,11 +181,11 @@ static inline void _print_xccdf_testresults(struct xccdf_benchmark *bench, const
 	xccdf_result_iterator_free(res_it);
 }
 
-static inline void _print_xccdf_benchmark(struct xccdf_benchmark *bench, const char *prefix)
+static inline void _print_xccdf_benchmark(struct xccdf_benchmark *bench, const char *prefix, void (*print_one_profile)(const struct xccdf_profile *, const char *))
 {
 	_print_xccdf_status(xccdf_benchmark_get_status_current(bench), prefix);
 	printf("%sResolved: %s\n", prefix, xccdf_benchmark_get_resolved(bench) ? "true" : "false");
-	_print_xccdf_profiles(xccdf_benchmark_get_profiles(bench), prefix);
+	_print_xccdf_profiles(xccdf_benchmark_get_profiles(bench), prefix, print_one_profile);
 
 	struct xccdf_policy_model *policy_model = xccdf_policy_model_new(bench);
 	_print_xccdf_referenced_files(policy_model, prefix);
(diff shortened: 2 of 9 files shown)


utils/oscap-info.c
```diff
@@ -102,6 +101,28 @@ static inline void _print_xccdf_profiles(struct xccdf_profile_iterator *prof_it,
...
+
+static void _print_xccdf_profile_terse(const struct xccdf_profile *prof, const char *prefix)
+{
+		struct oscap_text_iterator *title_it = xccdf_profile_get_title(prof);
+		char *profile_title = oscap_textlist_get_preferred_plaintext(title_it, NULL);
+		oscap_text_iterator_free(title_it);
+		printf("%s:%s\n",
+			xccdf_profile_get_id(prof),
+			profile_title);
+		free(profile_title);
```
mpreisler: 2 tabs for indents instead of 1

utils/oscap-info.c
```diff
@@ -174,17 +195,156 @@ static inline void _print_xccdf_benchmark(struct xccdf_benchmark *bench, const c
...
+	return 0;
+}
+
+static int app_info_single_ds_all(struct ds_stream_index_iterator* sds_it, struct ds_sds_session *session, const struct oscap_action *action)
+{
+	const char *prefix = "\t\t";
+	struct ds_stream_index * stream = ds_stream_index_iterator_next(sds_it);
+	struct oscap_string_iterator* checklist_it = ds_stream_index_get_checklists(stream);
+
+	if (action->profile != 0)
```
mpreisler: I know it's technically the same thing but `NULL` would be more readable here.

utils/oscap-info.c
```diff
@@ -102,6 +101,28 @@ static inline void _print_xccdf_profiles(struct xccdf_profile_iterator *prof_it,
...
+			xccdf_profile_get_id(prof),
+			profile_title);
+		free(profile_title);
+}
+
+static inline void _print_xccdf_profiles(struct xccdf_profile_iterator *prof_it, const char *prefix, void (*print_one_profile)(const struct xccdf_profile *, const char *))
+{
+	if (prefix)
+		printf("%sProfiles:\n", prefix);
+	if (print_one_profile == 0)
```
mpreisler: NULL

utils/oscap-info.c
```diff
@@ -102,6 +101,28 @@ static inline void _print_xccdf_profiles(struct xccdf_profile_iterator *prof_it,
...
+		print_one_profile = & _print_xccdf_profile_default;
```
mpreisler: unnecessary space after `&`

utils/oscap-info.c
```diff
@@ -102,6 +101,28 @@ static inline void _print_xccdf_profiles(struct xccdf_profile_iterator *prof_it,
...
+	while (xccdf_profile_iterator_has_more(prof_it)) {
+		struct xccdf_profile * prof = xccdf_profile_iterator_next(prof_it);
```
mpreisler: coding style: space after `*`

matejak:
The `oscap info --profile <profile id> <datastream file>` functionality is now also supported.

jan-cerny:
I have found a minor issue with the new list of profiles: If the name of a profile contains end-of-line character, the format of output is broken, beacuse it creates a line that is not in format `<id>:<title>`. 
I think we'll need to check and repleace `\n` characters before we print it out.

Try: 
```
[jcerny@thinkpad openscap{info-improvement}]$ myoscap info --profiles /usr/share/xml/scap/ssg/content/ssg-rhel6-ds.xml                                                                                               
xccdf_org.ssgproject.content_profile_standard:Standard System Security Profile                            
xccdf_org.ssgproject.content_profile_CS2:Example Server Profile                                           
xccdf_org.ssgproject.content_profile_common:Common Profile for General-Purpose Systems                    
xccdf_org.ssgproject.content_profile_desktop:Desktop Baseline                                             
xccdf_org.ssgproject.content_profile_server:Server Baseline                                               
xccdf_org.ssgproject.content_profile_ftp-server:FTP Server Profile (vsftpd)                               
xccdf_org.ssgproject.content_profile_stig-rhel6-workstation-upstream:Upstream STIG for Red Hat Enterprise Linux 6 Workstation                                                                                        
xccdf_org.ssgproject.content_profile_stig-rhel6-server-gui-upstream:Upstream STIG for Red Hat Enterprise Linux 6 Server Running GUIs                                                                                 
xccdf_org.ssgproject.content_profile_stig-rhel6-server-upstream:Upstream STIG for Red Hat Enterprise Linux 6 Server                                                                                                  
xccdf_org.ssgproject.content_profile_usgcb-rhel6-server:United States Government Configuration Baseline (USGCB)                                                                                                      
xccdf_org.ssgproject.content_profile_rht-ccp:Red Hat Corporate Profile for Certified Cloud Providers (RH CCP)                                                                                                        
xccdf_org.ssgproject.content_profile_CSCF-RHEL6-MLS:CSCF RHEL6 MLS Core Baseline                          
xccdf_org.ssgproject.content_profile_C2S:C2S for Red Hat Enterprise Linux 6                               
xccdf_org.ssgproject.content_profile_pci-dss:PCI-DSS v3 Control Baseline for Red Hat Enterprise Linux 6   
xccdf_org.ssgproject.content_profile_nist-cl-il-al:CNSSI 1253 Low/Low/Low Control Baseline for Red Hat    
Enterprise Linux 6                                   
xccdf_org.ssgproject.content_profile_fisma-medium-rhel6-server:FISMA Medium for Red Hat Enterprise Linux 6                                                                                                           
xccdf_org.ssgproject.content_profile_pci-dss_centric:PCI-DSS v3 Control Baseline for Red Hat Enterprise Linux 6
```
Notice third line from the end.

jan-cerny:
I have some ideas to improve:

* If user specifies wrong profile ID, it could tell him. Eg.: `oscap info --profile skowkdlqjdb /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml` could print an error message that no such profile exists.

* In some other modules, we support short profile IDs. That means that user doesn't have to specify the long profile ID, eg. `--profile xccdf_org.ssgproject.content_profile_pci-dss`, but instead, he can type just the last part, eg. `--profile pci-dss`. This works eg. in `oscap xccdf eval`. It tries to find longest suffix match. I think it would be nice to support those shortened IDs here as well.

* The new options could work not only with source datastreams, but also with XCCDF files. I think it would be done in very similar way.

matejak:
So the issue with newlines is fixed and the command works on XCCDF files too.
The idea with invalid profile IDs - I think that it would be nice to have, but the added value is very low (one can see easily that there is no profile of such ID) and a clean implementation of this in C is not trivial.
"Intelligent" matching of IDs has a good value, but I see the C implementation cost as relatively high.

If we want those issues, I propose to make another issues for them.

utils/oscap-info.c
```diff
@@ -89,21 +91,103 @@ static inline void _print_xccdf_status(struct xccdf_status *status, const char *
...
-		struct oscap_text_iterator *title_it = xccdf_profile_get_title(prof);
-		char *profile_title = oscap_textlist_get_preferred_plaintext(title_it, NULL);
-		oscap_text_iterator_free(title_it);
-		printf("%s\tTitle: %s\n", prefix, profile_title);
-		free(profile_title);
-		printf("%s\t\tId: %s%s\n", prefix,
-			xccdf_profile_get_id(prof),
-			xccdf_profile_get_abstract(prof) ? " (abstract)" : "");
+		struct xccdf_profile *prof = xccdf_profile_iterator_next(prof_it);
+		if (strncmp(xccdf_profile_get_id(prof), profile_id, 1024) == 0)
```
jan-cerny: Why is this 1024? Shouldn't it be `strlen(profile_id)` instead?

utils/oscap-info.c
```diff
@@ -174,17 +261,254 @@ static inline void _print_xccdf_benchmark(struct xccdf_benchmark *bench, const c
...
+	struct ds_stream_index * stream = ds_stream_index_iterator_next(sds_it);
+	struct oscap_string_iterator* checklist_it = ds_stream_index_get_checklists(stream);
+
+	printf("\nStream: %s\n", ds_stream_index_get_id(stream));
+	printf("Generated: %s\n", ds_stream_index_get_timestamp(stream));
+	printf("Version: %s\n", ds_stream_index_get_version(stream));
+
+	while (oscap_string_iterator_has_more(checklist_it)) {
+		const char * id = oscap_string_iterator_next(checklist_it);
+		// printf("\tRef-Id: %s\n", id);
```
jan-cerny: Please remove it

utils/oscap-info.c
```diff
@@ -174,17 +261,254 @@ static inline void _print_xccdf_benchmark(struct xccdf_benchmark *bench, const c
...
+
+static void app_info_single_benchmark(struct xccdf_benchmark *bench, const struct oscap_action *action, struct oscap_source *source)
+{
+	if (action->show_profiles_only)
+		_print_single_benchmark_profiles_only(bench);
+	else
+	if (action->profile)
+		_print_single_benchmark_one_profile(bench, action->profile);
+	else
+	{
```
jan-cerny: Please improve the coding style

utils/oscap-info.c
```diff
@@ -174,17 +261,254 @@ static inline void _print_xccdf_benchmark(struct xccdf_benchmark *bench, const c
...
+{
+	int return_value;
+	if (action->show_profiles_only)
+		return_value = app_info_single_ds_profiles_only(sds_it, session, action);
+	else
+	if (action->profile)
+		return_value = app_info_single_ds_one_profile(sds_it, session, action->profile);
+	else
+		return_value = app_info_single_ds_all(sds_it, session, action);
+	return return_value;
```
jan-cerny: Please improve the coding style

utils/oscap-info.c
```diff
@@ -60,7 +60,9 @@ struct oscap_module OSCAP_INFO_MODULE = {
     .help = "Print information about a file\n"
     "\n"
     "Options:\n"
-    "   --fetch-remote-resources \r\t\t\t\t - Download remote content referenced by DataStream.\n",
+    "   --fetch-remote-resources \r\t\t\t\t - Download remote content referenced by DataStream.\n"
+    "   --profile <id>\r\t\t\t\t - Show profile info when displaying a DataStream.\n"
+    "   --profiles\r\t\t\t\t - Only show profiles when displaying a DataStream. Outputs them in a <id>:<title> format, one profile per line.\n",
```
jan-cerny: It would be nice to document this in man page. See `utils/oscap.8`

utils/oscap-info.c
```diff
@@ -89,21 +91,103 @@ static inline void _print_xccdf_status(struct xccdf_status *status, const char *
...
+
+	text_it = xccdf_profile_get_description(prof);
+	char *profile_description = oscap_textlist_get_preferred_plaintext(text_it, NULL);
+	oscap_text_iterator_free(text_it);
+	_remove_occurence_of_character_from_string(profile_description, '\n');
+	printf("%s\tDescription: %s\n", prefix, profile_description);
+	free(profile_description);
+
+	str_it = xccdf_profile_get_platforms(prof);
+	printf("%s\tPlatforms: ", prefix);
```
jan-cerny: When I try `oscap info --profile` on  SCAP Security Guide Datastreams, I always have this empty. Is it expected to be empty?

utils/oscap-info.c
```diff
@@ -89,21 +91,103 @@ static inline void _print_xccdf_status(struct xccdf_status *status, const char *
...
+	free(profile_title);
+
+	printf("%s\tId: %s%s\n", prefix,
+		xccdf_profile_get_id(prof),
+		xccdf_profile_get_abstract(prof) ? " (abstract)" : "");
+
+	text_it = xccdf_profile_get_description(prof);
+	char *profile_description = oscap_textlist_get_preferred_plaintext(text_it, NULL);
+	oscap_text_iterator_free(text_it);
+	_remove_occurence_of_character_from_string(profile_description, '\n');
```
jan-cerny: Personally, I wouldn't remove '\n' characters from profile description, because it sometimes leads to merging words of description. See this example:

```
[jcerny@thinkpad openscap{feature/info-improvement}]$ myoscap info --profile xccdf_org.ssgproject.content_profile_nist-800-171-cui  /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml 
Document type: Source Data Stream
Imported: 2017-08-29T22:29:14

Stream: scap_org.open-scap_datastream_from_xccdf_ssg-rhel7-xccdf-1.2.xml
Generated: (null)
Version: 1.2
Profile
        Title: Unclassified Information in Non-federal Information Systems and Organizations (NIST 800-171)
        Id: xccdf_org.ssgproject.content_profile_nist-800-171-cui
        Description: From NIST 800-171, Section 2.2:Security requirements for protecting the confidentiality of CUI in nonfederal information systems and organizations have a well-defined structure that consists of:(i) a basic security requirements section;(ii) a derived security requirements section.The basic security requirements are obtained from FIPS Publication 200, whichprovides the high-level and fundamental security requirements for federalinformation and information systems. The derived security requirements, whichsupplement the basic security requirements, are taken from the security controlsin NIST Special Publication 800-53.This profile configures Red Hat Enterprise Linux 7 to the NIST SpecialPublication 800-53 controls identified for securing Controlled UnclassifiedInformation (CUI).
        Platforms: 
[jcerny@thinkpad openscap{feature/info-improvement}]$ 
```
Notice  federalinformation, UnclassifiedInformation, SpecialPublication, whichsupplement , etc

utils/oscap-info.c
```diff
@@ -174,17 +261,254 @@ static inline void _print_xccdf_benchmark(struct xccdf_benchmark *bench, const c
...
+		struct oscap_source *xccdf_source = ds_sds_session_select_checklist(session, ds_stream_index_get_id(stream), id, NULL);
+		if (xccdf_source == NULL) {
+			oscap_string_iterator_free(checklist_it);
+			ds_stream_index_iterator_free(sds_it);
+			ds_sds_session_free(session);
+			return OSCAP_ERROR;
+		}
+
+		if (oscap_source_get_scap_type(xccdf_source) == OSCAP_DOCUMENT_XCCDF) {
+			struct xccdf_benchmark *bench = xccdf_benchmark_import_source(xccdf_source);
```
jan-cerny: I think you should free `bench`.

jan-cerny:
```
==9697== 25,510,363 (16 direct, 25,510,347 indirect) bytes in 1 blocks are definitely lost in loss record 657 of 657                                                                                                 
==9697==    at 0x4C2EB6B: malloc (vg_replace_malloc.c:299)                                                
==9697==    by 0x4E8906C: oscap_list_add (list.c:55)                                                      
==9697==    by 0x4EEBA29: xccdf_group_parse (rule.c:228)                                                  
==9697==    by 0x4EEB951: xccdf_content_parse (rule.c:64)                                                 
==9697==    by 0x4EDC40A: xccdf_benchmark_parse (benchmark.c:208)                                         
==9697==    by 0x4EDC56C: xccdf_benchmark_import_source (benchmark.c:69)                                  
==9697==    by 0x40E371: app_info_single_ds_one_profile (oscap-info.c:362)                                
==9697==    by 0x40E371: app_info_single_ds (oscap-info.c:469)                                            
==9697==    by 0x40E371: app_info_sds (oscap-info.c:504)                                                  
==9697==    by 0x40E371: app_info (oscap-info.c:615)                                                      
==9697==    by 0x408857: oscap_module_call (oscap-tool.c:261)                                             
==9697==    by 0x408857: oscap_module_process (oscap-tool.c:346)                                          
==9697==    by 0x40783E: main (oscap.c:80)   
```

matejak: That's interesting. There was no free before, and if I free the benchmark, I get the segmentation error.
utils/oscap-info.c:261 has a message - `xccdf_benchmark_free not needed, it si already freed by the policy!`, so it is probably better not to free it.

utils/oscap-info.c
matejak: I don't know whether to trust the arguments to be 0-terminated. The code above is safe and will work correctly all right if profile IDs have less than 1024 characters. I am happy to learn a better way how to handle the situation; however if those strings are safe, than I would use `strcmp` without worrying about length.

utils/oscap-info.c
matejak: This is a good point, maybe it is better to substitute newline with space instead of trashing it?

yuumasato:
> The idea with invalid profile IDs - I think that it would be nice to have, but the added value is very low (one can see easily that there is no profile of such ID) and a clean implementation of this in C is not trivial.
"Intelligent" matching of IDs has a good value, but I see the C implementation cost as relatively high.

I find it confusing that no error message is returned when passing invalid ID.
Besides, other commands show error when Profile ID is not found.
```
$ oscap xccdf eval --profile other-stig-profile /usr/share/xml/scap/ssg/content/ssg-rhel6-ds.xml
WARNING: This content points out to the remote resources. Use `--fetch-remote-resources' option to download them.
WARNING: Skipping https://www.redhat.com/security/data/oval/Red_Hat_Enterprise_Linux_6.xml file which is referenced from XCCDF content
No profile matching suffix "other-stig-profile" was found. Get available profiles using:
$ oscap info "/usr/share/xml/scap/ssg/content/ssg-rhel6-ds.xml"
```

It would be nice to keep consistent user experience across oscap sub commands.

matejak:
Cool, I see now that the profile suffix matching is already part of the openscap lib, so I will reuse it in the info module.

utils/oscap-info.c
jan-cerny: you need to use `xccdf_benchmark_free(bench)` instead of just `free(bench)`.

utils/oscap-info.c
jan-cerny: Maybe it is better to not use the _remove_occurence_from function on profile description at all instead of modyfying it? We don't need it so strongly for profile descriptions. Newline characters in profile description will not break @dahaic

utils/oscap-info.c
jan-cerny: They're safe, you don't have to handle this

utils/oscap-info.c
matejak: It seems that if the platform is empty, it applies to all platforms (which is usually the case). After consulting @dahaic , we have reached conclusion that we don't need platform and description is OK.

utils/oscap-info.c
matejak: Update: There are occasions when it doesn't work - even `xccdf_benchmark_free(bench)` can cause segmentation errors. See `utils/oscap-info.c:614`.

matejak:
So the code reuse is not so easy, because:
* There is no `ds_sds_session_set_profile_id_by_suffix` (as opposed to `xccdf_session_set_profile_id_by_suffix`).
* Loading an XCCDF session results in a strange (but harmless) error at runtime (see comments at `utils/oscap-info.c:471`).
* The `xccdf_benchmark_get_profile_by_id` function from `src/XCCDF/benchmark.c` doesn't seem to be present neither in oscap or the library.

Therefore, I would like to ask for this assistance:
* Am I using the session data structure correctly? If so, how to cope with the absence of ds_sds_session_set_profile_id_by_suffix`?
* What is that `lt-probe_system_info` error about?

utils/oscap-info.c
jan-cerny: @matejak I'm afraid we can't accept code with huge memory leaks or segmentation faults. This must be fixed properly. I think that `valgrind` tool should help us to fix the memory issues. I'll try to look into the problem.

utils/oscap-info.c
```diff
@@ -49,6 +49,27 @@
 
 #include "oscap-tool.h"
 
+// COPIED from src/XCCDF/benchmark.c
```
jan-cerny: Please don't copy code around.

utils/oscap-info.c
```diff
@@ -49,6 +49,27 @@
...
+struct xccdf_profile * xccdf_benchmark_get_profile_by_id(struct xccdf_benchmark *benchmark, const char *profile_id)
+{
+	struct xccdf_profile_iterator *profit = xccdf_benchmark_get_profiles(benchmark);
+	while (xccdf_profile_iterator_has_more(profit)) {
+		struct xccdf_profile *profile = xccdf_profile_iterator_next(profit);
+		if (profile == NULL) {
+			assert(profile != NULL);
+			continue;
+		}
+		if (strcmp(xccdf_profile_get_id(profile), profile_id)) {
```
jan-cerny: This is wrong. It should be `strcmp(...) == 0`. The original version in `src/XCCDF/benchmark.c` used to have `oscap_streq` here.  Due to this the function always returns the first profile from benchmark regardless of the ID :-) :-)

utils/oscap-info.c
```diff
@@ -49,6 +49,27 @@
 
 #include "oscap-tool.h"
 
+// COPIED from src/XCCDF/benchmark.c
+struct xccdf_profile * xccdf_benchmark_get_profile_by_id(struct xccdf_benchmark *benchmark, const char *profile_id)
```
jan-cerny: Please don't copy code around. That is considered a bad practice. 

Perhaps you had decide to copy because had troubles with including and compiling. That's because `xccdf_benchmark_get_profile_by_id`  isn't part of public API and therefore it's hidden for `utils`. You can make the original function public, by moving its declaration into the public API. I think you can move the declaration of this function from `src/XCCDF/item.h` to `src/XCCDF/public/xccdf_benchmark.h`. and  include public/xccdf_benchmark.h here (if its not already included here).

utils/oscap-info.c
```diff
@@ -469,14 +739,23 @@ bool getopt_info(int argc, char **argv, struct oscap_action *action)
...
-	while ((c = getopt_long(argc, argv, "o:i:", long_options, NULL)) != -1) {
+	while ((c = getopt_long(argc, argv, "o:i:p:", long_options, NULL)) != -1) {
 		switch(c) {
 			case 0: break;
+			case 'p':
+				action->profile = optarg;
+				break;
+			case 'n':
+				action->show_profiles_only = 1;
+				action->provide_machine_readable_output = 1;
```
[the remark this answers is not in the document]
jan-cerny: Yeah, it looks to me that provide_machine_readable_output is an overabundant variable.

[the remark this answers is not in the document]
matejak: I understand your comments, but check out my point of view:
The `oscap` app has two components: A UI that communicates with the user, and its inner logic. Right now, when the user says that it wants to see only profile names, it is clear to the person that the output will be machine-readable. The reason for this is that UI documentation says so (and it is a use case, users expect something like that).
However, this dependence between listing profiles and keeping all aspects of the output machine-readable makes no sense from the application logic standpoint. Using the `provide_machine_readable_output` variable makes the code more understandable and it also allows easier implementation of other machine-readable outputs.

utils/oscap-info.c
```diff
@@ -174,17 +281,255 @@ static inline void _print_xccdf_benchmark(struct xccdf_benchmark *bench, const c
...
+	return return_value;
+}
+
+static void _print_sds_header(const struct oscap_action *action)
+{
+	if (! action->provide_machine_readable_output) {
+		printf("Document type: Source Data Stream\n");
+		print_time(action->file);
+	}
+}
```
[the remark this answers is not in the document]
matejak: Although I like [Robert C. Martin's approach](https://sites.google.com/site/unclebobconsultingllc/one-thing-extract-till-you-drop), I agree that this extraction did not fit.

matejak:
All requests that have been raised to date have been addressed, so the PR is ready for another review.

utils/oscap-info.c
```diff
@@ -174,17 +261,263 @@ static inline void _print_xccdf_benchmark(struct xccdf_benchmark *bench, const c
...
+
+static void app_info_single_benchmark(struct xccdf_benchmark *bench, const struct oscap_action *action, struct oscap_source *source)
+{
+	if (action->show_profiles_only) {
+		_print_single_benchmark_profiles_only(bench);
+		xccdf_benchmark_free(bench);
+	} else if (action->profile) {
+		const char *profile_id = benchmark_get_profile_or_report_bad_id(bench, action->profile, action->file);
+		if (profile_id != NULL) {
```
jan-cerny: Why the suffix match isn't used for datastreams as well? With XCCDF it's nice, but not so much useful. The benefit of suffix match is when the ID is long, which happens in datastreams.

The behavior of "oscap info --profile ID DS" should be same as "oscap info --profile ID XCCDF". Please use this function with datastreams as well.

Compare:
```
[jcerny@thinkpad openscap{feature/info-improvement}]$ myoscap info --profile sss /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml                                                                                   
Document type: Source Data Stream                    
Imported: 2017-08-29T22:29:14                        

Stream: scap_org.open-scap_datastream_from_xccdf_ssg-rhel7-xccdf-1.2.xml                                  
Generated: (null)         
Version: 1.2              
[jcerny@thinkpad openscap{feature/info-improvement}]$ myoscap info --profile sss /usr/share/xml/scap/ssg/content/ssg-rhel7-xccdf.xml                                                                                
Document type: XCCDF Checklist                       
No profile matching suffix "sss" was found. Get available profiles using:                                 
$ oscap info "/usr/share/xml/scap/ssg/content/ssg-rhel7-xccdf.xml"                                        
[jcerny@thinkpad openscap{feature/info-improvement}]$ 
```
The first command output is inappropriate, it should have look as the second.

utils/oscap-info.c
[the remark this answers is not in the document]
jan-cerny: @matejak OK, thaks for explanation, sounds reasnobale.

utils/oscap-info.c
matejak: oops, I forgot about that.

src/XCCDF/public/xccdf_session.h
```diff
@@ -337,7 +337,7 @@ bool xccdf_session_set_profile_id(struct xccdf_session *session, const char *pro
  *@param profile_suffix unique profile ID or suffix of the ID of the profile to set
  *@returns 0 on success, 1 if profile is not found, and 2 if multiple matches are found.
  */
-int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix);
+oscap_profile_match_t xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const char *profile_suffix);
```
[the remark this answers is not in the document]
matejak: OK, I go for anonymous enums.

matejak:
Turned out that some requests were not implemented (thanks @jan-cerny for noticing), so hopefully everything is OK now. The PR somehow drifted from the original issue, so it is not trivial to keep track of todos.
One thing - I am not 100% sure where to put declarations, so there probably are some declarations that should be relocated.

jan-cerny:
@matejak Thank you very much. I have reviewed this again. I tested it a little. Everything works ok for me. Great contribution!

PR #845, opened 2017-09-14
Update oscap-xccdf.c

jlcharton:
Make it more ISO C compliant:
IBM XLC/C++: 1506-131 (W) Explicit dimension specification or initializer required for an auto or static array.
GCC -pedantic: error: array size missing in XCCDF_SUBMODULES

PR #877, opened 2017-09-22
Fix several arrays declaration

jlcharton:
Get rid of warnings about missing array size reported by at least XLC and GCC compilers.
Also remove the use of magic numbers in a previous fix in ./utils/oscap-xccdf.c (#845).


matusmarhefka:
@openscap-ci test this please.

PR #882, opened 2017-09-26
Coverity fixes 1

jan-cerny:
This pull request fixes some of new issues reported by Coverity. See commit messages.

### DIFF
--- a/src/CVSS/cvss.c
+++ b/src/CVSS/cvss.c
@@ -80,7 +80,7 @@ static struct cvss_metrics **cvss_impact_metricsptr(struct cvss_impact* impact, 
     }
 }
 
-struct cvss_impact *cvss_impact_new(void) { return calloc(1, sizeof(struct cvss_metrics)); }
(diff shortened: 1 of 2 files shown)


src/CVSS/cvss.c
```diff
@@ -80,7 +80,7 @@ static struct cvss_metrics **cvss_impact_metricsptr(struct cvss_impact* impact,
     }
 }
 
-struct cvss_impact *cvss_impact_new(void) { return calloc(1, sizeof(struct cvss_metrics)); }
+struct cvss_impact *cvss_impact_new(void) { return calloc(1, sizeof(struct cvss_impact)); }
```
mpreisler: heh, it pays off to use standard C functions instead of custom allocators, right? :-)

mpreisler:
ACK

src/CVSS/cvss.c
jan-cerny: @mpreisler Yes, I also have some reports about incorrect usage of `realloc`. I'm going to fix them in some of next pull requests.

PR #884, opened 2017-10-02
Coverity fixes 2 - Fix possible NULL pointer dereferencing

jan-cerny:
"xccdf_policy_get_benchmark" can return NULL. It can cause a segfault
when we pass the returned value to "xccdf_benchmark_get_version".
I add should check if benchmark is not NULL. I move the statement
to the begining of the block because at original place I have to
free profile_title and profile_description.

Addressing:
src/XCCDF_POLICY/xccdf_policy_remediate.c:692: dereference:
Dereferencing a pointer that might be null "benchmark" when calling
"xccdf_benchmark_get_version".


matejak:
I have checked it and didn't notice any rough edges.

PR #887, opened 2017-10-03
Coverity fixes 4

jan-cerny:
This pull request fixes some of new issues reported by Coverity. See commit messages.

### DIFF
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -510,8 +510,9 @@ int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const 
 	}
 
 	if (return_code == OSCAP_PROFILE_MATCH_OK) {
-		const bool search_result = xccdf_session_set_profile_id(session, full_profile_id);
-		assert(search_result);
+		if (!xccdf_session_set_profile_id(session, full_profile_id)) {
+			return_code = OSCAP_PROFILE_NO_MATCH;
+		}
 	}
 	return return_code;
 }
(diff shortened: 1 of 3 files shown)


src/XCCDF/xccdf_session.c
```diff
@@ -486,8 +486,7 @@ int xccdf_session_set_profile_id_by_suffix(struct xccdf_session *session, const
 		return 2;
 	} else {
 		const bool search_result = xccdf_session_set_profile_id(session, full_profile_id);
-		assert(search_result);
-		return 0;
+		return (search_result ? 0 : 1);
```
mpreisler: this variable is superfluous, why not just use the result of set_profile_id directly?

src/common/list.c
```diff
@@ -393,7 +393,7 @@ struct oscap_htable *oscap_htable_new1(oscap_compare_func cmp, size_t hsize)
 		return NULL;
 	t->hsize = hsize;
 	t->itemcount = 0;
-	t->table = calloc(hsize, sizeof(struct oscap_list_item *));
+	t->table = calloc(hsize, sizeof(struct oscap_htable_item *));
```
mpreisler: wow, just wow

:D

src/XCCDF/xccdf_session.c
jan-cerny: ok

src/common/list.c
jan-cerny: It didn't cause problems because size of pointer to a structure is same as size of pointer to a different structure.

jan-cerny:
I have amended the first commit so that there is no superfulous variable anymore.

PR #889, opened 2017-10-06
Update and Enhance DISA SRG and STIG links

redhatrises:
- Update DISA SRG link for Operating Systems
- Add DISA SRG link for Applications
- Add DISA STIG link for Applications

### DIFF
--- a/xsl/xccdf-report-impl.xsl
+++ b/xsl/xccdf-report-impl.xsl
@@ -471,10 +471,15 @@ Authors:
         <xsl:when test="starts-with($href, 'http://iase.disa.mil/stigs/cci/')">
             <xsl:text>DISA CCI</xsl:text>
         </xsl:when>
-        <xsl:when test="starts-with($href, 'http://iase.disa.mil/stigs/srgs/')">
+        <!-- SRG weblinks can be subject to change. Keep the old ones for compatibility and add any new ones -->
+        <xsl:when test="starts-with($href, 'http://iase.disa.mil/stigs/srgs/') or 
+                starts-with($href, 'http://iase.disa.mil/stigs/os/general/Pages/index.aspx') or
+                starts-with($href, 'http://iase.disa.mil/stigs/app-security/app-servers/Pages/general.aspx')">
             <xsl:text>DISA SRG</xsl:text>
         </xsl:when>
-        <xsl:when test="starts-with($href, 'http://iase.disa.mil/stigs/os/')">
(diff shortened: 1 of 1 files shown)


mpreisler:
This breaks old content.

xsl/xccdf-report-impl.xsl
```diff
@@ -471,7 +471,10 @@ Authors:
...
             <xsl:text>DISA CCI</xsl:text>
         </xsl:when>
-        <xsl:when test="starts-with($href, 'http://iase.disa.mil/stigs/srgs/')">
+        <xsl:when test="starts-with($href, 'http://iase.disa.mil/stigs/os/general/Pages/index.aspx')">
+            <xsl:text>DISA SRG</xsl:text>
+        </xsl:when>
+        <xsl:when test="starts-with($href, 'http://iase.disa.mil/stigs/app-security/app-servers/Pages/general.aspx')">
             <xsl:text>DISA SRG</xsl:text>
         </xsl:when>
         <xsl:when test="starts-with($href, 'http://iase.disa.mil/stigs/os/')">
```
mpreisler: conflict here

mpreisler:
Added some review notes, we need to be very careful to avoid conflicts and keep this supportable.

xsl/xccdf-report-impl.xsl
redhatrises: @mpreisler what is the conflict?

mpreisler: there are 2 branches that will match the URL and both result in different shortnames. I know this is deterministic behavior and the first will match but I don't think you wanted to do that...

redhatrises: suggestions? There is no longer a consistent URL for SRGs.

jan-cerny: Maybe nest it
```
<when .. starts-with 'http://iase.disa.mil/stigs/os/' ..
               <when test="startswith http://iase.disa.mil/stigs/os/general/Pages/index.aspx
                        DISA SRG
               </when>
              <otherwise>
                        DISA STIG
              </otherwise>
</when>
```

And don't remove `<xsl:when test="starts-with($href, 'http://iase.disa.mil/stigs/srgs/')">`

redhatrises:
Updated to use "or" with "starts-with()"

mpreisler:
ACK

PR #894, opened 2017-10-12
Creates the command line parameter --stig-viewer to output STIG Viewer compatible results

WesleyCeraso:
This commit creates new fields in xccdf_rule_item and xccdf_rule_result, both meant to store the stig rule id associated with that rule.

In order to get the stig rule id, when parsing a Rule, it looks for a reference with href="http://iase.disa.mil/stigs/Pages/stig-viewing-guidance.asp", and uses its value.

I could not find any tag in XCCDF standard where I can store the stig rule id in a ```<rule-result>```, it means there is no way to input the the stig rule id when remediating without the original XCCDF.

### DIFF
--- a/docs/manual/manual.adoc
+++ b/docs/manual/manual.adoc
@@ -524,13 +524,28 @@ $ oscap xccdf eval --benchmark-id benchmark_id --results-arf arf-results.xml sca
 
 ==== Result STIG Viewer
 
-You can also generate XCCDF result files readable by DISA STIG Viewer using the 
-`--stig-viewer` command-line argument. Note that the XCCDF input must contain references
-to the STIG Rule ID for each Rule. See link:http://iase.disa.mil/stigs/Pages/stig-viewing-guidance.aspx[DISA STIG Viewer]
+If you want to import the XCCDF scan results to DISA STIG Viewer but your Rule IDs don't
+match the DISA's ones, you can use the `--stig-viewer` command-line argument along with
+a special reference in your Rules to generate XCCDF result files that can be imported by
+DISA STIG Viewer.
 
 --------------------------------------------------------------------------------------
 $ oscap xccdf eval --profile stig-rhel7-disa --stig-viewer results-stig.xml ssg-rhel7-ds.xml
 --------------------------------------------------------------------------------------
+
+Each rule in the input XCCDF must contain a reference to its STIG Rule ID, and its 
+href attribute must be exactly "http://iase.disa.mil/stigs/Pages/stig-viewing-guidance.aspx".
+
+For example:
+----
+<Rule id="rpm_verify_permissions">
+  ...
+  <reference href="http://iase.disa.mil/stigs/Pages/stig-viewing-guidance.aspx">SV-86473r2_rule</reference>
+  ...
+</Rule>
+----
+
+For more information on DISA STIG Viewer click link:http://iase.disa.mil/stigs/Pages/stig-viewing-guidance.aspx[here].
 
 === Remediate System
 OpenSCAP allows to automatically remediate systems that have been found in a
--- a/docs/manual/manual.adoc
+++ b/docs/manual/manual.adoc
@@ -524,12 +524,12 @@ $ oscap xccdf eval --benchmark-id benchmark_id --results-arf arf-results.xml sca
 
 ==== Result STIG Viewer
 
-You can also generate XCCDF result files that uses the DISA STIG Rule IDs using the 
+You can also generate XCCDF result files readable by DISA STIG Viewer using the 
 `--stig-viewer` command-line argument. Note that the XCCDF input must contain references
-with the STIG Rule ID for each Rule.
+to the STIG Rule ID for each Rule. See link:http://iase.disa.mil/stigs/Pages/stig-viewing-guidance.aspx[DISA STIG Viewer]
 
 --------------------------------------------------------------------------------------
-$ oscap xccdf eval --benchmark-id benchmark_id --stig-viewer stig-viewer-result.xml scap-ds.xml
+$ oscap xccdf eval --profile stig-rhel7-disa --stig-viewer results-stig.xml ssg-rhel7-ds.xml
 --------------------------------------------------------------------------------------
 
 === Remediate System
--- a/docs/manual/manual.adoc
+++ b/docs/manual/manual.adoc
@@ -522,6 +522,16 @@ command-line argument. You can use `--results-arf` to generate a Result DataStre
 $ oscap xccdf eval --benchmark-id benchmark_id --results-arf arf-results.xml scap-ds.xml
 --------------------------------------------------------------------------------------
 
+==== Result STIG Viewer
+
+You can also generate XCCDF result files that uses the DISA STIG Rule IDs using the 
+`--stig-viewer` command-line argument. Note that the XCCDF input must contain references
+with the STIG Rule ID for each Rule.
+
+--------------------------------------------------------------------------------------
+$ oscap xccdf eval --benchmark-id benchmark_id --stig-viewer stig-viewer-result.xml scap-ds.xml
+--------------------------------------------------------------------------------------
+
 === Remediate System
 OpenSCAP allows to automatically remediate systems that have been found in a
 non-compliant state. For system remediation, an XCCDF file with instructions is
--- a/src/XCCDF/item.c
+++ b/src/XCCDF/item.c
@@ -480,7 +480,7 @@ xmlNode *xccdf_item_to_dom(struct xccdf_item *item, xmlDoc *doc, xmlNode *parent
 			break;
 		case XCCDF_RESULT:
 			xmlNodeSetName(item_node,BAD_CAST "TestResult");
-			if (parent) xccdf_result_to_dom(XRESULT(item), item_node, doc, parent);
+			if (parent) xccdf_result_to_dom(XRESULT(item), item_node, doc, parent, false);
 			break;
 		case XCCDF_GROUP:
 			xmlNodeSetName(item_node,BAD_CAST "Group");
@@ -1135,6 +1135,7 @@ void xccdf_rule_item_clone(struct xccdf_rule_item *clone, const struct xccdf_rul
 	clone->profile_notes = oscap_list_clone(item->profile_notes, (oscap_clone_func) xccdf_profile_note_clone);
 	clone->fixes = oscap_list_clone(item->fixes, (oscap_clone_func) xccdf_fix_clone);
 	clone->fixtexts = oscap_list_clone(item->fixtexts, (oscap_clone_func) xccdf_fixtext_clone);
+	clone->stig_rule_id = oscap_strdup(item->stig_rule_id);
 }
 
 void xccdf_group_item_clone(struct xccdf_item *parent, const struct xccdf_group_item * item)
--- a/src/XCCDF/item.c
+++ b/src/XCCDF/item.c
@@ -1135,7 +1135,6 @@ void xccdf_rule_item_clone(struct xccdf_rule_item *clone, const struct xccdf_rul
 	clone->profile_notes = oscap_list_clone(item->profile_notes, (oscap_clone_func) xccdf_profile_note_clone);
 	clone->fixes = oscap_list_clone(item->fixes, (oscap_clone_func) xccdf_fix_clone);
 	clone->fixtexts = oscap_list_clone(item->fixtexts, (oscap_clone_func) xccdf_fixtext_clone);
-	clone->stig_rule_id = oscap_strdup(item->stig_rule_id);
 }
 
 void xccdf_group_item_clone(struct xccdf_item *parent, const struct xccdf_group_item * item)
--- a/src/XCCDF/item.h
+++ b/src/XCCDF/item.h
@@ -101,8 +101,6 @@ struct xccdf_rule_item {
 	struct oscap_list *checks;
 	struct oscap_list *fixes;
 	struct oscap_list *fixtexts;
-	
-	char *stig_rule_id;
 };
 
 struct xccdf_group_item {
--- a/src/XCCDF/item.h
+++ b/src/XCCDF/item.h
@@ -101,6 +101,8 @@ struct xccdf_rule_item {
 	struct oscap_list *checks;
 	struct oscap_list *fixes;
 	struct oscap_list *fixtexts;
+	
+	char *stig_rule_id;
 };
 
 struct xccdf_group_item {
--- a/src/XCCDF/public/xccdf_benchmark.h
+++ b/src/XCCDF/public/xccdf_benchmark.h
@@ -773,11 +773,11 @@ OSCAP_DEPRECATED(int xccdf_result_export(struct xccdf_result *result, const char
 struct oscap_source *xccdf_result_export_source(struct xccdf_result *result, const char *filepath);
 
 /**
- * Export TestResult to oscap_source structure
+ * Export TestResult to oscap_source structure using STIG Rule IDs instead of the actual rule ids
  * @memberof xccdf_result
  * @returns newly created oscap_source or NULL on error
  */
-struct oscap_source *stig_rule_id_xccdf_result_export_source(struct xccdf_result *result, const char *filepath);
+struct oscap_source *xccdf_result_stig_viewer_export_source(struct xccdf_result *result, const char *filepath);
 
 /**
  * Resolve an benchmark.
--- a/src/XCCDF/public/xccdf_benchmark.h
+++ b/src/XCCDF/public/xccdf_benchmark.h
@@ -773,6 +773,13 @@ OSCAP_DEPRECATED(int xccdf_result_export(struct xccdf_result *result, const char
 struct oscap_source *xccdf_result_export_source(struct xccdf_result *result, const char *filepath);
 
 /**
+ * Export TestResult to oscap_source structure
+ * @memberof xccdf_result
+ * @returns newly created oscap_source or NULL on error
+ */
+struct oscap_source *stig_rule_id_xccdf_result_export_source(struct xccdf_result *result, const char *filepath);
+
+/**
  * Resolve an benchmark.
  * @returns whether the resolving process has been successful
  * @retval true on success
@@ -2169,6 +2176,10 @@ const char *xccdf_rule_get_impact_metric(const struct xccdf_rule *rule);
 /**
  * @memberof xccdf_rule
  */
+const char *xccdf_rule_get_stig_rule_id(const struct xccdf_rule *rule);
+/**
+ * @memberof xccdf_rule
+ */
 xccdf_role_t xccdf_rule_get_role(const struct xccdf_rule *rule);
 /**
  * @memberof xccdf_rule
@@ -2864,6 +2875,8 @@ bool xccdf_rule_set_multiple(struct xccdf_rule *item, bool newval);
 //bool xccdf_rule_set_selector(struct xccdf_rule *item, const char * selector);
 /// @memberof xccdf_rule
 bool xccdf_rule_set_impact_metric(struct xccdf_rule *item, const char *newval);
+/// @memberof xccdf_rule
+bool xccdf_rule_set_stig_rule_id(struct xccdf_rule *item, const char *newval);
 /// @memberof xccdf_rule
 bool xccdf_rule_set_role(struct xccdf_rule *item, xccdf_role_t newval);
 /// @memberof xccdf_rule
--- a/src/XCCDF/public/xccdf_benchmark.h
+++ b/src/XCCDF/public/xccdf_benchmark.h
@@ -2176,10 +2176,6 @@ const char *xccdf_rule_get_impact_metric(const struct xccdf_rule *rule);
 /**
  * @memberof xccdf_rule
  */
-const char *xccdf_rule_get_stig_rule_id(const struct xccdf_rule *rule);
-/**
- * @memberof xccdf_rule
- */
 xccdf_role_t xccdf_rule_get_role(const struct xccdf_rule *rule);
 /**
  * @memberof xccdf_rule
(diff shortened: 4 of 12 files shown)


src/XCCDF/item.h
```diff
@@ -344,6 +346,7 @@ struct xccdf_fixtext {
 
 struct xccdf_rule_result {
 	char *idref;
+	char *stig_rule_id;
```
mpreisler: this seems excessive to store something this specific in every rule result. are you sure we can't do this in any different way? Can't we reference the actual Rule using idref and just get the ID there?

utils/oscap-xccdf.c
```diff
@@ -1040,6 +1045,7 @@ bool getopt_xccdf(int argc, char **argv, struct oscap_action *action)
 	// options
 		{"output",		required_argument, NULL, XCCDF_OPT_OUTPUT},
 		{"results", 		required_argument, NULL, XCCDF_OPT_RESULT_FILE},
+		{"results-stig", 	required_argument, NULL, XCCDF_OPT_RESULT_FILE_STIG},
```
mpreisler: we probably should put stig-viewer there because results-stig is misleading. people might switch --results to --results-stig just because they are evaluating a stig profile

mpreisler:
Needs to be added to man page as well.

utils/oscap-xccdf.c
jan-cerny: +1

src/XCCDF/item.c
```diff
@@ -480,7 +480,7 @@ xmlNode *xccdf_item_to_dom(struct xccdf_item *item, xmlDoc *doc, xmlNode *parent
 			break;
 		case XCCDF_RESULT:
 			xmlNodeSetName(item_node,BAD_CAST "TestResult");
-			if (parent) xccdf_result_to_dom(XRESULT(item), item_node, doc, parent);
+			if (parent) xccdf_result_to_dom(XRESULT(item), item_node, doc, parent, false);
```
jan-cerny: Why is this done?

WesleyCeraso: Because I had to add the option to output the stig rule id instead of the real rule id. However, I don't want it to happen when I call it here.

utils/oscap-xccdf.c
WesleyCeraso: Ok, makes sense, but not even "results-stig-viewer"?

src/XCCDF/item.h
WesleyCeraso: That's a good ideia, it's done!

WesleyCeraso:
I changed the command line parameter name and added missing documentation.

docs/manual/manual.adoc
```diff
@@ -522,6 +522,16 @@ command-line argument. You can use `--results-arf` to generate a Result DataStre
...
 --------------------------------------------------------------------------------------
 
+==== Result STIG Viewer
+
+You can also generate XCCDF result files that uses the DISA STIG Rule IDs using the 
+`--stig-viewer` command-line argument. Note that the XCCDF input must contain references
+with the STIG Rule ID for each Rule.
+
+--------------------------------------------------------------------------------------
+$ oscap xccdf eval --benchmark-id benchmark_id --stig-viewer stig-viewer-result.xml scap-ds.xml
```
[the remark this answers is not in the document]
yuumasato: I used the `--profile` approach. What were you using in your tests @WesleyCeraso?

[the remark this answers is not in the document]
WesleyCeraso: @yuumasato I used the --profile as well. It was my mistake to copy the command line from results-arf and only change the parameter.

src/XCCDF/item.h
```diff
@@ -101,6 +101,8 @@ struct xccdf_rule_item {
 	struct oscap_list *checks;
 	struct oscap_list *fixes;
 	struct oscap_list *fixtexts;
+	
+	char *stig_rule_id;
```
mpreisler: isn't this already stored in `idents`? I don't understand the point of all the extra logic.

src/XCCDF/result.c
```diff
@@ -1067,16 +1077,21 @@ xmlNode *xccdf_target_identifier_to_dom(const struct xccdf_target_identifier *ti
...
 
 	xmlNode *result_node = xmlNewTextChild(parent, ns_xccdf, BAD_CAST "rule-result", NULL);
 
 	/* Handle attributes */
 	const char *idref = xccdf_rule_result_get_idref(result);
-	if (idref)
-		xmlNewProp(result_node, BAD_CAST "idref", BAD_CAST idref);
+	if (benchmark && use_stig_rule_id) {
+		struct xccdf_item *item = xccdf_benchmark_get_member(benchmark, XCCDF_RULE, idref);
+		const char *stig_rule_id = xccdf_rule_get_stig_rule_id(XRULE(item));
```
mpreisler: we can query this from idents and references without introducing extra methods and extra fields, that would make the patch way smaller

src/XCCDF/rule.c
```diff
@@ -37,6 +37,9 @@
 #include "common/assume.h"
 #include "common/debug_priv.h"
 
+//SSG constant <xsl:variable name="disa-stigs-viewer">http://iase.disa.mil/stigs/Pages/stig-viewing-guidance.aspx</xsl:variable>
```
mpreisler: please improve this comment, explain what it is, don't copy paste XSLT source

utils/oscap.8
```diff
@@ -116,6 +116,11 @@ Write XCCDF results into FILE.
 Writes results to a given FILE in Asset Reporting Format. It is recommended to use this option instead of --results when dealing with datastreams.
 .RE
 .TP
+\fB\-\-stig-viewer FILE\fR
+.RS
+Write XCCDF results with DISA STIG Rule IDs into FILE.
```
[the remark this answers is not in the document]
mpreisler: Agree with @yuumasato . I would also explain that it means all the Rule `id` attributes will be replaced by DISA STIG Viewer IDs. A link to DISA STIG viewer would also help.

WesleyCeraso:
@mpreisler and @yuumasato I have fixed what you have pointed out.

src/XCCDF/result.c
```diff
@@ -1067,16 +1080,32 @@ xmlNode *xccdf_target_identifier_to_dom(const struct xccdf_target_identifier *ti
...
 }
 
-xmlNode *xccdf_rule_result_to_dom(struct xccdf_rule_result *result, xmlDoc *doc, xmlNode *parent, const struct xccdf_version_info* version_info)
+xmlNode *xccdf_rule_result_to_dom(struct xccdf_rule_result *result, xmlDoc *doc, xmlNode *parent, const struct xccdf_version_info* version_info, struct xccdf_benchmark *benchmark, bool use_stig_rule_id)
 {
+	/* Handle attributes */
+	const char *idref = xccdf_rule_result_get_idref(result);
+	if (benchmark && use_stig_rule_id) {
+		const char *stig_rule_id = NULL;
+		struct xccdf_item *item = xccdf_benchmark_get_member(benchmark, XCCDF_RULE, idref);
```
mpreisler: assert item is not NULL

src/XCCDF/result.c
```diff
@@ -1067,16 +1080,32 @@ xmlNode *xccdf_target_identifier_to_dom(const struct xccdf_target_identifier *ti
...
+		struct oscap_reference_iterator *references = xccdf_item_get_references(XRULE(item));
+		while (oscap_reference_iterator_has_more(references)) {
+			struct oscap_reference *ref = oscap_reference_iterator_next(references);
+			if (strcmp(ref->href, DISA_STIG_VIEWER_HREF) == 0)
+				stig_rule_id = ref->title;
```
mpreisler: save time, break

src/XCCDF/result.c
```diff
@@ -1067,16 +1080,32 @@ xmlNode *xccdf_target_identifier_to_dom(const struct xccdf_target_identifier *ti
...
+		}
+		oscap_reference_iterator_free(references);
+
+		if (stig_rule_id)
+			idref = stig_rule_id;
+		else
+			return;
```
mpreisler: code style, plus this definitely needs a comment. It looks like we are avoiding export of rules that don't have a stig viewer ID, this should be explained here IMO

src/XCCDF/result.c
```diff
@@ -1067,16 +1080,32 @@ xmlNode *xccdf_target_identifier_to_dom(const struct xccdf_target_identifier *ti
...
+	}
+	
 	xmlNs *ns_xccdf = lookup_xccdf_ns(doc, parent, version_info);
 
 	xmlNode *result_node = xmlNewTextChild(parent, ns_xccdf, BAD_CAST "rule-result", NULL);
 
-	/* Handle attributes */
-	const char *idref = xccdf_rule_result_get_idref(result);
-	if (idref)
-		xmlNewProp(result_node, BAD_CAST "idref", BAD_CAST idref);
```
mpreisler: this change chances logic, idref was checked before the PR, after the PR it isn't

src/XCCDF/xccdf_session.c
```diff
@@ -414,6 +416,13 @@ bool xccdf_session_set_xccdf_export(struct xccdf_session *session, const char *x
 	return true;
 }
 
+bool xccdf_session_set_stigviewer_export(struct xccdf_session *session, const char *stig_rule_id_file)
+{
+	free(session->export.stig_rule_id_file);
+	session->export.stig_rule_id_file = oscap_strdup(stig_rule_id_file);
```
mpreisler: use just strdup

WesleyCeraso:
@mpreisler Updated!

src/XCCDF/xccdf_session.c
mpreisler: oh I forgot strdup is not standard C, nvm

src/XCCDF/result.c
```diff
@@ -1067,14 +1080,40 @@ xmlNode *xccdf_target_identifier_to_dom(const struct xccdf_target_identifier *ti
...
+		while (oscap_reference_iterator_has_more(references)) {
+			struct oscap_reference *ref = oscap_reference_iterator_next(references);
+			if (strcmp(ref->href, DISA_STIG_VIEWER_HREF) == 0) {
+				stig_rule_id = ref->title;
+				break;
+			}
+		}
+		oscap_reference_iterator_free(references);
+
+		if (!stig_rule_id)
```
yuumasato: Shouldn't criteria for not outputting the rule result be no `stig_rule_id` and rule result `notchecked`? See my comments in previous review.

utils/oscap-xccdf.c
```diff
@@ -157,6 +157,7 @@ static struct oscap_module XCCDF_EVAL = {
         "   --export-variables\r\t\t\t\t - Export OVAL external variables provided by XCCDF.\n"
         "   --results <file>\r\t\t\t\t - Write XCCDF Results into file.\n"
         "   --results-arf <file>\r\t\t\t\t - Write ARF (result data stream) into file.\n"
+        "   --stig-viewer <file>\r\t\t\t\t - Write XCCDF Results with DISA STIG Rule IDs into file.\n"
```
yuumasato: Can this be updated as well?

utils/oscap-xccdf.c
```diff
@@ -194,6 +195,7 @@ static struct oscap_module XCCDF_REMEDIATE = {
 			"  --fetch-remote-resources\r\t\t\t\t - Download remote content referenced by XCCDF.\n"
 			"  --results <file>\r\t\t\t\t - Write XCCDF Results into file.\n"
 			"  --results-arf <file>\r\t\t\t\t - Write ARF (result data stream) into file.\n"
+			"  --stig-viewer <file>\r\t\t\t\t - Write XCCDF Results with DISA STIG Rule IDs into file.\n"
```
yuumasato: Update description, please.

src/XCCDF/xccdf_session.c
```diff
@@ -419,7 +419,7 @@ bool xccdf_session_set_xccdf_export(struct xccdf_session *session, const char *x
 bool xccdf_session_set_stigviewer_export(struct xccdf_session *session, const char *stig_rule_id_file)
 {
 	free(session->export.stig_rule_id_file);
-	session->export.stig_rule_id_file = oscap_strdup(stig_rule_id_file);
+	session->export.stig_rule_id_file = strdup(stig_rule_id_file);
```
mpreisler: you need to use oscap_strdup, sorry about my misleading comment

src/XCCDF/result.c
```diff
@@ -1067,14 +1080,40 @@ xmlNode *xccdf_target_identifier_to_dom(const struct xccdf_target_identifier *ti
 	}
 }
 
-xmlNode *xccdf_rule_result_to_dom(struct xccdf_rule_result *result, xmlDoc *doc, xmlNode *parent, const struct xccdf_version_info* version_info)
+xmlNode *xccdf_rule_result_to_dom(struct xccdf_rule_result *result, xmlDoc *doc, xmlNode *parent, const struct xccdf_version_info* version_info, struct xccdf_benchmark *benchmark, bool use_stig_rule_id)
```
mpreisler: you can query the benchamrk from the result, you don't have to pass it to the function

src/XCCDF/xccdf_session.c
```diff
@@ -414,6 +416,13 @@ bool xccdf_session_set_xccdf_export(struct xccdf_session *session, const char *x
 	return true;
 }
 
+bool xccdf_session_set_stigviewer_export(struct xccdf_session *session, const char *stig_rule_id_file)
```
mpreisler: stig_rule_id_file sounds like it's a file of stig rule IDs where it's actually an XCCDF results file

src/XCCDF/xccdf_session.c
```diff
@@ -1265,14 +1274,15 @@ static int _build_xccdf_result_source(struct xccdf_session *session)
...
 		if (session->xccdf.result == NULL) {
 			// Attempt to export session before evaluation
 			oscap_seterr(OSCAP_EFAMILY_OSCAP, "No XCCDF results to export.");
 			return 1;
 		}
-		xccdf_benchmark_add_result(xccdf_policy_model_get_benchmark(session->xccdf.policy_model),
-				xccdf_result_clone(session->xccdf.result));
+
+		struct xccdf_result *benchmarks_result = xccdf_result_clone(session->xccdf.result);
+		xccdf_benchmark_add_result(xccdf_policy_model_get_benchmark(session->xccdf.policy_model),	benchmarks_result);
```
mpreisler: code style, there is a tab as you pass the benchmark. it's also not clear why you made this change because you don't even use this variable anywhere

src/XCCDF/public/xccdf_benchmark.h
```diff
@@ -773,6 +773,13 @@ OSCAP_DEPRECATED(int xccdf_result_export(struct xccdf_result *result, const char
 struct oscap_source *xccdf_result_export_source(struct xccdf_result *result, const char *filepath);
 
 /**
+ * Export TestResult to oscap_source structure
+ * @memberof xccdf_result
+ * @returns newly created oscap_source or NULL on error
+ */
+struct oscap_source *stig_rule_id_xccdf_result_export_source(struct xccdf_result *result, const char *filepath);
```
mpreisler: This is a method so please name it accordingly, it should start with the "class" prefix -  xccdf_result_

src/XCCDF/public/xccdf_benchmark.h
```diff
@@ -773,6 +773,13 @@ OSCAP_DEPRECATED(int xccdf_result_export(struct xccdf_result *result, const char
 struct oscap_source *xccdf_result_export_source(struct xccdf_result *result, const char *filepath);
 
 /**
+ * Export TestResult to oscap_source structure
```
mpreisler: great docs...

mpreisler: needs better docs

src/XCCDF/result.c
WesleyCeraso: Leys say we a have a checked rule with no stig id, what is the point in keeping it on the Stig Viewer output? We would just get the popup complaining about it.

WesleyCeraso: Regarding the documentation, there is a little notice in manual.adoc

WesleyCeraso:
@mpreisler I have improved the code a little bit, there were some left overs from several different attempts. Now it does not duplicate the results nor add it to benchmark.

mpreisler:
@WesleyCeraso this change causes the test suite to fail at several places. Can you reproduce it on your machine or is it just jenkins?

src/XCCDF/xccdf_session.c
```diff
@@ -97,6 +97,7 @@ struct xccdf_session {
 	struct {
 		char *arf_file;				///< Path to ARF file to export
 		char *xccdf_file;			///< Path to XCCDF file to export
+		char *xccdf_stig_viewer_file;		///< Path to STIG Viewer XCCDF file to export
```
mpreisler: does this ever get initialized? I think this is memory garbage in non debug builds. See xccdf_session_new_from_source

mpreisler: ok nvm, the whole thing uses calloc so it's all NULL. something else must be causing the test fails then.

WesleyCeraso:
@mpreisler I'm investigating it right now. I'll push again soon.

WesleyCeraso:
@mpreisler  I got confused and deleted code that was not meant to, the software really needs to duplicate the results and add it to the Benchmark, and it's not something I came up with, it was there previously, I had just edited the text.

mpreisler:
> @mpreisler I got confused and deleted code that was not meant to, the software really needs to duplicate the results and add it to the Benchmark, and it's not something I came up with, it was there previously, I had just edited the text.

my comment was about the fact that you added a variable and then used it for no reason, it was only used once (the variable).

Now the resulting code is slightly different than the original, is there any reason for it? Could you please rather go back in history via `git rebase -i` and avoid this change? that would make things easier to review and I would be more confident this doesn't break anything.

WesleyCeraso:
@mpreisler Actually it was used in two points, and these points are the same now. However I'll rebase to make it better.

src/XCCDF/result.c
yuumasato: Motivation for this is to make it easier to spot mistakes in DISA Profiles.

> Leys say we a have a checked rule with no stig id, what is the point in keeping it on the Stig Viewer output?

In this scenario, somebody might have forgotten to assign a STIG ID to this Rule, or it may have been selected by mistake.

src/XCCDF/result.c
```diff
@@ -1067,14 +1080,40 @@ xmlNode *xccdf_target_identifier_to_dom(const struct xccdf_target_identifier *ti
...
+
+		struct xccdf_item *item = xccdf_benchmark_get_member(benchmark, XCCDF_RULE, idref);
+		if (!item)
+			return NULL;
+
+		const char *stig_rule_id = NULL;		
+		struct oscap_reference_iterator *references = xccdf_item_get_references(XRULE(item));
+		while (oscap_reference_iterator_has_more(references)) {
+			struct oscap_reference *ref = oscap_reference_iterator_next(references);
+			if (strcmp(ref->href, DISA_STIG_VIEWER_HREF) == 0) {
```
mpreisler: use oscap_reference_get_href instead

WesleyCeraso:
@mpreisler It uses the getter functions now.

jan-cerny:
What remains here to be done? I guess it is for 1.2.16 milestone, isn't it?

dahaic:
@yuumasato can you update your review? :)

yuumasato:
Code looks ok. Output of scan generated with `--stig-viewer` option can be nicely imported in DISA STIG Viewer.

In previous review I mentioned concern on following two scenarios:
- An evaluated Rule without `stigid` will be left out of output generated by `--stig-viewer`.
- A Rule with `stigid` that is not selected will not be evaluated and result in `notselected`, but still be output by `--stig-viewer` option. In this case DISA STIG Viewer input the Rule as Not Reviewed. 

I'm okay with this, as I see a use case in which someone would like to evaluate a non-STIG Profile (an unrelated Profile or maybe a super set or subset of a STIG Profile) and compare it against DISA STIG checklist.

One thing I miss is how the references in content should look like for `--stig-viewer` option to work. The link in the Manual points to DISA STIG Viewer but it doesn't explain how I can add it to my own content.

It maybe also be worthwhile to explain why one might need to use `--stig-viewer` option. I guess that if  content user is using already has its Rule's IDs the same as STIG Rule IDs, it will output results compatible with DISA STIG Viewer without `--stig-viewer` option.

WesleyCeraso:
@yuumasato I added more info to the manual. Please check if it's helpful.

docs/manual/manual.adoc
```diff
@@ -522,6 +522,31 @@ command-line argument. You can use `--results-arf` to generate a Result DataStre
...
+For example:
+----
+<Rule id="rpm_verify_permissions">
+  ...
+  <reference href="http://iase.disa.mil/stigs/Pages/stig-viewing-guidance.aspx">SV-86473r2_rule</reference>
+  ...
+</Rule>
+----
+
+See link:http://iase.disa.mil/stigs/Pages/stig-viewing-guidance.aspx[DISA STIG Viewer].
```
yuumasato: It looks like this link will give more information about how to prepare the contents for `--stig-viewer` output.
Suggestion: `See following link for information on what is DISA STIG Viewer`

docs/manual/manual.adoc
```diff
@@ -522,6 +522,31 @@ command-line argument. You can use `--results-arf` to generate a Result DataStre
...
+
+--------------------------------------------------------------------------------------
+$ oscap xccdf eval --profile stig-rhel7-disa --stig-viewer results-stig.xml ssg-rhel7-ds.xml
```
yuumasato: This command line example looks a bit loose and out of context.
Suggest to move this up, closer to mention of `you can use the '--stig-viewer'`, or add a sentence before the example.

WesleyCeraso:
@yuumasato I have applied your suggestions.

yuumasato:
@WesleyCeraso Thanks
Ack.

PR #896, opened 2017-10-16
Ansible variables to come first before tasks

mpreisler:
In ansible playbooks it is customary for variables to be specified in a list separate from tasks. That way Ansible users can tailor the playbooks without any external tools:

Playbook after this change:
```
---
###############################################################################
#
# Ansible remediation role for profile xccdf_org.ssgproject.content_profile_pci-dss
# Profile Title:  PCI-DSS v3 Control Baseline for Red Hat Enterprise Linux 7
# Profile Description:
# This is a *draft* profile for PCI-DSS v3.
#
# Benchmark ID:  xccdf_org.ssgproject.content_benchmark_RHEL-7
# Benchmark Version:  0.1.36
#
# XCCDF Version:  1.2
#
# This file was generated by OpenSCAP 1.2.16 using:
# 	$ oscap xccdf generate fix --profile xccdf_org.ssgproject.content_profile_pci-dss --template urn:xccdf:fix:script:ansible sds.xml 
#
# This script is generated from an OpenSCAP profile without preliminary evaluation.
# It attempts to fix every selected rule, even if the system is already compliant.
#
# How to apply this remediation role:
# $ ansible-playbook -i "192.168.1.155," playbook.yml
# $ ansible-playbook -i inventory.ini playbook.yml
#
###############################################################################


 - hosts: all
   vars:
      inactivity_timeout_value: 900
      var_accounts_maximum_age_login_defs: 90
      var_account_disable_post_pw_expiration: 90
      var_password_pam_dcredit: -1
      var_password_pam_minlen: 7
      var_password_pam_ucredit: -1
      var_password_pam_lcredit: -1
      var_auditd_max_log_file: 6
      var_auditd_max_log_file_action: rotate
      var_auditd_space_left_action: email
      var_auditd_admin_space_left_action: single
      var_auditd_action_mail_acct: root
      sshd_idle_timeout_value: 900
   tasks:
    - name: "Read permission of GPG key directory"
      stat:
        path: /etc/pki/rpm-gpg/
      register: gpg_key_directory_permission
      check_mode: no
      tags:
        - ensure_redhat_gpgkey_installed
        - high_severity
        - restrict_strategy
        - medium_complexity
        - medium_disruption
...
```

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -598,10 +598,17 @@ static inline int _xccdf_policy_rule_generate_fix_ansible(const char *template, 
 		return 1;
 	}
 
+	// ovector sizing:
+	// 2 elements are used for the whole needle,
+	// 4 elements are used for the 2 capture groups
+	// pcre documentation says we should allocate a third extra for additional
+	// workspace.
+	// (3 * 2) * (3 / 2) = 9
+	int ovector[9];
+
 	const size_t fix_text_len = strlen(fix_text);
 	int start_offset = 0;
 	while (true) {
-		int ovector[30];
 		const int match = pcre_exec(re, NULL, fix_text, fix_text_len, start_offset,
 				0, ovector, sizeof(ovector) / sizeof(ovector[0]));
 		if (match == -1)
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -696,6 +696,7 @@ static inline int _xccdf_policy_rule_generate_fix(struct xccdf_policy *policy, s
 			_write_remediation_to_fd_and_free(output_fd, template, remediation_part);
 		}
 
+		free(fix_text);
 		pcre_free(re);
 #else
 		// TODO: Implement the post-process for posix regex as well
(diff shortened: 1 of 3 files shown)


mpreisler:
This solution will repeat variables if they are used multiple times. I have looked into it and the result is still a valid playbook. The last variable definition wins.

Still, it's suboptimal that we don't do something nicer. To solve all of these issues we would have to buffer everything or at least remember a list of variables we have already written out.

jan-cerny:
> This solution will repeat variables if they are used multiple times. I have looked into it and the result is still a valid playbook. The last variable definition wins.

> Still, it's suboptimal that we don't do something nicer. To solve all of these issues we would have to buffer everything or at least remember a list of variables we have already written out.

We already have "buffered" a list of all the rules that contain a remeditaion (`rules_to_fix`). In your PR you already iterate through this list twice. First to extract the variables, second to print the scripts.

However, `_xccdf_policy_rule_generate_fix` does kind of complicated logic with finding applicable CPEs, suitable fixes, etc, which is done twice, in both iteration.

I think we should instead of a list of rules create two lists:
* a list of fixes
* a list of variables used in them
I guess these list could be generated at once in a single iteration over XCCDF results.

jan-cerny:
Also, it can be more readable to put the regex stuff into a separate function, and  describe meaning of the array indices.

mpreisler:
> I think we should instead of a list of rules create two lists:
> 
> a list of fixes
> a list of variables used in them
> I guess these lists could be generated at once in a single iteration over XCCDF results.

This would be a more or less a complete rewrite of how remediation generating is done.

mpreisler:
> Also, it can be more readable to put the regex stuff into a separate function, and describe meaning of the array indices.

will do

jan-cerny:
Awesome! Thanks.

PR #770, opened 2017-06-29
Fixes crash when deallocating rbt_node in windows

WesleyCeraso:
Fixes OpenSCAP/scap-workbench#129 and OpenSCAP/scap-workbench#61

Signed-off-by: Wesley Ceraso Prudencio <wcerasop@redhat.com>

PR #897, opened 2017-10-17
CVRF Parsing and Export

WesleyCeraso:
Somehow GitHub stopped fetching new commits from Katerina's branch. As I had to create a new PR, I  also rebased it to get rid of conflicts.


WesleyCeraso:
@mpreisler I have actually found a serious flaw, however it's not exclusive of this PR. We have been using free on memory allocated though libxml2, on linux their implementation uses malloc, so its not a big deal. However, on Windows, it can cause crashes similar to https://github.com/OpenSCAP/openscap/pull/770 and it can impact @jan-cerny work.

When we clone such pointers we use malloc, so it's not possible to replace every free with xmlFree because the memory could have been allocated in both ways.

As I said, the same problem occurs in other parts of the lib, so I think we should not try to fix it here in this PR.

So, what do you think, are ready to merge this PR?

WesleyCeraso:
Oh, it took awhile but GitHub finally updated the original pull request, still, I'll stick to this one.

mpreisler:
> So, what do you think, are we ready to merge this PR?

I will defer to you. There is a bunch of comments in the old PR by you that are not fixed, are you planning to fix them?

mpreisler:
> @mpreisler As I told you, they were more like a To Do list, and I checked every one of them, although I have not responded for each.

OK, this was not clear to me as.

Very well then, let's merge this.

jan-cerny:
Wow :+1:

PR #907, opened 2017-11-15
Run sql57 probe tests only if sql57 probe is build

jan-cerny:
The test should be skipped if the probe is not build.


jan-cerny:
Maybe we could do the same in future for all probe tests

mpreisler:
ACK

PR #910, opened 2017-11-15
Remove posix regex leftovers from master branch

jan-cerny:
Support for posix regular expression in master branch was removed long
time ago in 8bcad0850db80d379ca70a3daf293108af78c91e.
Bbut some code appeared again in master, probably by merging maint-1.2
branch. We need to be more careful when we merge maint-1.2 into master.


mpreisler:
ACK

PR #914, opened 2017-11-20
[Do not merge] Pack some CMake modules in our repository

### LINKED ISSUE #906: Cmake can't build swig bindings on RHEL6 and 7
CMake Error at swig/python2/CMakeLists.txt:9 (swig_add_library):
  Unknown CMake command "swig_add_library".

It looks like "UseSWIG" is a new module in CMake 3.0
https://cmake.org/cmake/help/v3.0/module/UseSWIG.html

See:
https://jenkins.open-scap.org/job/openscap-master/10/
https://jenkins.open-scap.org/job/openscap-master/11/

jan-cerny:
`GNUInstallDirs` and `UseSWIG` CMake Modules are not available in older versions of CMake. This PR wants to upload them in our repository.

fixes #906


jan-cerny:
This is wrong.

1) The file /usr/share/cmake/Modules/UseSWIG.cmake exists on RHEL 7, but the problem is that the variable `SWIG_USE_FILE` is wrong, so cmake cannot find the module. See https://cmake.org/Bug/view.php?id=13771

2) GNUInstallDirs is available on my RHEL 7.4 machine. Perhaps it is not needed to include it for now.

I will close this and first fix SWIG on RHEL7.

PR #915, opened 2017-11-20
Fix building SWIG bindings on RHEL7

jan-cerny:
Fixes couple of issues:
1. CMake in RHEL7 isn't able to set ${SWIG_USE_FILE} variable due to
a bug in FindSWIG module. We need to look for the UseSWIG module by
specifying its name directly in 'include' command.
2. CMake older than 3.8 doesn't know swig_add_library,
it defines swig_add_module instead, that does a similar thing.
3. CMake doesn't create 'build/swig/src' directory, but it instructs
swig to save its output there. That leads to a fail of swig, due to a
"no such file or directory error". This commit solves this by moving
files from 'swig/src/' to parent directory ('swig/').

this is part of #906


jan-cerny:
This fail is expected, because Jenkins is not configured to run cmake. Jenkins tries to run ./configure, make check etc, but that commands aren't available on master branch anymore.

A review would be much appreciated :-)

matejak:
Seems to work, bindings are built and then correctly installed.

PR #921, opened 2017-11-29
Enables probe systemdunitproperty to work offline

WesleyCeraso:
This implementation allows the probe to retrieve in offline mode only
the property UnitFileState.

### DIFF
--- a/src/OVAL/probes/unix/linux/systemdunitproperty.c
+++ b/src/OVAL/probes/unix/linux/systemdunitproperty.c
@@ -31,9 +31,12 @@
 #endif
 
 #include <probe-api.h>
+#include <probe/probe.h>
 #include <string.h>
 #include "probe/entcmp.h"
 #include "systemdshared.h"
+
+const char *UNIT_FILE_STATE = "UnitFileState";
 
 static int get_all_properties_by_unit_path(DBusConnection *conn, const char *unit_path, int(*callback)(const char *name, const char *value, void *arg), void *cbarg)
 {
@@ -174,6 +177,7 @@ struct unit_callback_vars {
 	SEXP_t *se_unit;
 	SEXP_t *se_property;
 	SEXP_t *item;
+	char *offline_unit_state;
 };
 
 static int property_callback(const char *property, const char *value, void *cbarg)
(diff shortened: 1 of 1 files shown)


src/OVAL/probes/unix/linux/systemdunitproperty.c
```diff
@@ -251,6 +259,44 @@ static int unit_callback(const char *unit, void *cbarg)
 	return 0;
 }
 
+void get_unit_file_state_in_offline_mode(const char *unit, struct unit_callback_vars *vars)
+{
```
jan-cerny: variable `unit` is not used

src/OVAL/probes/unix/linux/systemdunitproperty.c
```diff
@@ -263,38 +309,55 @@ int probe_main(probe_ctx *ctx, void *probe_arg)
...
+		// Allows the probe to retrieve property UnitFileState even if systemd is not accessible
+		SEXP_t *temp_unit_entity, *temp_property_entity;
+		const char *cstr_unit_entity = SEXP_string_cstr(temp_unit_entity = probe_ent_getval(unit_entity));
+		SEXP_free(temp_unit_entity);
+		const char *cstr_property_entity = SEXP_string_cstr(temp_property_entity = probe_ent_getval(property_entity));
+		SEXP_free(temp_property_entity);
+		
+		if(strcmp(cstr_property_entity, UNIT_FILE_STATE) != 0) {
+			dbus_error_free(&dbus_error);
+			SEXP_t *msg = probe_msg_creat(OVAL_MESSAGE_LEVEL_INFO, "DBus connection failed, could not identify systemd units.");
```
jan-cerny: This talks about DBus connection, but we don't do DBus connection here

jan-cerny:
We're looking for a volunteer to take this pull request, I think only minor changes are needed. Any help would be much appreciated :heart:

src/OVAL/probes/unix/linux/systemdunitproperty.c
```diff
@@ -251,6 +259,44 @@ static int unit_callback(const char *unit, void *cbarg)
 	return 0;
 }
 
+void get_unit_file_state_in_offline_mode(const char *unit, struct unit_callback_vars *vars)
+{
+	char systemctl_call[] = "systemctl list-unit-files | grep -P '\\S+\\.service'";
```
jan-cerny: This ***cannot** be here. We can't call external programs!

matejak:
I can adopt this PR.

src/OVAL/probes/unix/linux/systemdunitproperty.c
matejak: Is replacing this by calls to the DBUS a viable alternative?

jan-cerny:
That's great!

src/OVAL/probes/unix/linux/systemdunitproperty.c
jan-cerny: I think we  have to implement something filesystem based.

We already do something like that eg. in process probe, where we read form '/proc/' directory to get attributes of processes.

You can try to have a look how systemd implements `systemctl list-unit-files`.

From a quick look I found this:
https://github.com/systemd/systemd/blob/fbd0b64f44a7bae678137d67ad5da202f6b8d809/src/systemctl/systemctl.c

However I don't know if their code can be reused, that needs to be investigated.

matejak:
Code in this PR can't be used for the task, because we can't call `systemctl`. The proper way is to do the same as `systemd`, preferably using `systemd` routines.

PR #923, opened 2017-12-06
Added stig viewer tests

matejak:
The `--stig-viewer` option got its test case - the output is compared to a known output that is OK by a Python script.


jan-cerny:
Hi. I have run the test and I have reviewed this pull request. It looks good to me. Thank you for your awesome contribution!

mpreisler:
This PR introduced a new required dependency on a maintenance branch. We don't use python for tests and this PR introduces that.

PR #927, opened 2017-12-08
Patches to build on Windows using Visual Studio without probes

jan-cerny:
~~This is a work in progress.~~

This pull request will introduce patches that will slowly lead us to Microsoft Visual Studio. Those patches address portability issues, namely difference between Windows and POSIX environments, lack of support of C99 features in Visual Studio compiler, different libraries.

For details about changes please see the respective commit messages.

~~This cannot be merged before #925 is merged.~~ EDIT: it has already been merged


CMakeLists.txt
```diff
@@ -251,7 +251,9 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 configure_file("config.h.in" "config.h")
 add_definitions(-DHAVE_CONFIG_H)
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pipe -std=c99 -W -Wall -Wnonnull -Wshadow -Wformat -Wundef -Wno-unused-parameter -Wmissing-prototypes -Wno-unknown-pragmas -D_GNU_SOURCE -D_POSIX_C_SOURCE=200112L")
+if (${CMAKE_C_COMPILER_ID} STREQUAL "GNU")
+	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pipe -std=c99 -W -Wall -Wnonnull -Wshadow -Wformat -Wundef -Wno-unused-parameter -Wmissing-prototypes -Wno-unknown-pragmas -D_GNU_SOURCE -D_POSIX_C_SOURCE=200112L")
```
mpreisler: do we want to do something similar for MSVC but with MSVC specific flags?

jan-cerny: @mpreisler I don't know yet, MSVC has a lot of options, but so far I haven't identified any that we need.

jan-cerny:
I have added more patches. This PR needs to be rebased on master after #925 is merged.

mpreisler:
#925 has been merged, please rebase.

jan-cerny:
I have rebased this.

jan-cerny:
I need to include a patch for CMakeLists that will build libopenscap on Windows as a static library.

mpreisler:
@jan-cerny don't do that, instead fix the dynamic lib linking

jan-cerny:
e361c37f55fa1c039f5a2f4616d489e67bac9ae6 is wrong fix, it doesn't build on Linux. I need to come with a better fix for the semicolon issue.

jan-cerny:
I have removed e361c37 with wrong fix from this PR.
I have replaced it by d435002.

jan-cerny:
@mpreisler OK, I'll look into this issue.

CMakeLists.txt
```diff
@@ -264,7 +265,9 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 configure_file("config.h.in" "config.h")
 add_definitions(-DHAVE_CONFIG_H)
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pipe -std=c99 -W -Wall -Wvla -Wnonnull -Wshadow -Wformat -Wundef -Wno-unused-parameter -Wmissing-prototypes -Wno-unknown-pragmas -D_GNU_SOURCE -D_POSIX_C_SOURCE=200112L")
+if (${CMAKE_C_COMPILER_ID} STREQUAL "GNU")
```
mpreisler: doesn't clang have the same flags?

jan-cerny:
@openscap-ci test this please

mpreisler:
Please rebase, OSCAP_API has been merged.

jan-cerny:
I have rebased this PR on master branch.

mpreisler:
@jan-cerny what's missing in this PR before the "do not merge" label is removed?

jan-cerny:
@mpreisler I am not sure if it's correct to use `#ifdef _WIN32` everywhere, if it shouldn't be `#ifdef _MSC_VER` instead , to not break cross compilation. But I haven't tried that yet.

mpreisler:
> @mpreisler I am not sure if it's correct to use #ifdef _WIN32 everywhere, if it shouldn't be #ifdef _MSC_VER instead , to not break cross compilation. But I haven't tried that yet.

Both have different meaning. If it's Windows specific code use _WIN32, if it's code that only works on MSVC use _MSC_VER. Perhaps I am misunderstanding the question?

jan-cerny:
@mpreisler The problem is that I need to verify if I didn't break cross compilation by changes proposed in this pull request. I think that there can be some problems related to `#ifdef _WIN32`. But I haven't tried that yet.

jan-cerny:
I have verified that this PR does not break cross compilation on F27.

Now I need to resolve conflicts with the master branch,

jan-cerny:
I have resolved a conflict with the base branch and I think I should remove "do not merge" labels.

src/common/util.h
```diff
@@ -36,7 +36,12 @@
 #define __attribute__nonnull__(x) assert((x) != NULL)
 #endif
 
-/*
+#ifdef _MSC_VER
+#define OSCAP_HIDDEN_START
```
mpreisler: when we now have OSCAP_API does it even make sense to keep these macros? can't we just replace everything with OSCAP_API? If there is no OSCAP_API the symbol would be invisible.

jan-cerny: Great idea! We can do it according to https://gcc.gnu.org/wiki/Visibility

However, on GCC we now use `_Pragma("GCC visibility push(hidden)")` , and the article says:

> #pragma GCC visibility is stronger than -fvisibility; it affects extern declarations as well. -fvisibility only affects definitions, so that existing code can be recompiled with minimal changes. This is more true for C than C++; C++ interfaces tend use classes, which are affected by -fvisibility. 

Do you think it is a problem?

src/common/util.h
```diff
@@ -444,4 +451,41 @@ OSCAP_API char *oscap_realpath(const char *path, char *resolved_path);
  */
 OSCAP_API char *oscap_basename(char *path);
 
+/**
+ * Return directory component of a path
+ * @param path path
+ * @return dirname component of path
```
mpreisler: pretty bad docs... do I free what it returns? does it dupe the input or does it reuse it and destroy it?

src/common/util.h
mpreisler: Yes, it is. My suggestion is to get rid of OSCAP_HIDDEN_START everywhere. You can use OSCAP_API even for GCC to make symbols visible.

jan-cerny: @mpreisler Yes, I understand that. I meant: Do you think that what the quotation from the article says is a problem?

src/common/util.h
jan-cerny: OK

jan-cerny:
* I have improved the comment for `oscap_basename` and `oscap_dirname`.
* I have found that there was `OSCAP_API` at some places where it cannot be. Adding this macro in an automated way was a huge mistake. I have removed the problematic occurrences of `OSCAP_API`.
* `OSCAP_API` now works with GCC as well.
* `OSCAP_HIDDEN_START` and `OSCAP_HIDDEN_END` don't exist anymore.

mpreisler:
ACK

PR #928, opened 2017-12-08
Use oscap_strdup instead of just strdup

mpreisler:
strdup is not a C standard function, it is different on different platforms. On Windows you are supposed to use _strdup instead. Therefore we have a small wrapper called `oscap_strdup`. However we are not using that wrapper ocnsistently and in many cases we still use strdup directly. This causes a lot of issues with @jan-cerny 's work on the Windows port.

I will not be fixing strdup usage in probes in this PR. That will have to be done in a different PR.


jan-cerny:
Aha
I haven't seen this

:1st_place_medal:

jan-cerny:
Great contribution! Thank you very much!

PR #929, opened 2017-12-08
Use oscap_strdup everywhere.

jan-cerny:
MSVC deprecates `strdup()` and recommends `_strdup()` instead.
To handle this situation we already have `oscap_strdup()`. However we didn't use `oscap_strdup()` everywhere, but there were still some occurrences of `strdup()`. This pull request replace remaining occurrences of `strdup()` to `oscap_strdup()`.

We will need to use `oscap_strdup()` also in probes, but if it remains hidden and static inline, it will not be visible symbol in libopenscap.so and probes will not be able to find it. That would cause many linker errors when building probes. I have seen them.

For more details please see the respective commit messages.

### DIFF
--- a/src/OVAL/oval_state.c
+++ b/src/OVAL/oval_state.c
@@ -154,7 +154,7 @@ struct oval_state *oval_state_new(struct oval_definition_model *model, const cha
 	state->operator = OVAL_OPERATOR_UNKNOWN;
 	state->subtype = OVAL_SUBTYPE_UNKNOWN;
 	state->comment = NULL;
-	state->id = strdup(id);
+	state->id = oscap_strdup(id);
 	state->notes = oval_collection_new();
 	state->contents = oval_collection_new();
 	state->model = model;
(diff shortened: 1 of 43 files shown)


mpreisler:
lol

#928 

So this needs to go into maint-1.2 IMO, at least most of it.

src/OVAL/oval_state.c
```diff
@@ -219,15 +219,15 @@ void oval_state_set_subtype(struct oval_state *state, oval_subtype_t subtype)
...
+	oval_collection_add(state->notes, (void *)oscap_strdup(notes));
 }
 
 void oval_state_set_comment(struct oval_state *state, char *comm)
 {
 	__attribute__nonnull__(state);
 	if (state->comment != NULL)
 		free(state->comment);
-	state->comment = comm == NULL ? NULL : strdup(comm);
+	state->comment = comm == NULL ? NULL : oscap_strdup(comm);
```
mpreisler: oscap_strdup handles NULL already. you can collapse the logic as I do in the other PR #928

mpreisler:
As a side-note I think you overdid it a little bit (a lot actually) with the amount of commits in this PR.

PR #930, opened 2017-12-08
Introduce oscap_realpath to improve code portability

jan-cerny:
Windows doesn't know `realpath()`. Instead they have a similar function `_fullpath()`. This patch introduces a wrapper function `oscap_realpath()` that wraps either `realpath()` or `_fullpath()` depending on the platform. The PR also replaces every `realpath()` by `oscap_realpath()`.


mpreisler:
ACK

PR #932, opened 2017-12-09
 Introduce oscap_basename

jan-cerny:
We use `basename()` on a few places across our codebase. However on Visual Studio there is no such function. Most similar functions are `_splitpath()` or `_splitpath_s()` that can break pathname into components. To avoid many `#ifdef` section introduce `oscap_basename()`.

Then it replaces all occurences of `basename()` by `oscap_basename()`.


mpreisler:
ACK

Issue #909, opened 2017-11-15
Static analysis fixes


matejak:
The static analysis reported some errors ranging from missing includes to errors that could cause crash:
[coverity-warnings.log](https://github.com/OpenSCAP/openscap/files/1474819/coverity-warnings.log)

evgenyz:
We should check is some of these problems still exists. And close bug If they are obsolete.

matejak:
As this is an ongoing code quality effort, I close it as the original issue has been fixed.

PR #933, opened 2017-12-11
Fixed most painful issues found by Coverity scan.

matejak:
Coverity scan of 1.2.16 revealed some problems (see #909 ). This PR aims to fix the most harmful ones.

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -731,7 +731,8 @@ static inline int _xccdf_policy_rule_generate_fix(struct xccdf_policy *policy, s
 		// Ansible is special because we have two output modes, variable and
 		// task mode. In both cases we have to do post processing. 
 
-		ret = _xccdf_policy_rule_generate_fix_ansible(template, output_fd, fix_text, ansible_variable_mode);
+		// We can't recover the return value (it is rewritten by a later call) to _write_fix_footer_to_fd
+		_xccdf_policy_rule_generate_fix_ansible(template, output_fd, fix_text, ansible_variable_mode);
 		free(fix_text);
 	}
 	else {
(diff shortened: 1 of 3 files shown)


jan-cerny:
Great :+1:

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -731,7 +731,8 @@ static inline int _xccdf_policy_rule_generate_fix(struct xccdf_policy *policy, s
 		// Ansible is special because we have two output modes, variable and
 		// task mode. In both cases we have to do post processing. 
 
-		ret = _xccdf_policy_rule_generate_fix_ansible(template, output_fd, fix_text, ansible_variable_mode);
+		// We can't recover the return value (it is rewritten by a later call) to _write_fix_footer_to_fd
+		_xccdf_policy_rule_generate_fix_ansible(template, output_fd, fix_text, ansible_variable_mode);
```
yuumasato: Shouldn't we do the same handling as in line740 below and return if `ret !=0`? As the code which was taken out into `_xccdf_policy_rule_generate_fix_ansible` was directly returning.

utils/oscap-cvrf.c
```diff
@@ -151,7 +151,9 @@ static int app_cvrf_export(const struct oscap_action *action) {
 		if (oscap_err())
 			fprintf(stderr, "%s %s\n", OSCAP_ERR_MSG, oscap_err_desc());
 
-	oscap_source_free(import_source);
+	/* TODO: Refactor, cvrf_index_parse_xml (called by oscap_source_new_from_file) frees its argument as an unexpected side-effect.
+	 * oscap_source_free(import_source);
+	 */
```
yuumasato: Maybe intention in https://github.com/OpenSCAP/openscap/blob/2265c50484f76baf91e4420b247e8e04a0c34cf9/src/CVRF/cvrf_priv.c#L2335 was to `free(buffer)`, allocated by `oscap_source_get_raw_memory()`?

I don't understand where `buffer`  is used in [`cvrf_index_parse_xml`](https://github.com/OpenSCAP/openscap/blob/2265c50484f76baf91e4420b247e8e04a0c34cf9/src/CVRF/cvrf_priv.c#L2323).
Maybe to parse content put in `buffer`? Where does parsing occur?

PR #934, opened 2017-12-13
Add OSCAP_API macro

jan-cerny:
OSCAP_API is a macro that marks every public function in OpenSCAP API.
Functions marked with OSCAP_API will be exported to the shared library
interface.

### DIFF
--- a/src/OVAL/probes/SEAP/public/sm_alloc.h
+++ b/src/OVAL/probes/SEAP/public/sm_alloc.h
@@ -26,6 +26,7 @@
 
 #include "config.h"
 #include "src/common/debug_priv.h"
+#include "oscap_export.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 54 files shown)


matejak:
Seems OK to me, but e.g. http://lucumr.pocoo.org/2013/8/18/beautiful-native-libraries/ suggests to 
`#define OSCAP_EXPORT __attribute__((visibility("default")))` for `gcc>=4.0`, some libraries define the analogous macro to `extern` on Apple platforms.

jan-cerny:
@matejak Thank you for your suggestion. Different compilers have different macros. We can improve the definitions in oscap_export.h easily in future.

src/common/public/oscap_export.h
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * OSCAP_EXPORT macro
+ *
+ * OSCAP_EXPORT is a macro that marks every public function in OpenSCAP API.
+ * Functions marked with OSCAP_EXPORT will be exported to the shared library
+ * interface.
+ */
+#if defined(_WIN32) && defined(_MSC_VER)
+#define OSCAP_EXPORT __declspec(dllexport)
```
mpreisler: this won't work, you need to change between dllexport and dllimport based on whether you are building the lib or building something using it.

that's also why most people choose macro name such as OSCAP_API and not OSCAP_EXPORT, what the macro does changes depending on whether you are building DLL or using its headers and linking to it.

jan-cerny:
The problem is that this broke swig bindings when I run it on Fedora.
The issue is:

```
[ 80%] Swig source
/home/jcerny/openscap/src/common/../../src/common/public/oscap.h:62: Error: Syntax error in input(1).
make[2]: *** [swig/python2/CMakeFiles/_openscap_py.dir/build.make:64: swig/python2/openscapPYTHON_wrap.c] Error 1
make[1]: *** [CMakeFiles/Makefile2:2988: swig/python2/CMakeFiles/_openscap_py.dir/all] Error 2
make: *** [Makefile:163: all] Error 2
[jcerny@thinkpad build{dynamic_library}]$ 
```

It mentions line 62 in `oscap.h`, which is:

```
OSCAP_EXPORT void oscap_init(void);
```

From this I think that the macros before functions break SWIG.

src/common/public/oscap_export.h
jan-cerny: We don't build anything that is using it.

mpreisler:
With smart macro conditions you can get around the swig error as well as the issue I mentioned.

Most projects approach it like this, for DLL build they define a macro (via CLI arguments) called OSCAP_LIBRARY_BUILD, then in oscap_export.h they ifdef and if this macro is defined they declspec export, if it's not defined they declspec import.

Add a macro condition that detects that swig is generating from the header. See http://www.swig.org/Doc1.3/Preprocessor.html

If SWIG is generating define OSCAP_API as empty.

jan-cerny:
@mpreisler OK, thank you for your suggestion. I have discussed this with @matejak and we agreed we should have a similiar definition like in http://lucumr.pocoo.org/2013/8/18/beautiful-native-libraries/

At the same time, I wonder why it breaks swig, because I run that on Fedora, so the macro OSCAP_EXPORT should be an empty string on Fedora now.

jan-cerny:
I have changed the patch according to feedback.

I have also solved the problem of broken SWIG. Firstly I thought it would be fixed by adding some `include` to "*.i" file, but I tried various options and none of them helped. Finally I found that it's necessary to add just a definition of OSCAP_API to the "*.i" file.

jan-cerny:
Thanks @matejak for the great article, it really helped a lot :+1:

src/OVAL/probes/SEAP/public/sm_alloc.h
```diff
@@ -35,36 +36,36 @@ extern "C" {
 
 #if !defined(HAVE_POSIX_MEMALIGN)
 # if defined (HAVE_MEMALIGN)
-extern int posix_memalign (void ** __memptr, size_t __alignment, size_t __size);
+OSCAP_API extern int posix_memalign (void ** __memptr, size_t __alignment, size_t __size);
```
mpreisler: woah, are you sure about this?

mpreisler: actually I myself am not sure but I find it suspicious that something this low level is exported, it's also something that is defined in other SOs with the same symbol name so this might cause clashes, and it's defined as extern. I am basically asking if you had good reasons to explicitly export it.

jan-cerny: That's because this is a header file in a public directory. I haven't noticed it's in rbt_common.h. Are there some public headers that aren't public?

jan-cerny: I don't know. I think we will need to review symbols in the public API before we release 1.3.0, because there can be more functions in public files that should be private.

For now I'll add a patch that removes OSCAP_API from this line. I don't want to do bigger changes of header files in this PR, because it's already a big PR.

jan-cerny:
I have added a patch that removes OSCAP_API from posix_menalign.

jan-cerny:
@openscap-ci test this please

jan-cerny:
sm_alloc.h is completely weird file, maybe it should be private everything in sm_alloc.h

mpreisler:
ACK

Further refinement can happen later IMO.

PR #940, opened 2017-12-15
Remove OSCAP_GSYM macro

jan-cerny:
OSCAP_GSYM is a macro that adds prefix `___G_` to its argument. It's used for marking some global variables. I haven't found any reason for using it. In my opinion this macro makes code less readable and complicates debugging.

### DIFF
--- a/src/OVAL/oval_probe.c
+++ b/src/OVAL/oval_probe.c
@@ -50,7 +50,7 @@
 #include "oval_probe_ext.h"
 #include "collectVarRefs_impl.h"
 
-oval_probe_meta_t OSCAP_GSYM(__probe_meta)[] = {
+oval_probe_meta_t __probe_meta[] = {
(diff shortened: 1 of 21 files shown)


mpreisler:
I thought the macro was for prefixing to avoid clashes. The variables it prefixes are shared globally and not just in the local compile unit.

mpreisler:
@jan-cerny I don't feel strongly about this change. I definitely don't like removing the prefixing on global symbols, that just increases chances of weird clashes.

Whether the macro stays or not I think we should keep the prefixes.

jan-cerny:
@mpreisler I wanted to remove this because when I saw the macro for a first time, I didn't know what it means and I had to investigate it to understand it. After I removed it, they are only normal global variables.

mpreisler:
I will reiterate:

> Whether the macro stays or not I think we should keep the prefixes.

jan-cerny:
The docs are written in https://github.com/OpenSCAP/openscap/pull/961

PR #943, opened 2017-12-19
Add remediation tests.

matejak:
This PR adds tests to check that remediation roles have:
- [x] profile metadata (title and description).

### DIFF
--- a/tests/API/XCCDF/unittests/test_remediation_simple.sh
+++ b/tests/API/XCCDF/unittests/test_remediation_simple.sh
@@ -8,9 +8,11 @@ result=$(mktemp -t ${name}.out.XXXXXX)
 stderr=$(mktemp -t ${name}.out.XXXXXX)
 ret=0
 
+input_xml="$srcdir/${name}.xccdf.xml"
+
 rm -f test_file
 # First make sure that this OVAL fails, then scan again with --remediate
-$OSCAP xccdf eval --results $result $srcdir/${name}.xccdf.xml 2> $stderr || ret=$?
+$OSCAP xccdf eval --results $result "$input_xml" 2> $stderr || ret=$?
 [ $ret -eq 2 ]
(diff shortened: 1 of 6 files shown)


tests/API/XCCDF/unittests/test_remediation_simple.sh
```diff
@@ -43,4 +45,19 @@ assert_exists 1 '//score'
 assert_exists 1 '//score[text()="100.000000"]'
 
 rm test_file
-rm $result
+
+$OSCAP xccdf generate fix --fix-type bash --profile '' --output "$result" "$input_xml" 2> $stderr
```
yuumasato: We should generate remediation role for specific Profile.
I guess this only works because Rule `xccdf_moc.elpmaxe.www_rule_1` is selected by default.

yuumasato: It is probably fitting into `default` Profile, but than it is strange that is outputs title and description for Profile `xccdf_moc.elpmaxe.www_profile_common`

mpreisler: If there is only one profile then "" will only match one profile and it will all work. I agree that listing the profile or at least the "common" suffix would improve readability here.

matejak: I have added another profile to the XML as suggested by @yuumasato and modified the oscap call. I don't see the reasons why to add another test - I guess that the test is still pretty minimalistic.

mpreisler:
> I think there should be new test for this, hijacking test_remediation_simple may not be good.

100% agree. Plus our long term goal is to improve testing of openscap by splitting up the mega tests.

matejak:
OK, I have split this test, so there are new files and the old ones are untouched.

PR #944, opened 2018-01-03
Make command in bash role header in line with --help

dahaic:
Point is to make more explicit it's a placeholder name - using
the same name --help use might do the trick.


matejak:
Thank you for your contribution! We have agreed that this change, although small, is beneficial to the project.

PR #948, opened 2018-01-08
Split test_mitre.sh

mpreisler:
Split the one test into 4:
- test linux probes
- test ind probes
- test unix probes
- test OVAL features

Unfortunately the linux probes test takes by far the most time on most machines because of rpmverifyfile. Still, I think this is a step in the right direction.

PR #947, opened 2018-01-08
Split test_ds.sh

mpreisler:
Some of the slowest and largest tests in our test suite. Large tests are hard to fix because re-running them takes forever. Furthermore if something breaks it is harder to figure out why because the test tests everything imaginable...

before:
```
~/d/openscap/build $ ctest -j 4 -R DS/
Test project /home/mpreisle/d/openscap/build
    Start 34: DS/test_ds.sh
    Start 36: DS/signed/all.sh
    Start 37: DS/validate/all.sh
    Start 35: DS/ds_sds_index/all.sh
1/4 Test #35: DS/ds_sds_index/all.sh ...........   Passed    0.30 sec
2/4 Test #37: DS/validate/all.sh ...............   Passed    1.70 sec
3/4 Test #36: DS/signed/all.sh .................   Passed    2.06 sec
4/4 Test #34: DS/test_ds.sh ....................   Passed  103.34 sec

100% tests passed, 0 tests failed out of 4

Total Test time (real) = 103.35 sec

```

after:
```
~/d/openscap/build $ ctest -j 4 -R DS/
Test project /home/mpreisle/d/openscap/build
    Start 36: DS/test_sds_compose_split.sh
    Start 35: DS/test_rds.sh
    Start 34: DS/test_ds_misc.sh
    Start 37: DS/test_sds_eval.sh
1/9 Test #37: DS/test_sds_eval.sh ...............   Passed   17.41 sec
    Start 41: DS/signed/all.sh
2/9 Test #34: DS/test_ds_misc.sh ................   Passed   20.01 sec
    Start 42: DS/validate/all.sh
3/9 Test #41: DS/signed/all.sh ..................   Passed    2.71 sec
    Start 38: DS/test_sds_fix_from_results.sh
4/9 Test #38: DS/test_sds_fix_from_results.sh ...   Passed    2.01 sec
5/9 Test #42: DS/validate/all.sh ................   Passed    2.60 sec
    Start 39: DS/test_sds_fix_from_source.sh
    Start 40: DS/ds_sds_index/all.sh
6/9 Test #40: DS/ds_sds_index/all.sh ............   Passed    0.40 sec
7/9 Test #39: DS/test_sds_fix_from_source.sh ....   Passed    0.60 sec
8/9 Test #35: DS/test_rds.sh ....................   Passed   57.11 sec
9/9 Test #36: DS/test_sds_compose_split.sh ......   Passed   75.90 sec

100% tests passed, 0 tests failed out of 9

[...]

PR #949, opened 2018-01-08
Use ctest logs instead of separate log files for testing

mpreisler:
This is mostly for discussion at this point. Our test "framework" pipes stderr output from tests into log files. The user is expected to look into that log file to figure out why a test failed. ctest is setup to gather output from all tests and it can be set to output the entire test output when an error is encountered via env var CTEST_OUTPUT_ON_FAILURE. I think that may be more usable for developers than having to go on jenkins and look for log files.

If we get consensus we need to get rid of the test_init calls completely or at least remove the log file name argument before merging.

### DIFF
--- a/tests/API/CCE/test_api_cce.sh
+++ b/tests/API/CCE/test_api_cce.sh
@@ -46,7 +46,7 @@ function test_api_cce_search_non_existing {
 }
 
 # Testing.
-test_init "test_api_cce.log" 
+test_init
 
 if [ -z ${CUSTOM_OSCAP+x} ] ; then
     test_run "test_api_cce_smoke" test_api_cce_smoke
--- a/tests/API/CPE/dict/test_api_cpe_dict.sh
+++ b/tests/API/CPE/dict/test_api_cpe_dict.sh
@@ -115,7 +115,7 @@ function test_api_cpe_dict_import_official_v23(){
 
 # Testing.
 
-test_init "test_api_cpe_dict.log"
+test_init
 
 if [ -z ${CUSTOM_OSCAP+x} ] ; then
     test_run "test_api_cpe_dict_smoke" test_api_cpe_dict_smoke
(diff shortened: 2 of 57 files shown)


matusmarhefka:
+1 for this. It is very annoying that there is only PASS/FAIL result stated in the final test output for each test case. It would be great to append content of `CTEST_OUTPUT_ON_FAILURE` into the final test output so one does not need to connect to jenkins and look for specific log when some test fails/errors.

matejak:
I recall that when tests on RHEL6, the output was more informative. However, in the long run, I definitely vote for outsourcing some of our functionality to external test helpers s.a. `ctest`.

mpreisler:
> I recall that when tests on RHEL6, the output was more informative.

Not exactly sure what you mean by this, the output is exactly the same, the only difference is that it is not longer piped into separate files but into `ctest` instead.

matejak:
I just wanted to point out that ther [RHEL6 test run output](https://jenkins.open-scap.org/view/Pull%20Request%20Jobs/job/openscap-pull-requests/423/consoleFull) was more informative than [Fedora test run output](https://jenkins.open-scap.org/view/Pull%20Request%20Jobs/job/openscap-pull-requests/421/consoleFull). Although one still needed logs for complete tests diagnostics, the report was more granular.

mpreisler:
Ah, I see what you mean. Splitting up tests and using ctest brings us closer to rhel6 times but not quite there. And I do agree that the new automake output sucks for tests.

jan-cerny:
This is really great and I like that we leverage the power of CMake.

Also, in future, we could split the huge tests, `all.sh`, to many small separate test cases.

mpreisler:
> Also, in future, we could split the huge tests, all.sh, to many small separate test cases.

No question about that, that is already happening. #947 #948

mpreisler:
I think we have consensus here, so I will go ahead and fix up the `test_init` calls, then this can be merged.

mpreisler:
This is now ready for review and hopefully merging.

jan-cerny:
Looks good to me.

I think we should add `--output-on-failure` to `ctest` on Jenkins, which will show the output of failed tests in Jenkins log.

PR #951, opened 2018-01-09
Update cpe_name.h (CPE name patch retarget)

mpreisler:
This is #824 except it targets the master branch. Everything else is the same.


jan-cerny:
Thanks for opening this in the master branch. I agree that main-1.2 was not a suitable place.

@openscap-ci test this please

mpreisler:
@openscap-ci test this please

PR #952, opened 2018-01-09
Update oscap-ssh to support --oval-results

### LINKED ISSUE #863: [RFE] Support --oval-results in oscap-ssh
SSG Test Suite relies on `oscap-ssh` command and currently it doesn't get the OVAL results file from remote machine.
The OVAL results file would help debug OVAL definitions, tests scenarios, and remediations tested in SSG Test Suite.

dahaic:
With this switch, oscap-ssh will run oscap with --oval-results
and download all generated oval xmls to current directory, mimicking
behaviour of oscap tool itself.

Scan on remote machine is now triggered in temp directory to not
leave artifacts in $HOME.

Fixes #863

### DIFF
--- a/utils/oscap-ssh
+++ b/utils/oscap-ssh
@@ -42,7 +42,7 @@ function usage()
     echo "  --tailoring-file"
     echo "  --tailoring-id"
     echo "  --cpe (external OVAL dependencies are not supported yet!)"
-    # --oval-results, --sce-results and --check-engine-results are not supported
+    # --sce-results and --check-engine-results are not supported
     # use --results-arf instead
     echo "  --oval-results"
     echo "  --results"
@@ -226,6 +226,9 @@ if [ "$LOCAL_DIRECTIVES_PATH" != "" ]; then
 fi
 
 echo "Starting the evaluation..."
+# changing directory because of --oval-results support. oval results files are
+# dumped into PWD, and we can't be sure by the file names - we need controlled
+# environment
 ssh -o ControlPath="$MASTER_SOCKET" $SSH_ADDITIONAL_ARGS -p "$SSH_PORT" "$SSH_HOST" "cd $REMOTE_TEMP_DIR; $OSCAP_SUDO oscap ${args[*]}"
 OSCAP_EXIT_CODE=$?
 echo "oscap exit code: $OSCAP_EXIT_CODE"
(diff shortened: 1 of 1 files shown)


utils/oscap-ssh
```diff
@@ -238,6 +243,9 @@ fi
 if [ "$TARGET_SYSCHAR" != "" ]; then
     scp -o ControlPath="$MASTER_SOCKET" -P "$SSH_PORT" "$SSH_HOST:$REMOTE_TEMP_DIR/syschar.xml" "$TARGET_SYSCHAR" || die "Failed to copy the OVAL syschar file back to local machine!"
 fi
+if [ "$OVAL_RESULTS" == "yes" ]; then
+    scp -o ControlPath="$MASTER_SOCKET" -P "$SSH_PORT" "$SSH_HOST:$REMOTE_TEMP_DIR/*-oval.xml.result.xml" "./" || die "Failed to copy OVAL result files back to local machine!"
```
mpreisler: oval file names don't necessarily have to end with `-oval.xml`, I suggest to instead use `*.result.xml`. Technically it doesn't even have to have the xml extension.

dahaic: I started with just result.xml, it felt too vague, so I added more - you are right though. SSG naming convention is not something to hardcode in oscap.

utils/oscap-ssh
```diff
@@ -221,7 +226,7 @@ if [ "$LOCAL_DIRECTIVES_PATH" != "" ]; then
 fi
 
 echo "Starting the evaluation..."
-ssh -o ControlPath="$MASTER_SOCKET" $SSH_ADDITIONAL_ARGS -p "$SSH_PORT" "$SSH_HOST" "$OSCAP_SUDO oscap ${args[*]}"
+ssh -o ControlPath="$MASTER_SOCKET" $SSH_ADDITIONAL_ARGS -p "$SSH_PORT" "$SSH_HOST" "cd $REMOTE_TEMP_DIR; $OSCAP_SUDO oscap ${args[*]}"
```
yuumasato: A comment on why we change directory would be nice.

utils/oscap-ssh
```diff
@@ -44,6 +44,7 @@ function usage()
     echo "  --cpe (external OVAL dependencies are not supported yet!)"
     # --oval-results, --sce-results and --check-engine-results are not supported
```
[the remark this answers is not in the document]
dahaic: fixed

utils/oscap-ssh
dahaic: good point :) fixed

yuumasato:
Ack, thanks.

PR #955, opened 2018-01-12
CVE: IDs are now 4-7 digits long

vbatts:
Reviewing the NVD feeds from
https://nvd.nist.gov/vuln/data-feeds#CVE_FEED reflects like
`CVE-2016-1000031`, `CVE-2017-18016` and `CVE-2017-1000439`.

I'm sure this xsd needs to be vetted in other places, but currently
`oscap cve validate ./nvdcve-2.0-modified.xml` fails with

```shell
File './nvdcve-2.0-modified.xml' line 65535: Element '{http://scap.nist.gov/schema/vulnerability/0.4}cve-id': [facet 'pattern'] The value 'CVE-2017-1000367' is not accepted by the pattern 'CVE-([1,2])\d{3}-\d{4}'.
File './nvdcve-2.0-modified.xml' line 65535: Element '{http://scap.nist.gov/schema/vulnerability/0.4}cve-id': 'CVE-2017-1000367' is not a valid value of the atomic type '{http://scap.nist.gov/schema/cve/0.1}cveNamePatternType'.
```

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

PR #958, opened 2018-01-15
Cve numbers

matejak:
Retarget of #955

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -63,6 +63,7 @@ Steve Grubb <sgrubb@redhat.com>
 Tomas Heinrich <theinric@redhat.com>
 Trey Henefield <trey.henefield@ultra-ats.com>
 V. Vinay <vvinay@hpe.com>
+Vincent Batts <vbatts@hashbangbash.com>
 Watson Sato <wsato@redhat.com>
 Watson Yuuma Sato <wsato@redhat.com>
 Wesley Ceraso Prudencio <wcerasop@redhat.com>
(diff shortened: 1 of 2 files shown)


mpreisler:
Hmm, isn't this XSD coming from an official source? I don't think we should retroactively change an old official CVE XSD.

Also consider that there are tons of CVE patterns all over the codebase. We should probably review them all to make sure everything is in check.

mpreisler:
> Hmm, isn't this XSD coming from an official source? I don't think we should retroactively change an old official CVE XSD.

It seems the CVE XSD is not official so I think we can change it.

vbatts:
@mpreisler i wondered the same and search around, but couldn't determine

jan-cerny:
@vbatts @mpreisler I have found this file on NIST web https://nvd.nist.gov/schema/
I'm not sure if it's official.

redhatrises:
There appears to be an update to this file at https://nvd.nist.gov/schema/cve_0.1.1.xsd. Not sure if there is an even newer version than what I found.

mpreisler:
@matejak I'd prefer to use the official update than change it ourselves, do you agree?

vbatts:
oh cool. I can open that PR

vbatts:
though their https://scap.nist.gov/schema/nvd/vulnerability_0.4.xsd still references the 0.1 version

matejak:
@mpreisler I did some research that indicated that fix submitted by @vbatts was right on the money, but I see that it is more complex. So go ahead - you know much more about how to do this right.

mpreisler:
I am merging this because this is an improvement (IMO) and doesn't break anything. Adding the new official CVE XSD would be cleaner but all sorts of other XSDs still use the old one so the other option would be to patch many other XSDs to reference the new one.

vbatts:
sounds good. When I said I would open a PR, now I'm finding that there is
references that I would want some input on, before shoving them in.
I'lll hold off on putting the cve-0.1.1.xsd in.

PR #962, opened 2018-01-18
Add pretask to ansible roles to check the ansible version

redhatrises:
Asserts that the ansible version is greater than or equal to 2.3 for the ansible roles generation.


mpreisler:
There is nothing in OpenSCAP that requires Ansible 2.3 or higher AFAIK. Why did you choose 2.3?

redhatrises:
Yeah... you are right... I was thinking that since SSG needs ansible>=2.3, and OpenSCAP generates the ansible playbooks for SSG that it would make sense to add this pretask here in OpenSCAP. But, maybe it doesn't make sense.

redhatrises:
Closing as @mpreisler comments are spot on.

Issue #954, opened 2018-01-12
oval results missing in html results for many sysctl checks


ghost:
#### Description of problem:
Running oscap xccdf eval with the following command line results in missing oval results for some checks:
`oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_nist-800-171-cui --oval-results --report remote-report.html /usr/share/xml/scap/ssg/content/ssg-centos7-ds.xml`

This happens whether the checks pass or fail.

The checks with no oval results are:
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_default_send_redirects
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_all_send_redirects
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_ip_forward
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_all_accept_source_route
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_all_accept_redirects
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_all_secure_redirects
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_all_log_martians
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_default_log_martians
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_default_accept_source_route
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_default_accept_redirects
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_default_secure_redirects
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_icmp_echo_ignore_broadcasts
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_icmp_ignore_bogus_error_responses
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_tcp_syncookies
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_all_rp_filter
xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_conf_default_rp_filter

xccdf_org.ssgproject.content_rule_sysctl_kernel_ipv6_disable
xccdf_org.ssgproject.content_rule_sysctl_net_ipv6_conf_all_accept_source_route
xccdf_org.ssgproject.content_rule_sysctl_net_ipv6_conf_all_accept_ra
[...]

PR #964, opened 2018-01-19
[WIP] OVAL details in HTML report fixes

### LINKED ISSUE #916: HTML report doesn't show OVAL details for a specific rule
When I run this command:
```
$ oscap xccdf eval --profile xccdf_scap-workbench_profile_customized \
--results results.xml --oval-results --report report.html \
--tailoring-file ssg-rhel7-ds-tailoring.xml \
/usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml
```
I got a HTML report with no OVAL details for the rule `xccdf_org.ssgproject.content_rule_sysctl_kernel_dmesg_restrict`

That is a bug, because:
* I have specified `--oval-results` which implies OVAL details in HTML report
* A collected item is collected in the oval results file, yet this item **isn't** displayed in HTML report.

screenshot from 2017-11-21 14-12-12

Please notice no OVAL details on the screenshot.

You can download tailoring file here:
[ssg-rhel7-ds-tailoring.xml.zip](https://github.com/OpenSCAP/openscap/files/1491396/ssg-rhel7-ds-tailoring.xml.zip)

You can download my HTML report here:
[report.html.zip](https://github.com/OpenSCAP/openscap/files/1491401/report.html.zip)

I am using  openscap-1.2.14-2.el7.x86_64  and scap-security-guide-0.1.33-6.el7_4.noarch on Red Hat Enterprise Linux 7.4.

Any help would be very appreciated.

### LINKED ISSUE #953: [RFE] Update html output of --oval-results to include extended definitions and AND/OR logic
See https://github.com/OpenSCAP/scap-security-guide/issues/2527 for context:

I found 2 issues with the --oval-results flag that makes it not very useful in troubleshooting.

1. It does not show extended definitions.  So anything failing because RHEL7 is not installed (run on a CentOS system) is not showing that not being on RHEL is the issue.
2. It does not show AND/OR logic, so it will list all failures, some of which look quite contradictory

For example, the xccdf datastream file for the rule 'Set Boot Loader Password' has the following logic:
```
<ns3:criteria operator="OR">
  <ns3:criterion comment="Pass if /boot/grub2/grub.cfg does not exist" test_ref="oval:ssg-test_bootloader_grub_cfg:tst:1" />
   <ns3:criteria operator="AND">
    <ns3:criterion comment="make sure a password is defined in /boot/grub2/grub.cfg" test_ref="oval:ssg-test_bootloader_password:tst:1" />
    <ns3:criterion comment="make sure a superuser is defined in /boot/grub2/grub.cfg" test_ref="oval:ssg-test_bootloader_superuser:tst:1" />
    </ns3:criteria>
    </ns3:criteria>
```
In that case, if /boot/grub2/grub.cfg does not exist the test automatically passes.

However, if one was to just read the html file, it claims that 3 rules failed:
1. /boot/grub2/grub.cfg does not exist
2. make sure a password is defined in /boot/grub2/grub.cfg
3. superuser is defined in /boot/grub2/grub.cfg files. Superuser is not root, admin, or administrator

As you can see, rule 1 and rule 2 contradict each other by simultaneously requiring that the file does exist and does not exist.  Since this is not a quantum system and this is not schrodingers grub, this doesn't make sense until you actually look at the xccdf datastream manually in a text editor and realize the presence of the AND/OR logic.

Further, if there are AND/OR logic plus extended definitions (like the RHEL7 rule), you will have to use a much more manual approach to figure out what was actually causing a rule to fail such as `oscap xccdf eval --verbose INFO` and save the results to a text file and then grep the file for `evaluated as` to see which of several checks failed and caused the overall rule evaluation to fail.  Then you would have to open the xccdf datastream in a text editor and see the specific AND/OR logic to see which of those failures are disregarded or included in the final evaluation.

It would be great if the --oval-results flag would include the AND/OR logic and list extended definitions in the html results so the longer troubleshooting approach would not be necessary.

jan-cerny:
* If an OVAL definition was extended using 'extend_definition', there were
no details shown in the HTML report for the refereced definition(s).
This commit adds OVAL details for the extended definition to HTML report.
Fixes #916, #954.

* This pull request also adds basic AND/OR logic - Fixes #953, but only partially, because it ignores negations.

I need to solve negations before it will be finished.

I also would like to improve it graphically. It is terribly ugly. Do you have any suggestions how to improve it, please?

- [ ] Extended definitions are shown in the report.
- [ ] AND/OR logic:.

Before:

screenshot-2018-1-19 xccdf_org open-scap_testresult_default-profile openscap evaluation report

After:

screenshot-2018-1-19 xccdf_org open-scap_testresult_default-profile openscap evaluation report 1

### DIFF
--- a/xsl/xccdf-report-impl.xsl
+++ b/xsl/xccdf-report-impl.xsl
@@ -674,11 +674,7 @@ Authors:
             </xsl:choose>
         </xsl:variable>
         <span class="label label-default"><abbr title="OVAL details taken from {$details_origin}">OVAL details</abbr></span>
-        <div class="panel panel-default">
-            <div class="panel-body">
-                <xsl:copy-of select="$details"/>
-            </div>
-        </div>
+        <xsl:copy-of select="$details"/>
     </xsl:if>
 </xsl:template>
 
--- a/xsl/xccdf-report-oval-details.xsl
+++ b/xsl/xccdf-report-oval-details.xsl
@@ -44,34 +44,25 @@ Authors:
 
 <xsl:template mode='brief' match='ovalres:oval_results'>
     <xsl:param name='definition-id' />
-    <xsl:param name='result'/>
     <ul>
-        <xsl:apply-templates select='key("oval-definition", $definition-id)' mode='brief'>
-            <xsl:with-param name='result' select='$result'/>
-        </xsl:apply-templates>
+        <xsl:apply-templates select='key("oval-definition", $definition-id)' mode='brief'/>
     </ul>
 </xsl:template>
 
 <xsl:template mode='brief' match='ovalres:extend_definition'>
-    <xsl:param name='result'/>
-    <xsl:apply-templates select='key("oval-definition", @definition_ref)' mode='brief'>
-        <xsl:with-param name='result' select='$result'/>
-    </xsl:apply-templates>
+    <xsl:apply-templates select='key("oval-definition", @definition_ref)' mode='brief'/>
 </xsl:template>
 
 
 <xsl:template mode='brief' match='ovalres:criterion'>
-    <xsl:param name='result'/>
     <li>
         <xsl:apply-templates select='key("oval-test", @test_ref)' mode='brief'>
             <xsl:with-param name='title' select='key("oval-testdef", @test_ref)/@comment'/>
-            <xsl:with-param name='result' select='$result'/>
         </xsl:apply-templates>
     </li>
 </xsl:template>
 
 <xsl:template mode='brief' match='ovalres:criteria'>
-    <xsl:param name='result'/>
     <li>
         <p>
             <span class="label label-info">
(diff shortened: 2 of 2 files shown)


mpreisler:
Looks really cool so far.

jan-cerny:
If an expression is complex, it adds a lot of nested rectangles and a lot of borders.
It could be better visualized as a tree.

jan-cerny:
To reduce the borders, I have tried to show the results as a list, that can contain nested lists. List item is either a logical operator or a OVAL test. I show pass/ fail at every node.

To make it readable at least a little, I make the OVAL items dump hidden by default, and they can be show on a click.
screenshot-2018-2-6 xccdf_org open-scap_testresult_default-profile openscap evaluation report
screenshot-2018-2-6 xccdf_org open-scap_testresult_default-profile openscap evaluation report 1

jan-cerny:
However, for some OVAL rules that are more complex the visualisation is still terribly terrible
screenshot-2018-2-6 xccdf_org open-scap_testresult_xccdf_org ssgproject content_profile_common openscap evaluation report

yuumasato:
In this case particular case I see a few definitions that are often extended, so a graph can be a better representation. It maybe hard to implement though.

jan-cerny:
Hi, The graph is a beautiful idea, i had it already in my mind. The definitions basically form a tree, where root is the definitions, nodes are operators or extended definitions and leafs are OVAL tests. I tried to represeent this hierarchy be the lists that are nested. We can use some CSS+JavaScript to make it look like a true graph, but the structure is already there. 

My concern was that this takes a lot of space and it is hard to orient. But I think a graph would take even more space in the report than this. Also most of the rules consists only of one test, so for them that would be an overkill.

What I think about is something like a clickable tree, that user can expand the nodes that he is interested in. Eg. https://www.jstree.com/demo/

Problem is that forces user to click with mouse even more. We need to find a solution that scales from the 1-test rules to complex rules with many tests. If we hide passed tests and show only failed tests it would not help in the example above. Maybe the extend_definitions should be collapsed by default.
We can also set some threshold on number of tests, eg. 6, if this threshold is exceeded all the tree nodes would be collapsed.

yuumasato:
The clickable tree is interesting.
The threshold for expanding the tree is interesting also.
About hiding passed tests, I think they should be just collapsed.

And another idea, if an OR operator results in PASS, it could be collapsed, even if some of the OR'ed definitions failed the final result is PASS.
This could save some space, in particular case above, the "64bit architecture" definitions that failed but are under an OR operator that resulted in PASS would be collapsed.

redhatrises:
@jan-cerny I like your idea about having a collapsible tree. I also like @yuumasato's idea about collapsing the OVAL result if it passes and expand the failing ones.

matejak:
test this please

matejak:
test this please

matejak:
test this please

mpreisler:
Something like http://js.cytoscape.org/demos/images-breadthfirst-layout/

It's not actually a tree, it's a digraph, you can represent it with input/output edges. It should be simple to dedup the nodes even at the stage of generating the HTML from the XSLT template.

PR #965, opened 2018-01-19
Enable Ninja Build

jan-cerny:
Ninja Build didn't work when 'compat' object was empty.
We should build and link 'compat' object only if it contains
at least 1 function.
Addressing:
`CMake Error: CMake can not determine linker language for target: compat_object`


mpreisler:
@openscap-ci test this please

PR #967, opened 2018-01-24
Fix oscap-docker to clean up temp image.

matejak:
This fixes a known issue: https://bugzilla.redhat.com/show_bug.cgi?id=1454637

The code contained calls to cleanup functions, but they were not effective, because the entity that kept track of temp images was left out of the process. Specifically, the `DockerMount` instance was not passed to the `_cleanup_by_path` and a new one was constructed there instead.
I have modified the code so that the cleanup code path is preserved for the `oscap_chroot` method. The code path for `scan_cve` and `scan` has changed - creation of the new `DockerMount` was avoided by passing the existing instance. I have tested the `scan` path and it performs according to expectations.

### DIFF
--- a/utils/oscap_docker_python/oscap_docker_util.py
+++ b/utils/oscap_docker_python/oscap_docker_util.py
@@ -155,7 +155,8 @@ class OscapHelpers(object):
             sys.stderr.write(oscap_stderr.decode("utf-8") + "\n")
 
             # Clean up
-            self._cleanup_by_path(chroot_path)
+            DM = DockerMount("/tmp")
+            self._cleanup_by_path(chroot_path, DM)
 
             sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


jan-cerny:
I have discovered that if the evaluation fails there is still a leftover image.

I run a wrong command:
`./oscap-docker container 9bf4619216e7 oval eval fsdfsfd`

It returns
```
Command: oscap oval eval fsdfsfd failed!
Command returned exit code 1.                       
OpenSCAP Error: Unable to open file: 'fsdfsfd' [oscap_source.c:284] 
```
But it does not clean up, as we can easily verify:
```
docker images -a               
REPOSITORY                         TAG                 IMAGE ID            CREATED             SIZE      
<none>                             <none>              61bc826ae06b        6 seconds ago       196 MB 
```

I think that it should clean up in this case as well.

utils/oscap_docker_python/oscap_docker_util.py
```diff
@@ -155,7 +155,8 @@ def oscap_chroot(self, chroot_path, target, *oscap_args):
             sys.stderr.write(oscap_stderr.decode("utf-8") + "\n")
 
             # Clean up
-            self._cleanup_by_path(chroot_path)
+            DM = DockerMount("/tmp")
```
jan-cerny: I think this shouldn't be here, it can be a reason of my problem in my comment below.

jan-cerny: This code path runs if oscap returns exit code 1.

matejak: I see, this actually needs to be changed. I will refactor the error handling.

matejak:
Update on the fix:
The error @jan-cerny came across was caused by the code skipping the cleanup section (due to sys.exit call). I have wrapped it in a try..finally block instead. The finally block contains the cleanup code and is executed even if sys.exit is called: https://stackoverflow.com/questions/7709411/why-finally-block-is-executing-after-calling-sys-exit0-in-except-block

jan-cerny:
Thank you for a great contribution.

PR #977, opened 2018-02-07
Fixed string length bug related to ansible remediations.

matejak:
The previous code allocated wrong amount of memory for the remediation fix body.
If the second remediation body was longer than body of the first one, there was not enough of memory allocated. The first remediation was always handled correctly, and third, fourth etc. remediations usually got much more memory than they needed.
Fixes https://github.com/OpenSCAP/scap-security-guide/issues/2593

Update: The issue was caused by assuming that the "return value" of `pcre_exec` `ovector` is relative to `start_offset`. However, we have verified that it contains (absolute) indices of the `fix_text` string.

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -643,9 +643,12 @@ static inline int _xccdf_policy_rule_generate_fix_ansible(const char *template, 
 			}
 		}
 		else {
-			char *remediation_part = malloc((ovector[0] + 1) * sizeof(char));
-			memcpy(remediation_part, &fix_text[start_offset], ovector[0]);
-			remediation_part[ovector[0]] = '\0';
+			// Remarks: ovector doesn't contain values relative to start_offset, it contains
+			// absolute indices of fix_text.
+			const int length_between_matches = ovector[0] - start_offset;
+			char *remediation_part = malloc((length_between_matches + 1) * sizeof(char));
+			memcpy(remediation_part, &fix_text[start_offset], length_between_matches);
+			remediation_part[length_between_matches] = '\0';
 			if (_write_remediation_to_fd_and_free(output_fd, template, remediation_part) != 0) {
 				pcre_free(re);
 				return 1;
--- a/tests/API/XCCDF/unittests/test_generate_fix_ansible_vars_ds.xml
+++ b/tests/API/XCCDF/unittests/test_generate_fix_ansible_vars_ds.xml
@@ -99,6 +99,12 @@
   tags:
     - foo
     - bar
+
+- name: XCCDF Value www_value_val3 # promote to variable
+  set_fact:
+    www_value_val3: "dummy value"
+  tags:
+    - always
     </fix>
     <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5">
       <check-export export-name="oval:ssg:var:1" value-id="xccdf_com.example.www_value_val1"/>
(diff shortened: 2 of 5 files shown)


jan-cerny:
`tests/API/XCCDF/unittests` fail can be legitimate

yuumasato:
@matejak The function should actually remove the task `XCCDF Value ... #promote to variable` which was consumed during variable mode.

matejak:
I have modified the new behaviour to match the original one for the first match, which used to be discarded.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -643,9 +643,14 @@ static inline int _xccdf_policy_rule_generate_fix_ansible(const char *template,
 			}
 		}
 		else {
-			char *remediation_part = malloc((ovector[0] + 1) * sizeof(char));
-			memcpy(remediation_part, &fix_text[start_offset], ovector[0]);
-			remediation_part[ovector[0]] = '\0';
+			int fix_match_length = ovector[1] - ovector[0];
+			if (ovector[0] == 0) {
+				// Don't output anything if this is the first match (i.e. at the beginning of the fix_text).
+				fix_match_length = 0;
```
jan-cerny: Wouldn't it be better in this case if you don't do malloc, memcpy, _write_remediation at all?

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -643,9 +643,14 @@ static inline int _xccdf_policy_rule_generate_fix_ansible(const char *template,
 			}
 		}
 		else {
-			char *remediation_part = malloc((ovector[0] + 1) * sizeof(char));
-			memcpy(remediation_part, &fix_text[start_offset], ovector[0]);
```
yuumasato: I believe the `malloc`ed  and copyed size should be `ovector[0] - start_offset +1`

yuumasato:
A test with multiple Ansible values in a single fix would be interesting, we don't have such test.

src/XCCDF_POLICY/xccdf_policy_remediate.c
matejak: I agree, kudos for figuring out the true intention of the code!

matejak:
The test that is part of this PR fails nicely with unpatched `oscap`.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -643,9 +643,12 @@ static inline int _xccdf_policy_rule_generate_fix_ansible(const char *template,
...
-			remediation_part[ovector[0]] = '\0';
+			// Remarks: ovector doesn't contain values relative to start_offset, it contains
+			// absolute indices of fix_text.
```
yuumasato: That is very helpful comment, ;)

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -643,9 +643,12 @@ static inline int _xccdf_policy_rule_generate_fix_ansible(const char *template,
...
+			const int length_between_matches = ovector[0] - start_offset;
+			char *remediation_part = malloc((length_between_matches + 1) * sizeof(char));
+			memcpy(remediation_part, &fix_text[start_offset], length_between_matches);
+			remediation_part[length_between_matches] = '\0';
```
yuumasato: I noticed some lines with just indentation is left where the ` #promote to variable` task was.
As it doesn't break syntax, and it looks like it is common in other places of the generated playbook, I consider it a non blocker.

PR #978, opened 2018-02-12
Build Windows with common probe code -- easy part

jan-cerny:
I try to build OpenSCAP on Windows using Visual Studio with code that is shared by every probe - eg. SEXP, SEAP, SCH and related modules.

Also enables 2 basic probes - system_info and family, because we need at least those two to be able to see if the probe subsystem works or not.

This is a work in progress PR. Doesn't work at all, because the communication between probes needs to be implemented. Doesn't build yet, because all the probes now share the same `main` function, which MSVC can't handle.

What remains to be done and I am currently aware of:
- ~~create a Windows version of `sch_pipedata_t` data type~~ Instead of this we can solve communication with probes by prototype proposed in #981 .
- ~~replace socketpair(), fork(), etc. with similar things https://msdn.microsoft.com/en-us/library/ms682425.aspx https://msdn.microsoft.com/en-us/library/windows/desktop/aa365152(v=vs.85).aspx~~ Instead of this we can solve communication with probes by prototype proposed in #981 .
- ~~add main() for every probe~~ Is not necessary to do
- ~~solve multiple definitions of `probe_init()`~~ Is not necessary to do
- replace weird allocations and alignments with standard `malloc`, `realloc`, `free` - mainly get rid of `sm_alloc.c` -- will be proposed in a separate PR
- replace weird macros `assume_d`, `assume_r` by normal code everywhere
- fix compiler warnings
- ~~EDIT: change SEXP parser to not use dynamically computed goto labels (https://gcc.gnu.org/onlinedocs/gcc/Labels-as-Values.html), because that is a GCC-specific feature.  Whole `SEXP_parse() ` https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/OVAL/probes/SEAP/sexp-parser.c#L247 has to be rewritten~~ already done and merged upstream in #979 
There can be more.


jan-cerny:
9fec3ee418dd34fda6b7337b2381c52346639eda introduces a bug, I need to amend on this patch.

dahaic:
@jan-cerny what do you think about having feature branch on OpenSCAP/openscap that would host this effort? Would it be useful to you to have option to merge partial solutions based on logical blocks, instead of having huge PR that has to wrap everything?

mpreisler:
> effort? Would it be useful to you to have option to merge partial solutions based on logical blocks, instead of having huge PR that has to wrap everything?

+1 this. I would love if you took any small chunks that make sense even separately and just PR them to maint-1.2 or master where appropriate. Reviewing a huge PR of hundreds of commits (which this is going to morph into) takes a lot of effort.

jan-cerny:
@dahaic @mpreisler 

What does it mean to have a feature branch to host an effort? This PR doesn't introduce any feature. How do you imagine that branch? What should that branch contain?

Of course I plan to have separate pull requests instead of having only one big.

Moreover, I think I will do the `SEXP_parse()` rewrite in a new separate pull request, even if it is not logically separated, because it is a big function.

I think the review can be easier for you if you check it commit by commit. Click on Commits, pick first one, and then keep clicking on Next. This technique helps me with reviewing bigger PRs. I think it is easier to focus on small changes instead of trying to understand the complete diff. For this reason I create small commits.

src/OVAL/probes/SEAP/sexp-datatype.c
```diff
@@ -58,7 +57,9 @@ void SEXP_datatypeTbl_free(SEXP_datatypeTbl_t *t)
 
 int SEXP_datatypeTbl_init (SEXP_datatypeTbl_t *t)
 {
-        assume_d(t != NULL, -1);
+	if (t == NULL) {
+		return -1;
+	}
```
mpreisler: This changes the code btw, assume_d is only turned into this when NDEBUG is not defined. Maybe you know this and are okay with it but the commit message suggests you just replaced macro with code and expect no change so I am pointing this out, **THIS IS DIFFERENT**!

jan-cerny: Thank you, I didn't know that. What are the reasons for checking some conditions only with debug symbols? Maybe it can improve our safety if we check some conditions always.

jan-cerny: Maybe the answer would be that the code is faster, but I think the difference is not so big.

src/OVAL/probes/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ if(ENABLE_PROBES)
 
 		add_oscap_probe(probe_family "independent/family.c")
 
+		if(NOT WIN32)
 		add_oscap_probe(probe_textfilecontent "independent/textfilecontent.c")
```
matejak: What to do about indentation here? I agree that adding a tab to all lines until `endif()` could cause merge conflicts, but on the other hand, there are not multiple branches with actively developed CMake content. So I suggest to increase indentation of the if block.

src/OVAL/oval_probe.c
```diff
@@ -498,9 +497,13 @@ void oval_probe_meta_list(FILE *output, int flags)
 		output = stdout;
 
 	probe_dir = oval_probe_ext_getdir();
-	assume_d(probe_dir != NULL, /* void */);
```
matejak: This code will behave the same as the macro when not in debug mode. The macro would `abort()` in order to trigger the debugger. The functionality is not preserved on purpose.

src/OVAL/probes/probe/entcmp.c
```diff
@@ -36,8 +36,12 @@
 #include <string.h>
 #include <stdarg.h>
 #include <errno.h>
+#ifdef _WIN32
```
matejak: By skimming through `winsock2.h`, I got the impression that it exposes different API. Therefore, I would shift this platform-dependency one level up - instead of having `#ifdef`s in the source file, have the naturally cross-platform part of `entcmp.c` stay there, separate the unix-only part to `entcmp-unix.c` and introduce a windows-only `entcmp-w32.c`. Then, conditionally insert the right `.c` file into the compilation by CMake.

src/OVAL/oval_probe.c
```diff
@@ -514,7 +514,11 @@ void oval_probe_meta_list(FILE *output, int flags)
 
 			if (flags & OVAL_PROBEMETA_LIST_DYNAMIC) {
 				dI("Checking access to \"%s\"", probe_path);
+#ifdef _WIN32
```
matejak: What about having an `#ifdef` section at the top of the file and declare `X_OK` to zero?

src/OVAL/probes/SEAP/_seap-scheme.h
```diff
@@ -26,7 +26,11 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#ifdef _WIN32
```
matejak: Just a remark - [stackoverflow](https://stackoverflow.com/questions/341817/is-there-a-replacement-for-unistd-h-for-windows-visual-c) also suggests that `process.h` also has some stuff as `unistd.h`.

src/OVAL/probes/probe/icache.c
```diff
@@ -334,16 +345,22 @@ probe_icache_t *probe_icache_new(void)
 
 static int __probe_icache_add_nolock(probe_icache_t *cache, SEXP_t *cobj, SEXP_t *item, pthread_cond_t *cond)
 {
-        assume_d((cond == NULL) ^ (item == NULL), -1);
```
matejak: `!((cond == NULL) ^ (item == NULL))` can be rewritten to `(cond == NULL && item == NULL) || (cond != NULL && item != NULL)`

src/OVAL/probes/SEAP/generic/strbuf.c
```diff
@@ -282,6 +284,7 @@ size_t strbuf_fwrite (FILE *fp, strbuf_t *buf)
 
 ssize_t strbuf_write (strbuf_t *buf, int fd)
 {
+#ifdef HAVE_UIO_H
```
matejak: I prefer to have functions inside `#ifdef` blocks than vice versa.

src/OVAL/probes/SEAP/sexp-datatype.c
```diff
@@ -110,12 +110,16 @@ SEXP_datatypePtr_t *SEXP_datatype_get (SEXP_datatypeTbl_t *t, const char *k)
                 if (d->dt_flg & SEXP_DTFLG_LOCALDATA) {
                         struct SEXP_datatype_extptr *eptr = NULL;
 
+#ifdef HAVE_POSIX_MEMALIGN
```
matejak: Would it be possible to declare a `get_possibly_aligned_memory` with platform-dependent implementation and call that one from now on, so there are no further `#ifdef`s in the code?

mpreisler: +100, make it inlined too please

src/OVAL/probes/SEAP/sch_pipe.c
```diff
@@ -275,12 +279,15 @@ ssize_t sch_pipe_recv (SEAP_desc_t *desc, void *buf, size_t len, uint32_t flags)
         sch_pipedata_t *data;
 	ssize_t         ret;
 
-        assume_d (desc != NULL, -1, errno = EFAULT;);
```
matejak: This is a duplicate code with `sch_pipe_send`. Deduplicating may be out of scope, but at least put a TODO comment here.

src/OVAL/oval_probe_ext.c
```diff
@@ -1141,13 +1141,14 @@ int oval_probe_ext_abort(SEAP_CTX_t *ctx, oval_pd_t *pd, oval_pext_t *pext)
 	switch (dsc->scheme) {
 	case SCH_PIPE:
 	{
```
matejak: Those `#if...` blocks make the code harder to understand.

src/OVAL/probes/probe/main.c
```diff
@@ -189,14 +191,17 @@ int main(int argc, char *argv[])
 	oscap_set_verbose(verbosity_level, verbose_log_file, true);
 
 	if ((errno = pthread_barrier_init(&OSCAP_GSYM(th_barrier), NULL,
```
matejak: Why not to use const ints to name the argument variables and discard those comments?
Then, this call to `pthread_barrier_init` that seems to have different number of arguments across platforms could be wrapped in an inline function. And you could use the preprocessor to pick between the two definitions and leave the call itself be without preprocessor.

src/OVAL/probes/probe/main.c
```diff
@@ -228,7 +228,11 @@ int main(int argc, char *argv[])
 	 * Initialize SEAP stuff
 	 */
 	probe.SEAP_ctx = SEAP_CTX_new();
-	probe.sd       = SEAP_openfd2(probe.SEAP_ctx, STDIN_FILENO, STDOUT_FILENO, 0);
+#ifdef _WIN32
+	probe.sd = SEAP_openfd2(probe.SEAP_ctx, _fileno(stdin), _fileno(stdout), 0);
```
matejak: What about defining the `STDIN_FILENO` etc. for W32 close to the include directive? [stacko says](https://stackoverflow.com/questions/15102992/what-is-the-difference-between-stdin-and-stdin-fileno) that it is defined in `unistd.h` on unixes.

src/OVAL/probes/CMakeLists.txt
jan-cerny: OK

src/OVAL/probes/probe/entcmp.c
jan-cerny: I don't prefer creating 2 almost identical files that will differ only in 1-2 lines. That will be harder to maintain, because if somebody will fix or change this code in future, he will have to change the same code at 2 places, and he will very likely forget to change one of them.

src/OVAL/oval_probe.c
jan-cerny: Great! If we remove every weird construction, we will have less likely the Heisenbug.

src/OVAL/oval_probe.c
jan-cerny: Thank you, I agree, that is definitely better.

src/OVAL/probes/probe/icache.c
jan-cerny: OK

src/OVAL/probes/SEAP/generic/strbuf.c
jan-cerny: OK

src/OVAL/probes/SEAP/sexp-datatype.c
jan-cerny: Maybe it would be even better to get rid of memory hacks.

src/OVAL/oval_probe_ext.c
jan-cerny: OK

src/OVAL/probes/SEAP/sch_pipe.c
jan-cerny: OK

src/OVAL/probes/probe/main.c
jan-cerny: OK

src/OVAL/probes/SEAP/sexp-datatype.c
mpreisler: it's not a memory hack

src/OVAL/oval_probe.c
mpreisler: Wait what? I am really lost here. So do we know and acknowledge that we are changing behavior with these changes? Also, why even bother changing this in the same PR that already has tons of changes? There is nothing in these assume macros that doesn't work on Windows.

Please, please, please, if we decide to do something about this let's do it in a separate PR with a nice description of what is happening. This seems like a change that doesn't belong here.

jan-cerny:
@matejak Thank you very much for your review.

I have fixed the wrong indentation in src/OVAL/probes/CMakeLists.txt.
I have declared X_OK at the beginning of src/OVAL/oval_probe.c.
I have replaced a bitwise XOR in src/OVAL/probes/probe/icache.c by no bit operations.
I have moved ifdefs outside of function strbuf_fwrite.
I have added a TODO comment on duplicate code in SEAP pipe schema, but I don't know how to solve it. 
I have made the code in oval_probe_ext.c more readable.
I have removed conditional compilation inside a function parameters list in main().
I have defined STDIN_FILENO and STDOUT_FILENO on Windows and I have also included unistd.h in main.c.

I have rebased the pull request on the top of master's head.

I have removed the commit that tried to change posix_memalign from this pull request. I think the problem with allocating aligned memory needs a more serious solution.
For example, on Windows, memory allocated aligned by a special function, has to be freed by a special function.
I would prefer to work on that problem in a separate pull request.

src/OVAL/probes/probe/entcmp.c
matejak: I meant something else than creating almost files, but since this can be done any time later, I am fine with either way for now.

jan-cerny: @matejak What was your idea then?

src/OVAL/oval_probe.c
jan-cerny: @mpreisler I am sorry, I don't understand at all what do you want. What do you suggest to change?

src/OVAL/probes/probe/entcmp.c
matejak: My idea was to keep the common stuff in one file and the platform-specific stuff in another two files. Therefore, this is not about having multiple almost identical files, but it is probably more about implementing a function with platform-independent interface, but platform-specific internals in those two files.

jan-cerny: @matejak OK, but is there something platform specific in the `entcmp.c`?

src/OVAL/oval_probe.c
jan-cerny: @mpreisler I think it was not clear why I replace these macros. The problem is that Visual Studio 2017 can't compile them. Visual Studio 2017 fails and produces errors. It complains either that it expects an expression or doesn't expect an operator when expanding these macros. I found it more understandable and beneficial to replace them by normal code than wrapping them into `#ifndef _WIN32 ... #endif` sections. Is it OK?

CMakeLists.txt
```diff
@@ -263,6 +264,10 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
 configure_file("config.h.in" "config.h")
 add_definitions(-DHAVE_CONFIG_H)
+if (MSVC)
+	add_definitions(-D_CRT_SECURE_NO_WARNINGS)
+	add_definitions(-D_CRT_NONSTDC_NO_WARNINGS)
```
mpreisler: are you sure about this? aren't both of these valuable warnings?

mpreisler:
Since you removed usage of SEXP_vfree, remove it's definition as well. Otherwise it's gonna creep back in.

src/OVAL/probes/SEAP/sexp-manip.c
```diff
@@ -1869,39 +1867,15 @@ void SEXP_free (SEXP_t *s_exp)
...
-                sm_free (s_exp);
-        }
-        return;
-}
-#endif
-
-#if defined(NDEBUG)
-void __SEXP_vfree (int n, SEXP_t *s_exp, ...)
-#else
-void __SEXP_vfree (const char *file, uint32_t line, const char *func, int n, SEXP_t *s_exp, ...)
```
jan-cerny: @mpreisler SEXP_vfree is already removed here.

mpreisler:
Please don't remove any of hte assume stuff. Let's do that separately if we want to do it, it's not part of the Windows effort. I consider that a blocker for merging this PR.

jan-cerny:
> Please don't remove any of hte assume stuff. Let's do that separately if we want to do it, it's not part of the Windows effort. I consider that a blocker for merging this PR.

I think that it is part of the Windows effort, because Microsoft Visual Studio 2017 cannot compile them.

CMakeLists.txt
jan-cerny: I am, I'll add a comment to there.

src/OVAL/oval_probe.c
mpreisler: without looking more into it I am not sure but what exactly does it complain about? I'd suggest to fix the macros instead of removing them? also consider using commit message such as "replaced assume_d macros because they cause build issues in MSVC2017..." instead of  "replaced weird macros with normal code"

jan-cerny:
I have added a long comment and I have rebased this pull request on the top of the current OpenSCAP upstream master head.

src/OVAL/oval_probe.c
jan-cerny: @mpreisler 

For example, in this particular case that we comment, `oval_probe.c`:

```
Error (active)	E0029	expected an expression	oval_object	c:\devel\openscap\src\OVAL\oval_probe.c	505		

Warning	C4003	not enough actual parameters for macro 'assume'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	505		

Warning	C4003	not enough actual parameters for macro '__LB'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	505		

Warning	C4003	not enough actual parameters for macro '__terminate'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	505		

Warning	C4003	not enough actual parameters for macro '__RB'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	505		

Error	C2059	syntax error: ')'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	505		

Error	C2065	'__cont505': undeclared identifier	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	505		

Error	C2059	syntax error: 'else'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	505		

Error	C2059	syntax error: '}'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	505		

Error	C2065	'probe_dir': undeclared identifier	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	507		

```

And then about 10 errors in the rest of the file, because this weird macro causes that Microsoft Visual Studio 2017 cannot understand the rest of file. For example:

```
Error	C2099	initializer is not a constant	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	507		

Error	C2059	syntax error: 'if'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	508		

Error	C2059	syntax error: 'for'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	512		

Error	C2143	syntax error: missing '{' before '<'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	512		

Error	C2059	syntax error: '<'	oval_object	C:\devel\openscap\src\OVAL\oval_probe.c	512		
```
and so on.

When I replace the macro by `if(...)` as I propose in this patch, all of those errors and warnings disappear.

A very similar situation is at all other occurrences of assume macro.

jan-cerny:
@openscap-ci test this please

src/OVAL/oval_probe.c
mpreisler: I briefly looked at it and assume is a variadic macro. We don't use that aspect of it almost anywhere so can't you just make it non-variadic? It should be a pretty quick fix in `src/common/assume.h`.

jan-cerny:
@openscap-ci test this please

mpreisler:
ACK

PR #979, opened 2018-02-17
Rewrite SEXP parser

jan-cerny:
SEXP is Lisp serial expression. They are used across our codebase. They are parsed using `SEXP_parse()` function.

The problem is that this function uses many goto whose targets are computed. This is called "labels as values." See https://gcc.gnu.org/onlinedocs/gcc/Labels-as-Values.html This feature is supported by GCC and Clang compilers. However Visual Studio 2017 does not support it. Visual Studio 2017 says that it does not want `&&` operator there.

Therefore, to be able to compile probe subsystem on Windows using Visual Studio 2017, it is necessary to avoid using "labels as values". I have done that easily by converting the parser to an usual finite automaton.


mpreisler:
@openscap-ci test this please

src/OVAL/probes/SEAP/sexp-parser.c
```diff
@@ -475,211 +504,199 @@ SEXP_t *SEXP_parse (const SEXP_psetup_t *psetup, char *buffer, size_t buflen, SE
...
-                if (e_dsc.p_bufoff >= spb_len)
+                switch (dfa_state) {
+                case S_START:
+                        /*
+                         * Allocate an empty S-exp. The value or type will be assigned in the
+                         * subparser.
+                         */
+                        assume_d (e_dsc.s_exp == NULL, NULL); /* no leaks */
+                        e_dsc.s_exp = SEXP_new ();
+                        dfa_state = S_NO_SEXP_ALLOC;
```
mpreisler: indents

mpreisler:
Great work! The indents are a little out of whack but that should be an easy fix.

I wish some of the switches were not replaced because switches are easier for me to read than nested ifs but I won't lobby for that too much :-)

jan-cerny:
@mpreisler You are right that the switches were more readable, and I agree it was not necessary to change them. But it was hard for me to orient in that, because there were nested switches, and multiple break statements, which could refer to different switches. I though I would be less likely to make a mistake in later phase if I convert the switches first.

I will look into the indents problem. The whole file was intended by 8 spaces, so I decided to keep that, to avoid mixing tabs and spaces and help my orientation in that.

mpreisler:
> I will look into the indents problem. The whole file was intended by 8 spaces, so I decided to keep that, to avoid mixing tabs and spaces and help my orientation in that.

openscap code style is tabs for indents, let's do that for all new code

src/OVAL/probes/SEAP/sexp-parser.c
jan-cerny: Unfortunately, this whole file was indented by 8 spaces. If we follow this indentation style, the result looks correct, even if the diff is weird. Do you think it would be better to change the indents in this file or in this function to tabs?

mpreisler: Use tabs for indentation in new code. It's easier to just do it than fret about it too much IMO. Yes, sometimes that will be weird but over time we will converge the codebase to one code style.

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

mpreisler:
ACK

PR #990, opened 2018-03-01
Increased robustness of configure script regarding to Python.

matejak:
When requiring `python2` bindings and no `python2` interpreter is available, the configure script will abort with an error message.

The `configure.ac` template section for `python3` has been slightly refactored.

### DIFF
--- a/ac_probes/configure.ac.tpl
+++ b/ac_probes/configure.ac.tpl
@@ -540,32 +540,48 @@ if test "x${perl_bind}" = xyes; then
 fi
 
 # (AM_PATH_PYTHON) cannot be used for multiple Python version at once
-if test "x${python3_bind}" = xyes; then
-	AC_PATH_PROG([PYTHON3], [python3])
-	AC_PATH_PROG([PYTHON3_CONFIG], [python3-config], [no])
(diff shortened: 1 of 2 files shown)


mpreisler:
ACK

matejak:
Unfortunately, this breaks configure runs on systems that don't have `python2-config`, but only `python-config` for Python2 such as RHEL6. (BTW RHEL6 `python-devel` provides `python2.6-config`, but doesn't have `python2-config`). This is likely to break jenkins test runs (e.g. tests of #992). However, recent systems have `python-config`, which is symlink for `python3-config` and this PR is meant to address those.
So this PR fixes new systems, but breaks old systems that don't have `python2-config`. The question is what to do about this. Possibilities:
- Do nothing: Easiest, the behavior won't change for users of new systems. As usage of old systems will decline, this problem will become less and less important. We have to do something about RHEL6 Jenkins tests though.
- Disable building Python bindings by default: Easy, people on those older systems won't notice anything unless they decide that they want to build those.
- Introduce logic that will determine what version of Python is `python-config` related to: Moderately difficult and probably closest to the proper solution.

jan-cerny:
I think we still have to support RHEL6 including Python bindings. I would prefer solution 3.

Issue #986, opened 2018-02-27
oval_enumerations.c:667: warning: comparison of unsigned expression < 0


jlcharton:
Hi,

Taking a closer look at this warning, the `if (subtype < 0)` seems broken to me since the subtype variable is initialized to OVAL_SUBTYPE_UNKNOWN (which equals 0) at line 611 and never change above line 667.
So I believe the intended code was `if (subtype_s < 0)` instead.
But then, considering the definitions of the OVAL_SUBTYPE_XXXX_MAP arrays there is still a problem since then oscap_string_to_enum returns OVAL_SUBTYPE_UNKNOWN as default value.

So in my opinion either the default value in OVAL_SUBTYPE_XXXX_MAP arrays should be -1 or the if statement at line 667 should be `if (subtype_s == OVAL_SUBTYPE_UNKNOWN)`.
Can you confirm this?

Regards,
Jean-Louis

matejak:
Thank you for your findings! It is a difficult question for us. The code in question was written 4 years ago by @isimluk and it may take some time for us to find out the true intention of the code.
Here is the link to the incriminated piece for reference: https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/OVAL/oval_enumerations.c#L667

jlcharton:
Further investigations make me think that
```
	int subtype_s = oscap_string_to_enum(map, tagname);
	if (subtype < 0) {
		dW("Unknown OVAL family subtype: %s", tagname);
		subtype = OVAL_ENUMERATION_INVALID;
	}
	else {
		subtype = subtype_s;
        }
```
can simply be replaced by
```
        subtype = oscap_string_to_enum(map, tagname);
        if (subtype == OVAL_SUBTYPE_UNKNOWN) {
                dW("Unknown OVAL family subtype: %s", tagname);
        }
```

Moreover, at the beginning of the function, tagname should be test for NULL after the call to xmlTextReaderLocalName. That is:
```
        char *tagname = (char *)xmlTextReaderLocalName(reader);
        char *endptr = (tagname != NULL) ? strrchr(tagname, '_') : NULL;
```

J-L

jan-cerny:
@jlcharton  I think you are correct. The `if (subtype < 0) {` condition at line 667 makes no sense to me. Try to replace it, and if it works, feel free to open a pull request :smiley:

jan-cerny:
@isimluk That's fine, there's no need to excuse. Thank you very much for looking at this issue.

[...]

PR #991, opened 2018-03-02
Fix unknown subtype handling in oval_subtype_parse

jlcharton:
See #986 discussion.


jan-cerny:
This is perfect! Thank you very much!

PR #993, opened 2018-03-12
[WIP] Add search functions to xccdf-guide-impl.xsl

redhatrises:
Adds similar functionality from xccdf-report-impl.xsl to search rules.

### DIFF
--- a/xsl/xccdf-guide-impl.xsl
+++ b/xsl/xccdf-guide-impl.xsl
@@ -597,6 +597,19 @@ Authors:
     <div id="guide-tree">
         <h2>Checklist</h2>
      
+        <div class="row">
+            <div class="col-sm-6">
+                <div class="input-group">
+                    <input type="text" class="form-control" placeholder="Search through the guide" id="search-input" oninput="ruleSearch()"/>
+
+                    <div class="input-group-btn">
+                         <button class="btn btn-default" onclick="ruleSearch()">Search</button>
+                    </div>
+                </div>
+                <p id="search-matches"></p>
+            </div>
+        </div>
+
         <table class="treetable table table-bordered">
             <tbody>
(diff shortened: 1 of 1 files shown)


xsl/xccdf-guide-impl.xsl
```diff
@@ -597,6 +597,18 @@ Authors:
     <div id="guide-tree">
```
matejak: There are five opening `div` tags, but only four closing ones until the end of the file.

mpreisler:
The javascript function shouldn't be copied like this, we should instead change the ruleSearch function to accomodate both cases. Copying code makes it less maintainable.

yuumasato:
@openscap-ci test this please

matejak:
Tests of `oscap` on RHEL6 are broken as of 3/2018, so the failure has nothing to do with the PR itself.

matejak:
@openscap-ci test this please

matejak:
Now, `make check` seems to be failing on an item that is related to the PR (it fails in `tests/API/XCCDF/unittests` and `tests/API/XCCDF/guide`).

yuumasato:
@redhatrises Hello,

Can the search built in browser resolve the need for a search function?
Maybe more meaningful way to search rule is through IDs, and we can just expose them.

yuumasato:
By clicking "Show all result detail" the rule IDs are shown and one can search using the browser search function.

redhatrises:
Sorry, but this is about syncing our reports and guides formats to have similar workflows, reopening.

jan-cerny:
The code here doesn't work, it needs to implement new functions in our JavaScript modules. @redhatrises We have discussed this with the team and we think that it's not necessary to keep this pull request. If you prefer, please open an issue.

PR #994, opened 2018-03-13
Remove unused code in SEXP

jan-cerny:
This PR removes sexp-template.c and sexp-template.h which is never used anywhere.

Also this removes parts of sexp-datatype.c that are never executed. The removal includes both occurrences of `posix_memalign` in this file, therefore we won't have to port them to Windows.


mpreisler:
Can you describe how you know the parts of sexp-datatype.c will never be executed when using OpenSCAP?

jan-cerny:
Yes, I can.

SEXP_datatype_add is always called with passing `d` as NULL.
```
$ git grep '\<SEXP_datatype_add\>'
src/OVAL/probes/SEAP/_sexp-datatype.h:SEXP_datatypePtr_t *SEXP_datatype_add(SEXP_datatypeTbl_t *t, char *name, SEXP_datatype_t *d, void *l);
src/OVAL/probes/SEAP/sexp-datatype.c:SEXP_datatypePtr_t *SEXP_datatype_add(SEXP_datatypeTbl_t *t, char *n, SEXP_datatype_t *d, void *l)
src/OVAL/probes/SEAP/sexp-manip.c:                t = SEXP_datatype_add (&g_datatypes, k, NULL, NULL);
src/OVAL/probes/SEAP/sexp-manip.c:                t = SEXP_datatype_add (&g_datatypes, k, NULL, NULL);
src/OVAL/probes/SEAP/sexp-parser.c:                        dsc->s_exp->s_type = SEXP_datatype_add (&g_datatypes, name, NULL, NULL);
src/OVAL/probes/SEAP/sexp-parser.c:                dsc->s_exp->s_type = SEXP_datatype_add (&g_datatypes, name, NULL, NULL);
```
Therefore I can remove code guarded by `d!=NULL` in SEXP_datatype_add.

The flag SEXP_DTFLPG_LOCALDATA was set only in sexp-template.c, which is never called, so I remove it as well. Therefore this flag is never set, so I can remove the code that depends on if this flag is set.

mpreisler:
@jan-cerny make it impossible to break this assumption then. Remove the possibility of passing d in SEXP_datatype_add

mpreisler:
ACK

PR #996, opened 2018-03-14
Don't repeatedly call getenv in SEXP_VALIDATE

mpreisler:
This makes oscap scans way way faster in debug mode. On my machine it makes it roughly 3 times faster.

This issue was found by @jan-cerny , see #981


jan-cerny:
@openscap-ci test this please

jan-cerny:
That's great

PR #997, opened 2018-03-20
Refactored the Python handling in build scripts.

matejak:
This PR should fix the handling of different Python major versions in different environments for good (see discussion at the end of #990).

I have took the `python.m4` macro, refactored it and tested that it works. If conditions permit, I will try to push this to upstream Automake.

Highlights of this PR are:
* Correct and relatively easy handling of Python2 and Python3 interpreters and environments.
* Greedy build of Python bindings - if possible, they are built, if not possible, no fatal error is raised. The explicit choice to build or not to build them is preserved - in that case, if a binding can't be built for whatever reason, an error is raised.
* The "preferred_python" for oscap-docker is now selected by checking for the Atomic module availability, prioritizing Python3 over Python2. If there is not a suitable interpreter found, ~an error is raised (this is maybe too harsh)~ a message is output and an available interpreter is selected without raising an error.

### DIFF
--- a/ac_probes/configure.ac.tpl
+++ b/ac_probes/configure.ac.tpl
@@ -565,9 +565,12 @@ dnl
 dnl $1: Python major version
 m4_define([_ATTEMPT_TO_SET_PREFERRED_PYTHON_FOR_OSCAP_DOCKER],
 	[AS_IF([test "$HAVE_PYTHON$1" = yes],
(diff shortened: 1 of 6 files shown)


jan-cerny:
this could be related: https://lists.gnu.org/archive/html/autoconf/2010-01/msg00050.html

yuumasato:
Looks good!
@matejak  Is it still WIP?

yuumasato:
Ack.

Issue #998, opened 2018-03-22
System info probe returns confusing data


jan-cerny:
System info probe in OpenSCAP 1.2.16 returns some values in OVAL results `system_info` element that I think could be improved to better inform the consumers of OVAL results.

In normal mode, when we scan local host, we get 
```
<system_info>
  <os_name>Linux</os_name>
  <os_version>#1 SMP Wed Feb 28 17:53:39 UTC 2018</os_version>
  <architecture>x86_64</architecture>
  <primary_host_name>thinkpad</primary_host_name>
  <interfaces>
     ...
```
I found weird that we return this date as OS version, but it is consistent with output of `uname --kernel-version`. Maybe we could add some information about Linux distribution used.

In offline mode, when scanning a RHEL7.4 virtual domain using oscap-vm:
```
<system_info>
  <os_name>Red</os_name>
  <os_version>Hat Enterprise Linux Server (3.10.0-693.el7.x86_64) 7.4 (Maipo)</os_version>
  <architecture>x86_64</architecture>
  <primary_host_name>localhost.localdomain</primary_host_name>
  <interfaces/>
</system_info>
```
Notice not only that the os name and version is broken, but it is not consistent with output in online mode (see above). Moreover, network interfaces information are missing. I also don't like that the probe is parsing /boot/grub/ files.

In offline scanning of Docker containers, it returns unknown.
```
<system_info>
  <os_name>Unknown</os_name>
  <os_version>Unknown</os_version>
  <architecture>Unknown</architecture>
  <primary_host_name>Unknown</primary_host_name>
  <interfaces/>
</system_info>
```
Some data aren't possible to get from an container, but if we provide data about Linux kernel on the host, we could maybe return it the same when scanning a container, because containers share the same kernel with the host.

OVAL specification in  Core System Characteristics schema is quite unspecific. 

> The required os_name elements describes the operating system of the machine the data was collected on.

> The required os_version elements describe the operating system version of the machine the data was collected on.

evgenyz:
@jan-cerny 

Any idea why we are using --kenel-version to fill *os*_version field? And why it is not, at least, --kernel-release?

Wouldn't it be more sane to use information from '/etc/os-release'?

evgenyz:
@jan-cerny 

Also, is there any particular reason to get os name/version information from `/boot/grub2/grubenv` in the offline mode? Why not `/etc/os-release' again?

jan-cerny:
I have no idea.

[...]

PR #999, opened 2018-03-22
Change offline mode in system info probe to not use chroot()

jan-cerny:
We prefer `PROBE_OFFLINE_OWN` which doesn't use `chroot()`, for various reasons, namely that users have to be granted by `CAP_SYS_CHROOT` to be allowed to call `chroot()`

This pull request replaces CHROOT offline mode by OWN offline mode in system info probe. System info probe won't call chroot() in offline mode anymore, but it will add a prefix to file paths instead.

The supported mode should be advertised by probes in probe_offline_mode() function, and not in probe_init function(). That's because the offline mode flags PROBE_OFFLINE.* are checked in main() after probe_offline_mode() is called, but before probe_init() is called, therefore setting the flags in probe_init doesn't have the desired effect. In fact, system_info probe supports only PROBE_OFFLINE_CHROOT mode, not PROBE_OFFLINE_ALL, as it was set here before.

During my work on this I have found some issues or suggestions, but I don't fixed them. I have described them in https://github.com/OpenSCAP/openscap/issues/998 and they can be fixed in other PRs.

This is also related to Windows support, because it will help us to move to "single process" architecture.


src/OVAL/probes/independent/system_info.c
```diff
@@ -353,15 +353,25 @@ static ssize_t __sysinfo_saneval(const char *s)
...
 	int rc = PCRE_ERROR_NOMATCH;
 	int len;
 
-	fp = fopen("/boot/grub2/grub.cfg", "r");
+	const char *grubcfg_path = "/boot/grub2/grub.cfg";
+	if (oscap_probe_root != NULL) {
+		char grubcfg_path_with_root[PATH_MAX]; // PATH_MAX includes terminating '\0' byte
+		size_t oscap_probe_root_len = strlen(oscap_probe_root);
+		strncpy(grubcfg_path_with_root, oscap_probe_root, PATH_MAX);
+		strncpy(grubcfg_path_with_root + oscap_probe_root_len, grubcfg_path, PATH_MAX - oscap_probe_root_len);
```
yuumasato: Lines 367 to 369 could be turned into a function.

yuumasato:
@openscap-ci test this please

jan-cerny:
I think how to add a test for offline mode into system_info probe to upstream test suite. Could a test create a fake directory structure that contains `/boot/grub2/grubenv` and `/etc/hostname`?

src/OVAL/probes/independent/system_info.c
```diff
@@ -353,15 +353,25 @@ static ssize_t __sysinfo_saneval(const char *s)
 	return (ssize_t)real_length;
 }
 
-static char * _offline_chroot_get_menuentry(int entry_num)
+static char *_offline_chroot_get_menuentry(const char *oscap_probe_root, int entry_num)
```
mpreisler: rename the function because it no longer does chroot

mpreisler:
> I think how to add a test for offline mode into system_info probe to upstream test suite. Could a test create a fake directory structure that contains /boot/grub2/grubenv and /etc/hostname?

sure, in `/tmp` or some such

src/OVAL/probes/independent/system_info.c
```diff
@@ -353,6 +353,21 @@ static ssize_t __sysinfo_saneval(const char *s)
 	return (ssize_t)real_length;
 }
 
+static FILE *_fopen_with_prefix(const char *prefix, const char *path)
+{
+	FILE *fp;
+	if (prefix != NULL) {
+		char path_with_prefix[PATH_MAX]; // PATH_MAX includes terminating '\0' byte
+		size_t prefix_len = strlen(prefix);
```
mpreisler: const

src/OVAL/probes/independent/system_info.c
```diff
@@ -353,6 +353,21 @@ static ssize_t __sysinfo_saneval(const char *s)
...
+		strncpy(path_with_prefix, prefix, PATH_MAX);
+		strncpy(path_with_prefix + prefix_len, path, PATH_MAX - prefix_len);
```
mpreisler: cound't this be done in an easier way with `oscap_asprintf`? or are you doing it like this because you are worried of it running over PATH_MAX?

jan-cerny: It could.

jan-cerny:
I have changed the code to use oscap_sprintf. I have also added a very simple test.

jan-cerny:
After experience from #1001, #1002, https://bugzilla.redhat.com/show_bug.cgi?id=1556988
I was thinking if it is safe to not use chroot() in system_info probe.

My opinion is it is safe to not use chroot in system_info probe. Unlike textfilecontent* and rpm* probes, all the paths in system info probes are hardcoded. System info probe doesn't implement any OVAL object, where you could provide any path. It just replies to a request "give me system info". This request has no parameters.

Do you agree with this?

mpreisler:
> My opinion is it is safe to not use chroot in system_info probe. Unlike textfilecontent* and rpm* probes, all the paths in system info probes are hardcoded. System info probe doesn't implement any OVAL object, where you could provide any path. It just replies to a request "give me system info". This request has no parameters.
> 
> Do you agree with this?

yeah, makes sense to me

PR #1000, opened 2018-03-24
Improve offline mode

jan-cerny:
In this pull request I propose to make offline mode less confusing.

The probes consistently advertise offline modes they support by defining a `probe_offline_mode_supported` function, specifically in the return value of this function. If this function is not defined in the probe C file, then the default value of PROBE_OFFLINE_NONE is used, meaning probe doesn't support any offline mode.

I fixed the problem that probes set offline mode after it was checked in probe_main, which meant it took no effect.

I fixed the bug that the probe called `chroot` even if it didn't support the PROBE_OFFLINE_CHROOT mode.

I remove global variables and replace them by adding the variables into probe and probe context structures.

I hope the logic in `probe_main` function is now more understandable.

This PR is also related to Windows support, because it should help to fix the broken offline mode in #981.


jan-cerny:
Pull request #1000!  :birthday: :champagne: :fireworks: :smile:

matejak:
Check out the conflict, it is not trivial to fix.

jan-cerny:
@matejak I have rebased on the top of current upstream maint-1.2 and fixed the conflict. This was a problem of separate pull requests.

matejak:
Good job of getting rid of global variables and cleaning up the code afterwards.

PR #1002, opened 2018-03-26
Fix offline scan in probes

### LINKED ISSUE #1001: OpenSCAP textfilecontent54 probe does not work in offline mode
Hello, I use openscap-1.2.16-1.fc27.x86_64 on Fedora 27.

I have a serious problem because I can't retrieve data from text files using `oscap-chroot`.

I use this content:
[reproducer.oval.zip](https://github.com/OpenSCAP/openscap/files/1845120/reproducer.oval.zip)
This OVAL file is a textfilecontent54 test that requires that a `hello.txt` exists in root directory.

I use these steps:
```
mkdir /tmp/xxx
echo "Hello" > /tmp/xxx/hello.txt
sudo oscap-chroot /tmp/xxx oval eval --results results.xml reproducer.oval.xml
```

The `results.xml` doesn't contain data about `hello.txt`.
```
 74         <collected_objects>
 75           <object id="oval:x:obj:1" version="1" flag="does not exist"/>
 76         </collected_objects>
```

Other problem is that it starts working if a file with the same name exists on the host filesystem. In other words, OpenSCAP checks if the given file exists on the host filesystem and then it reads data from a file with same name in guest filesystem.

Example:
```
mkdir /tmp/zzz
echo "Hello" > /tmp/zzz/hello.txt
sudo  touch /hello.txt
sudo oscap-chroot /tmp/zzz oval eval --results results.xml reproducer.oval.xml
```

Now the OVAL results contain the data from `/tmp/zzz` , we can verify that we see "Hello" there, but the file in `/` is empty. But it required the file in `/` to exist in order to get data from `/tmp/zzz`. That I think is completely wrong.

```
 82       <ind-sys:textfilecontent_item id="1075811" status="exists">
 83             <ind-sys:filepath>/tmp/zzz//hello.txt</ind-sys:filepath>
 84             <ind-sys:path>/tmp/zzz/</ind-sys:path>
 85             <ind-sys:filename>hello.txt</ind-sys:filename>
 86             <ind-sys:pattern>^.*$</ind-sys:pattern>
 87             <ind-sys:instance datatype="int">1</ind-sys:instance>
 88             <ind-sys:line>^.*$</ind-sys:line>
 89             <ind-sys:text>Hello</ind-sys:text>
 90           </ind-sys:textfilecontent_item>
 91         </system_data>
```

This should be a problem if we scan a container and we want to check configuration compliance of something that is installed only in the container but is not present on the host. Then OpenSCAP isn't able to get the data.

The problem looks to be introduced by https://github.com/OpenSCAP/openscap/commit/908d002c68e43a3d3c3bede128c535fbee815a10 Unfortunately, when we reviewed that, we haven't  considered the `oval_fts_open` function, which still checks the host filesystem in offline mode.

jan-cerny:
The textfilecontent(54) probes were not able to read from file in the guest filesystem if a file with the same name doesn't exist on host filesystem.

Fixes #1001.

This PR first reverts commit 908d002c68e43a3d3c3bede128c535fbee815a10 and then it adds new commits that change the offline mode in the probes, namely: 
* textfilecontent
* textfilecontent54
* filehash
* filehash58
* xmlfilecontent
* file
* fileexetendedattribute
* selinuxsecuritycontext

The problem with the commit  908d002c68e43a3d3c3bede128c535fbee815a10 is that it ignores how oval_fts functions work. They traverse the host filesystem, handle OVAL file behaviors, check if the filesystem is remote, check recurse directions, ... 

The revert commit will revert us back to use `chroot()`.

I tried to find some way how to enable offline mode in textfilecontent(54) probes without doing `chroot()`.

The basic idea is very simple: You just add a prefix to the filepath that you want to check, eg. if your container is mounted in `/tmp/abcdef` and you want to check `/etc/passwd` there, you just check `/tmp/abcdef/etc/passwd`.

However, it is not so easy as it seems, for multiple reasons:
* The filepath in OVAL can be specified by a regular expression. The regular expressions can be very complex. You have to understand them and modify the regular expression by the prefix correctly.
* The filepath in OVAL can be specified in a variable, which is handled by SEXPs, so we have to modify how OVAL variables are handled, too.
* The paths can be obtained by some functions, eg. glob_to_regex. We have to make sure glob_to_regex returns a prefixed path.
* The paths could be read from another files, returned by other definitions, etc etc. and we would have to make sure they are prefixed, but not multiple times. 
* The OVAL_fts compares the path it checks with oval_probe_ent using generic functions. They can handle various cases, like OVAL sets, OVAL variables, regexes, OVAL filters, and so on. We would have to modify all this comparisons to be able to handle the filepath prefix - either always add it or always strip it. That's a  change of whole SEXP module.
* We would also have to make sure the prefix doesn't appear in the collected OVAL item. Otherwise OVAL states don't work (which btw doesn't work now in 1.2.16).
* Moreover we would have to make sure the filepath prefix doesn't appear the OVAL results documents and HTML reports. It would be terribly confusing for users if they see some `/tmp/abcdef` prefixes there, because they expect the container scan is completely isolated from the host.

I think it is possible to do that properly, but I expect that to be a very big task due to aforementioned reasons.

For more details, please see #1001 .

### DIFF
--- a/src/OVAL/probes/independent/textfilecontent54.c
+++ b/src/OVAL/probes/independent/textfilecontent54.c
@@ -260,7 +260,7 @@ static int process_file(const char *prefix, const char *path, const char *file, 
 	 * to return 'FTS_SL' and the presence of a valid target has to
 	 * be determined with stat().
 	 */
-	whole_path_with_prefix = oscap_sprintf("%s%s", prefix ? prefix : "", whole_path);
+	whole_path_with_prefix = oscap_path_join(prefix, whole_path);
 	if (stat(whole_path_with_prefix, &st) == -1)
 		goto cleanup;
 	if (!S_ISREG(st.st_mode))
--- a/src/OVAL/probes/independent/textfilecontent54.c
+++ b/src/OVAL/probes/independent/textfilecontent54.c
@@ -354,6 +354,7 @@ int probe_offline_mode_supported()
 
 void *probe_init(void)
 {
+  probe_setoption(PROBEOPT_OFFLINE_MODE_SUPPORTED, PROBE_OFFLINE_CHROOT);
   return NULL;
 }
 
@@ -375,8 +376,6 @@ int probe_main(probe_ctx *ctx, void *arg)
 #endif
 	OVAL_FTS    *ofts;
 	OVAL_FTSENT *ofts_ent;
-	char path_with_root[PATH_MAX + 1];
-	unsigned int root_len = 0;
 
         (void)arg;
 
@@ -505,22 +504,12 @@ int probe_main(probe_ctx *ctx, void *arg)
 	}
 #endif
 
-	path_with_root[PATH_MAX] = '\0';
-	if (ctx->offline_mode & PROBE_OFFLINE_OWN) {
-		strncpy(path_with_root, getenv("OSCAP_PROBE_ROOT"), PATH_MAX);
-		root_len = strlen(path_with_root);
-
-		if (path_with_root[root_len - 1] == FILE_SEPARATOR)
-			--root_len;
-	}
-
 	if ((ofts = oval_fts_open(path_ent, file_ent, filepath_ent, bh_ent, probe_ctx_getresult(ctx))) != NULL) {
 		while ((ofts_ent = oval_fts_read(ofts)) != NULL) {
 			if (ofts_ent->fts_info == FTS_F
 			    || ofts_ent->fts_info == FTS_SL) {
-				strncpy(path_with_root + root_len, ofts_ent->path, PATH_MAX - root_len);
 				// todo: handle return code
-				process_file(path_with_root, ofts_ent->file, &pfd);
+				process_file(ofts_ent->path, ofts_ent->file, &pfd);
 			}
 			oval_ftsent_free(ofts_ent);
 		}
--- a/src/OVAL/probes/independent/textfilecontent54.c
+++ b/src/OVAL/probes/independent/textfilecontent54.c
@@ -187,7 +187,13 @@ static SEXP_t *create_item(const char *path, const char *filename, char *pattern
 	if (oval_schema_version_cmp(over, OVAL_SCHEMA_VERSION(5.6)) < 0) {
 		se_filepath = NULL;
 	} else {
-		se_filepath = SEXP_string_newf("%s%c%s", path, FILE_SEPARATOR, filename);
+		const size_t path_len = strlen(path);
+		/* Avoid 2 slashes */
+		if (path_len >= 1 && path[path_len - 1] == FILE_SEPARATOR) {
+			se_filepath = SEXP_string_newf("%s%s", path, filename);
+		} else {
+			se_filepath = SEXP_string_newf("%s%c%s", path, FILE_SEPARATOR, filename);
+		}
 	}
 
         item = probe_item_create(OVAL_INDEPENDENT_TEXT_FILE_CONTENT, NULL,
@@ -220,12 +226,12 @@ struct pfdata {
 #endif
 };
 
-static int process_file(const char *path, const char *file, void *arg)
+static int process_file(const char *prefix, const char *path, const char *file, void *arg)
 {
 	struct pfdata *pfd = (struct pfdata *) arg;
 	int ret = 0, path_len, file_len, cur_inst = 0, fd = -1, substr_cnt,
 		buf_size = 0, buf_used = 0, ofs = 0, buf_inc = 4096;
-	char *whole_path = NULL, *buf = NULL;
+	char *whole_path = NULL, *whole_path_with_prefix = NULL, *buf = NULL;
 	SEXP_t *next_inst = NULL;
 	struct stat st;
 
@@ -254,12 +260,13 @@ static int process_file(const char *path, const char *file, void *arg)
 	 * to return 'FTS_SL' and the presence of a valid target has to
 	 * be determined with stat().
 	 */
-	if (stat(whole_path, &st) == -1)
+	whole_path_with_prefix = oscap_sprintf("%s%s", prefix ? prefix : "", whole_path);
+	if (stat(whole_path_with_prefix, &st) == -1)
 		goto cleanup;
 	if (!S_ISREG(st.st_mode))
 		goto cleanup;
 
-	fd = open(whole_path, O_RDONLY);
+	fd = open(whole_path_with_prefix, O_RDONLY);
 	if (fd == -1) {
 		SEXP_t *msg;
 
@@ -343,6 +350,7 @@ static int process_file(const char *path, const char *file, void *arg)
 	free(buf);
 	if (whole_path != NULL)
 		free(whole_path);
+	free(whole_path_with_prefix);
 
 	return ret;
 }
@@ -350,12 +358,6 @@ static int process_file(const char *path, const char *file, void *arg)
 int probe_offline_mode_supported()
 {
 	return PROBE_OFFLINE_OWN;
-}
-
-void *probe_init(void)
-{
-  probe_setoption(PROBEOPT_OFFLINE_MODE_SUPPORTED, PROBE_OFFLINE_CHROOT);
-  return NULL;
 }
 
 int probe_main(probe_ctx *ctx, void *arg)
@@ -503,12 +505,14 @@ int probe_main(probe_ctx *ctx, void *arg)
 		goto cleanup;
 	}
 #endif
-	if ((ofts = oval_fts_open(NULL, path_ent, file_ent, filepath_ent, bh_ent, probe_ctx_getresult(ctx))) != NULL) {
+	const char *prefix = getenv("OSCAP_PROBE_ROOT");
+
+	if ((ofts = oval_fts_open(prefix, path_ent, file_ent, filepath_ent, bh_ent, probe_ctx_getresult(ctx))) != NULL) {
 		while ((ofts_ent = oval_fts_read(ofts)) != NULL) {
 			if (ofts_ent->fts_info == FTS_F
 			    || ofts_ent->fts_info == FTS_SL) {
 				// todo: handle return code
-				process_file(ofts_ent->path, ofts_ent->file, &pfd);
+				process_file(prefix, ofts_ent->path, ofts_ent->file, &pfd);
 			}
 			oval_ftsent_free(ofts_ent);
 		}
--- a/src/OVAL/probes/independent/textfilecontent54.c
+++ b/src/OVAL/probes/independent/textfilecontent54.c
@@ -503,8 +503,7 @@ int probe_main(probe_ctx *ctx, void *arg)
 		goto cleanup;
 	}
 #endif
-
-	if ((ofts = oval_fts_open(path_ent, file_ent, filepath_ent, bh_ent, probe_ctx_getresult(ctx))) != NULL) {
+	if ((ofts = oval_fts_open(NULL, path_ent, file_ent, filepath_ent, bh_ent, probe_ctx_getresult(ctx))) != NULL) {
 		while ((ofts_ent = oval_fts_read(ofts)) != NULL) {
 			if (ofts_ent->fts_info == FTS_F
 			    || ofts_ent->fts_info == FTS_SL) {
--- a/src/OVAL/probes/independent/textfilecontent54.c
+++ b/src/OVAL/probes/independent/textfilecontent54.c
@@ -507,7 +507,7 @@ int probe_main(probe_ctx *ctx, void *arg)
 #endif
 	const char *prefix = getenv("OSCAP_PROBE_ROOT");
 
-	if ((ofts = oval_fts_open(prefix, path_ent, file_ent, filepath_ent, bh_ent, probe_ctx_getresult(ctx))) != NULL) {
+	if ((ofts = oval_fts_open_prefixed(prefix, path_ent, file_ent, filepath_ent, bh_ent, probe_ctx_getresult(ctx))) != NULL) {
 		while ((ofts_ent = oval_fts_read(ofts)) != NULL) {
 			if (ofts_ent->fts_info == FTS_F
 			    || ofts_ent->fts_info == FTS_SL) {
--- a/src/OVAL/probes/independent/xmlfilecontent.c
+++ b/src/OVAL/probes/independent/xmlfilecontent.c
@@ -320,7 +320,7 @@ int probe_main(probe_ctx *ctx, void *arg)
 	pfd.filename_ent = filename_ent;
         pfd.ctx = ctx;
 
-	if ((ofts = oval_fts_open(NULL, path_ent, filename_ent, filepath_ent, behaviors_ent, probe_ctx_getresult(ctx))) != NULL) {
+	if ((ofts = oval_fts_open_prefixed(NULL, path_ent, filename_ent, filepath_ent, behaviors_ent, probe_ctx_getresult(ctx))) != NULL) {
 		while ((ofts_ent = oval_fts_read(ofts)) != NULL) {
 			process_file(ofts_ent->path, ofts_ent->file, &pfd);
 			oval_ftsent_free(ofts_ent);
--- a/src/OVAL/probes/independent/xmlfilecontent.c
+++ b/src/OVAL/probes/independent/xmlfilecontent.c
@@ -320,7 +320,7 @@ int probe_main(probe_ctx *ctx, void *arg)
 	pfd.filename_ent = filename_ent;
         pfd.ctx = ctx;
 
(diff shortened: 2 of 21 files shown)


jan-cerny:
I would like to add a test for this PR, that would prevent issue #1001, but I think it would be better to do that in a separate pull request after https://github.com/OpenSCAP/openscap/pull/999 is merged, because https://github.com/OpenSCAP/openscap/pull/999 introduces a new directory `offline_mode` under `tests`, and I think the test for this PR would perfectly fit to the `offline_mode` directory.

mpreisler:
What about changing process_file to take prefix and the path separately? As long as we keep it separate there is no need to alter any regexes (that's an impossible problem) or filter anything for the results. I think we can fix this properly instead of reverting.

jan-cerny:
How that would help? The `process_file` function is called only if the file was found by `oval_fts_open` and `oval_fts_read`.

mpreisler:
> How that would help? The process_file function is called only if the file was found by oval_fts_open and oval_fts_read.

We also need to improve oval_fts_*. But overall, this is a problem that can be solved and can be solved fairly easily. I don't think we should give up on this and do chroot here.

jan-cerny:
oval_fts_read traverses the filesystem according to OVAL file behaviors. For the files it finds it converts them to SEXP and then it comapres the SEXP with the SEXPs that are parameters of the function. These SEXPs however don't necessarily contain the path, but can reference a variable, filter, ...

I can think that we add a prefix to the paths that are traversed and before every comparison we strip the prefix and after every comparison we add the prefix back, but I don't know if it could work with the regexes and so on

mpreisler:
we need to pass a prefix (default to NULL) to the function and whenever we do any file opening reading, etc we prefix before we open the file. Let's not do any postprocess preprocess manipulation to avoid polluting the result files.

jan-cerny:
I am not sure if I should do offline mode in sysctl probe. The sysctl probe reads from `/proc` and in mounted containers and mounted VMs there isn't `/proc` so it won't be able to get any data.

mpreisler:
@jan-cerny What's the relationship between #1000 and #1002? Which should be merged first? Is this still WIP or not? It looks like it's ready to merge but you keep adding commits.

mpreisler:
> I am not sure if I should do offline mode in sysctl probe. The sysctl probe reads from /proc and in mounted containers and mounted VMs there isn't /proc so it won't be able to get any data.

Separate PR.

mpreisler:
@jan-cerny Also, please fix the description of this PR, it's outdated and misleading now.

jan-cerny:
@mpreisler The relationship is that the PR #1000 improves the offline mode in generic common code, but PR #1002 fixes specific probes (starting by issue #1001).

None of them should be merged first.

It was not WIP. I think it is ready as well.

jan-cerny:
I'm going to rebase this on the top of the current upstream maint-1.2 and resolve the conflicts.

jan-cerny:
I have rebased the pull request on the top of upstream maint-1.2 branch. I have resolved the merge conflicts.

src/OVAL/probes/independent/textfilecontent54.c
```diff
@@ -254,12 +260,13 @@ static int process_file(const char *path, const char *file, void *arg)
 	 * to return 'FTS_SL' and the presence of a valid target has to
 	 * be determined with stat().
 	 */
-	if (stat(whole_path, &st) == -1)
+	whole_path_with_prefix = oscap_sprintf("%s%s", prefix ? prefix : "", whole_path);
```
mpreisler: I wish you declared the variable here and initialized it at the same time. We don't have to follow C89. I don't care enough to hold up this PR because of it though.

src/OVAL/probes/independent/xmlfilecontent.c
```diff
@@ -128,12 +128,18 @@ static int process_file(const char *path, const char *filename, void *arg)
 		whole_path[path_len] = FILE_SEPARATOR;
 		++path_len;
 	}
+	dI("%s", path);
```
mpreisler: leftover debug message. If you want it there add some actual message.

src/OVAL/probes/independent/xmlfilecontent.c
```diff
@@ -128,12 +128,18 @@ static int process_file(const char *path, const char *filename, void *arg)
...
 
 	memcpy(whole_path + path_len, filename, filename_len + 1);
 
-	/* evaluate xpath */
+	if (prefix == NULL) {
+		doc = xmlParseFile(whole_path);
+	} else {
+		char *path_with_prefix = oscap_sprintf("%s%s", prefix, whole_path);
```
mpreisler: comment why you don't have to insert PATH_SEPARATOR in between. this is not obvious

src/OVAL/probes/oval_fts.c
```diff
@@ -102,17 +102,28 @@ static OVAL_FTSENT *OVAL_FTSENT_new(OVAL_FTS *ofts, FTSENT *fts_ent)
 	ofts_ent = oscap_talloc(OVAL_FTSENT);
 
 	ofts_ent->fts_info = fts_ent->fts_info;
+	const size_t shift = ofts->prefix ? strlen(ofts->prefix) : 0;
```
mpreisler: comment what this is for

src/OVAL/probes/oval_fts.c
```diff
@@ -651,6 +662,11 @@ static int process_pattern_match(const char *path, pcre **regex_out)
 #undef TEST_PATH2
 
 OVAL_FTS *oval_fts_open(SEXP_t *path, SEXP_t *filename, SEXP_t *filepath, SEXP_t *behaviors, SEXP_t* result)
+{
+	return oval_fts_open_prefixed(NULL, path, filename, filepath, behaviors, result);
+}
+
+OVAL_FTS *oval_fts_open_prefixed(const char *prefix, SEXP_t *path, SEXP_t *filename, SEXP_t *filepath, SEXP_t *behaviors, SEXP_t* result)
```
mpreisler: <3

src/OVAL/probes/oval_fts.c
```diff
@@ -811,6 +827,12 @@ OVAL_FTS *oval_fts_open(SEXP_t *path, SEXP_t *filename, SEXP_t *filepath, SEXP_t
 		paths[0] = strdup("/");
 	}
 
+	if (prefix != NULL) {
+		char *path_with_prefix = oscap_sprintf("%s%s", prefix, paths[0]);
```
mpreisler: again, comment why you are *sure* you don't need FILE_SEPARATOR here.

src/OVAL/probes/oval_fts.c
```diff
@@ -811,6 +827,12 @@ OVAL_FTS *oval_fts_open(SEXP_t *path, SEXP_t *filename, SEXP_t *filepath, SEXP_t
...
+		free((void *) paths[0]);
+		paths[0] = path_with_prefix;
+	}
+	dI("Opening file '%s'.", paths[0]);
```
mpreisler: this is twice in the same function.

src/OVAL/probes/unix/file.c
```diff
@@ -272,17 +274,25 @@ static int file_cb (const char *p, const char *f, void *ptr)
 	if (f == NULL) {
 		st_path = p;
 	} else {
-		snprintf (path_buffer, sizeof path_buffer, "%s/%s", p, f);
+		const size_t p_len = strlen(p);
+		/* Avoid 2 slashes */
+		if (p_len >= 1 && p[p_len - 1] == FILE_SEPARATOR) {
+			snprintf(path_buffer, sizeof path_buffer, "%s%s", p, f);
+		} else {
+			snprintf(path_buffer, sizeof path_buffer, "%s/%s", p, f);
```
mpreisler: Use FILE_SEPARATOR instead of hardcoded /

src/OVAL/probes/unix/fileextendedattribute.c
```diff
@@ -97,20 +99,31 @@ static int file_cb (const char *p, const char *f, void *ptr)
 	if (f == NULL) {
 		st_path = p;
 	} else {
-		snprintf (path_buffer, sizeof path_buffer, "%s/%s", p, f);
+		const size_t p_len = strlen(p);
+		/* Avoid 2 slashes */
+		if (p_len >= 1 && p[p_len - 1] == FILE_SEPARATOR) {
+			snprintf(path_buffer, sizeof path_buffer, "%s%s", p, f);
+		} else {
+			snprintf(path_buffer, sizeof path_buffer, "%s/%s", p, f);
```
mpreisler: /

src/OVAL/probes/unix/fileextendedattribute.c
```diff
@@ -97,20 +99,31 @@ static int file_cb (const char *p, const char *f, void *ptr)
...
+		}
 		st_path = path_buffer;
 	}
 
         SEXP_init(&xattr_name);
 
+	char *st_path_with_prefix = oscap_sprintf("%s%s", prefix ? prefix : "", st_path);
```
mpreisler: no separator?

mpreisler:
I found a lot of minor issues. I am especially concerned whether you have a guarantee that `prefix` ends with a slash and that it's the right slash. If you really assume that add an assert or comment.

src/OVAL/probes/independent/xmlfilecontent.c
jan-cerny: OK

src/OVAL/probes/oval_fts.c
jan-cerny: ?

src/OVAL/probes/unix/file.c
jan-cerny: OK, but It was there

src/OVAL/probes/oval_fts.c
mpreisler: https://www.internetslang.com/_26lt_3B3-meaning-definition.asp

src/OVAL/probes/unix/file.c
mpreisler: IMO just make it consistent. I don't care if it's changed here or in another PR. I would change it here to save effort.

src/OVAL/probes/oval_fts.c
jan-cerny: What is wrong?

src/OVAL/probes/independent/xmlfilecontent.c
jan-cerny: I don't know, the `whole_path` should come from `FTSENT fts_path`, which is described in `man 3 fts` as:

> The  path  for  the file relative to the root of the traversal.  This path contains the path specified to fts_open() as a prefix. 

but in  my cases it was not a relative path, it started with `/`.

How would you understand that?

src/OVAL/probes/oval_fts.c
jan-cerny: I removed the second one.

src/OVAL/probes/independent/xmlfilecontent.c
mpreisler: Your reply is precisely why I am asking about this. To write robust code you need to be sure about this or insert logic that makes sure about this.

src/OVAL/probes/oval_fts.c
mpreisler: Nothing is wrong, I just commented that I like how you handled avoiding ABI/API break. <3 is a ASCII art for a heart.

src/OVAL/probes/independent/xmlfilecontent.c
jan-cerny: I think it would be useful if we have something like `os.path.join` in Python that would merge 2 parhs.

mpreisler: Yes, especially if it's in common and can be used all over the codebase.

src/common/util.c
```diff
@@ -229,3 +230,30 @@ char *oscap_expand_ipv6(const char *input)
 
 	return ret;
 }
+
+char *oscap_path_join(const char *path1, const char *path2)
+{
+	if (path1 == NULL) {
+		return oscap_strdup(path2);
```
mpreisler: path2 can also be NULL at this point, right?

src/common/util.c
```diff
@@ -229,3 +230,30 @@ char *oscap_expand_ipv6(const char *input)
...
+	}
+	if (path2 == NULL) {
+		return oscap_strdup(path1);
+	}
+	size_t path1_len = strlen(path1);
+	size_t path2_len = strlen(path2);
```
mpreisler: const

src/common/util.c
```diff
@@ -229,3 +230,30 @@ char *oscap_expand_ipv6(const char *input)
...
+	size_t path2_shift = 0;
+	while (path1_len >= 1 && path1[path1_len - 1] == PATH_SEPARATOR) {
+		path1_len--;
+	}
+	while (path2_shift < path2_len && path2[path2_shift] == PATH_SEPARATOR) {
+		path2_shift++;
+	}
+	path2_len -= path2_shift;
+	const size_t joined_path_len = path1_len + 1 + path2_len;
```
mpreisler: this is a nitpick but for readability I would except the +1 to be at the end because it is for the \0 terminator which is at the end.

src/common/util.h
```diff
@@ -390,6 +390,15 @@ OSCAP_HIDDEN_END;
 /// Print to a newly allocated string using varialbe arguments.
 char *oscap_sprintf(const char *fmt, ...);
 
+/**
+ * Join 2 paths intelligently
+ * Caller is responsible to free the returned pointer.
```
mpreisler: responsible for..

src/common/util.h
```diff
@@ -390,6 +390,15 @@ OSCAP_HIDDEN_END;
 /// Print to a newly allocated string using varialbe arguments.
 char *oscap_sprintf(const char *fmt, ...);
 
+/**
+ * Join 2 paths intelligently
```
mpreisler: intelligently?!

src/common/util.h
```diff
@@ -390,6 +390,15 @@ OSCAP_HIDDEN_END;
...
+ * Caller is responsible to free the returned pointer.
+ * @param path1 first path
+ * @param path2 second path
+ * @return Join of path1 and path2.
```
mpreisler: These docs need improvement for sure

src/common/util.c
jan-cerny: yes, but oscap_strdup checks if its argument is NULL

src/common/util.h
jan-cerny: yes

src/common/util.c
jan-cerny: I think it is changed below

tests/offline_mode/test_textfilecontent54.sh
```diff
@@ -0,0 +1,58 @@
...
+
+    rm -rf "$temp_dir"
+    rm -f "$result"
+}
+
+# Testing.
+
+test_init "test_textfilecontent54.log"
+
+test_run "test_textfilecontent54" test_textfilecontent54
```
mpreisler: this is not a good name for this test. it should have offline in the name, otherwise it's called exactly the same as another test that is not offline

src/common/util.c
mpreisler: ah, ok

tests/offline_mode/test_textfilecontent54.sh
mpreisler: (try looking for this test in jenkins results)

mpreisler: Actually I am pretty sure this test is never run, could you double check this?

mpreisler: Nvm, I had a typo in my grep string:

```
Making check in offline_mode
make[3]: Entering directory `/home/jenkins/workspace/openscap-pull-requests/openscap-1.2.17/_build/tests/offline_mode'
make  check-TESTS
make[4]: Entering directory `/home/jenkins/workspace/openscap-pull-requests/openscap-1.2.17/_build/tests/offline_mode'

----------------------------------------------------------------------
+ test_system_info                                            [ PASS ]
--------------------------------------------------
See tests/offline_mode/test_system_info.log.

PASS: test_system_info.sh

----------------------------------------------------------------------
+ test_textfilecontent54                                      [ PASS ]
--------------------------------------------------
See tests/offline_mode/test_textfilecontent54.log.
```

I still wish both tests had better names.

jan-cerny: I checked and it is in Jenkins output.

I meant the `offline_mode` directory should contain only tests that are related only to offline mode.

But if you wish I will rename it.

src/common/util.c
jan-cerny: Yes, both path1_len and path2_len are changed a few lines below.

mpreisler:
I have done some limited testing and this seems to work well. ACK.

Great work, Jan!

PR #1003, opened 2018-03-27
Avoid passing NULL parameters to oscap_seterr

jlcharton:
While some platforms (e.g. Linux) accept stuff like printf("%s\n", NULL) others (e.g. Solaris) don't like this and raise a SIGSEGV.


mpreisler:
I would have preferred if you declared a variable and assigned on the same line but I don't think it's important enough to hold up this PR.

Thank you for your contribution!

PR #1004, opened 2018-03-27
Add Windows support for system_info probe

jan-cerny:
The system_info probe retreives operating system name, operating
system version, names of network interfaces, their MAC and IP
addresses. On Windows, we can get these information using Windows
API calls and then populate them into OVAL item.

See the example OVAL Results Document.
[results.zip](https://github.com/OpenSCAP/openscap/files/1853105/results.zip)


src/OVAL/probes/independent/system_info.c
```diff
@@ -535,7 +681,35 @@ static char * _offline_chroot_get_hname(void)
...
+
+	hmodule = LoadLibraryW(L"ntdll.dll");
+	RtlGetVersion = (fRtlGetVersion *)GetProcAddress(hmodule, "RtlGetVersion");
+
+	memset(&osverinfo, 0, sizeof(OSVERSIONINFOEXW));
+	osverinfo.dwOSVersionInfoSize = sizeof(OSVERSIONINFOEXW);
+	RtlGetVersion(&osverinfo);
+
+	char *version = malloc(VERSION_LEN);
+	snprintf(version, VERSION_LEN, "%d.%d.%d", osverinfo.dwMajorVersion, osverinfo.dwMinorVersion, osverinfo.dwBuildNumber);
```
mpreisler: any reason not to use oscap_sprintf?

mpreisler:
A lot of this code breaks our code conventions, mainly the variable names.

src/OVAL/probes/independent/system_info.c
jan-cerny: I always forget about existence of this function :smile:

jan-cerny:
> A lot of this code breaks our code conventions, mainly the variable names.

How it should look like?

mpreisler:
Labeled do not merge becuase this is blocked by another PR.

jan-cerny:
I have rebased this on the top of master branch and I have resolved conflicts with the master branch.

mpreisler:
@openscap-ci test this please

mpreisler:
> How it should look like?

It would be really strenuous to mention every single instance of it breaking our code conventions, so here is one example:
```
PIP_ADAPTER_ADDRESSES pAddresses = NULL;
PIP_ADAPTER_ADDRESSES pCurrAddresses = NULL;
```
We do not use camelcase variable names. It's not really a big deal, I guess if you took the code from other sources cite them properly, make sure licensing works and then we can keep them in their original conventions. If you wrote this from scratch then why are you using WinAPI conventions?

src/OVAL/probes/independent/system_info.c
```diff
@@ -295,6 +305,142 @@ static int get_ifs(SEXP_t *item)
...
+	ULONG outBufLen = WORKING_BUFFER_SIZE;
+	ULONG family = AF_UNSPEC; // both IPv4 and IPv6
+	ULONG flags = GAA_FLAG_INCLUDE_PREFIX;
+	ULONG iterations = 0;
+
+	/* Try to get information about network adapters and their adresses */
+	do {
+		pAddresses = malloc(WORKING_BUFFER_SIZE);
+		dwRetVal = GetAdaptersAddresses(family, flags, NULL, pAddresses, &outBufLen);
+		if (dwRetVal == ERROR_BUFFER_OVERFLOW) {
```
mpreisler: if this fails once why would it ever work again in any other try? shouldn't you increase buffer size if it fails or something like that?

src/OVAL/probes/independent/system_info.c
```diff
@@ -295,6 +305,142 @@ static int get_ifs(SEXP_t *item)
...
+	while (pCurrAddresses != NULL) {
+		/* MAC address*/
+		char *mac_address_str;
+		if (pCurrAddresses->PhysicalAddressLength != 0) {
+			/* n bytes of MAC address will be printed as:
+			(2 * n) hexa numbers + (n - 1) delimiters + 1 terminating null byte = 3 * n */
+			const size_t mac_address_str_len = pCurrAddresses->PhysicalAddressLength * 3;
+			mac_address_str = malloc(mac_address_str_len);
+
+			for (int i = 0; i < (int)pCurrAddresses->PhysicalAddressLength;
```
mpreisler: if length is unsigned and i is also unsigned why not just use unsigned instead of casting here?

src/common/util.c
```diff
@@ -338,3 +339,13 @@ char *oscap_path_join(const char *path1, const char *path2)
 	joined_path[joined_path_len] = '\0';
 	return joined_path;
 }
+
+#ifdef _WIN32
+char *oscap_wstr_to_str(wchar_t *wstr)
```
mpreisler: wstr and wchar have a ton of definitions, I suggest to prefix this with win to make it clear this is windows only. wchar can have different widths depending on the OS and compiler.

src/common/util.c
```diff
@@ -338,3 +339,13 @@ char *oscap_path_join(const char *path1, const char *path2)
...
+{
+	size_t required_size = WideCharToMultiByte(CP_UTF8, 0, wstr, -1, NULL, 0, NULL, NULL);
```
mpreisler: const

src/common/util.c
jan-cerny: +1

jan-cerny:
I see, I should use variable names in our conventions. I usually named them after the function parameters that I have found in docs. I will change that.

I am not sure about the data types, though. Some of the Windows data types are just aliases for C dtata types. For example `DWORD` is defined as `typedef unsigned long DWORD` but I don't know if it is preferred to use `unsigned long` or `DWORD` in the code that interacts with Windows API.

src/OVAL/probes/independent/system_info.c
jan-cerny: Nice catch, thatk you. I think here  should be outBufLen passed to malloc as parameter.

jan-cerny:
I have also realized that I should refactor printing the error messages and use there wide character strings there. I will try to change that.

mpreisler:
> For example DWORD is defined as typedef unsigned long DWORD but I don't know if it is preferred to use unsigned long or DWORD in the code that interacts with Windows API.

Use Windows typedefs in this case, using a different typedef is not a coding style, that's different code with potentially different behavior on different platform/OS.

mpreisler:
@openscap-ci test this please

src/OVAL/probes/independent/system_info_probe.c
```diff
@@ -565,11 +731,39 @@ int system_info_probe_main(probe_ctx *ctx, void *arg)
...
-	hname = oscap_strdup(unknown);
+	WCHAR computer_name_wstr[MAX_COMPUTERNAME_LENGTH + 1];
+	DWORD computer_name_len = MAX_COMPUTERNAME_LENGTH + 1;
+	GetComputerNameW(computer_name_wstr, &computer_name_len);
+
+	SYSTEM_INFO si;
+	GetNativeSystemInfo(&si);
+	const char *arch;
+	switch (si.wProcessorArchitecture) {
+	case PROCESSOR_ARCHITECTURE_AMD64:
```
jan-cerny: Maybe it would be better to put this out from main function and create something like `static const char *get_windows_architecture(void)`.

jan-cerny: done

jan-cerny:
I have improved the code and I have rebased it on master. I have changed variables and functions names, refactored the error messages, and I have improved the code style.

jan-cerny:
@openscap-ci test this please

mpreisler:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

mpreisler:
ACK

PR #1006, opened 2018-04-06
Create and remove temporary directory on Windows

jan-cerny:
This commit fixes `oscap_acquire_temp_dir` which was broken on Windows.

First issue was that it was not able to create a temp directory if the parent directory didn't exist. The second issue was that it created a directory with always same name.

The commit also provides a Windows implementation of `oscap_acquire_cleanup_dir`, which was completely missing, and which removes the temporary directory created by `oscap_acquire_temp_dir`.

These functions are used when working with SCAP DataStreams.


mpreisler:
ACK

PR #1007, opened 2018-04-06
Add Windows Registry probe

jan-cerny:
This probe implements OVAL Windows Registry test according
to OVAL specification. The Windows Registry is queried using
Windows API calls. The commit also introduces new directory
for Windows probes and modifies CMakeLists to build them.

The registry_test in OVAL Definitions Schema:
https://oval.mitre.org/language/version5.11/ovaldefinition/documentation/windows-definitions-schema.html#registry_test
The registry_item in OVAL System Characteristics Schema:
https://oval.mitre.org/language/version5.11/ovalsc/documentation/windows-system-characteristics-schema.html#registry_item

PR #1009, opened 2018-04-13
Add accesstoken Windows probe

jan-cerny:
This probe implements the accesstoken_test defined in OVAL Windows Definitions Schema. The probe is able to enumerate privileges for local and global users and groups.

The accesstoken_test in OVAL Windows Definitions Schema:
https://oval.mitre.org/language/version5.10/ovaldefinition/documentation/windows-definitions-schema.html#accesstoken_test

The accesstoken_item in OVAL Windows System Characteristics Schema:
https://oval.mitre.org/language/version5.10/ovalsc/documentation/windows-system-characteristics-schema.html#accesstoken_item

Marking as do not merge because it is blocked by #1007.


mpreisler:
ACK

PR #1010, opened 2018-04-13
Update "Build OpenSCAP Using Visual Studio" section in User Manual

jan-cerny:
It is already possible to build family and system info probe using Microsoft Visual Studio 2017. We don't need to tell developers contributing on Windows that they have to build OpenSCAP without probes.


mpreisler:
ACK

PR #1012, opened 2018-04-18
Fix oscap_dirname on Windows

jan-cerny:
oscap_dirname must work same as dirname(). However on Windows
it works different. We need to fix it according to 'man 3 dirname':
If path is a null pointer or points to an empty string, then both
dirname() and basename() return the string ".".
If path does not contain a slash, dirname() returns the string ".".


mpreisler:
ACK

PR #1017, opened 2018-04-20
Adding with-sce-script-dir flag to configure.ac. Fixes #1016

### LINKED ISSUE #1016: Configure flag for sce script directory
The SCE feature currently stores and executes shell scripts in /tmp. Some security benchmarks require /tmp to be non-executable, causing the scan to fail.

A solution would be to add a parameter to the configure.ac to customize the directory used by SCE for storage and execution of scripts.

JennyUWaterloo:
When configure is run with --enable-sce, add a new parameter --with-sce-script-dir to set the directory path for SCE to store and execute script files (default /tmp). Configure will append -DSCE_SCRIPT_DIR to CFLAGS (P_tmpdir is always defined by stdio.h, which is included by curl.h. Therefore we added SCE_SCRIPT_DIR to prevent the provided cflag from being overwritten.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1152,15 +1152,15 @@ AC_ARG_ENABLE([sce],
      esac],[sce=no])
 
 
-AC_ARG_WITH([sce-script-dir],
-     [AS_HELP_STRING([--with-sce-script-dir],
+AC_ARG_WITH([oscap-temp-dir],
+     [AS_HELP_STRING([--with-oscap-temp-dir],
      [use different temporary directory to execute sce scripts [default=/tmp]])],
      [],
-     [with_sce_script_dir="/tmp"])
+     [with_oscap_temp_dir="/tmp"])
 
 if test "x${sce}" = xyes; then
   AC_DEFINE([ENABLE_SCE], [1], [compilation of script check engine enabled])
-  CFLAGS="$CFLAGS -DSCE_SCRIPT_DIR=\\\"${with_sce_script_dir}\\\"" # double escape needed for compilation on some systems
+  CFLAGS="$CFLAGS -DOSCAP_TEMP_DIR=\\\"${with_oscap_temp_dir}\\\"" # double escape needed for compilation on some systems
 fi
 
 AC_ARG_ENABLE([util-oscap],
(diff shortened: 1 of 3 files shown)


mpreisler:
@openscap-ci test this please

JennyUWaterloo:
@openscap-ci test this please

mpreisler:
@openscap-ci add to whitelist

configure.ac
```diff
@@ -1160,7 +1160,7 @@ AC_ARG_WITH([sce-script-dir],
 
 if test "x${sce}" = xyes; then
   AC_DEFINE([ENABLE_SCE], [1], [compilation of script check engine enabled])
-  CFLAGS="$CFLAGS -DSCE_SCRIPT_DIR=\"${with_sce_script_dir}\""
+  CFLAGS="$CFLAGS -DSCE_SCRIPT_DIR=\\\"${with_sce_script_dir}\\\""
```
matejak: Please add a comment why the double escape is needed, it is not obvious.

JennyUWaterloo: It's not obvious to me either. The project compiled on my ubuntu container with single escaped quotes, but failed to compile on CI (https://jenkins.open-scap.org/job/openscap-pull-requests/612/console - ```error: 'tmp' undeclared```) until the quotes were double escaped.

matejak: This is interesting. I am not 100% sure about what the cause may be. People are quite confused in this discussion too:
https://stackoverflow.com/questions/2410976/how-to-define-a-string-literal-in-gcc-command-line/15227471
The more robust, but also more arcane solution would be to write some macro magic to stringify the incoming macro, but I am fine if you just include a comment saying that the double escape is a tricky manner, very likely required in connection with jenkins CI.
Last but not least, please add the `configure.ac` code changes to `ac_probes/configure.ac.tpl` (I usually use `vimdiff` to find the right place where to add the lines.)

JennyUWaterloo:
@matejak I've applied the requested changes

matejak:
@JennyUWaterloo thank you for your great observation! I see that something like `OSCAP_TEMP_DIR` should work fine, so feel free to go ahead with the rename.

matejak:
Great contribution, thank you for your perseverance! It might have been frustrating that this PR was not so straightforward, but it was all for the good reasons. Thank you!

mpreisler:
@jan-cerny this will need a careful merge into master and we will need to redo this with cmake.

jan-cerny:
@mpreisler That's awesome! Thanks.

PR #1025, opened 2018-04-27
Remove '\r' characters from help output

### LINKED ISSUE #1023: OpenSCAP options are not shown in --help output
Hello, I often use OpenSCAP build from current master branch on Windows 10.

I want to run a help to see which command line options are available. But I don't see the options, only their descriptions.

Please see the attached screenshot for `oscap xccdf eval --help`.

screenshot from 2018-04-26 09-39-13

The problem is caused by `\r` in the help strings. For example: https://github.com/OpenSCAP/openscap/blob/maint-1.2/utils/oscap-xccdf.c#L154

What is your opinion on this problem? Should  `\r` be removed everywhere completely?

jan-cerny:
Instead of a hack with '\r\t\t\t\t' we will indent the help options manually.
Fixes #1023.

### DIFF
--- a/utils/oscap-cvrf.c
+++ b/utils/oscap-cvrf.c
@@ -58,8 +58,8 @@ static struct oscap_module CVRF_EVALUATE_MODULE = {
 	.opt_parser = getopt_cvrf,
 	.func = app_cvrf_evaluate,
 	.help = "Options:\n"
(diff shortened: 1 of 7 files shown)


mpreisler:
ACK

PR #1026, opened 2018-04-27
openSUSE CPE update

abergmann:
I've updated the openSUSE CPEs to match the updated entries inside the official NIST CPE dictionary.

### DIFF
--- a/cpe/openscap-cpe-dict.xml
+++ b/cpe/openscap-cpe-dict.xml
@@ -165,6 +165,14 @@
             <title xml:lang="en-us">openSUSE Leap 42.2</title>
             <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.opensuse:def:422</check>
       </cpe-item>
+      <cpe-item name="cpe:/o:opensuse:leap:42.3">
+            <title xml:lang="en-us">openSUSE Leap 42.3</title>
+            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.opensuse:def:423</check>
+      </cpe-item>
+      <cpe-item name="cpe:/o:opensuse:leap:15.0">
+            <title xml:lang="en-us">openSUSE Leap 15.0</title>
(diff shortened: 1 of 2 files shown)


mpreisler:
@openscap-ci add to whitelist

mpreisler:
We might want to keep the old CPEs because people could be relying on them.

jan-cerny:
@mpreisler But if the old CPEs are not official, they cannot be found in https://nvd.nist.gov/products/cpe/search ?

abergmann:
The entry `cpe:/o:novell:leap:42.2` is not listed inside the official CPE dictionary (_I just noticed_). And `cpe:/o:novell:leap:42.1` has a `deprecated_by="cpe:/o:opensuse:leap:42.1"` set.

Both, openSUSE Leap 42.1 and 42.2, reached already End-of-Life (EOL), so only openSUSE Leap 42.3 is currently active.

Therefore (IMHO) the impact of changing the CPEs is quite limited, as the CPE for openSUSE Leap 42.3 was just added by this pull-request.

jan-cerny:
This is interesting.
We maybe should keep the `cpe:/o:novell:leap:42.1` due to backwards compatibility

abergmann:
I've also pushed [scap-security-guide/commit/b491a77](https://github.com/OpenSCAP/scap-security-guide/commit/b491a7779c20be9a3bcd65bc45132df46d2d94fc) that changed the openSUSE Leap 42.1 CPE from "novell" to "opensuse" and it was already accepted.

The "novell" CPEs where never used on the openSUSE side. So my aim was (and still is) to get rid of all incorrect syntax.

jan-cerny:
@abergmann OK, thank you for clarification.

I would merge this.  Do you agree @mpreisler ?

abergmann:
@mpreisler, any update here? Please let me know if something needs to be adjusted.

jan-cerny:
@mpreisler Which of the solutions do you prefer?

matusmarhefka:
Please keep the old novell CPEs, even though they are deprecated we want to retain backward compatibility in `maint-1.2` branch.

mpreisler:
@openscap-ci add to whitelist

mpreisler:
Thank you! Sorry this took so long to review.

Issue #939, opened 2017-12-15
test probes/rpmverifyfile/all.sh fails on RHEL6 on master


jan-cerny:
The OVAL result for `capabilities_differ` is `not performed`.

I think there is some problem with incorrect usage C preprocessor macros in the rpmverifyfile_probe or in the CMake script that detects RPM version.

jan-cerny:
```
Definition oval:x:def:1: true
Evaluation done.
+ rpmverifyfile probe test with OVAL 5.11.1                   [ FAIL ]
Failed: expected count: 1, real count: 0, xpath: 'oval_results/results/system/oval_system_characteristics/system_data/lin-sys:rpmverifyfile_item/lin-sys:capabilities_differ[text()="not performed"]'
RESULT: FAILED

TEST: rpmverifyfile probe test with OVAL 5.11
Definition oval:x:def:1: true
Evaluation done.
+ rpmverifyfile probe test with OVAL 5.11                     [ PASS ]
RESULT: PASSED
```

jan-cerny:
This is a big problem for every upstream pull request on master on Jenkins!

jan-cerny:
We don't focus on RHEL6 in master, therefore we remove the BLOCKER label.

evgenyz:
Changing the milestone as 1.3 is not supposed to be released for RHEL6/7.

jan-cerny:
We don't build master branch and maint-1.3 branch on RHEL 6 anymore.

PR #1027, opened 2018-04-30
Add IP addresses in XCCDF results on Windows

jan-cerny:
I found this code in my `git stash`, so I think it would be useful
to create a commit from it and share it on Github.
There is a second place where we collect IP addresses of the
scanned machine. These are saved to XCCDF results and are shown
in HTML report. So this change is very visible.
I understand that XCCDF and OVAL are different things, however
I don't like that this copies a lot of code from the system info
probe.

### DIFF
--- a/src/XCCDF/result.c
+++ b/src/XCCDF/result.c
@@ -28,8 +28,19 @@
 #include <string.h>
 
 #ifdef _WIN32
+ /* By defining WIN32_LEAN_AND_MEAN we ensure that Windows.h won't include
+  * winsock.h, which would conflict with symbols from WinSock2.h.
+  */
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#include <WinSock2.h>
+#include <Iphlpapi.h>
+#include <Windows.h>
+#include <ws2def.h>
 #include <io.h>
-#include <windows.h>
+#include <winternl.h>
+#include <io.h>
 #include <lmcons.h>
 #else
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


mpreisler:
@openscap-ci test this please

jan-cerny:
#939 happened again

mpreisler:
@openscap-ci test this please

src/XCCDF/result.c
```diff
@@ -28,8 +28,19 @@
...
+  * winsock.h, which would conflict with symbols from WinSock2.h.
+  */
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#include <WinSock2.h>
+#include <Iphlpapi.h>
+#include <Windows.h>
+#include <ws2def.h>
+#include <io.h>
```
mpreisler: included twice

src/XCCDF/result.c
```diff
@@ -28,8 +28,19 @@
...
 #ifdef _WIN32
+ /* By defining WIN32_LEAN_AND_MEAN we ensure that Windows.h won't include
+  * winsock.h, which would conflict with symbols from WinSock2.h.
+  */
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#include <WinSock2.h>
+#include <Iphlpapi.h>
+#include <Windows.h>
```
mpreisler: why capital W?

src/XCCDF/result.c
jan-cerny: ok, thank you

src/XCCDF/result.c
jan-cerny: I don't know.

jan-cerny:
I have removed second io.h and made windows.h small.

PR #1028, opened 2018-05-04
Python 3 compatibility fixes

jan-cerny:
This PR changes our Python scripts to work in both Python 2 and 3.

### DIFF
--- a/tests/sce/script_tester.py
+++ b/tests/sce/script_tester.py
@@ -64,6 +64,9 @@ import subprocess
 
 for file_path in files:
     result = subprocess.call(["./%s" % (file_path)])
-    result_text = xccdf_reverse_result_map[result] if result in xccdf_reverse_result_map else "Unknown exit code %i" % (result)
+    if result in xccdf_reverse_result_map:
+        result_text = xccdf_reverse_result_map[result]
+    else:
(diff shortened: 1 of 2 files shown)


tests/sce/script_tester.py
```diff
@@ -52,18 +52,18 @@
...
 # (this is especially useful for bash scripts)
-for key, value in xccdf_result_map.iteritems():
+for key, value in xccdf_result_map.items():
     os.environ[key] = str(value)
 
 import subprocess
 
 for file_path in files:
     result = subprocess.call(["./%s" % (file_path)])
-    result_text = xccdf_reverse_result_map[result] if xccdf_reverse_result_map.has_key(result) else "Unknown exit code %i" % (result)
```
matejak: This is perfectly valid, but I personally dislike the non-trivial form ternary operators, as it is difficult to see what is going on (as opposed to an if statement). I don't insist on changing it, but you may give it a second thought.

jan-cerny: I like this suggestion very much. I will change it.

matejak:
ACK, thank you for your efforts towards keeping the codebase up-to-date with latest PYthon.

PR #1029, opened 2018-05-09
Add CI / coverage support for OpenSCAP [WIP]

matejak:
## PR aim

This PR does two things:
- enables the Travis CI support for OpensSCAP, and
- adds partial support for coverage output.

Partial support means that:
- not all source code has coverage support yet,
- not all tests have coverage support yet, and
- coverage output can't be analyzed by e.g. sonar, but you can view it locally in your web browser.

## PR background

The CI support is quite straightforward.
On the coverage side, things are more complicated.
The build system uses recursive make, which is [considered harmful](http://aegis.sourceforge.net/auug97.pdf) anyway. Here, it complicated the enablement of the coverage support, so I have converted the `common` and `source` subdirectories setups into non-recursive. Similarly, I have converted `unittests`  and `variable_instance` in `API/XCCDF` directory.

In order to get test coverage output, make sure you have `lcov` installed, and run the configure with `--enable-code-coverage`. After OpenSCAP builds, run `make check-code-coverage` instead of `make check`, and after it finishes, check out results, typically by running `xdg-open openscap-1.2.17-coverage/index.html` in your build root.

The transition from recursive make tests has the benefit of reducing existing boilerplate code, as well as enabling parallel runs of test suites. This work can be performed independently from this PR though.


jan-cerny:
I like this effort very much.

I am just not sure if changes in Makefiles are a good use of energy when master is already moved to CMake.

Please make sure it will not break RHEL6.

mpreisler:
> I am just not sure if changes in Makefiles are a good use of energy when master is already moved to CMake.

+1 we looked at some of this in the past and decided to live with the issues and focus on master cmake port.

matejak:
I did more research and it turns out that recursive builds don't matter at all. Thanks @jan-cerny  for the RTFM reminder :-) Recursive tests, however, may still be a problem.
Anyway, I am now investigating a simpler way how to provide coverage data that will come as a separate PR.

matejak:
Yes, this is not relevant as #1031 has been merged.

PR #1030, opened 2018-05-10
Remove basic Python implementation of oval_probes.c

jan-cerny:
This example doesn't work in any version of Python, is very outdated,
and doesn't demonstrate good practices. We have decided to remove it.


dahaic:
Thank you @jan-cerny good catch.

PR #1031, opened 2018-05-14
Add CI / coverage support for OpenSCAP (simple setup)

matejak:
This PR adds the CI / coverage checks capabilities to OpenSCAP using simpler methods than those used in #1029. Thanks @jan-cerny for pointing out the simpler way of how to enable coverage during the build!
However, tests based on recursive `make` don't produce coverage files in a single directory, so a script has been added to generate those `.gcov` files from `.gcno` and `.gcda` and to move them to the build directory, where they can be picked up by the analyzer.


jan-cerny:
The analysis is really great!

PR #1032, opened 2018-05-16
Add AppVeyor CI badge and configuration

jan-cerny:
This will add:
* an AppVeyor icon which will show the status of last build on AppVeyor, where we build WIndows builds.
* AppVeyor project configuration in YAML format

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,23 @@
+version: 1.0.{build}
+branches:
+  only:
+  - master
+image: Visual Studio 2017
+configuration: Release
+clone_folder: c:\projects\openscap
+install:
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -0,0 +1,23 @@
+version: 1.0.{build}
```
mpreisler: why not just {build}?

mpreisler: or master-{build} or something like that. The 1.0 is confusing.

jan-cerny:
@openscap-ci test this please

mpreisler:
@jan-cerny could you set up appveyor not to post comments but instead set commit status? similar to what jenkins does? posting comments will blow up my notifications.

mpreisler:
ACK

PR #1037, opened 2018-05-24
Drop Python 2 support from master branch

### LINKED ISSUE #1034: Move to Python3 shebangs
Python 2 will be sunset in 2020: https://pythonclock.org

Some modern Linux distributions will not support Python 2 anymore.
Our code is compatible with both Python 2 and 3.
However, the shebangs often contain python2.
The downstream packagers will have to patch these code.

For example, see: https://github.com/OpenSCAP/openscap/blob/maint-1.2/utils/scap-as-rpm#L1

Do you want to change shebangs to `#!/usr/bin/python3`?

Another problem is that we sometimes use "#!/usr/bin/env` which is forbidden by Fedora: https://fedoraproject.org/wiki/Packaging:Guidelines#Shebang_lines

jan-cerny:
This PR tries to move to Python 3 in master branch, drops Python 2 support and sometimes cleans the code up.

The changes include:
* Updating shebangs to `#!/usr/bin/python3` (fixes #1034 )
* Update README and User Manuals
* Remove SWIG bindings for Python 2
* Remove fallback solutions, that are now in Python 3 by default, eg. `argparse` or `subprocess.check_output`

Do we want to disable RHEL6/7 builds for the master branch?


jan-cerny:
This is a do not merge because it needs merge maint-1.2 into master and then rebase.

jan-cerny:
I have merged maint-1.2 branch in the master branch and I have rebased this pull request on the top of the master branch.

CMakeLists.txt
```diff
@@ -86,8 +86,8 @@ if(PROCPS_FOUND)
 	check_include_file(proc/devname.h HAVE_PROC_DEVNAME_H)
 endif()
 
-find_package(PythonLibs 2 EXACT)
-find_package(PythonInterp)
+find_package(PythonInterp 3)
+find_package(PythonLibs 3 EXACT)
```
mpreisler: why was the order changed, does it matter? if so please comment

tests/bindings/test_python.sh
```diff
@@ -4,11 +4,12 @@
 #   Martin Preisler <mpreisle@redhat.com>
 
 . $builddir/tests/test_common.sh
+set -x
```
mpreisler: leftover debug code

utils/scap-as-rpm
```diff
@@ -25,106 +25,15 @@ import locale
...
-        error.output = output
-        raise error
-    return output
-
-
-if hasattr(subprocess, "check_output"):
-    # if available we just use the real function
-    subprocess_check_output = subprocess.check_output
-
+import sys
```
mpreisler: pep is gonna flag this, why not top level import?

tests/bindings/test_python.sh
jan-cerny: thank you

CMakeLists.txt
jan-cerny: OK, I'll copy the commit message.

utils/oscap_docker_python/get_cve_input.py
```diff
@@ -15,14 +15,8 @@
...
 
-try:
-    # Python2 imports
-    import urlparse
-    import urllib2 as urllib
-except ImportError:
-    #Python3 imports
-    import urllib.parse as urlparse
-    import urllib.request as urllib
+import urllib.parse as urlparse
```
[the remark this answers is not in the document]
jan-cerny: Aha, I thought that `urlparse` is a preferred name.

jan-cerny:
I have removed the `set -x`, copied the commit message into the code as a comment and I have used `urllib.parse` and `urllib.requests` without renaming them to `urlparse` and `urllib`.

jan-cerny:
I think we need to start using Python 3 also in tests XCCDF that tests SCE.

jan-cerny:
I have found that we already use Python 3 in tests that test SCE. However, we did not use Python 3 in remediation tests. Therefore, I have changed the `test_remediate_python` and `test_remediate_python_subs` which are part of `tests/API/XCCDF/unittests/`. This way we will be able to test remediations also on systems that do not contain Python 2.

mpreisler:
@jan-cerny this breaks openscap master on rhel6 and rhel7. That is probably fine but we need to adjust jenkins to deal with it before we merge this.

jan-cerny:
@mpreisler Yes, that was a reason of my question. Which would be the proper solution?
a) install Python 3 on RHEL 6 and on RHEL 7 node
b) do not build master's nightlys and master's pull requests on RHEL 6 node and RHEL 7 node
@dahaic @matusmarhefka @matejak @yuumasato What is your opinion on this problem?

matejak:
We have came to an agreement within our group that as OpenSCAP 1.3  won't be available on RHEL6 neither on RHEL7, we will got the simple way, using just Fedora for the CI.

jan-cerny:
@matejak Thanks for your reply! I will look into the problem on Jenkins.

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

matejak:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

CMakeLists.txt
```diff
@@ -86,8 +86,13 @@ if(PROCPS_FOUND)
...
 
-find_package(PythonLibs 2 EXACT)
-find_package(PythonInterp)
+# According to CMake documentation:
+# https://cmake.org/cmake/help/latest/module/FindPythonInterp.html
+# If calling both find_package(PythonInterp) and find_package(PythonLibs),
+# call find_package(PythonInterp) first to get the currently active Python
+# version by default with a consistent version of PYTHON_LIBRARIES.
+find_package(PythonInterp 3)
+find_package(PythonLibs 3 EXACT)
```
mpreisler: what is the reason for EXACT?

matejak: Difficult to tell, but the check finds any Python 3.y.z, so it is probably OK and doesn't make a difference. In theory, `find_package(PythonLibs 3)` should try to find Python libraries that are compatible with Python 3, and as Python 2 neither Python 4 are compatible, the behavior of the check is somehow expected.

CMakeLists.txt
```diff
@@ -251,12 +256,7 @@ endif()
 option(ENABLE_VALGRIND "enables Valgrind memory testing in the test-suite" FALSE)
 
 # ---------- LANGUAGE BINDINGS
-
-# TODO: Shouldn't we only enable this if python2 was found?
-option(ENABLE_PYTHON2 "if enabled, the python2 swig bindings will be built" TRUE)
-
-# TODO: Shouldn't we only enable this if python3 was found?
-option(ENABLE_PYTHON3 "if enabled, the python3 swig bindings will be built" FALSE)
+cmake_dependent_option(ENABLE_PYTHON3 "if enabled, the python3 swig bindings will be built" ON "PYTHONINTERP_FOUND;SWIG_FOUND" OFF)
```
mpreisler: this also needs to depend on pythonlibsfound

CMakeLists.txt
mpreisler: well yeah but if EXACT is not needed let's not even have it there

mpreisler:
Would be great to get rid of the merge commits here, rebase, fix conflicts and get this merged. This makes building on Windows less annoying.

matejak:
This one needs rebasing etc., so see #1070

PR #1040, opened 2018-05-25
Don't shout about pthread when configuring

mpreisler:
It's confusing in the configure output... Is this a debugging leftover, @jan-cerny ?


jan-cerny:
@mpreisler Most likely it is. I am sorry. Thank you very much!

jan-cerny:
@openscap-ci test this please

PR #1046, opened 2018-06-01
Tests fixes

matejak:
Fixed bugs in test scripts:
- Fixed unexpected behavior when command doesn't fail (`test_object_component_type.sh`)
- Replaced some exit statements by return statements, so tests skipping is handled properly.

### DIFF
--- a/tests/API/OVAL/unittests/test_object_component_type.sh
+++ b/tests/API/OVAL/unittests/test_object_component_type.sh
@@ -2,10 +2,11 @@
 
 stderr=`mktemp`
 
-set -e
 set -o pipefail
 
-$OSCAP oval eval $srcdir/test_object_component_type.oval.xml 2> $stderr || ret=$?
+$OSCAP oval eval $srcdir/test_object_component_type.oval.xml 2> $stderr
+ret=$?
+set -e
 [ $ret -eq 1 ]
 
 grep -q "Entity [']something_bogus['] has not been found in textfilecontent_item (id: [0-9]\+) specified by object [']oval:oscap:obj:10[']." $stderr
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 
 before_install:
   - sudo apt-get -qq update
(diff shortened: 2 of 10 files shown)


tests/API/OVAL/unittests/test_object_component_type.sh
```diff
@@ -2,10 +2,11 @@
 
 stderr=`mktemp`
 
-set -e
 set -o pipefail
 
-$OSCAP oval eval $srcdir/test_object_component_type.oval.xml 2> $stderr || ret=$?
+$OSCAP oval eval $srcdir/test_object_component_type.oval.xml 2> $stderr
+ret=$?
+set -e
```
matusmarhefka: I would suggest to leave `set -e`  where it was and rather change the call to oscap in the following way:

`$OSCAP oval eval $srcdir/test_object_component_type.oval.xml 2> $stderr || ret=$?`

matejak: It was like that before, but then, the `ret` would not be defined if `oscap` finished without error. My change allows `oscap` to finish with any code it wants, storing it in the variable using a minimal number of steps.

matusmarhefka: IMO this test is expected to fail, if `oscap` finishes without error it is unexpected state in this test (see OVAL file). These constructions with test in `[]` under the command are used in the test suite to explicitly state what is the expected return code (when `set -e` is defined):
`[ $ret -eq 1 ]`

Your change of the code does not change behavior in any way (if `oscap` command passes test fails), it can be merged but the rest of tests are written in a way I described above.

matejak: The problem is that although it shouldn't, `oscap` can return zero (it actually did, that's how I have discovered this error). Consequently, `[ $ret -eq 1 ]` failed in the original code, but for a wrong reason - because `ret` was not defined. My change ensures that `ret` is always defined using the standard procedure of asking the return code unconditionally after the command ends. As the command is expected to fail, I had to deal with the `set -e` somehow. Putting `&& ret=$? || ret=$?` past the command would make possible to capture zero and any non-zero return codes, but I rather went for moving the `set -e` :-)

matusmarhefka: This is correct and as I said I agree with merging this. Actually it makes code more readable. In future we should rewrite and unify these code constructs in the test library to make the code consistent.

PR #1047, opened 2018-06-01
Include an example how to apply generated playbook on localhost

mpreisler:
I use this all the time and it's not obvious to use the -c local


mildass:
ACK.

PR #1048, opened 2018-06-04
Make Travis tests less troublesome.

matejak:
Prevent harmless or expected failiures to result in failed test run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,11 @@ before_script:
   - cd build
 
 script:
-  - ../configure CFLAGS="--coverage -ftest-coverage -fprofile-arcs" LDFLAGS=-lgcov --enable-debug && build-wrapper-linux-x86-64 --out-dir bw-output make all
+  - ../configure CFLAGS="--coverage -ftest-coverage -fprofile-arcs" LDFLAGS=-lgcov --enable-debug
+  - build-wrapper-linux-x86-64 --out-dir bw-output make all || true  # Will always fail builds on forked repositories.
   - make -k check VERBOSE=true || { test "$SEND_COVERAGE_EVEN_IF_TESTS_FAIL" = yes && curl -s https://codecov.io/bash > cov.sh && bash cov.sh -x "$GCOV"; }
-  - bash ../tests/recursively_generate_gcov.sh .
-  - cd .. && sonar-scanner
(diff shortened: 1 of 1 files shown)


mpreisler:
ACK when jenkins finishes

matejak:
And Travis, to be really sure that it works :-)

.travis.yml
```diff
@@ -13,10 +13,10 @@ before_script:
   - cd build
 
 script:
-  - ../configure CFLAGS="--coverage -ftest-coverage -fprofile-arcs" LDFLAGS=-lgcov --enable-debug && build-wrapper-linux-x86-64 --out-dir bw-output make all
+  - ../configure CFLAGS="--coverage -ftest-coverage -fprofile-arcs" LDFLAGS=-lgcov --enable-debug && build-wrapper-linux-x86-64 --out-dir bw-output make all || true  # Will always fail on external PRs.
```
mpreisler: could we split configure step and build step into two lines, makes it much easier to read. we can insist on configure passing and maybe `|| true` only the build part

.travis.yml
```diff
@@ -13,10 +13,10 @@ before_script:
...
   - make -k check VERBOSE=true || { test "$SEND_COVERAGE_EVEN_IF_TESTS_FAIL" = yes && curl -s https://codecov.io/bash > cov.sh && bash cov.sh -x "$GCOV"; }
-  - bash ../tests/recursively_generate_gcov.sh .
-  - cd .. && sonar-scanner
+  - bash ../tests/recursively_generate_gcov.sh . || true  # Failiures that occur here are harmless
```
mpreisler: typo

.travis.yml
```diff
@@ -13,10 +13,10 @@ before_script:
...
+  - cd .. && sonar-scanner || true  # Will always fail on external PRs.
```
mpreisler: since we only have external PRs, does it make sense to even run this? I assume you mean forks.

mpreisler:
Note for later: This will need a merge into master and then another PR in master to change it to use cmake and ctest.

.travis.yml
matejak: It makes sense, as merge also triggers a build, which won't fail.

.travis.yml
matejak: Good idea!

PR #1049, opened 2018-06-04
Add gcc-c++ to dependencies

cipherboy:
- The build fails without `gcc-c++` installed on the system:

```
$ cmake ..
-- The C compiler identification is GNU 8.1.1
-- The CXX compiler identification is unknown
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
CMake Error at CMakeLists.txt:3 (project):
  No CMAKE_CXX_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.


-- Configuring incomplete, errors occurred!
See also "/home/ascheel/GitHub/openscap/openscap/build/CMakeFiles/CMakeOutput.log".
See also "/home/ascheel/GitHub/openscap/openscap/build/CMakeFiles/CMakeError.log".
```
- Also clarifies that the package list is for Fedora / RHEL, as next section identifies Ubuntu packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,17 +37,17 @@ cd openscap
 
 2) To build the library you will also need the following build dependencies
 (some of these are optional, if they are not detected, openscap will be compiled
-without respective optional features):
-```
-cmake \
-dbus-devel GConf2-devel libacl-devel libblkid-devel libcap-devel libcurl-devel \
-libgcrypt-devel libselinux-devel libxml2-devel libxslt-devel make openldap-devel \
-pcre-devel perl-XML-Parser perl-XML-XPath perl-devel python-devel rpm-devel swig \
-bzip2-devel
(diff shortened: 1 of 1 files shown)


mpreisler:
@openscap-ci add to whitelist

mpreisler:
C++ is only needed for the dpkg probe but I think implementing the logic that checks for C++ only if user also wants the dpkg probe would be too complex, let's just require c++, it will be on most dev systems already anyway.

Issue #1022, opened 2018-04-26
Remove occurences of OVAL_PROBE_DIR in master


jan-cerny:
In our master branch, we don't build probes as separated binaries anymore. It is not needed to specify a probe directory using `OVAL_PROBE_DIR` anymore. It would be nice to remove it from our code and from CMakeLists.txt in order to simplify the code. Also, `oval_probe_ext.c` which uses this variable could be simplified.

jan-cerny:
Fixed in #1050

PR #1050, opened 2018-06-04
Remove occurrences of OVAL_PROBE_DIR in master

cipherboy:
This removes the need for OVAL_PROBE_DIR environment variable
from the build scripts, oval_probe_ext.c, and related code.
Breaks ABI/API compatibility, as oval_probe_ext_getdir()
is also removed.

Closes: #1022 

Note that this doesn't update the [documentation](https://github.com/OpenSCAP/openscap/blob/master/docs/manual/manual.adoc#623-testing-library).


src/OVAL/oval_probe_ext.c
```diff
@@ -781,8 +769,8 @@ int oval_probe_sys_handler(oval_subtype_t type, void *ptr, int act, ...)
 			break;
 		}
 
-		probe_urilen = snprintf(probe_uri, sizeof probe_uri, "%s://%s/%s",
-				OVAL_PROBE_SCHEME, probe_dir, oval_subtype_get_text(type));
+		probe_urilen = snprintf(probe_uri, sizeof probe_uri, "%s://%s",
```
cipherboy: Would appreciate feedback on whether this and the similar change on line 835 is the correct change; since `OVAL_PROBE_DIR` is going away, I removed `oval_pext_t->probe_dir` as well. To my knowledge, probe_dir was only used by the OVAL_PROBE_DIR related code, hence removing this seems like a necessary simplification. The following checks on line 775 seems a little extraneous as well, as the `probe_uri` now only consists of `OVAL_PROBE_SCHEME` and the subtype, but I'd like feedback there too.

yuumasato: The removal of  `oval_pext_t->probe_dir` looks ok, it seems to be only used for `probe_uri`.
About the checks on 775, yes, the length of `probe_uri` will now be determined by factors under our control, so we can be less worried about it, but there is no ham in keep checking it.

cipherboy:
@yuumasato I'm not aware of what context this PR fits in the larger plan for `openscap`; if additional changes are coming, and the documentation section will end up being significantly restructured, then I would prefer to leave it and let someone more informed rewrite it. Otherwise, I can remove the references in the documentation.

PR #1052, opened 2018-06-04
The probes compile options all depend on ENABLE_PROBES

mpreisler:
Without this change it will report UNIX probes enabled even though ENABLE_PROBES is OFF.

```
-- OVAL:
-- base probe support: FALSE
-- SEAP msgid bit-size: 32
-- independent probes: ON
-- unix probes: ON
-- linux probes: ON
-- solaris probes: OFF
```

After this change:
```
-- OVAL:
-- base probe support: FALSE
-- SEAP msgid bit-size: 32
-- independent probes: OFF
-- unix probes: OFF
-- linux probes: OFF
-- solaris probes: OFF
```


cipherboy:
@mpreisler I pulled your branch and tested it locally; LGTM. I didn't check Windows though.

yuumasato:
Thanks for checking this @cipherboy.
Looks good to me also.
As there are some probes that work for Windows, there is no need to strictly test build with disabled probes on windows. 
Ack.

PR #1053, opened 2018-06-05
OpenSCAP master uses cmake instead of automake

mpreisler:
An attempt to make travis work on openscap master.


mpreisler:
@matejak I think I will need your help here. How do I get `build-wrapper-linux-x86-64` to appear on travis?

matejak:
@mpreisler This will appear there automagically - it is provided by the Sonar integration.

mpreisler:
> @mpreisler This will appear there automagically - it is provided by the Sonar integration.

ok, so what's missing here then? can we just merge this?

matejak:
As we can see in the Travis output, compilation uses `cmake` correctly. However, you have removed the `make check`, which builds test dependencies and then executes tests. Just put `make` command before `ctest`, so everything builds even if build wrapper is not present.

mpreisler:
> make check, which builds test dependencies and then executes tests

There is no `make check` in the master branch.

matejak:
I meant `make && ctest` instead of just `ctest`.

cipherboy:
@mpreisler A suggestion might be to create common CI build scripts (e.g.: [cryptominisat](https://github.com/msoos/cryptominisat/blob/master/scripts/travis-cmake.sh)). For simple builds they are put inline (e.g., [krb5](https://github.com/krb5/krb5/blob/master/.travis.yml#L24)), but I think having different build instructions on different branches could warrant a CI build script. 

One issue with this is if it is a different script for each, merging between master/maint could get tricky w.r.t. the script, but you could probably use one common script shared between the two.

mpreisler:
> I meant make && ctest instead of just ctest.

Doesn't the line above with the wrapper build everything? In master tests are built as part of the overall build.

matejak:
> Doesn't the line above with the wrapper build everything?

It does, when the wrapper is available. It may not be available, so we have to make sure that everything is built later. `make test` will execute test scripts, but it doesn't guarantee that all tests and their deps are built before (unlike `make check`).

mpreisler:
I have mixed in a minor fix. We weren't linking to ACL_LIBRARY. Hope that's not a big deal.

matejak:
Good work! It is all build-system related, so it is fine.
The `build-wrapper` thing is somehow confusing, one basically needs to make sure that tests will run even if the build supervised by the wrapper fails.

Issue #768, opened 2017-06-28
Alert user to proprietary extensions in SCAP content


shawndwells:
Sometimes certain content providers use their own proprietary extension (instead of standard
OVAL), you can tell that from content by seeing something along the lines of

    <xccdf:check system="http://cisecurity.org/check">

Perhaps, we can make OpenSCAP report this ^ element to the user. OpenSCAP could output something like:

~Warning: Skipping rule {#X} that requires proprietary extension to evaluate. Please ask you content vendor to provide SCAP/OVAL instead.~

    Warning: Skipping rule {#X} that requires a non-standard extension to evaluate. Please ask you content vendor to provide SCAP/OVAL instead.

(An idea from @isimluk, I'm only capturing it here as an RFE)

jan-cerny:
Thanks @isimluk for the idea. Thanks @shawndwells for reporting.

PR #1054, opened 2018-06-05
Warn about non-registered check systems

cipherboy:
When evaluating XCCDF, if a check uses a system that is not registered (and no other checks can be selected), emit a warning. 

Closes: #768


mpreisler:
@shawndwells what do you think about the wording?

cipherboy:
Closing in favor of #1055.

PR #1055, opened 2018-06-05
Warn about non-registered check systems

cipherboy:
When evaluating XCCDF, if a check uses a system that is not registered (and no other checks can be selected), emit a warning. (Targets maint-1.2 versus master)

Closes: #768 
Replaces: #1054 

Per @mpreisler's comment on #1054:
@shawndwells what do you think about the wording?

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy.c
+++ b/src/XCCDF_POLICY/xccdf_policy.c
@@ -28,6 +28,7 @@
 #endif
 
 #include <stdlib.h>
+#include <stdio.h>
 #include <string.h>
 #include <libgen.h>
 
@@ -572,11 +573,32 @@ _xccdf_policy_rule_get_applicable_check(struct xccdf_policy *policy, struct xccd
 			// If the refined selector does not match, checks without selector shall be used.
 			candidate_it = xccdf_rule_get_checks_filtered(rule, NULL);
 		}
+
+		bool print_general_warning = false;
+		bool print_oval_warning = false;
+		char *warning_check_system;
 		// Check Processing Algorithm -- Check.System
 		while (xccdf_check_iterator_has_more(candidate_it)) {
 			struct xccdf_check *check = xccdf_check_iterator_next(candidate_it);
-			if (_xccdf_policy_is_engine_registered(policy, (char *) xccdf_check_get_system(check)))
+			if (_xccdf_policy_is_engine_registered(policy, (char *) xccdf_check_get_system(check))) {
 				result = check;
+			} else if (strcmp("http://oval.mitre.org/XMLSchema/oval-definitions-5", check->system) == 0) {
+				print_oval_warning = true;
+			} else {
+				print_general_warning = true;
+				warning_check_system = check->system;
+			}
+		}
+
+		// Only print a warning if we didn't select a check but could've otherwise.
+		if (print_oval_warning) {
+			printf("WARNING: Skipping rule that uses OVAL but is possibly malformed; "
+			       "an incorrect content reference prevents this check from being evaluated.\n");
+		} else if (print_general_warning && result == NULL) {
+			printf("WARNING: Skipping rule that requires an unregistered check system "
+			       "or incorrect content reference to evaluate. "
+			       "Please consider providing a valid SCAP/OVAL instead of %s\n",
+				warning_check_system);
 		}
 		xccdf_check_iterator_free(candidate_it);
 	}
--- a/tests/API/XCCDF/parser/test_proprietary_extensions.xml
+++ b/tests/API/XCCDF/parser/test_proprietary_extensions.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Benchmark xmlns="http://checklists.nist.gov/xccdf/1.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="RHEL-6" xsi:schemaLocation="http://checklists.nist.gov/xccdf/1.1 xccdf-1.1.4.xsd" resolved="0" xml:lang="en-US">
+  <status date="2011-12-08">draft</status>
+  <title xml:lang="en-US">Sample XCCDF using proprietary extensions</title>
+  <description xml:lang="en-US">Proprietary extensions WILL NOT appear the XCCDF report</description>
+  <version>0.1</version>
+  <model system="urn:xccdf:scoring:default"/>
+  <model system="urn:xccdf:scoring:flat"/>
+  <Group id="bash-passer" hidden="false">
+    <title xml:lang="en-US">Check with a hint of Mint</title>
+    <description xml:lang="en-US">Mint tea consumed during the making of this test</description>
+    <Rule id="rule-1000" selected="true" weight="10.000000">
+      <title xml:lang="en-US">TEST (proprietary)</title>
+      <check system="https://www.bigelowtea.com/Teas/Flavors/Mint">
+        <check-content-ref href="mint.tea"/>
+      </check>
+    </Rule>
+  </Group>
+</Benchmark>
--- a/tests/API/XCCDF/parser/Makefile.am
+++ b/tests/API/XCCDF/parser/Makefile.am
@@ -22,10 +22,11 @@ TESTS_ENVIRONMENT = \
 		OSCAP_FULL_VALIDATION=1 \
 		$(top_builddir)/run
 
-TESTS = test_api_xccdf.sh
+TESTS = test_api_xccdf.sh test_extensions.sh
 check_PROGRAMS = test_api_xccdf
 
 test_api_xccdf_SOURCES = test_api_xccdf.c
 
-EXTRA_DIST = test_api_xccdf.sh xccdf11.xml xccdf12.xml xccdf11-results.xml
+EXTRA_DIST = test_api_xccdf.sh test_extensions.sh xccdf11.xml xccdf12.xml xccdf11-results.xml \
+             test_known_extensions.xml test_malformed_extensions.xml test_proprietary_extensions.xml
 
--- a/tests/API/XCCDF/parser/test_extensions.sh
+++ b/tests/API/XCCDF/parser/test_extensions.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 Red Hat Inc., Durham, North Carolina.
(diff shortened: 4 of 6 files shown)


mpreisler:
> Merging #1055 into maint-1.2 will decrease coverage by <.01%.
> The diff coverage is 100%.

does not compute!

cipherboy:
@mpreisler Agreed: https://codecov.io/gh/OpenSCAP/openscap/pull/1055/changes#L116 ?

matejak:
This is a bit tricky. The message is OK - it says that 100% of changes are covered by tests. However, if tests change, or the amount of covered code shrinks, the total coverage may decrease.
The problem is that those tests are not passing (because they are runing on Ubuntu), so even adding a test may reduce the test coverage (e.g. when it is executed first and fails, so other tests in the batch are skipped). Or who knows, there may be an issue with the test coverage generation/collection. I guess that I can disable PR comments until we figure out what exactly going on.

cipherboy:
I'm not quite sure why Jenkins failed on this one. I wasn't able to find a failing test case, and it doesn't look like a syntax error?

-- Alex

matejak:
There is a
`make[7]: *** No rule to make target `test_extensions.sh', needed by `test_extensions.sh.log'.`
The `make check` command is probably passed `-k`, so it doesn't stop right after the failiure and all tests are executed. Check out `EXTRA_DIST` in `Makefile.am`, the test script is not distributed in the package.

matejak:
Could you just rebase against updated upstream `maint-1.2`? It should recalculate the coverage to a green state.

cipherboy:
@matejak  Several unrelated test cases are failing on travis, however the ones introduced in this PR pass:

https://travis-ci.org/OpenSCAP/openscap/builds/388917553#L4659

mpreisler:
@openscap-ci test this please

cipherboy:
Again looks to be `asciidoctor`:

```
lookup cache used 8495/65536 hits=62280 misses=8987
finished...
cd openscap-1.2.18/docs/manual && asciidoctor -b html5 manual.adoc
/bin/sh: asciidoctor: command not found
make[2]: *** [dist-hook] Error 127
make[2]: Leaving directory `/home/jenkins/workspace/openscap-pull-requests'
make[1]: *** [distdir] Error 2
make[1]: Target `dist-gzip' not remade because of errors.
make[1]: Leaving directory `/home/jenkins/workspace/openscap-pull-requests'
make: *** [dist] Error 2
make: Target `distcheck' not remade because of errors.
Build step 'Execute shell' marked build as failure
Set build name.
New build name is 'wpe - #675'
An attempt to send an e-mail to empty list of recipients, ignored.
Setting status of 1a7fc8f25c844b6c4162911de7bf096513534032 to FAILURE with url https://jenkins.open-scap.org/job/openscap-pull-requests/675/ and message: ' '
Finished: FAILURE
```

mpreisler:
@openscap-ci test this please

I have fixed that (I hope)

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -572,12 +573,26 @@ _xccdf_policy_rule_get_applicable_check(struct xccdf_policy *policy, struct xccd
 			// If the refined selector does not match, checks without selector shall be used.
 			candidate_it = xccdf_rule_get_checks_filtered(rule, NULL);
 		}
+
+		int print_warning = 0;
```
mpreisler: I also miss the 90s but let's use `bool` :-D

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -572,12 +573,26 @@ _xccdf_policy_rule_get_applicable_check(struct xccdf_policy *policy, struct xccd
...
+				print_warning = 1;
+				warning_check_system = check->system;
+			}
 		}
+
+		// Only print a warning if we didn't select a check but could've otherwise.
+		if (print_warning && result == NULL)
+			printf("WARNING: Skipping rule that requires an unregistered check system "
+			       "or incorrect content reference to evaluate. "
+			       "Please consider providing SCAP/OVAL with valid references instead: %s\n",
```
mpreisler: How about:
Please consider providing a valid SCAP/OVAL check instead of '%s'.

tests/API/XCCDF/parser/test_proprietary_extensions.xml
```diff
@@ -0,0 +1,19 @@
...
+  <description xml:lang="en-US">Proprietary extensions WILL NOT appear the XCCDF report</description>
+  <version>0.1</version>
+  <model system="urn:xccdf:scoring:default"/>
+  <model system="urn:xccdf:scoring:flat"/>
+  <Group id="bash-passer" hidden="false">
+    <title xml:lang="en-US">Check with a hint of Mint</title>
+    <description xml:lang="en-US">Mint tea consumed during the making of this test</description>
+    <Rule id="rule-1000" selected="true" weight="10.000000">
+      <title xml:lang="en-US">TEST (proprietary)</title>
+      <check system="https://www.bigelowtea.com/Teas/Flavors/Mint">
```
mpreisler: :D

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -572,12 +573,26 @@ _xccdf_policy_rule_get_applicable_check(struct xccdf_policy *policy, struct xccd
...
+		bool print_warning = 0;
+		char *warning_check_system;
 		// Check Processing Algorithm -- Check.System
 		while (xccdf_check_iterator_has_more(candidate_it)) {
 			struct xccdf_check *check = xccdf_check_iterator_next(candidate_it);
-			if (_xccdf_policy_is_engine_registered(policy, (char *) xccdf_check_get_system(check)))
+			if (_xccdf_policy_is_engine_registered(policy, (char *) xccdf_check_get_system(check))) {
 				result = check;
+			} else {
+				print_warning = 1;
```
mpreisler: true

cipherboy:
@mpreisler Oops, that was me being sloppy. Fixed.

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -572,12 +573,26 @@ _xccdf_policy_rule_get_applicable_check(struct xccdf_policy *policy, struct xccd
...
+				warning_check_system = check->system;
+			}
 		}
+
+		// Only print a warning if we didn't select a check but could've otherwise.
+		if (print_warning && result == NULL)
+			printf("WARNING: Skipping rule that requires an unregistered check system "
+			       "or incorrect content reference to evaluate. "
+			       "Please consider providing a valid SCAP/OVAL instead of %s\n",
+				warning_check_system);
```
shawndwells: Missed @mpreisler and @cipherboy's original ping. What's the nuance of "unregistered check system" vs "unsupported check system"?

mpreisler: OpenSCAP may support a check system - say SCE - but the plugin may not be installed at the time.

cipherboy:
Further, due to the way this section of code is structured, a check's
system may be registered and supported (e.g. OVAL), but the check's content
reference may be invalid (if the referenced OVAL is missing / doesn't
exist). That's why it also has the clause "incorrect content reference"...
In this case, it's not "unsupported", so unregistered was a compromise
wording that is perhaps slightly weaker than unsupported but has similar
intent.

src/XCCDF_POLICY/xccdf_policy.c
shawndwells: Given the current texts, how would a user identify if the rule doesn't have a check component at all versus if it's just something that hasn't been implemented yet?

If a user opens a support ticket, how would someone help troubleshoot the difference? Would throwing a unique error/warning for each condition help decrease support burdens?

cipherboy: @shawndwells A [rule](https://github.com/cipherboy/openscap/blob/4fb933b860305dfefd6e1ab8c053b75403edb671/src/XCCDF_POLICY/xccdf_policy.c#L568) will have multiple [checks](https://github.com/cipherboy/openscap/blob/4fb933b860305dfefd6e1ab8c053b75403edb671/src/XCCDF_POLICY/xccdf_policy.c#L570). If [a check](https://github.com/cipherboy/openscap/blob/4fb933b860305dfefd6e1ab8c053b75403edb671/src/XCCDF_POLICY/xccdf_policy.c#L570) would've been selected if [it was registered](https://github.com/cipherboy/openscap/blob/4fb933b860305dfefd6e1ab8c053b75403edb671/src/XCCDF_POLICY/xccdf_policy.c#L582), it will [print a warning](https://github.com/cipherboy/openscap/blob/4fb933b860305dfefd6e1ab8c053b75403edb671/src/XCCDF_POLICY/xccdf_policy.c#L585) if it wasn't able to select an [appropriate check](https://github.com/cipherboy/openscap/blob/4fb933b860305dfefd6e1ab8c053b75403edb671/src/XCCDF_POLICY/xccdf_policy.c#L591) because either (the check system is unregistered / proprietary / invalid / etc.) or (it was a supported check system (e.g., OVAL), but the reference to the specific check was invalid (`check-content-ref` pointed to a missing file / invalid path / etc)).


Thus, this message won't print if a rule doesn't have a check so that scenario won't happen. More likely, a warning (which won't halt execution) will be printed:
 - If the OVAL referenced check is wrong (in which case the warning will print the OVAL reference URL and thus can be identified that way). 
 - Or, if the system used a proprietary check system (in which case the URL will not be OVAL and will be something else, which will also be printed) and no valid check systems (OVALs) were supplied as alternatives. 

I'd have to check, but to my knowledge, a warning is not printed when a rule has no checks; perhaps this should be added in a separate PR.

mpreisler:
Hmm, I am looking at this again and I think this may be confusing if OVAL is disabled or the file was not found. Could we change the condition and the wording?

IMO it should only trigger if the check system is not OVAL, otherwise suggesting providing OVAL instead is misleading.

mpreisler:
> Do you think we should display a different warning for when the OVAL has a dangling reference?

IMO yes

cipherboy:
@mpreisler This should be updated now. :)

mpreisler:
@cipherboy fails test suite:
```
+ malformed check extension                                   [ FAIL ]
```

cipherboy:
@mpreisler This is updated and passes tests. Thanks for the reminder about `distcheck`!

PR #1063, opened 2018-06-13
Pkgconfig only non windows

mpreisler:
Pretty minor stuff


shawndwells:
ack once jenkins done

PR #1068, opened 2018-06-15
Added GitHub issue template

cipherboy:
Largely similar to the one in [scap-security-guide](https://github.com/OpenSCAP/scap-security-guide/blob/master/.github/issue_template.md), but includes directions of where to report issues at the top.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,35 @@
+- Is this an issue with SCAP Workbench?
+    - If so, report it here: https://github.com/OpenSCAP/scap-workbench/issues
+- Is this an issue with SCAP Security Guide (i.e., related to the content of scans, not the scanner proper)?
+    - If so, report it here: https://github.com/OpenSCAP/scap-security-guide/issues
+- Is this an issue during the OS installation process?
+    - If so, report it here: https://github.com/OpenSCAP/oscap-anaconda-addon/issues
+
+Thanks!
+
+
+#### Description of Problem:
+
+
+#### OpenSCAP Version:
+
+
+#### Operating System & Version:
+
+
+#### Steps to Reproduce:
+
+1. 
+2. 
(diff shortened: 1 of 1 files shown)


redhatrises:
Please don't use "here". Use the full URL.

.github/issue_template.md
```diff
@@ -0,0 +1,32 @@
+Is this an issue with SCAP Workbench? If so, head [here](https://github.com/OpenSCAP/scap-workbench/issues).
+Is this an issue with SCAP Security Guide (i.e., related to the content of scans, not the scanner proper)? If so, head [here](https://github.com/OpenSCAP/scap-security-guide/issues).
+Is this an issue during the OS installation process? IF so, head [here](https://github.com/OpenSCAP/oscap-anaconda-addon/issues).
```
mpreisler: capital IF

mpreisler:
Great stuff, found one minor typo.

cipherboy:
@redhatrises / @mpreisler Updated. :)

.github/issue_template.md
```diff
@@ -0,0 +1,35 @@
+- Is this an issue with SCAP Workbench?
+    - If so, report it here: https://github.com/OpenSCAP/scap-workbench/issues
+- Is this an issue with SCAP Security Guide (i.e., related to the content of scans, not the scanner proper)?
+    - If so, report it here: https://github.com/OpenSCAP/scap-security-guide/issues
+- Is this an issue during the OS installation process?
```
cipherboy: Would you prefer this says RHEL / etc? It isn't default on any OS so this should be a rare enough occurrence and depending on the issue, it might actually be an `openscap` issue, but I'm not sure what we want the desired behavior to be.

redhatrises: IMO it's fine as is.

PR #1070, opened 2018-06-19
Drop Python 2 support from master branch

### LINKED ISSUE #1034: Move to Python3 shebangs

matejak:
Based on #1037. This one is rebased, and contains minor fixes.
This PR tries to move to Python 3 in master branch, drops Python 2 support and sometimes cleans the code up.
The changes include:
* Updating shebangs to #!/usr/bin/python3 (fixes #1034 )
* Update README and User Manuals
* Remove SWIG bindings for Python 2
* Remove fallback solutions, that are now in Python 3 by default, eg. argparse or subprocess.check_output


matejak:
@mpreisler I have become Jan's partner in crime, so feel free to merge it as I am no longer independent.

README.md
```diff
@@ -37,21 +37,25 @@ cd openscap
 
 2) To build the library you will also need the following build dependencies
 (some of these are optional, if they are not detected, openscap will be compiled
-without respective optional features). On RHEL 7 / Fedora, the command to
```
mpreisler: Why was the note about RHEL7 / Fedora removed? Doesn't seem to be related to python2.

mpreisler:
Found one minor thing that I am curious about, other than that I think this is ready to be merged.

README.md
matejak: I thought that as there is only Python2 on RHEL7, I will take the mention about it out. Fedora also installs Python 2 by default.

mpreisler: But the rest of the packages are RHEL / Fedora specific. There is a separate section for ubuntu. It doesn't read to me like the note about RHEL 7 and Fedora is there just because of python2.

matejak: OK, I have put back the sentence with `RHEL / Fedora / CentOS` list.

mpreisler:
ACK

PR #1072, opened 2018-06-22
Minor fixes to avoid linker errors when built with disabled probes

mpreisler:


redhatrises:
Ack. I can now compile... Plus mpreisler acks this pr. :D

PR #1075, opened 2018-06-25
Fix building using MINGW

### LINKED ISSUE #1073: Cmake cannot find mingw32-winpthreads
#### Description of Problem:
Cmake cannot find mingw32-winpthreads

#### OpenSCAP Version:
master

Fedora 28

#### Steps to Reproduce:

1. mingw32-cmake -DENABLE_PYTHON3=FALSE -DENABLE_PROBES=FALSE -DENABLE_OSCAP_UTIL_DOCKER=FALSE ../
2. 
3. 

#### Actual Results:
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Threads (missing: THREADS_PTHREADS_WIN32_LIBRARY)
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindThreads.cmake:232 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:116 (find_package)

#### Expected Results:
Build compiles

redhatrises:
- Some build fixes for when using MINGW.


CMakeLists.txt
```diff
@@ -113,7 +113,9 @@ find_package(BZip2)
 set(CMAKE_THREAD_PREFER_PTHREAD)
 set(THREADS_PREFER_PTHREAD_FLAG)
 set(THREADS_USE_PTHREADS_WIN32 true)
+if(NOT MINGW)
```
mpreisler: don't we also need Threads with mingw?

redhatrises: @mpreisler if you skip this threads check when using mingw, the build completes successfully.

src/CMakeLists.txt
```diff
@@ -94,11 +94,11 @@ if (ENABLE_PROBES)
...
-endif()
-if (GCRYPT_FOUND OR NSS_FOUND)
-	list(APPEND OBJECTS_TO_LINK_AGAINST
-		$<TARGET_OBJECTS:crapi_object>
-	)
+	if (GCRYPT_FOUND OR NSS_FOUND)
+        list(APPEND OBJECTS_TO_LINK_AGAINST
+                $<TARGET_OBJECTS:crapi_object>
+        )
+	endif()
```
mpreisler: indents borked

src/OVAL/probes/CMakeLists.txt
```diff
@@ -1,13 +1,10 @@
...
-if(CRYPTO_FOUND)
-	add_subdirectory("crapi")
-endif()
-
-include_directories("." "public")
-
 if (ENABLE_PROBES)
+	if(CRYPTO_FOUND)
+        	add_subdirectory("crapi")
+	endif()
```
mpreisler: indents

src/XCCDF/result.c
```diff
@@ -34,8 +34,13 @@
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
 #endif
+#ifdef __MINGW32__
+#include <winsock2.h>
+#include <iphlpapi.h>
+#else
 #include <WinSock2.h>
 #include <Iphlpapi.h>
```
mpreisler: Windows is not case sensitive, let's use the right names here and avoid __MINGW32__.

It makes no sense to use the wrong names on Windows and the correct names in case we are on a case sensitive platform. Let's just use the correct names...

CMakeLists.txt
mpreisler: ok, but why? where does it get pthreads?

jan-cerny:
Does anybody have any idea why AppVeyor build status is not reported in this PR?
It failed:
https://ci.appveyor.com/project/OpenSCAP/openscap/build/master-56

CMakeLists.txt
jan-cerny: I have already commented on https://github.com/OpenSCAP/openscap/issues/1073. I think that would work only if you build OpenSCAP without probes and probe subsystem.

I think we need to look into cmake/FindThreads.cmake file instead.
There is a variable THREADS_USE_PTHREADS_WIN32, which is used in root CMakeLists.txt around line 115.

Try something like:

if(NOT MINGW)
   set(THREADS_USE_PTHREADS_WIN32 true)
endif()

If that does not help, we will most likely need to dive into cmake/FindThreads.cmake .

redhatrises:
@jan-cerny no sure why it is, but there are a few of these errors:
```
fatal error C1083: Cannot open include file
```

redhatrises:
@jan-cerny this is passing now for appveyor as well. I had accidentally removed a line.

CMakeLists.txt
```diff
@@ -108,7 +108,9 @@ find_package(BZip2)
 # PThread
 set(CMAKE_THREAD_PREFER_PTHREAD)
 set(THREADS_PREFER_PTHREAD_FLAG)
-set(THREADS_USE_PTHREADS_WIN32 true)
+if(NOT MINGW)
+	set(THREADS_USE_PTHREADS_WIN32 true)
```
mpreisler: don't we want to use pthreads? especially on mingw?

mpreisler:
it's really odd that you needed to add THREADS_USE_PTHREADS_WIN32 but only when not using mingw. why does that not work with mingw?

redhatrises:
@mpreisler I could be wrong, but I believe that `THREADS_USE_PTHREADS_WIN32` is looking for pthread implementation on a windows platform and not when using MINGW. When you run CMake with mingw and it skips `THREADS_USE_PTHREADS_WIN32`, it appears to find it. 

Output:
```
-- Looking for pthread_setname_np in pthread
-- Looking for pthread_setname_np in pthread - found
-- Looking for pthread_getname_np in pthread
-- Looking for pthread_getname_np in pthread - found
```

mpreisler:
> @mpreisler I could be wrong, but I believe that THREADS_USE_PTHREADS_WIN32 is looking for pthread implementation on a windows platform and not when using MINGW. When you run CMake with mingw and it skips THREADS_USE_PTHREADS_WIN32, it appears to find it.

We need to be sure about this. Experimental development leads to a hellscape...

jan-cerny:
I will try to investigate the solution further and verify if that works with both MINGW and MSVC 2017.

jan-cerny:
I confirm that this solution works on both Visual Studio 2017 on Windows and MINGW on F27 (cross-compilation). 

However, it works only because https://github.com/OpenSCAP/openscap/blob/master/cmake/FindThreads.cmake
is written wrong and doesn't work as we would expect. 

If we use MinGW and we define THREADS_USE_PTHREADS_WIN32 in our CMakeLists, then on https://github.com/OpenSCAP/openscap/blob/master/cmake/FindThreads.cmake#L91 it sets name to `pthreadsGC2`. No library `pthreadsGC2` exists on my system, so it fails. A correct name on this line would be `libpthread`.

But if we use MinGW and we don't define THREADS_USE_PTHREADS_WIN32 in our CMakeLists, then it goes to https://github.com/OpenSCAP/openscap/blob/master/cmake/FindThreads.cmake#L111 and there it looks for "pthread.h" which leads to finding pthread library successfully.

So we have 2 options:
1) we use @redhatrises solution and we provide a comment that explains the confusing condition
at https://github.com/OpenSCAP/openscap/pull/1075/files#diff-af3b638bc2a3e6c650974192a53c7291R111

2) We rework https://github.com/OpenSCAP/openscap/blob/master/cmake/FindThreads.cmake
Maybe it would be enough to rename `THREADS_USE_PTHREADS_WIN32` to something more sensible like `MSVC_THREADS_USE_PTHREADS_WIN32` and remove code around L91
Then the CMakeLists.txt will not contain confusing conditions.

@mpreisler Which do you prefer?

jan-cerny:
@mpreisler I looked at it again and I tend to the option 1, because it would be much faster to finish. What about you?

mpreisler:
@jan-cerny I strongly prefer option 2, we could also patch this for other people and submit a PR. I agree that the hack/workaround option 1 would be faster to implement.

jan-cerny:
@mpreisler OK, then should we close this PR and start a new one?

jan-cerny:
@mpreisler Actually it would be better to continue in this one, because there are other fixes in this PR which I find valuable, like fixing the includes and preprocessor directives.

cipherboy:
@redhatrises If you remove the thread-related changes, we'll merge it and fix threads support in a separate PR. :)

jan-cerny:
@openscap-ci test this please

jan-cerny:
LGTM.

PR #1078, opened 2018-06-26
Fix failing tests if CMake Build Type is release

### LINKED ISSUE #912: Some tests fail if CMake build type is Release
If build type is Release, it defines NDEBUG, which means that `assert()` calls in some of our tests do nothing, and then the test is not performed, and some results are not generated.

An example could be https://github.com/OpenSCAP/openscap/blob/master/tests/API/XCCDF/unittests/test_xccdf_overrides.c test, which uses `assert()`. If the build type is release, it defines NDEBUG, which means that those `assert()` macros are compiled tobasically no operation. Please see `man 3 assert`.
 That means that the checks are not performed. Moreover, in this example the file `test_xccdf_overrides.xml.out` is not created, because it is created iside an assert, and therefore the test fails without any report. The reason for that is that in bash test the test checks if the file exists. Please check https://github.com/OpenSCAP/openscap/blob/master/tests/API/XCCDF/unittests/test_xccdf_overrides.sh#L12

I have set CMake build type on Jenkins openscap-master job to Debug as temporary workaround.

This is the most serious problem, because some of the tests won't test anything with the released version.

Does anybody have any idea how to solve this issue?

jan-cerny:
Macro `assert` compiles to empty command if build type is Release.
That caused some of the tests failing in Release configuration.
We need to replace `assert` by a macro that is not dependent on
NDEBUG variable. This commit introduces `oscap_assert` and replaces
all occurrences of `assert` in tests by `oscap_assert`.

Fixes #912.

Furthermore, it provides the following changes in order to use consistently a single macro oscap_assert:
* Replace old assume in a sparse buffer test
* Replace assume by oscap_assert in oscap_common test
* Remove unused assume.h


matusmarhefka:
Great job.

PR #1084, opened 2018-06-27
Set RPATH on by default for OSX

redhatrises:


CMakeLists.txt
```diff
@@ -332,6 +332,9 @@ endif()
 set(OSCAP_TEMP_DIR "/tmp" CACHE STRING "use different temporary directory to execute sce scripts (default=/tmp)")
 
 # ---------- RPATHS for linking
+if(APPLE)
```
mpreisler: according to cmake docs you don't have to guard this with conditionals...

Also why did you add this special case that we barely support as the first line to read when looking at rpath handling?

redhatrises: @mpreisler cmake is generating a policy warning on MAC with RPATHs. I can remove the if, but should I move `set(CMAKE_MACOSX_RPATH ON)` elsewhere? Seemed to make sense to keep with rpath stuff.

mpreisler: it's just a variable so we can set it even when we are not on MacOS X, we do it for all the other rpath stuff. IMO just moved to be the last statement in the RPATH section.

redhatrises: @mpreisler moved it

mpreisler:
ACK

PR #1085, opened 2018-06-28
Add OSX build to master Travis configuration

redhatrises:
- Need to test OSX OpenSCAP builds for integration into SCAP Workbench.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,31 @@
-sudo: required
-
-before_install:
-  - sudo apt-get -qq update
-  - sudo apt-get install -y cmake lcov libdbus-1-dev libdbus-glib-1-dev libcurl4-openssl-dev libgcrypt-dev libselinux1-dev libgconf2-dev libacl1-dev libblkid-dev libcap-dev libxml2-dev swig libxml-parser-perl libxml-xpath-perl libperl-dev librpm-dev swig librtmp-dev xsltproc
-
 language: c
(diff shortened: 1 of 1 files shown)


redhatrises:
OSX fails because of using `ssize_t`. Not sure if there is an option to disable, or if there is a replacement of it in osx.

cipherboy:
@redhatrises See #1086. :)

redhatrises:
Yay! This build is passing now after rebase.

.travis.yml
```diff
@@ -1,25 +1,31 @@
...
+        - bash ../tests/recursively_generate_gcov.sh . || true  # Failures that occur here are harmless
+        - cd .. && sonar-scanner || true  # Will always fail builds on forked repositories.
+      after_success:
+        - curl -s https://codecov.io/bash > cov.sh && bash cov.sh -x "$GCOV"
+    - os: osx
+      osx_image: xcode8.3
+      before_script:
+        - cd build
+      script:
+        - cmake -DENABLE_PYTHON3=false -DENABLE_PROBES=false -DENABLE_OSCAP_UTIL_DOCKER=false ../
```
mpreisler: I wonder if it will build with probes

redhatrises: @mpreisler it fails..... unfortunately... It will however build with this:
```
cmake -DENABLE_PROBES=false
```

PR #1087, opened 2018-06-29
Remove custom allocation functions in SEAP module

### LINKED ISSUE #1077: Remove custom allocations in SEAP
In `src/OVAL/probes/SEAP/sm_alloc.c` we have some functions which are similar to malloc, realloc and calloc. Investigate if they can be removed and replaced by normal functions.

jan-cerny:
It removes `sm_alloc.c` and `sm_alloc.h` and all of the usages of functions from these files.

Custom functions and macros from `sm_alloc` module do not provide any significant value, but they worsen readability. These functions might confuse static analysis tools and other people.

Unfortunately, I have realized that this should be a very low priority for the alpha release.
However, I decided to finish that, because I already had most of it done when I realized this problem.

Fixes #1077


jan-cerny:
https://ci.appveyor.com/project/OpenSCAP/openscap/build/master-86

PR #1090, opened 2018-06-29
Remove deprecated symbols from master branch

### LINKED ISSUE #1088: Remove deprecated symbols from master branch
#### Description of Problem:

OpenSCAP API contains a very large amount of deprecated symbols.
If we remove them before we release 1.3.0-alpha1, we will be able to detect soon if the removal causes any problem.

This could be very useful in modern Linux distributions if they will contain 1.3.0-alpha1. They will be able to detect possible problems with integration of other components that depend on OpenSCAP API.

#### OpenSCAP Version:
current upstream master HEAD

every

#### Steps to Reproduce:

1. open the header files
2. 
3. 

#### Actual Results:

OpenSCAP API contains deprecated symbols

#### Expected Results:

OpenSCAP API doesn't contain deprecated symbols

nothing

jan-cerny:
In this PR I have removed deprecated symbols from master branch.

For further details and detailed description of the specific changes, please read the commit messages of the specific commit.

Fixes #1088


src/CPE/cpelang.c
```diff
@@ -125,11 +114,6 @@ char *cpe_lang_model_detect_version_priv(xmlTextReader *reader)
 	return oscap_strdup("2.3");
 }
 
-char * cpe_lang_model_detect_version(const char* file)
-{
-	return cpe_lang_model_detect_version_priv(NULL);
```
mpreisler: why not also remove the function above, is anything using it?

jan-cerny: cpe_lang_model_detect_version_priv is used only at src/source/oscap_source.c:387. It could be replaced by its code. On the other hand, it is a private function. I would prefer to define this string as a constant in some of the header files.

jan-cerny:
AppVeyor failed! https://ci.appveyor.com/project/OpenSCAP/openscap/build/master-98

jan-cerny:
AppVeyor is now OK! https://ci.appveyor.com/project/OpenSCAP/openscap/build/master-99
Resolved in 332983b.

jan-cerny:
I am going to check if some of the deprecated symbols are used in Workbench and OSCAP Anaconda Addon

jan-cerny:
I have verified that SCAP Workbench builds and runs with OpenSCAP built from the code in this PR.

PR #1092, opened 2018-07-03
Build Windows Installer in AppVeyor

jan-cerny:
This runs `cpack` and adds an AppVeyor Artifact.
See https://www.appveyor.com/docs/packaging-artifacts/

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,14 @@ before_build:
 - cmd: >-
     cd build
 
-    cmake -DENABLE_PYTHON3=FALSE -DCMAKE_TOOLCHAIN_FILE=c:/tools/vcpkg/scripts/buildsystems/vcpkg.cmake ..
+    cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_PYTHON3=FALSE -DCMAKE_TOOLCHAIN_FILE=c:/tools/vcpkg/scripts/buildsystems/vcpkg.cmake ..
 build:
   project: c:\projects\openscap\build\openscap.sln
   verbosity: minimal
+after_build:
+  - cmd: cpack
+artifacts:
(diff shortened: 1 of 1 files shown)


jan-cerny:
See an example in https://ci.appveyor.com/project/OpenSCAP/openscap/build/master-104/artifacts

mpreisler:
So.. this will store 8MB of installer for every commit we build on appveyor?

mpreisler:
Perhaps something like https://github.com/OpenSCAP/scap-workbench/pull/184/files would be less wasteful?

redhatrises:
I know it's a lot of work, but it would be nice to have a .wxs file to be able to customize and have greater control over the msi installation.

mpreisler:
@redhatrises agreed, it would make OpenSCAP look way more professional. We can steal the one in SCAP Workbench.

I think for a lot of people on Windows having a "portable ZIP" would also be nice. We can get both for very little added cost.

jan-cerny:
@mpreisler 

I can't find a way how to disable building artifacts on PRs. 

There is a way how to push artifacts to the GitHub Release
https://www.appveyor.com/docs/deployment/github/

jan-cerny:
@mpreisler I tried to investigate further.

If I understand it well, it would be not possible to push the Installer to the GitHub Release if it would not be specified in `artifacts` section in `appveyor.yml`, which leads to building the Installer in every PR.
See
https://blog.markvincze.com/setting-up-an-appveyor-pipeline-for-golang/

If you don't want to build the Windows Installer, feel free to close this PR.

cipherboy:
We decided that storing artifacts for each PR is fine and AppVeyor might delete old artifacts.

ACK from the team. Thanks @jan-cerny for the improvements! :)

PR #1093, opened 2018-07-03
[WIP] Update WIN CI builds

redhatrises:
- Enable python 3 builds on Windows

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,14 +5,27 @@ branches:
 image: Visual Studio 2017
 configuration: Release
 clone_folder: c:\projects\openscap
+
+environment:
+  matrix:
+    - PYTHON_VERSION: 36
+
+init:
+- set PATH=C:\Python%PYTHON_VERSION%;%PYTHON%\Scripts;%PATH%
+
 install:
+- choco install swig
+- choco install doxygen.install
 - cmd: vcpkg install curl libxml2 libxslt bzip2 pcre pthreads
+
 cache: c:\tools\vcpkg\installed\
+
 before_build:
 - cmd: >-
     cd build
 
-    cmake -DENABLE_PYTHON3=FALSE -DCMAKE_TOOLCHAIN_FILE=c:/tools/vcpkg/scripts/buildsystems/vcpkg.cmake ..
+    cmake -DCMAKE_TOOLCHAIN_FILE=c:/tools/vcpkg/scripts/buildsystems/vcpkg.cmake ..
+
 build:
   project: c:\projects\openscap\build\openscap.sln
   verbosity: minimal
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,6 +16,7 @@ init:
 install:
 - choco install swig
 - choco install doxygen.install
+- cmd: vcpkg update
 - cmd: vcpkg install curl libxml2 libxslt bzip2 pcre pthreads
 
 cache: c:\tools\vcpkg\installed\
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,8 +16,17 @@ init:
 install:
 - choco install swig
 - choco install doxygen.install
-- cmd: vcpkg update
+# update vcpkg
+- cmd: cd C:\tools\vcpkg
+- cmd: git pull
+- cmd: .\bootstrap-vcpkg.bat
+- cmd: vcpkg version
+# remove outdated versions
+- cmd: vcpkg remove --outdated --recurse
+# install required dependencies
 - cmd: vcpkg install curl libxml2 libxslt bzip2 pcre pthreads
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -5,14 +5,20 @@ branches:
...
 configuration: Release
 clone_folder: c:\projects\openscap
+
+environment:
+  matrix:
+    - PYTHON_VERSION: 36
+
 install:
+- set PATH=C:\Python%PYTHON_VERSION%;%PYTHON%\Scripts;%PATH%
 - cmd: vcpkg install curl libxml2 libxslt bzip2 pcre pthreads
```
mpreisler: also needs swig I think

redhatrises: Doesn't fail on build (kinda expect it to, but anyway). Is it Python that needs SWIG?

redhatrises:
Depends on #1096.

This build is still failing right now on Windows.

redhatrises:
Also depends on #1097

cipherboy:
@redhatrises There are conflicts in this PR. Do you mind updating this and rebasing off of master? :)

jan-cerny:
@redhatrises Is this still WIP?

redhatrises:
@jan-cerny unfortunately, yes.... there are some fun things with swig.

jan-cerny:
I will try to check what is wrong with the swig bindings and CMake.

jan-cerny:
AppVeyor CI is experiencing problems with downloading bzip2 source code from bzip.org.

jan-cerny:
Now AppVeyor CI is experiencing problems with connecting to gnu.org. :racehorse: :recycle: :registered: :relaxed:

cipherboy:
@jan-cerny FWIW, I can't connect to https://ftp.gnu.org/gnu/libiconv/ from Boston area either. :/ I seem to recall a post saying that gnu.org was down, perhaps this weekend... Wondering if they're still recovering. 

RIP 2018 FTP.GNU.ORG, purveyor of fine tarballs and other exquisite source code releases.

redhatrises:
I split out the OSX build updates to #1168 since the OSX updates should be good to go.

jan-cerny:
It looks like the bzip2 issue has been already resolved in vcpkg, so we need it to force it to update. See https://github.com/Microsoft/vcpkg/commit/3d87ec6b9fbb07bd299367b6bc89c4a271567233

jan-cerny:
We finally overcome the problem with bzip2 by forcing vcpkg to update its "port" files.

SWIG now runs, but not successfully. It reports the following error:

```
c:\projects\openscap\build\swig\python3\openscappython_wrap.c(67420): error C2065: 'XML_ERROR': undeclared identifier [c:\projects\openscap\build\swig\python3\_openscap_py.vcxproj]
```

jan-cerny:
The XML_ERROR is defined in 
https://github.com/OpenSCAP/openscap/blob/b84c4fce98330b9ed2ac942c6ac693d0e420ab31/src/OVAL/public/oval_types.h#L327-L329
has been excluded on Windows by 99c5f68529109ede238c46ee4e35a8baa3ac46fc

jan-cerny:
The `XML_ERROR` is not used anywhere. What if we remove it?

redhatrises:
I believe that adding the following to `openscap.i` should work.

```
#if defined(_WIN32) || defined(WIN32)
%ignore XML_ERROR;
#endif
```

jan-cerny:
@redhatrises OK, let's see what will happen :smile:

jan-cerny:
Ouch, now there are again some problems in `vcpkg` again. Maybe updating that to their upstream head is not OK. But they don't do any releases or versions.

redhatrises:
appveyor now fails because of vcpkg curl

jan-cerny:
I wonder why it doesn't reuse the cache like for the other PRs. Maybe that's because something changed in `appveyor.yml`? But it seems to me we don't do anything of the items described in https://www.appveyor.com/docs/build-cache/, so it shouldn't invalidate the cache. Maybe it's concerned about the new empty line on line 32, but that would be ridiculous :smile_cat:

matejak:
@openscap-ci Jenkins test this please

matejak:
test this please

ras0219-msft:
>But they don't do any releases or versions.

We recommend pinning to a particular vcpkg commit using git:
```
>git fetch origin
>git checkout <sha>
>.\bootstrap-vcpkg.bat
>vcpkg remove --outdated --recurse
>vcpkg install x y z 1 2 3
```

This will also pin all your dependency versions (and makes it easy to repro any problems outside -- just check out that version!). Whenever you want to update your dependencies, just check what the latest master from Microsoft/vcpkg is and change the SHA to that.

PR #1097, opened 2018-07-03
Use found CMAKE ${PYTHON_EXECUTABLE} for SWIG Python3 CMakeLists.txt

redhatrises:


mpreisler:
There are more cases of this, could you please fix it all?

redhatrises:
@mpreisler fixed them all in this file.

PR #1103, opened 2018-07-06
Set CMAKE_BUILD_TYPE if build is WIN32

redhatrises:
- Fixes issue with cmake -P cmake_install.cmake where cmake files were referencing Debug files


mpreisler:
NACK

this is an antipattern for sure...

PR #1104, opened 2018-07-06
Set CMAKE_BUILD_TYPE

redhatrises:
- When CMAKE_BUILD_TYPE is not specified, it is set to none which causes problems for
  builds on different operating systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,13 @@
 cmake_minimum_required(VERSION 2.6)
 
+# Inspired and referenced from https://blog.kitware.com/cmake-and-the-default-build-type
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+	message(STATUS "Setting build type to 'Release' as none was specified.")
+	set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
+	set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release"
+		"MinSizeRel" "RelWithDebInfo")
+endif()
+
 project("openscap")
 set(OPENSCAP_VERSION_MAJOR "1")
 set(OPENSCAP_VERSION_MINOR "3")


CMakeLists.txt
```diff
@@ -1,5 +1,12 @@
 cmake_minimum_required(VERSION 2.6)
 
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+  message(STATUS "Setting build type to 'Release' as none was specified.")
+  set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
+  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release"
+    "MinSizeRel" "RelWithDebInfo")
+endif()
```
mpreisler: bad indents, plus I am not sold on this

redhatrises:
Fixed indents. See https://blog.kitware.com/cmake-and-the-default-build-type/

jan-cerny:
Hi, I like this PR very much, because undefined `CMAKE_BUILD_TYPE` caused me problems on Windows many times.

I'm not sure if the Release should be the default value. When people develop, they usually want Debug, to be able to run OpenSCAP in a debugger. With Release being default, they will have to type the `-DCMAKE_BUILD_TYPE=Debug`  option explicitly, which is not so convenient. I would vote for changing the default to Debug.

mpreisler:
@jan-cerny I think we should optimize defaults for the majority of users, I presume most users want Release or RelWithDebugSymbols.

jan-cerny:
Aha, I thought that the users use RPM packages.

mpreisler:
Hmm, I feel like this was merged without my comments being addressed...

Do we have plans to implement the same thing with SCAP Workbench?

jan-cerny:
@mpreisler I'm sorry. I thought that indents were fixed and Release type is used. It is not clear to me which concern was not addressed. Perhaps it's discussed somewhere else?

Regarding SCAP Workbench, it depends if there will be problems when making Windows Installer on Windows.

PR #1109, opened 2018-07-10
Build HTML documentation using CMake

### LINKED ISSUE #1083: Add CMake targets to build all the HTML documentation
We forgot to port the dist-hook (https://github.com/OpenSCAP/openscap/blob/maint-1.2/Makefile.am#L95) from the root `Makefile.am` from the `maint-1.2` into the `master` branch. CPack should be extended to generate documentation files using `doxygen` and `asciidoctor` and we should also document in `README.md` that `doxygen` and `asciidoctor` packages are needed for CPack.

jan-cerny:
This adds a new CMake custom target, `docs`, that serves to
build all the HTML documentation. Users can run `make docs`.
The target generates API documentation in HTML format from
source code using Doxygen and generates both User and Developer
manuals in HTML format using Asciidoc.
Removes needless option ENABLE_DOXYGEN_DOCS, and replaces it by
checking if Doxygen is present. Also the `asciidoctor` program
is checked for presence.

Adds a section describing building HTML documentation to the
OpenSCAP Developer Manual.

Adds a target `changelog` that generates a ChangeLog from git history.

Installs the built HTML docs.

Builds also the Contribute docs.

Fixes #1083


jan-cerny:
I have searched the internet and I haven't found any meaningful way how to include the built artifacts to the source tarball using CPack source generators and make it dependent on the `docs` target. It looks like CPack source generator is not intended to do anything else than pack the source files.

I have found articles suggesting using an undocumented variable `CPACK_SOURCE_INSTALLED_DIRECTORIES` that is supposed to add additional directories into the tarball, but I haven't been successful in using it, because there are directories with same name in the source tree.

I think we have multiple options:

1) Do not insist on HTML documentation being included in the source tarball. That would be a difference from maint-1.2. Moreover, that would cause problems in Fedora packaging, because now Doxygen HTML documentation is shipped in `openscap-devel` RPM package.

2) Create a custom target, something like `dist`, that would simulate `make dist` from Automake. That would need to specify all the actions manually. It would be easy to add a dependency on `docs` target and therefore include the docs in the archive. On the other hand, this looks to me like a CMake anti-pattern, because packages are supposed to be generated automatically by CPACK. 

I'm in favor of 2nd solution, because of the consistency with maint-1.2 and using it in Fedora. However, we will be able to change the spec file if we decide to not ship Doxygen in openscap-devel anymore.

Do you have any suggestions?

jan-cerny:
@mpreisler As I mentioned in my previous comment in this PR, I was unable to find a way how to include the build HTML docs to the source tarball using CPack source generators.

mpreisler:
> @mpreisler As I mentioned in my previous comment in this PR, I was unable to find a way how to include the build HTML docs to the source tarball using CPack source generators.

We do not want to include built HTML docs in **source** (emphasis mine) tarball. It's a source tarball, why would it have any built stuff? People can just install doxygen and build it themselves.

jan-cerny:
It looks like simulating make dist from maint-1.2 is not needed and we could have a source tarball without HTML docs.

jan-cerny:
Thanks @mpreisler for clarification of the goal of #1083. 

I will remove the last 2 commits (fa88fec and 18e5fef), that try to exactly port `make dist` form maint-1.2.

Then I will build further work upon commit b88d6e1 which introduces `docs` target in CMake that will builds all the documentation.

jan-cerny:
I think it can be reviewed.

docs/CMakeLists.txt
```diff
@@ -7,8 +7,31 @@ if (ENABLE_DOXYGEN_DOCS)
...
+
+add_custom_target(docs
+    COMMENT "Generate documentation and manuals in HTML format"
+)
+
+if(ASCIIDOCTOR_PRESENT)
+    add_dependencies(docs user_manual developer_manual)
+endif()
+if(DOXYGEN_FOUND)
+    add_dependencies(docs doxygen_docs)
```
mpreisler: please reorg this so you don't repeat the conditions. add the docs target first, then if asciidoctor is present you define targets AND add depenedncies after

jan-cerny: it has been reorganized in 8031d58

CMakeLists.txt
```diff
@@ -427,6 +421,12 @@ if(NOT WIN32)
 	)
 endif()
 
+# changelog
+add_custom_target(changelog
+	COMMAND git log | sed '/^commit/d\; /^Merge/d' > "${CMAKE_BINARY_DIR}/ChangeLog"
```
mpreisler: hardcoded commands, you need to use find_program and find git and sed. This is not an issue on Linux because of standard paths but it would be an issue on OSX and especially Windows

docs/CMakeLists.txt
```diff
@@ -7,8 +11,40 @@ if (ENABLE_DOXYGEN_DOCS)
...
         COMMENT "Generating API documentation with Doxygen")
+    add_dependencies(docs doxygen_docs)
+    install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/html"
+        DESTINATION "${CMAKE_INSTALL_DOCDIR}"
+    )
+endif()
+
+if(ASCIIDOCTOR_PRESENT)
+    add_custom_target(user_manual
+        COMMAND asciidoctor "${CMAKE_CURRENT_SOURCE_DIR}/manual/manual.adoc" -o "${CMAKE_CURRENT_BINARY_DIR}/manual/manual.html"
```
mpreisler: use ASCIIDOCTOR_EXECUTABLE, don't hardcode the names of the commands!

CMakeLists.txt
jan-cerny: Wouldn't it be better to put it inside `if(NOT WIN32)` ?

jan-cerny:
I have changed it to use executables instead of commands.

docs/CMakeLists.txt
```diff
@@ -7,8 +11,40 @@ if (ENABLE_DOXYGEN_DOCS)
...
+    )
+    install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/manual"
+        DESTINATION "${CMAKE_INSTALL_DOCDIR}"
+    )
+    add_custom_target(developer_manual
+        COMMAND "${ASCIIDOCTOR_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}/developer/developer.adoc" -o "${CMAKE_CURRENT_BINARY_DIR}/developer/developer.html"
+        COMMENT "Generating OpenSCAP Developer Manual in HTML format"
+    )
+    install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/developer"
+        DESTINATION "${CMAKE_INSTALL_DOCDIR}"
```
mpreisler: IMO it doesn't make sense to install the developer manual for end-users

docs/CMakeLists.txt
```diff
@@ -7,8 +11,40 @@ if (ENABLE_DOXYGEN_DOCS)
...
+    )
+    add_custom_target(contribute_docs
+        COMMAND "${ASCIIDOCTOR_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}/contribute/contribute.adoc" -o "${CMAKE_CURRENT_BINARY_DIR}/contribute/contribute.html"
+        COMMAND "${ASCIIDOCTOR_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}/contribute/testing.adoc" -o "${CMAKE_CURRENT_BINARY_DIR}/contribute/testing.html"
+        COMMAND "${ASCIIDOCTOR_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}/contribute/versioning.adoc" -o "${CMAKE_CURRENT_BINARY_DIR}/contribute/versioning.html"
+        COMMENT "Generating contribute documentation in HTML format"
+    )
+    install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/contribute"
+        DESTINATION "${CMAKE_INSTALL_DOCDIR}"
```
mpreisler: same here

docs/CMakeLists.txt
mpreisler: we should probably build it but not install it and add a comment explaining why we are not installing it

jan-cerny: aha ok

docs/CMakeLists.txt
jan-cerny: aha ok

jan-cerny:
Hopefully it's now installing the documentation that makes sense.

PR #1110, opened 2018-07-11
Fix broken file name in install command in SWIG Perl bindings

jan-cerny:
A correct file name should be openscap_pm.pm, according to
the target name. Addressing:
```
CMake Error at swig/perl/cmake_install.cmake:61 (file):
  file INSTALL cannot find
  "/home/jcerny/openscap/build/swig/perl/openscap.pm".
  swig/cmake_install.cmake:42 (include)
  cmake_install.cmake:54 (include)
```


matejak:
ACK, I agree with the fix.

PR #1111, opened 2018-07-11
Do not include build-win32 directory in the source tarball

jan-cerny:
Similarly to build directory, the build-win32 directory contains
build binaries and therefore it shouldn't be a part of source
tarball.


matejak:
ACK

PR #1115, opened 2018-07-11
Update OS_defines to use oscap_platforms.h

redhatrises:


redhatrises:
@openscap-ci test this please

src/OVAL/fts_sun.h
```diff
@@ -30,6 +30,7 @@
  */
 
 #include "config.h"
+#include "oscap_platforms.h"
```
mpreisler: why? it's gonna be included as part of config.h

redhatrises: @mpreisler haha... you are right. Should I use `config.h` for the others instead of `oscap_platforms.h`?

redhatrises: @mpreisler ^

mpreisler: every c file should include config as its first include and no header file should include it, that's the general rule

redhatrises: @mpreisler fixed.

compat/oscap_platforms.h
```diff
@@ -21,6 +21,7 @@
 #undef OS_FREEBSD
 #undef OS_LINUX
 #undef OS_SOLARIS
+#undef OS_SUN
```
mpreisler: this is never used, why do we even define it?

src/common/public/oscap_export.h
```diff
@@ -23,6 +23,8 @@
 #ifndef OSCAP_API_H
 #define OSCAP_API_H
 
+#include "oscap_platforms.h"
```
mpreisler: NACK! you can't include it in a public header, this will cause fatal compiler errors to anyone using the openscap API

src/common/public/oscap_debug.h
```diff
@@ -24,6 +24,7 @@
 #define OSCAP_DEBUG_H_
 
 #include "oscap_export.h"
+#include "oscap_platforms.h"
```
mpreisler: NACK. this is a public header

src/OVAL/public/oval_types.h
```diff
@@ -324,7 +324,7 @@ typedef enum {
 	XML_VALIDITY_WARNING = 1,
 	XML_VALIDITY_ERROR = 2,
 	XML_WARNING = 3,
-#ifndef _WIN32
+#ifndef OS_WINDOWS
```
mpreisler: this is never going to be defined at this stage

jan-cerny:
@redhatrises Could you please rebase this on the top of master branch, resolve the conflicts and address the issues found by @mpreisler ?

src/common/public/oscap_export.h
redhatrises: @mpreisler the goal is that `oscap_platforms.h` would define the os for all the code. Is this not possible at all to do with public headers which have OS defines in them?

jan-cerny:
@redhatrises There are still private header files included in 'oval_definitions.h'. That needs to be removed.

redhatrises:
@mpreisler @jan-cerny fixed requested issues.

matejak:
@redhatrises I would like to proceed towards merging this PR, but it is difficult to do so without knowing the intention behind it. Could you please add a description to the issue so I can get started with the final review?

redhatrises:
@matejak sorry... been busy with other tasks the past few weeks. Added description.

matejak:
@redhatrises I can re-review the formal aspect of the PR, but it is still not clear to me what is the intention. Is it about fixing some issue, or is it about unifying OS defines and making them more readable, or maybe is it about something else? I can see the changes in contents, but in order to provide useful feedback or to acknowledge that there is nothing to do any more, one needs to know what is the true intention of the PR, i.e. how it aims to change the project.
So please, be more verbose and share some more background information with us.

jan-cerny:
@matejak I think that there are many `#ifdef`s in our code, they're defined at multiple places and their definitions are not consistent. IMO the purpose would be to concentrate those definitions to a single place and then use them consistently all over the code base.

@redhatrises Could you fix the description and resolve conflicts please?

compat/oscap_platforms.h
```diff
@@ -35,6 +35,7 @@
 #elif defined(__linux__) || defined(__GNU__) || defined(__GLIBC__)
```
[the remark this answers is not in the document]
redhatrises: > Concerning __GNU__, I don't know whether it is not a typo. __GNUC__ is defined iff gcc is used as a compiler, and there is an indication that __GNU__ is defined on the GNU Hurd system, but was this the original author's intention?

Could have been. Removed it per your suggestion.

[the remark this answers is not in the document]
redhatrises: Also introduced `OS_LINUX_WITH_GLIBC` and made it a part of `OS_LINUX` since we don't have it in the code either.

compat/oscap_platforms.h
```diff
@@ -49,6 +50,8 @@
 # if defined(Macintosh) || defined(macintosh) || defined(__APPLE__) && defined(__MACH__)
```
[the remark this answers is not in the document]
redhatrises: Went a head and `||` `defined(__APPLE__)` last. PR should also be rebased.

redhatrises:
Rebased PR. The remaining `_WIN32` are in the public API which I believe cannot be changed.

matejak:
Great, you are right about occurrences of `_WIN32` in public headers. Therefore, I am happy to merge this PR.

redhatrises:
Thanks @matejak!

PR #1116, opened 2018-07-12
Prepare automated release tools for first alpha release from master branch

### LINKED ISSUE #1125: openscap master is reported as 1.3.0 isntead of 1.3.0_alpha1 in pkgconfig
```
$ cmake ..
-- The C compiler identification is GNU 8.1.1
-- The CXX compiler identification is GNU 8.1.1
-- Check for working C compiler: /usr/lib64/ccache/cc
-- Check for working C compiler: /usr/lib64/ccache/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/lib64/ccache/c++
-- Check for working CXX compiler: /usr/lib64/ccache/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done                                                                           
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.4.2")                                                    
-- Checking for module 'libopenscap>=1.2.0'                                                                        
--   Package 'libopenscap', required by 'virtual:world', not found                                                 
CMake Error at /usr/share/cmake/Modules/FindPkgConfig.cmake:418 (message):                                         
  A required package was not found                                                                                 
  /usr/share/cmake/Modules/FindPkgConfig.cmake:585 (_pkg_check_modules_internal)                                   
  CMakeLists.txt:22 (pkg_check_modules)                                                                            
                                                                                                                   
                                                                                                                   
-- Configuring incomplete, errors occurred!                                                                        
See also "/home/mpreisle/d/scap-workbench/build/CMakeFiles/CMakeOutput.log".                                       
mpreisle@localhost ~/d/scap-workbench/build $ cmake ..
-- Checking for module 'libopenscap>=1.2.0'                                                                        
--   Found libopenscap, version 1.3.0                                                                              
-- Configuring done                                                                                                
-- Generating done                                                                                                 
-- Build files have been written to: /home/mpreisle/d/scap-workbench/build
```

libopenscap.pc:
```
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include/openscap

Name: libopenscap
Description: An open source library enabling integration of the SCAP line of standards
URL: https://www.open-scap.org
Version: 1.3.0
Libs: -L${libdir} -lopenscap
Cflags: -I${includedir}

```

jan-cerny:
I started to release a first alpha version from the master branch.

I have tried to run release tools on our master branch to discover "unexpected" problems with the release process.

Prior to that, I have already updated openscap-abi-check to CMake in https://github.com/OpenSCAP/openscap-abi-check/commit/51818c42c8d7bfdcfc6a8e0db207145383580948

In this PR, I fix 2 things in release tools that are blocking the release.

1)  Change version to 1.3.0-alpha1
Change 'previous_version' variable to be able to run the
`up_to_compliance_check.sh` release script which failed,
because ABI compliance checker could not find sources
tarball for 1.3.0.

2) Update calling ABI compliance checker with release tools
The release tools need to be able to run ABI compliance check
between branches.
Also fixes the bug that printf command printed current version
instead of previous version.

The reason of changing version to `1.3.0-alpha1` is that if we would release `1.3.0` directly, then we would need to maintain ABI/API compatibility with that release. With `-alpha1` suffix, we can make multiple alpha or even beta releases before we release the sharp version.

This PR also fixes #1125


release_tools/versions.sh
```diff
@@ -1,5 +1,5 @@
-version=1.3.0
-previous_version=1.3.0
+version=1.3.0-alpha1
+previous_version=1.2.17
```
cipherboy: Why do we want to check ABI compat. against the `1.2.x` series if we don't care about ABI compat. within the `1.3.x` series? :)

jan-cerny: It's an important part of release process to check API/ABI compatibility with the previous released version. We have an ABI/API compliance checker and we produce and publish reports where users can easily see which symbols were added, removed, or changed.

jan-cerny:
We must replace `-` in version by `_` or something like that, because `-` is a forbidden character by RPM. `rpmlint` says: 
```
Illegal char '-' (0x2d) in: Version:        1.3.0-alpha1
```

jan-cerny:
@openscap-ci test this

jan-cerny:
@openscap-ci test this please

release_tools/release.sh
```diff
@@ -102,11 +102,12 @@ check_abi()
...
     rpm -q abi-compliance-checker || sudo dnf install abi-compliance-checker
     test -d openscap-abi-check || git clone https://github.com/OpenSCAP/openscap-abi-check
+    openscap_git_branch=$(git branch | grep '^\* ' | cut -f2 -d' ')
     (
         cd openscap-abi-check
-        perl run_check.pl "$previous_version" "maint-$version_major_minor"
+        perl run_check.pl "$previous_version" "$openscap_git_branch"
         echo 'Read the report, if there is an ABI issue (some symbols were changed / removed), fix it before proceeding further!'
-        printf "%s\n" "Then, Change directory to '$(pwd)' and add the reports. Then, commit with the appropriate message:" "" "cd $(pwd)" "git add reports/${version}_maint-$version_major_minor.html'" "git commit -m 'Report before $2 release'"
+        printf "%s\n" "Then, Change directory to '$(pwd)' and add the reports. Then, commit with the appropriate message:" "" "cd $(pwd)" "git add reports/${previous_version}_${openscap_git_branch}.html'" "git commit -m 'Report before $2 release'"
```
matusmarhefka: This function is called from `release_tools/up_to_compliance_check.sh` without arguments, therefore we should fix the `$2` variable.

jan-cerny:
The `_alpha1` string in the version breaks the Windows installer build. See https://ci.appveyor.com/project/OpenSCAP/openscap/build/master-188#L848
 
It has a problem with candles.
```
Problem running WiX candle. Please check 'C:/projects/openscap/build/_CPack_Packages/win32/WIX/wix.log' for errors.
```

I was able to reproduce this problem on my Windows 10 virtual Machine. I looked into the logs as the error message suggested. In the logs I found:

```
C:\devel\openscap\build\_CPack_Packages\win32\WIX\main.wxs(8) : error CNDL0108 : 
The Product/@Version attribute's value, '1.3.0_alpha1', is not a valid version.  Legal
version values should look like 'x.x.x.x' where x is an integer from 0 to 65534.
```

From this, I guess Windows Installer cannot contain produce a version that contains letters and underscores.

I have discussed the possible solutions with @matusmarhefka . We think that it could be acceptable to change the version only for the purposes of the Windows Installer. I'm going to investigate if we can have such a workaround in our CMakeLists.

jan-cerny:
The version format is restricted by MSI format. See https://docs.microsoft.com/en-us/windows/desktop/Msi/productversion

release_tools/versions.sh
```diff
@@ -1,5 +1,5 @@
-version=1.3.0
-previous_version=1.3.0
+version=1.3.0_alpha1
+previous_version=1.2.17
```
mpreisler: shouldn't this be 1.3.0_alpha1?

jan-cerny: @mpreisler No, it shouldn't. 1.3.0_alpha1 it has not been released yet. There is no `openscap-1.3.0_alpha1.tar.gz` tarball published on Github and there is no tag `1.3.0_alpha1` in git.

mpreisler: why would we compare to 1.2.17? of course it won't be compatible... by design...

jan-cerny: @mpreisler Because 1.2.17 is the latest released version.

mpreisler: well OK but comparing ABI to it makes zero sense

cipherboy:
Why not comment it out for this release until 1.3.1?

release_tools/versions.sh
jan-cerny: @mpreisler Why?

mpreisler: because of course it won't be compatible, it's a different branch

jan-cerny: @mpreisler In the report it is easily visible which symbols were removed and added.

PR #1118, opened 2018-07-12
fileextendedattribute_probe requires xattr header, only enable it if …

mpreisler:
Prevents:
```
[1/122] Building C object src/OVAL/probes/unix/CMakeFiles/unix_probes_object.dir/fileextendedattribute_probe.c.o
FAILED: src/OVAL/probes/unix/CMakeFiles/unix_probes_object.dir/fileextendedattribute_probe.c.o 
/usr/lib64/ccache/cc -DHAVE_CONFIG_H -DOSCAP_BUILD_SHARED -I../compat -I../src -I../src/common -I../src/common/public -I../src/CCE/public -I../src/CPE/public -I../src/CVE/public -I../src/CVRF/public -I../src/CVSS/public -I../src/DS/public -I../src/OVAL/public -I../src/OVAL/probes/public -I../src/OVAL/probes/SEAP -I../src/OVAL/probes/SEAP/public -I../src/OVAL -I../src/source/public -I../src/XCCDF -I../src/XCCDF/public -I../src/XCCDF_POLICY -I../src/XCCDF_POLICY/public -I. -I/usr/include/libxml2 -I../src/OVAL/probes/. -pipe -std=c99 -W -Wall -Wnonnull -Wshadow -Wformat -Wundef -Wno-unused-parameter -Wmissing-prototypes -Wno-unknown-pragmas -D_GNU_SOURCE -D_POSIX_C_SOURCE=200112L -fPIC -MD -MT src/OVAL/probes/unix/CMakeFiles/unix_probes_object.dir/fileextendedattribute_probe.c.o -MF src/OVAL/probes/unix/CMakeFiles/unix_probes_object.dir/fileextendedattribute_probe.c.o.d -o src/OVAL/probes/unix/CMakeFiles/unix_probes_object.dir/fileextendedattribute_probe.c.o   -c ../src/OVAL/probes/unix/fileextendedattribute_probe.c
../src/OVAL/probes/unix/fileextendedattribute_probe.c:44:10: fatal error: attr/xattr.h: No such file or directory
 #include <attr/xattr.h>
          ^~~~~~~~~~~~~~
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,7 @@ check_include_file(shadow.h HAVE_SHADOW_H)
 check_include_file(sys/systeminfo.h HAVE_SYS_SYSTEMINFO_H)
 check_include_file(getopt.h HAVE_GETOPT_H)
 check_include_file(sys/uio.h HAVE_UIO_H)
+check_include_file(attr/xattr.h HAVE_XATTR_H)
 
 # HAVE_ATOMIC_BUILTINS
 check_c_source_compiles("#include <stdint.h>\nint main() {uint16_t foovar=0; uint16_t old=1; uint16_t new=2;__sync_bool_compare_and_swap(&foovar,old,new); return __sync_fetch_and_add(&foovar, 1); __sync_fetch_and_add(&foovar, 1);}" HAVE_ATOMIC_BUILTINS)
@@ -205,7 +206,7 @@ cmake_dependent_option(OPENSCAP_PROBE_INDEPENDENT_XMLFILECONTENT "Independent xm
 # UNIX PROBES
 cmake_dependent_option(OPENSCAP_PROBE_UNIX_DNSCACHE "Unix dnscache probe" ON "ENABLE_PROBES_UNIX" OFF)
 cmake_dependent_option(OPENSCAP_PROBE_UNIX_FILE "Unix file probe" ON "ENABLE_PROBES_UNIX" OFF)
-cmake_dependent_option(OPENSCAP_PROBE_UNIX_FILEEXTENDEDATTRIBUTE "Unix fileextendedattribute probe" ON "ENABLE_PROBES_UNIX" OFF)
+cmake_dependent_option(OPENSCAP_PROBE_UNIX_FILEEXTENDEDATTRIBUTE "Unix fileextendedattribute probe" ON "ENABLE_PROBES_UNIX; HAVE_XATTR_H" OFF)
 cmake_dependent_option(OPENSCAP_PROBE_UNIX_GCONF "Unix gconf probe" ON "ENABLE_PROBES_UNIX; GCONF_FOUND" OFF)
 cmake_dependent_option(OPENSCAP_PROBE_UNIX_INTERFACE "Unix interface probe" ON "ENABLE_PROBES_UNIX" OFF)
 cmake_dependent_option(OPENSCAP_PROBE_UNIX_PASSWORD "Unix password probe" ON "ENABLE_PROBES_UNIX" OFF)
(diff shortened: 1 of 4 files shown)


redhatrises:
OSX uses `<sys/xattr.h>`

mpreisler:
> OSX uses <sys/xattr.h>

That's a valid point but the current version of that file doesn't support OSX. I guess we should look for two headers and use the one that was found in the future. That way we don't even have to check for platofrm but just check which header was found.

Do you want me to support OSX in this pull request?

redhatrises:
This is going into the master branch, right? Support for OSX should be included in this PR. `<sys/xattr.h>` in the master branch. https://github.com/OpenSCAP/openscap/blob/master/src/OVAL/probes/unix/fileextendedattribute_probe.c#L45

CMakeLists.txt
```diff
@@ -205,7 +206,7 @@ cmake_dependent_option(OPENSCAP_PROBE_INDEPENDENT_XMLFILECONTENT "Independent xm
 # UNIX PROBES
 cmake_dependent_option(OPENSCAP_PROBE_UNIX_DNSCACHE "Unix dnscache probe" ON "ENABLE_PROBES_UNIX" OFF)
 cmake_dependent_option(OPENSCAP_PROBE_UNIX_FILE "Unix file probe" ON "ENABLE_PROBES_UNIX" OFF)
-cmake_dependent_option(OPENSCAP_PROBE_UNIX_FILEEXTENDEDATTRIBUTE "Unix fileextendedattribute probe" ON "ENABLE_PROBES_UNIX" OFF)
+cmake_dependent_option(OPENSCAP_PROBE_UNIX_FILEEXTENDEDATTRIBUTE "Unix fileextendedattribute probe" ON "ENABLE_PROBES_UNIX; (HAVE_SYS_XATTR_H OR HAVE_ATTR_XATTR_H)" OFF)
```
jan-cerny: Where are those 2 variables defined?

cipherboy: @jan-cerny: https://github.com/OpenSCAP/openscap/pull/1118/files#diff-57fb65583e7c90f3400179b1a4319f42R78 (config.in)

jan-cerny: @cipherboy 

Not really, In fact that's an usage of these variables. CMake uses them to generate `config.h` from `config.h.in`. 

The problem is they need to be defined somewhere in CMakeLists.txt to make this `cmake_dependent_option` working. I'm afraid that now the fileextendedattribute probe does not build at all.

Most likely there should be on line 151:
```
check_include_file(sys/xattr.h HAVE_SYS_XATTR_H)
check_include_file(attr/xattr.h HAVE_ATTR_XATTR_H)
```

jan-cerny:
Also, I think we should strongly prefer `sys/xattr.h` over `attr/xattr.h` on Linux as well. http://lists.nongnu.org/archive/html/acl-devel/2012-04/msg00001.html

CMakeLists.txt
mpreisler: hmm, I am really puzzled by this because I have it locally, I guess I just didn't commit that hunk

docs/developer/developer.adoc
```diff
@@ -46,7 +46,7 @@ On RHEL / Fedora / CentOS, the command to install the build dependencies is:
 ----
 sudo yum install \
 cmake dbus-devel GConf2-devel libacl-devel libblkid-devel libcap-devel libcurl-devel \
-libgcrypt-devel libselinux-devel libxml2-devel libxslt-devel make openldap-devel \
```
jan-cerny: `attr/xattr.h` does not exist anymore in Fedora 28 and newer. Instead, we should use `sys/xattr.h`. The file `sys/xattr.h` is provided by `glibc-headers` RPM package. 
Please compare F27: https://koji.fedoraproject.org/koji/rpminfo?rpmID=10561854 with F28: https://koji.fedoraproject.org/koji/rpminfo?rpmID=14495196 . Notice that `/usr/include/attr/xattr.h` is not present  anymore in  `libattr-devel` RPM package in F28.

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

PR #1119, opened 2018-07-12
Add header guard to oscap_platforms.h

mpreisler:
Also fixed the copyright date because @redhatrises has just written this, it's not from 2017...


jan-cerny:
@openscap-ci test this please

jan-cerny:
Great catch! Thanks.

PR #1120, opened 2018-07-12
Move dev_to_tty to compat

mpreisler:
It's also used in process probe and it was causing linker errors.

```
FAILED: tests/API/SEAP/test_api_sexp_ID 
: && /usr/lib64/ccache/cc -pipe -std=c99 -W -Wall -Wnonnull -Wshadow -Wformat -Wundef -Wno-unused-parameter -Wmissing-prototypes -Wno-unknown-pragmas -D_GNU_SOURCE -D_POSIX_C_SOURCE=200112L -O2 -DNDEBUG  -rdynamic tests/API/SEAP/CMakeFiles/test_api_sexp_ID.dir/test_api_sexp_ID.c.o  -o tests/API/SEAP/test_api_sexp_ID  -Wl,-rpath,/home/mpreisle/d/openscap/build/src src/libopenscap.so.8.13.0 -lxml2 -lxslt -lexslt -lpcre -lcurl -lpthread -lbz2 -lgcrypt && :
src/libopenscap.so.8.13.0: undefined reference to `dev_to_tty'
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```


src/OVAL/probes/unix/process58-devname.h
```diff
@@ -1,12 +0,0 @@
-
```
jan-cerny: src/OVAL/probes/unix/process58_probe.c and src/OVAL/probes/unix/process_probe.c use this file

compat/CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ if(NOT HAVE_GETOPT_H)
 	list(APPEND COMPAT_SOURCES "getopt.c")
 endif()
 
+if(NOT HAVE_DEV_TO_TTY)
+	list(APPEND COMPAT_SOURCES "dev_to_tty.c")
```
jan-cerny: It probably shouldn't be added on Windows.

compat/dev_to_tty.c
```diff
@@ -30,7 +30,7 @@
 #include <fcntl.h>
 #include <unistd.h>
```
jan-cerny: MSVC has problems with this file (it doesn't know `unistd.h`, doesn't support `restrict` keyword, dosn't have `readlink`, etc.). But it is used only in Unix probes, so it should be enough to remove it  from Windows builds.

mpreisler: fixed

jan-cerny:
It still doesn't build on Ubuntu. https://travis-ci.org/OpenSCAP/openscap/jobs/404322544#L1896

jan-cerny:
I have found that the Travis CI currently is not configured to install `libprocps-dev`, which provides the header file that provides `dev_to_tty`. That means that the `compat` replacement for this function is built on Travis. It builds sucessfully, but then it is not able to link libopenscap.so.

jan-cerny:
@openscap-ci test this please

PR #1122, opened 2018-07-13
Check as much of the library symbols with their actual found paths

mpreisler:
This fixes the build for people who have libraries in /opt, have custom
build environments, SCL, Windows and other platforms where not
everything is in standard paths.

I fixed it for all but pthread because the FindThreads cmake module is really weird and doesn't define the found paths. I think we should replace the whole FindThreads module with something better.


jan-cerny:
@openscap-ci test this please

PR #1124, opened 2018-07-16
Add libprocps-dev to Travis CI and Ubuntu dependencies

jan-cerny:
libprocps-dev provides `proc/devname.h`, which provides `dev_to_tty`
function, which is used by process probe and process58 probe.
It is better to use the system library version than the custom
version from `compat` if the system version is available.
The Travis CI will now install this package and we will recommend
users to do that as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
       sudo: required
       before_install:
         - sudo apt-get -qq update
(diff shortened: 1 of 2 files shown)


mpreisler:
Why not test the compat module on at least one of the platforms?

cipherboy:
@mpreisler Isn't that scope creep? :) I agree that testing what we provide is a good thing...

But, I agree with @jan-cerny's comment that the system version should be preferred to the `compat` version. That should be an independent change from testing `compat` interface. :)

mpreisler:
> @mpreisler Isn't that scope creep? :) I agree that testing what we provide is a good thing...

nah, it's radical scope implosion :D the testing gets done unless this PR gets merged

jan-cerny:
I didn't want any scope.

jan-cerny:
@mpreisler Feel free to close this PR if you think it is not useful change. I think there are benefits of both ways.

PR #1127, opened 2018-07-19
Remove CMAKE_BUILD_TYPE=Release

jan-cerny:
After we have made a default build type Release in da92c1b8a3
we don't need this in AppVeyor anymore.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@ before_build:
 - cmd: >-
     cd build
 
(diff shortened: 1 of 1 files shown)


jan-cerny:
I have tested that with this change Windows Installer is still able to successfully install OpenSCAP on Windows 10 and the installed OpenSCAP runs without any crash.

redhatrises:
Ack

PR #1129, opened 2018-07-20
Add test for filehash58 offline mode.

matejak:
This PR adds a test that tests expected pass and failiure of the `filehash58` probe when doing offline scanning.

### DIFF
--- a/tests/probes/filehash58/test_probes_filehash58.sh
+++ b/tests/probes/filehash58/test_probes_filehash58.sh
@@ -38,9 +38,59 @@ function test_probes_filehash58 {
 	ret_val=1
     fi
 
+    # The file was created as a side-effect of test_probes_filehash58.xml.sh
     [ $ret_val -eq 0 ] && rm -f /tmp/test_probes_filehash58.tmp
 
     return $ret_val
+}
+
+
+# $1: The chroot directory
+function test_probes_filehash58_chroot {
+
+    probecheck "filehash58" || return 255
+    require "sha1sum" || return 255
+
+    local ret_val=0;
+    local DF="$srcdir/check_filehash_simple.xml"
+
+    absolute_probe_root=$(cd "$1" && pwd)
+
+    # oscap-chroot is not readily available during test run, so we use oscap + env var setting.
+    result_keyword=$(OSCAP_PROBE_ROOT="$absolute_probe_root" "$OSCAP" oval eval "$DF" | grep oval_test_has_hash | grep -o '\w*$')
+    
+    [ "$result_keyword" == "$2" ] && return 0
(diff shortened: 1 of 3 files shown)


tests/probes/filehash58/test_probes_filehash58.sh
```diff
@@ -43,10 +43,65 @@ function test_probes_filehash58 {
...
+    local ret_val=0;
+    local DF="$srcdir/check_filehash_simple.xml"
+
+    absolute_probe_root=$(cd "$1" && pwd)
+
+    # oscap-chroot is not readily available during test run, so we use oscap + env var setting.
+    result_keyword=$(OSCAP_PROBE_ROOT="$absolute_probe_root" "$OSCAP" oval eval "$DF" | grep oval_test_has_hash | grep -o '\w*$')
+    
+    [ "$result_keyword" == "$2" ] && return 0
+    # vvv This is more a test error than a failiure or "warning" vvv
```
jan-cerny: failiure

tests/probes/filehash58/test_probes_filehash58.sh
```diff
@@ -43,10 +43,65 @@ function test_probes_filehash58 {
...
+	local ret_val=0
+
+	mkdir -p fail
+	echo bar > fail/oval-test
+
+	test_probes_filehash58_chroot fail false
+	ret_val=$?
+	rm -rf fail
+
+	[ $ret_val -eq 0 ] && rm -f /tmp/$FUNCNAME.tmp
```
jan-cerny: Where is this directory created?

tests/probes/filehash58/test_probes_filehash58.sh
matejak: Good catch, fixed by rebase.

tests/probes/filehash58/test_probes_filehash58.sh
matejak: Another good catch, I was puzzled by the first function of this test suite, fixed by rebase.

jan-cerny:
@matejak Thank you very much. However, now we will have to merge `maint-1.2` to master and then port this test to CTest.

tests/probes/filehash58/test_probes_filehash58.sh
```diff
@@ -38,15 +38,69 @@ function test_probes_filehash58 {
...
 
     return $ret_val
 }
 
+
+# $1: The chroot directory
+function test_probes_filehash58_chroot {
+
+    probecheck "filehash58" || return 255
+    require "sha1sum" || return 255
```
mpreisler: `sha1sum` is not actually used in the script, so why do we skip it if it's not found?

PR #1130, opened 2018-07-20
tests/API/XCCDF/default_cpe: test rewrite

matusmarhefka:
* Removed old XCCDF files (new Fedora versions >24 were not even
  tested) and replaced with XCCDF templates which are used to
  create XCCDF files with CPEs matching the system on which test
  suite is ran.
* From now, the test works with /etc/system-release-cpe to determine
  the CPE of a system instead of using {fedora,redhat}-release RPM
  package (as the scanner does when determining a system CPE).


dahaic:
Thanks @matusmarhefka, the test is much cleaner now. The code coverage dropped is in the code not related to the changes, so it's not a big deal. ACK

PR #1131, opened 2018-07-23
Use suffix for alpha release

redhatrises:
- Fixes issue with scap-workbench which wants integers to comparing and checking the oscap release
  version


redhatrises:
See https://github.com/OpenSCAP/scap-workbench/pull/191

jan-cerny:
We have discussed this with @mpreisler and we think we should introduce a new variable OPENSCAP_VERSION_SUFFIX, which will be empty for regular releases. For alpha and beta releases OPENSCAP_VERSION_SUFFIX would contain the suffix and the OPENSCAP_VERSION_PATCH would be set to 0.

redhatrises:
Changed the PR for suffix versioning... but this maybe not what you envisioned.

CMakeLists.txt
```diff
@@ -11,14 +11,15 @@ endif()
...
-if(WIN32)
-	set(OPENSCAP_VERSION_PATCH "0")
+set(OPENSCAP_VERSION_PATCH "0")
+set(OPENSCAP_VERSION_SUFFIX "alpha2")
+
+if(OPENSCAP_VERSION_SUFFIX AND NOT WIN32)
+	set(OPENSCAP_VERSION "${OPENSCAP_VERSION_MAJOR}.${OPENSCAP_VERSION_MINOR}.${OPENSCAP_VERSION_PATCH}.${OPENSCAP_VERSION_SUFFIX}")
 else()
-	set(OPENSCAP_VERSION_PATCH "0_alpha2")
+	set(OPENSCAP_VERSION "${OPENSCAP_VERSION_MAJOR}.${OPENSCAP_VERSION_MINOR}.${OPENSCAP_VERSION_PATCH}")
```
yuumasato: @redhatrises With introduction of suffix, can't the versions be the same for win and linux? 
As in the CMakeLists it only fetches up to the [third token](https://github.com/OpenSCAP/scap-workbench/blob/v1-2/CMakeLists.txt#L39), I believe the suffix should be ignored by scap-workbench.

redhatrises: @yuumasato not for Windows. It requires integers only for versioning.

CMakeLists.txt
```diff
@@ -11,14 +11,15 @@ endif()
...
 set(OPENSCAP_VERSION_MINOR "3")
-#TODO: Remove this before final 1.3.0
-#FIXME: Windows Installer allows only integers in version name.
-if(WIN32)
-	set(OPENSCAP_VERSION_PATCH "0")
+set(OPENSCAP_VERSION_PATCH "0")
+set(OPENSCAP_VERSION_SUFFIX "alpha2")
+
+if(OPENSCAP_VERSION_SUFFIX AND NOT WIN32)
+	set(OPENSCAP_VERSION "${OPENSCAP_VERSION_MAJOR}.${OPENSCAP_VERSION_MINOR}.${OPENSCAP_VERSION_PATCH}.${OPENSCAP_VERSION_SUFFIX}")
```
mpreisler: NACK, this is not how this should be handled, same version on all platforms!

redhatrises: @mpreisler can we then change `_alpha2` to `.9999`? Can't use a string in versioning for some platforms e.g. `windows`.

jan-cerny: I think this applies only to the Windows installer. In other places we could use the full string.

redhatrises: So, recommendations for path forward?

yuumasato: @redhatrises actually shouldn't the suffix be added after a dash instead of a dot?

yuumasato: Well, it looks like rpm won't like the `-` in the version.

redhatrises: > @redhatrises actually shouldn't the suffix be added after a dash instead of a dot? Would windows build accept strings after the dash?

@yuumasato no. http://wixtoolset.org/documentation/manual/v3/xsd/wix/simple_type_versiontype.html

yuumasato: > NACK, this is not how this should be handled, same version on all platforms!

It seems inevitable if this is a Windows characteristic.

Idea for the windows installer:
How about passing as the version`1.3.0` to `msi` installer, but 
- Set up `oscap -v` to show "correct" version, i.e `1.3.0_alpha2`,
- Show a warning during installation saying that this installs an alpha version,
- And also set CPACK_PACKAGE_NAME "OpenSCAP Alpha2"

jan-cerny:
I have updated it.

jan-cerny:
I have verified that with the patches in this pull request I am able to sucessfully build SCAP Workbench 1.2.0 (obtained from https://github.com/OpenSCAP/scap-workbench/releases/download/1.2.0/scap-workbench-1.2.0.tar.bz2) on a fresh Fedora 28 virtual machine. 

Note: The build `scap-workbench` binary really links to new `libopenscap`, as can be seen from.
```
ldd ./scap-workbench | grep openscap
        libopenscap.so.23 => /lib64/libopenscap.so.23 (0x00007ff0cb275000)
```

CMakeLists.txt
jan-cerny: Changing CPACK_PACKAGE_NAME is not a good solution, because the CPACK_PACKAGE_VERSION is appended to the package name anyway.
I think for now we can survive that the information will not be displayed in the Windows Installer.

yuumasato:
> I think for now we can survive that the information will not be displayed in the Windows Installer.

This is not ideal but will move things forward.
OpenSCAP version will still be a bit confusing without the suffix, but SCAP Workbench will build.
ACK.

PR #1133, opened 2018-07-24
Using AsciiDoc instead of Asciidoctor to generate HTML documentation

jan-cerny:
We need to be able to generate HTML documentation using Asciidoc instead of Asciidoctor, because on some systems Asciidoctor is not available.

The asciidoc command does not support certain syntax that is
supported by asciidoctor command. Fortunately, we can use syntax
that is compatible with both programs. Most notable changes are:
- Do not use triple backquotes for monospaced text, use single
  backquotes that work in both programs.
- Use ---- for source code listing everywhere because they
  are rendered differently in asciidoc than indented blocks,
  whereas in asciidoctor they looked the same as indented blocks.
- Use explicit wrapping in source code listing because asciidoc
  doesn't wrap long lines implicitly in source code listings,
  which causes ugly overlows and extremely long lines that are
  hard to read.
- Fix SSG link.

Then, we need to change CMakeLists to be able to use asciidoc. 
Explicitely set the HTML5 backend because it is not default in asciidoc.
We have to make the target directory before we call the program,
because asciidoc doesn't create directories automatically.

Finally, we removed the mention of `asciidoctor` from Developers Manual and provided instructions to install `asciidoc`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,7 +166,7 @@ endif()
 mark_as_advanced(ENV_PRESENT VALGRIND_PRESENT)
 find_program(ENV_PRESENT env)
 find_program(VALGRIND_PRESENT valgrind)
-find_program(ASCIIDOCTOR_EXECUTABLE NAMES asciidoc asciidoctor)
+find_program(ASCIIDOC_EXECUTABLE asciidoc)
 find_program(SED_EXECUTABLE sed)
 find_program(GIT_EXECUTABLE git)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,7 +166,7 @@ endif()
 mark_as_advanced(ENV_PRESENT VALGRIND_PRESENT)
 find_program(ENV_PRESENT env)
 find_program(VALGRIND_PRESENT valgrind)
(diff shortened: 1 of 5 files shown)


jan-cerny:
I have updated the PR to use only `asciidoc`.

cipherboy:
These changes look good. :) I verified that all the code blocks look like code blocks, that it builds fine with asciidoc installed and that the new files look fine.

One question: [On line 71 of the changed file](https://github.com/OpenSCAP/openscap/pull/1133/files#diff-9e863917501abb474f28b7012df461e5R71) you add a `--` line (whereas your codeblock changes were `----`)... but you didn't add them later. What do the `--` do? It looks from [here](https://powerman.name/doc/asciidoc) that it is for use in lists, but wasn't exactly sure their purpose.

I'm ACK-ing it anyways as the output file looks good, which is what matters. :P


-- Alex

jan-cerny:
@cipherboy That is  a "List Item Continuation". That allows you to put a larger block as a list item which is well indented. See section 17.7 here: http://www.methods.co.nz/asciidoc/chunked/ch17.html

cipherboy:
Ahhh okie! :) Looks good then, thanks for your changes! :)

PR #1135, opened 2018-07-30
Remove CCE API

jan-cerny:
Also removes a test for this part of OpenSCAP API and removes
CCE from the documentation. Reasons for dropping:
- No third party application is using it.
- The headers for CCE are not installed on system, even if openscap is compiled with ENABLE_CCE.
- No CLI utility provided by openscap uses the CCE API.


jan-cerny:
The AppVeyor fail is not related to the change proposed in this PR.

docs/manual/manual.adoc
```diff
@@ -14,7 +14,6 @@
 :scap:  http://scap.nist.gov/
 :nist:  http://www.nist.gov/
 :cpe:   https://cpe.mitre.org/
-:cce:   https://cce.mitre.org/
```
mpreisler: we might still want to include this link because we use CCEs, we just don't use the API we have written for them

docs/manual/manual.adoc
```diff
@@ -44,7 +43,7 @@ http://scap.nist.gov/timeline.html[Release cycle] in order to provide a
 consistent and repeatable revision workflow. The `oscap` mainly processes
 the {xccdf}[XCCDF] which is a standard way of expressing a checklist content and
 defines security checklists. It also combines with other specifications such as
-{cpe}[CPE], {cce}[CCE] and {oval}[OVAL] to create a SCAP-expressed checklist that
```
mpreisler: again, keep this

mpreisler:
It's mostly fine, just found a few cases where you (IMO) went a bit too far and removed links and mentions of CCE. We still use and support CCE in content, just not through our CCE API.

docs/manual/manual.adoc
jan-cerny: ok, makes sense

jan-cerny:
I have removed the removal of CCE from docs.

PR #1136, opened 2018-07-30
Remove validate-xml submodule from CPE, OVAL, XCCDF modules

jan-cerny:
The 'validate-xml' and 'validate' submodules provide the same functionality.
Their output is the same. Therefore it makes sense to keep only one of them.
This commit also replaces all ocurrences of 'validate-xml' by 'validate'
in the test suite.


yuumasato:
Looks good, ACK!

PR #1137, opened 2018-07-30
Remove 'oscap oval list-probes' submodule

jan-cerny:
Users can get a list of probes using `oscap --version` instead.
The extra submodule is not needed in the TUI.
Also, the options of this submodule stopped to be useful after
probes have become part of main oscap process.


matejak:
LGTM, failing tests are unrelated, so merging.

PR #1138, opened 2018-07-30
Remove deprecated option --sce-results

jan-cerny:
This option has been deprecated and has been replaced by --check-engine-results,
which is more generic. With upcoming major release we can remove it.


matejak:
Merging, as it LGTM and the AppVeyor failing test is unrelated. Kudos for not forgetting about man pages!

PR #1139, opened 2018-07-30
Update links between documentations

jan-cerny:
The compilation instructions have been moved from README to OpenSCAP
Developer Manual. This commit updates the links which referenced
those instructions.


yuumasato:
Nice catch! Ack.

PR #1144, opened 2018-07-31
Remove --probe-root option

jan-cerny:
Dropping `--probe-root` option from `oscap oval eval`. It only sets
an environment variable which users can set themselves. Moreover,
it clashes with utilities like oscap-chroot or oscap-docker.

### DIFF
--- a/src/OVAL/public/oval_session.h
+++ b/src/OVAL/public/oval_session.h
@@ -175,7 +175,6 @@ OSCAP_API int oval_session_load(struct oval_session *session);
  *
  * @memberof oval_session
  * @param session an \ref oval_session
- * @param probe_root FIXME:
  * @param id id of an OVAL Definition
  * @param result variable to write the result into
  *
@@ -183,7 +182,7 @@ OSCAP_API int oval_session_load(struct oval_session *session);
  * @retval 1 on an internal error (use \ref oscap_err_desc or \ref
  * oscap_err_get_full_error to get more details)
  */
-OSCAP_API int oval_session_evaluate_id(struct oval_session *session, char *probe_root, const char *id, oval_result_t *result);
+OSCAP_API int oval_session_evaluate_id(struct oval_session *session, const char *id, oval_result_t *result);
 
 /**
  * Evaluate OVAL Definitions. Optionally you can set a callback function which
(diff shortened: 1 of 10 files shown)


utils/oscap-oval.c
```diff
@@ -363,12 +362,12 @@ int app_evaluate_oval(const struct oscap_action *action)
...
 	/* evaluation */
 	if (action->id) {
-		if ((oval_session_evaluate_id(session, action->probe_root, action->id, &eval_result)) != 0)
+		if ((oval_session_evaluate_id(session, NULL, action->id, &eval_result)) != 0)
 			goto cleanup;
 		printf("Definition %s: %s\n", action->id, oval_result_get_text(eval_result));
 	}
 	else {
-		if ((oval_session_evaluate(session, action->probe_root, app_oval_callback, NULL)) != 0)
+		if ((oval_session_evaluate(session, NULL, app_oval_callback, NULL)) != 0)
```
mpreisler: You should IMO also refactor `oval_session_evaluate`, do we need probe_root as an arg there?

jan-cerny: We don't need it. But I think it would break the API,

jan-cerny: aha

jan-cerny:
I have removed unneeded parameter probe_root.

src/OVAL/public/oval_session.h
```diff
@@ -194,15 +193,14 @@ OSCAP_API int oval_session_evaluate_id(struct oval_session *session, char *probe
...
- * @param probe_root FIXME:
  * @param fn a callback function
  * @param arg an optional argument for your callback function
  *
  * @retval 0 on success
  * @retval 1 on an internal error (use \ref oscap_err_desc or \ref
  * oscap_err_get_full_error to get more details)
  */
-OSCAP_API int oval_session_evaluate(struct oval_session *session, char *probe_root, agent_reporter fn, void *arg);
+OSCAP_API int oval_session_evaluate(struct oval_session *session, agent_reporter fn, void *arg);
```
jan-cerny: I think we could also remove  `void *arg` because it is NULL at the only place where it is called.

mpreisler: IMO that's scope creep, it's used for the agent_reporter so let's keep it at least in this PR.

jan-cerny: OK

jan-cerny:
@openscap-ci test this please

PR #1140, opened 2018-07-30
Clean up implementation of strto_ functions

cipherboy:
Because `strtoll`/`strtoull` take a null-terminated
C-string, keep the `strto_` functions as wrappers,
but explicitly null-terminate inputs. This is useful
in the parser, where inputs to `strto_`* are length
terminated substrs of the input.

Fixes the SEAP parser test case failure on gcc 8.1.1
by removing the internal implementation and using `strtoll`
instead.

Signed-off-by: Alexander Scheel <ascheel@redhat.com>

I can redo this PR against `maint-1.2` if you prefer, but I removed a number of unimplemented functions in the header (I'm not sure if it is public or not), so I proposed it against `master`.

PR #1145, opened 2018-07-31
Remove SEXP_parser

cipherboy:
Remove the SEXP parser (`SEXP_parse`) and related code from OpenSCAP. While SEXP data types are used extensively internally, the parser (from `char *` -> `SEXP_t *`) is only used within the test suite for serialization purposes.

There have been and are numerous bugs in the parser: segfaults (#1142), undefined behavior (#1140). Several have yet to be fixed. Rather than trying to fix the parser, removing it seems to be the best course of action.

The two commits in this series first remove its use in the one test case that needed it (`oval_fts_list.c`), and then the other commit removes the source and all other uses of it.


cipherboy:
Rebased again.

jan-cerny:
I will review the code part and @dahaic is going to review the tests.

src/OVAL/probes/SEAP/seap-descriptor.h
```diff
@@ -82,7 +80,7 @@ typedef struct {
 #define SEAP_MAX_OPENDESC 128
 #define SDTABLE_REALLOC_ADD 4
 
-int          SEAP_desc_add (SEAP_desctable_t *sd_table, SEXP_pstate_t *pstate, SEAP_scheme_t scheme, void *scheme_data);
+int          SEAP_desc_add (SEAP_desctable_t *sd_table, SEAP_scheme_t scheme, void *scheme_data);
```
jan-cerny: I would prefer to fix the indents to make the new code comply with our coding style guide, but that§s just a nitpick.

src/OVAL/probes/SEAP/seap-descriptor.c
```diff
@@ -40,8 +39,8 @@ SEAP_desctable_t *SEAP_desctable_new (void)
         return(t);
 }
 
-int SEAP_desc_add (SEAP_desctable_t *sd_table, SEXP_pstate_t *pstate,
-                         SEAP_scheme_t scheme, void *scheme_data)
+int SEAP_desc_add (SEAP_desctable_t *sd_table, SEAP_scheme_t scheme,
```
jan-cerny: I think that in future we will also remove SEAP scheme.

src/OVAL/probes/SEAP/seap-descriptor.h
cipherboy: I'm happy to fix this in a separate PR, but I'm not actually sure what the coding style guide is for OpenSCAP. IIRC `scap-workbench` uses QT style code, but I've seen so many different styles within OpenSCAP I'm not sure which is authoritative. 

Also, I'd prefer to fix the entire file at once rather than fixing just the line, especially since the next line has the same style and should be updated for consistency too. 

That's me though. If you want me to do it here, just tell me what the style guide is and I'll fix stuff :)

jan-cerny: We use Linux Kernel style https://www.kernel.org/doc/html/v4.10/process/coding-style.html
However some of the old code is not using it. We prefer to fix just the line that is being changed for another reason. We don't prefer big patches that change code style in whole file, because downstream Linux distributions and similar people have downstream patches which would become inapplicable. That's what we try to follow.

cipherboy: Okay, thanks! I have updated this. :)

src/OVAL/probes/SEAP/seap-descriptor.c
cipherboy: While I was here, I updated this line to remove the space.

cipherboy:
I also updated the instance of the function call on `src/OVAL/probes/SEAP/seap.c` line 138 to remove the space between the function name and the parenthesis. I believe that should be it of already-changed-lines which need their style updated, but I could be wrong as I didn't do a thorough review/read through of the diff. :)


Thanks for posting the style guide @jan-cerny!

jan-cerny:
@cipherboy you're great

jan-cerny:
@dahaic Do you have any feedback on the test part of this PR?

tests/API/probes/fts.sh
```diff
@@ -50,142 +50,145 @@ while read args; do
...
+5 "$ROOT/d1/d12/f121" \
+"-1" "symlinks and directories" "none" "all" \
 d1/d12/f121,
 
 test2 \
 '' '' \
-'((filepath :operation 11) "^'$ROOT'/d1/.*/f1111")' \
-'((behaviors :max_depth "-1" :recurse "symlinks and directories" :recurse_direction "none" :recurse_file_system "all"))' \
+'' '' \
+11 "^$ROOT/d1/.*/f1111" \\
```
dahaic: double \ it's a nitpick, I know

cipherboy:
@dahaic  Okay, this is updated. I added a new public method for symmetry, `oval_operation_from_text`, that accesses the values of `OVAL_OPERATION_MAP`, and then I updated the test cases and documentation to use that.

jan-cerny:
This is really awesome. I have run the tests and they all pass.

cipherboy:
Thanks @jan-cerny! :) Hopefully this is the last the SEXP_parser will trouble us. XD

PR #1146, opened 2018-07-31
Removed an unused static fn from CVRF

mpreisler:
```
[36/323] Building C object src/CVRF/CMakeFiles/cvrf_object.dir/cvrf_eval.c.o
../src/CVRF/cvrf_eval.c:278:20: warning: ‘get_rpm_name_from_cvrf_product_id’ defined but not used [-Wunused-function]
 static const char *get_rpm_name_from_cvrf_product_id(struct cvrf_session *session, const char *product_id) {
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```


jan-cerny:
@openscap-ci test this please

PR #1147, opened 2018-07-31
The function was freeing unowned memory using const pointers

mpreisler:
```
[37/323] Building C object src/CVRF/CMakeFiles/cvrf_object.dir/cvrf_priv.c.o
../src/CVRF/cvrf_priv.c: In function ‘cvrf_score_set_to_dom’:
../src/CVRF/cvrf_priv.c:1684:7: warning: passing argument 1 of ‘free’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  free(base);
       ^~~~
```


jan-cerny:
That's great!

PR #1148, opened 2018-08-01
Do not build the documentation on every build

jan-cerny:
Makes the 'docs' an independent target, which does not depend on
'ALL' pseudotarget. This brings us back fast incremental rebuilds,
which is useful for developing and debugging. Users can still build
the documentations easily using 'make docs'.


matusmarhefka:
ACK.

PR #1149, opened 2018-08-01
Reduce warnings

cipherboy:
A bunch of random small fixes to reduce warnings on `master`. I haven't tested their applicability against the `maint-1.2` branch.

All of these were found in gcc output.

I'd appreciate feedback on the `snprintf` related changes; it'd seem like we'd want to raise an error rather than truncating, but I'm not entirely sure which errors, etc. Thoughts?


cipherboy:
@openscap-ci test this please

cipherboy:
It looks like the murmur3 changes are failing, so I've dropped them from this patch series.

cipherboy:
Looks like the RPM changes aren't liked by Travis: https://travis-ci.org/OpenSCAP/openscap/jobs/410907166#L1788

jan-cerny:
I think that Travis has a really old librpm that doesn't have yet a replacement for the deprecated function.

cipherboy:
I'll postpone the RPM changes until later. This'll probably need build-time detection and gets complicated.

jan-cerny:
Thank you.

Issue #938, opened 2017-12-15
Some public symbols in public headers might not be public


jan-cerny:
Before we release OpenSCAP 1.3.0 we need to review our public functions, because it might turn out that some of the public functions shouldn't be public, or some of them are in wrong header files. 

For example https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/OVAL/probes/SEAP/public/sm_alloc.h#L38

I think it is important to do this review because after we release 1.3.0 we will have to maintain the public API again and it will be forbidden to break it.

yuumasato:
@jan-cerny I believe this should a blocker then

jan-cerny:
@yuumasato Do you think that this blocks only the final release or it blocks the alpha release as well?

I think it would be very useful to have this finished in alpha, because then we can catch possible problems with other components soon.

yuumasato:
@jan-cerny I agree with you. I even think that alpha can be a bit bold on removing symbols, and beta more conservative and fixing problems with symbols.

jan-cerny:
It doesn't necessarily block the first alpha release.

evgenyz:
We had our chance with `1.4` but probably we won't be able to work on that ever. Closing.

PR #1150, opened 2018-08-03
[do not merge] [WIP] Move SEXP, SEAP and probe internals out from public API

jan-cerny:
This pull request is a first part of our API cleanup before the 1.3.0 release.

The SEXP and probe modules are mostly implementation details. they should not be a part of public API. After inclusion of probes into libopenscap, we don't need them in the public API anymore. 
Therefore, we move the symbols to private headers.

For more details, please see the commit messages of respective commits.

This PR also will help us to fix issue #938 .


cipherboy:
@jan-cerny Do you want to wait until #1145 is merged / rebase on top of #1145? This will conflict and there is little point in moving the parser from public -> private only to have it removed entirely :)

jan-cerny:
@cipherboy Yes, I would like to wait until #1145 is merged. Then, I would rebase my PR on the top of master. :)

jan-cerny:
I'm going to rebase on the top of master branch and resolve all the conflicts.

jan-cerny:
After merging https://github.com/OpenSCAP/openscap/pull/1165 and rebasing this pull request on the top of master branch, I'm not able to build some tests anymore.

Specifically, I have terrible problems with compiling `oval_fts_list.c` and `test_api_probes_smoke.c` from `tests/API/probes/`. These tests test some functions that won't be public anymore. Therefore linking their binaries with libopenscap.so is not enough to build them. I tried to add some C files from the source tree, but they have so many dependencies that I ended up with building most of the OpenSCAP library twice.

@cipherboy @mpreisler Do you have any opinions on this problem?

We have already discussed this with @dahaic and @matejak.

My opinion was that I will try to modify the tests to reduce their dependencies, but I haven't been successful much, because the FUTs bring those dependencies.

One of the suggestions was to make those symbols public, but undocumented, so that people will not know how to use them, so they will avoid them. I'm afraid that it would be more difficult to maintain the ABI then, because the changes that would made in the undocumented public symbols will affect the ABI.

Another solution could be to drop these tests. But they test key parts of the code, eg. they test FTS, which is a backend of all our filesystem related probes.

Another solution is to introduce public ABI and private ABI, but I haven't found anything on the internet how to do that.

cipherboy:
@jan-cerny I'm not quite sure how this works in practice in openscap...

But iirc there's a way to strip symbols from the generated code artifacts. So you can build the test suite, check it against the unstripped library, and generate the "final" library from the unstripped library by stripping unwanted symbols. 

This is roughly what `krb5` does:
 - Maintain a list of explicit exports, e.g. [libgssapi_krb5.exports](https://github.com/krb5/krb5/blob/master/src/lib/gssapi/libgssapi_krb5.exports)
 - Automagic/CMagic it into `--version-script`: https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html

Because this is a linking stage command, we can duplicate our object creations (one without the `--version-script`), and generate a `*-test.so` and link all of our tests against that. That'd have all of the symbols and then we can keep our unit tests as is, and we can further prune our output symbols.

Thoughts? It'd require maintaining an explicit list of exports, but 1) that'd be a good practice to keep it explicit, and 2) we can mostly auto-generate it up front from your new public exports tool and then prune the things we don't want, and/or from the installed headers.

jan-cerny:
I think that's a nice solution. I'm afraid that we would forget to update that list, so I think the list should be generated automatically from the source code. The symbols will still be labeled by OSCAP_API macro in the headers, but the macro will be defined as empty string.

@mpreisler Do you have any opinion on this topic?

jan-cerny:
This article "How To Write Shared Libraries" might be helpful, especially Section 2.2.5. https://www.akkadia.org/drepper/dsohowto.pdf

jan-cerny:
This PR is blocked by #1178 and #1176 .

jan-cerny:
This PR is blocked by #1178

jan-cerny:
This PR blocks #1182

jan-cerny:
This PR needs to be reworked because the symbol visibility and test linking problems will not be solved the way that was proposed in #1178 .

jan-cerny:
We now need to resolve builds of problematic tests that were disabled temporarily.

test_api_seap_string
test_api_seap_number
test_api_seap_list
test_api_sexp_deepCmp
test_api_seap_concurency
test_api_probes_smoke
oval_fts_list

jan-cerny:
I have split out the SEXP list functions and SEXP list recursive functions out to separate files.

jan-cerny:
I have resolved the conflicts and rebased this on the top of our master's head.

matejak:
jenkins test this please

matejak:
jenkins test this please

matejak:
jenkins test this please

matejak:
jenkins test this please

jan-cerny:
This PR is going to be abandoned after the non-controvesial patches will be splitted out.

PR #1151, opened 2018-08-03
add a safety \n before the end script comment.

msmeissn:
This helps if someone uses <fix></fix> without a terminating newline


jan-cerny:
Thank you!

PR #1153, opened 2018-08-09
Skip sysctl probe test if no SysV services are found

jan-cerny:
If the list of services returned by `chkconfig --list` is empty,
then the generated OVAL file would be invalid. Since there would
not be anything to test in that situation, it would be better
to skip the test instead. This situation can happen on modern
Linux distributions that have only native systemd services.

### DIFF
--- a/tests/probes/runlevel/test_probes_runlevel.sh
+++ b/tests/probes/runlevel/test_probes_runlevel.sh
@@ -25,7 +25,7 @@ function test_probes_runlevel_A {
     require "egrep" || return 255
     require "awk" || return 255
 
-    local services_list=`get_services_list`
+    local services_list="$(get_services_list)"
     [ -n "$services_list" ] || return 255
 
     local ret_val=0;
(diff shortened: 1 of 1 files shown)


tests/probes/runlevel/test_probes_runlevel.sh
```diff
@@ -25,6 +25,9 @@ function test_probes_runlevel_A {
     require "egrep" || return 255
     require "awk" || return 255
 
+    local services_list=`get_services_list`
```
cipherboy: Do you mind using the preferred `local services_list="$(get_services_list)"` syntax? iirc that's what `shellcheck` recommends.

cipherboy: (However, the script does use backticks a lot so up to you).

cipherboy:
I haven't validated the functionality of this on a system for which it _is_ empty, but it does not affect a system for which it isn't empty.

tests/probes/runlevel/test_probes_runlevel.sh
jan-cerny: Yes, I'll change that.

mpreisler:
Shouldn't this be done against `maint-1.2`?

jan-cerny:
Updated to maint-1.2

cipherboy:
Looks good, @jan-cerny, thanks! :)

PR #1154, opened 2018-08-09
Move Probe handler to private header

jan-cerny:
The probe handler is an internal implementation detail and it should not be included in the public API.
This PR proposes to move the probe handler to a private header file.

To remove the probe handler from the public API , we need first to get rid of `oval_probe_session_sethandler ` from public module OVAL_SESSION. This function just returns -1.  The function is never used. This removal breaks API.

We also remove unused function `oval_probe_session_close` which is also not used anywhere.


cipherboy:
This is not a thorough review but I verified that tests pass locally and that the changes look sane. I'm not sure if anyone else has comments though. 

Thanks for the cleanup @jan-cerny! :)

mpreisler:
ACK

PR #1155, opened 2018-08-09
Fix some warnings in maint-1.2

jan-cerny:
This PR fixes some of the annoying warnings by compiler.

For more details and addressing, please see the commit messages of specific commits.

### DIFF
--- a/src/OVAL/probes/independent/sql.c
+++ b/src/OVAL/probes/independent/sql.c
@@ -186,8 +186,9 @@ static int dbURIInfo_parse(dbURIInfo_t *info, const char *conn)
 			if (*(tok) != '=') goto __fail;		\
 			else (dst) = strdup((tok) + 1);		\
 		}						\
-		else dE("Unrecognized token: '%s'", (tok)-1);		\
-	while(0)
+		else {	\
+			dE("Unrecognized token: '%s'", (tok)-1);		\
+		}
 
 #define matchitem2(tok, first, rest1, dst1, rest2, dst2)		\
 	case first:							\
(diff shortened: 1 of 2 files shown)


src/OVAL/probes/SEAP/sexp-manip.c
```diff
@@ -1912,7 +1912,7 @@ int SEXP_datatype_set_nth (SEXP_t *list, uint32_t n, const char *name)
         SEXP_val_t v_dsc;
         SEXP_t    *s_nth;
 
-	if (list == NULL || name == NULL) {
+	if (list == NULL) {
```
cipherboy: I was looking at fixing this, but I wasn't sure that this'd always work. In particular, I _think_ `clang` supports the `nonnull` modifier, but I'm not sure if MSFT's `cc` does. Can you validate that it does?

cipherboy:
Ah one additional question: Except for the struct change (which I assume changes ABI/API)... do most of these apply to `maint-1.2`? I'm guilty of this as well I think. :D

src/OVAL/probes/SEAP/sexp-manip.c
jan-cerny: It looks like it does not. Maybe it would be better to remove the modifier and keep this condition instead.

cipherboy: I was thinking about disabling this particular warning for that file with `cmake`. The reason why I'd suggest this is because we still get the `nonnull` compile time checks for supported compilers and we can enforce non-null for compilers which don't support it. The overhead of leaving this in the code is small. 

The compiler check may not be of enough value to keep though. Thoughts?

jan-cerny:
Updated to maint-1.2.

jan-cerny:
AppVeyor fail is expected, beacuse AppVeyor can't run on maint-1.2. This PR was originally based on master, so AppVeyor continues to trying building it.

cipherboy:
LGTM, thanks for doing this against `maint-1.2`, @jan-cerny! :)

Tests pass locally and warnings appear to be reduced.

PR #1156, opened 2018-08-09
[WIP] [HELP WANTED] Container builds

cipherboy:
This adds several new container builds to Travis CI for PR gating:

 - gcc 4
 - gcc 5
 - gcc 6
 - gcc 7
 - ggc latest
 - clang (`fedora:rawhide`)
 - SSG Integration test

I'm not sure if we want all of these, but they're added anyways. I'm guessing between rhel6, rhel7, older Ubuntu/Debian, and new Fedoras, a majority of these compilers are still being supported / used frequently in the wild.

The integration test might be preferable to run as a nightly test as it does take a while to run, longer than the other jobs, but that's up for discussion. 

Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,6 +101,15 @@ matrix:
       script:
         - docker run openscap_$BASE_IMAGE:latest
     - os: linux
+      env: BASE_IMAGE="fedora_clang"
+      sudo: required
+      services:
+        - docker
+      before_install:
+        - docker build --no-cache --tag openscap_$BASE_IMAGE:latest -f Dockerfiles/$BASE_IMAGE .
+      script:
+        - docker run openscap_$BASE_IMAGE:latest
+    - os: linux
       env: BASE_IMAGE="fedora_ssg"
       sudo: required
       services:
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,51 @@ matrix:
       script:
         - cmake -DENABLE_PROBES=false ../
         - make -j 4
+    - os: linux
+      env: BASE_IMAGE="gcc_latest"
+      sudo: required
+      services:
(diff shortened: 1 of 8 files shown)


cipherboy:
This is broken for a number of reasons, mostly due to `travis-ci`. In particular, scheduling both the first job and any of the containers on the same system appear to cause it to go boom.

cipherboy:
Fedora Build (`fedora_clang`) on TravisCI: 
https://paste.fedoraproject.org/paste/u~k0TCgucjnwSL0D9xM6xg

Locally (`fedora_clang` via docker commands in `.travis.yml`):
https://paste.fedoraproject.org/paste/BJVUqL6-p08pNzIdFotBig

So I'm really confused why this is happening. Does anyone have any suggestions?

jan-cerny:
It looks like it's a problem in Travis CI, we think that discussing the problem with Travis customer support is only way to proceed.

matusmarhefka:
What about moving this to our Jenkins Fedora node? It would just require to set up docker and we can better debug issues as we have the control over nodes. Downside is that it would take much more CPU time.

Another solution might be to move it to CentOS CI but that probably cannot be used for PR gating.

jan-cerny:
We have re-executed the job and we will decide based on the latest results.

@matusmarhefka Yes, we can run that in the new RHEL7 node in our Jenkins.

cipherboy:
Either my code changes are bad or `master` doesn't build... But it seems like master should build. I'm fine closing that if that'd be preferable, else you can cannibalize this and take what you need. :)

cipherboy:
I'm going to close this. I don't delete branches and this PR can serve as a template to anyone else desiring to add container builds on Travis.

PR #1159, opened 2018-08-10
Disable MITRE tests by default

cipherboy:
Gated around `ENABLE_MITRE`. Since `tests/mitre/*` seem fairly unstable, doing this seems like a benefit for stability. 

`Signed-off-by: Alexander Scheel <ascheel@redhat.com>`


cipherboy:
(Travis failure unrelated `apt-get install gcc-7` failed), restarting that job.

jan-cerny:
I think it would be better to find out why MITRE tests are unstable instead of disabling it. These tests provide many valuable checks. Is it a problem in the test suite or a bug in OpenSCAP?

cipherboy:
@jan-cerny It is a bug in the tests. They are not applicable to be run on all systems.

Here is the output of running the tests on my system: https://paste.fedoraproject.org/paste/qxOljEnKjSn09IcxKlGiDg

In particular, this is the failing test:
```
+ linux-def_inetlisteningservers_test.xml                     
TEST: linux-def_inetlisteningservers_test.xml
Definition oval:org.mitre.oval.test:def:789: false
Evaluation done.
RESULT: FAILED
```

That test uses [`linux-def_inetlisteningservers_test.xml`](https://github.com/openscap/openscap/blob/master/tests/mitre/linux-def_inetlisteningservers_test.xml). 

Line 15 has an important note:

```xml
        <note>It is important to note that the values used in this test definition are specific to a particular machine, and may not necessarily apply to your system. As a result, in order to have the definition return a result of 'true', you must either change these values to the ones found on your particular system, or you must configure your system to use these values.</note>
```

In particular, it expects port 25 to be open by default (lines 55-60):
```xml
  <states>
    <inetlisteningservers_state id="oval:org.mitre.oval.test:ste:653" version="2" comment="This inetlisteningservers_state represents an inetlisteningservers_item which has a protocol equal to 'tcp', a local_address equal to '127.0.0.1', and a local_port equal to '25'." xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5#linux">
      <protocol>tcp</protocol>
      <local_address>127.0.0.1</local_address>
      <local_port datatype="int" operation="equals">25</local_port>
    </inetlisteningservers_state>
```

(I for one do not run a SMTP server, and running it on port 25 (iirc requires unencrypted, no TLS SMTP) seems like a Bad Idea (TM)).

I suppose we could introduce additional complexity to run a server on a free port, have it report which port it is listening on, and then update the OVAL definition to test for that port and validate the results, but then we might also need to verify that that port is reachable and not firewalled, etc. (I suppose `localhost` _shouldn't_  be firewalled, but meh). 







A separate and not failing issue on my Fedora box is:

In [`test_mitre_linux_probes.sh`](https://github.com/openscap/openscap/blob/master/tests/mitre/test_mitre_linux_probes.sh), lines 15 and 16 shouldn't be enabled on all platforms by default and should only be enabled on platforms with RPM, like how `dpkg` is gated on lines 31-33, but perhaps less brittle. 

```bash
test_run "linux-def_rpminfo_test.xml" test_mitre linux-def_rpminfo_test.xml "true"
test_run "linux-def_rpmverify_test.xml" test_mitre linux-def_rpmverify_test.xml "true"
```

Looking more at the other tests, a lot of them seem highly specific and do not work generically. They need significant environmental support and aren't necessarily useful as a general purpose, cross-platform unit tests or functionality tests. Perhaps setting up the right environment as part of containerized builds would be useful for enabling tests to be run in a CI or personal environment, but for shipped, defaulting to enabled tests, I think we'll get a lot of false positives.

I'm sure they're useful tests, if and when environment support is present, but perhaps they should either have better feature detection before being enabled or better yet, automatic scripts/containers/Kickstarts/etc. to set up a conforming environment.

CMakeLists.txt
```diff
@@ -299,6 +299,8 @@ endif()
 
 option(ENABLE_VALGRIND "enables Valgrind memory testing in the test-suite" FALSE)
 
+option(ENABLE_MITRE "enables MITRE tests -- prone to random failures" FALSE)
```
jan-cerny: So, I think the failures are not random, but the tests require a very specific environment.

jan-cerny:
Awesome write-up. I think it would be valuable to summarize it and add a new paragraph to the OpenSCAP testing guide https://github.com/OpenSCAP/openscap/blob/master/docs/contribute/testing.adoc where it would describe the environment setup needed to run these tests.

cipherboy:
@jan-cerny Updated the comment! I also fixed some spelling mistakes in the `CMakeLists.txt`, added a Docker container for testing using the Mitre tests, and some documentation on both. 

I think there are additional dependencies, but I haven't yet tried building on Ubuntu and verifying whether or not that works, but perhaps that can be done in a later PR.

cipherboy:
@jan-cerny Okay, this is updated and ready for review :)

docs/contribute/testing.adoc
```diff
@@ -287,3 +299,33 @@ $ less Testing/Temporary/LastTest.log
 Results from testing will be printed on the stdout and detailed log file with
 your test results can be found in the `Testing/Temporary/LastTest.log` file.
 
+== Running the MITRE tests
+
+The MITRE tests (in `tests/mitre`) are an example of how not to write tests.
```
jan-cerny: I think we shouldn't put this evaluation here, because MITRE is an external contributor and we are happy that they provided us the tests.

docs/contribute/testing.adoc
```diff
@@ -287,3 +299,33 @@ $ less Testing/Temporary/LastTest.log
...
+$ cmake -DENABLE_MITRE=TRUE ..
+$ make
+$ ctest --output-on-failure -R mitre
+----
+
+To run the containerized MITRE tests (which installs a SMTP server and tests
+it):
+
+----
+$ cd scap-security-guide/
```
jan-cerny: It should be `cd openscap`

docs/contribute/testing.adoc
cipherboy: Ah sorry. Fair point, I'll update.

docs/contribute/testing.adoc
cipherboy: :) I think I've contributed to SSG too much.

PR #1160, opened 2018-08-10
Travis improvements

cipherboy:
A few Travis improvements copied from #1156. Install dependencies via the `apt` addon seems less error prone than the previous method.

Also includes a documentation fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,32 @@ language: c
 matrix:
   include:
     - os: linux
-      sudo: required
-      before_install:
-        - sudo apt-get -qq update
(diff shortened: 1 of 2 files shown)


mpreisler:
I wonder if we should do the same thing for SSG travis apt-get installs.

cipherboy:
`SSG` Travis doesn't use a native Travis instance. It uses Docker: https://github.com/OpenSCAP/scap-security-guide/blob/master/.travis.yml#L15-L19

mpreisler:
Ah, right...

PR #1162, opened 2018-08-14
Introduce a script for comparison of symbols in SO and public headers

jan-cerny:
This commit introduces a new short Python script that compares the exported
symbols in libopenscap.so with symbols marked as OSCAP_API in public header
files. The script prints out 2 lists:
* a list of symbols that are exported in binary, but are not marked as public
* a list of symbols that are marked as public, but are not exported in binary

### DIFF
--- a/tests/check_public_api.py
+++ b/tests/check_public_api.py
@@ -1,0 +1,113 @@
+#!/usr/bin/python3
+
+import subprocess
+import tempfile
+import os
+import re
+
+
+def get_public_symbols_from_shared_object(prefix):
+    """
+    Find public (exported) symbols in a shared library (*.so) by calling
+    the 'nm' command.
+    :param prefix: CMAKE_INSTALL_PREFIX
+    :return: a set of public symbols
+    """
+    so_path = os.path.join(prefix, "lib64", "libopenscap.so")
+    nm_command = ["nm", "--demangle", "--dynamic", "--defined-only", so_path]
+    nm_output_bytes = subprocess.check_output(nm_command)
+    nm_output = nm_output_bytes.decode(encoding="utf-8")
+    symbols = [line.split()[2] for line in nm_output.splitlines()]
+    return set(symbols)
+
+
+def get_public_headers(prefix):
+    """
+    It is useful to check if all the public headers are actually installed under $PREFIX/include/openscap.
+    :param prefix: CMAKE_INSTALL_PREFIX
+    :return: a set of public headers file paths
+    """
+    public_headers = []
+    include_path = os.path.join(prefix, "include", "openscap")
+    for root, dirs, files in os.walk(include_path):
+        for f in files:
+            full_path = os.path.join(root, f)
+            public_headers.append(full_path)
+    return set(public_headers)
+
+
+def _get_public_symbols_from_header(header_path):
+    symbols = []
+    prototype_regex = re.compile(r"^[ \t]*OSCAP_API[^;]*;", re.MULTILINE)
+    with open(header_path, "r") as header_file:
+        content = header_file.read()
+    prototypes = prototype_regex.findall(content)
+    function_name_regex = re.compile(r"([a-zA-Z0-9_]+)\s*\(.*\)")
+    for p in prototypes:
+        p = p.replace("\n", " ")
+        f_match = function_name_regex.search(p)
+        f_name = f_match.group(1) if f_match else "unknown"
+        symbols.append(f_name)
+    return symbols
+
+
+def get_public_symbols_from_headers(headers):
+    """
+    Find public (exported) symbols in header files by finding symbols
+    marked as OSCAP_API.
+    :param headers: a list of public headers file paths
+    :return: a set of public symbols of public headers file paths
+    """
+    symbols = []
+    for h in headers:
+        symbols.extend(_get_public_symbols_from_header(h))
+    return set(symbols)
+
+
+def analyze(src_dir):
+    cwd = os.getcwd()
+    #so_symbols = set()
+    #header_symbols = set()
+    with tempfile.TemporaryDirectory() as prefix:
+        build_dir = os.path.join(src_dir, "build")
+        os.chdir(build_dir)
+        cmake_command = ["cmake",
+                         "-DENABLE_PYTHON3=FALSE",
+                         "-DENABLE_PERL=FALSE",
+                         "-DCMAKE_INSTALL_PREFIX=" + prefix,
+                         "-DENABLE_TESTS=FALSE",
+                         ".."]
+        subprocess.run(cmake_command)
+        make_command = ["make"]
+        subprocess.run(make_command)
+        # TODO: remove make docs after https://github.com/OpenSCAP/openscap/issues/1161 is resolved
+        make_docs_command = ["make", "docs"]
+        subprocess.run(make_docs_command)
+        make_install_command = ["make", "install"]
+        subprocess.run(make_install_command)
+        public_headers = get_public_headers(prefix)
+        so_symbols = get_public_symbols_from_shared_object(prefix)
+        header_symbols = get_public_symbols_from_headers(public_headers)
+    os.chdir(cwd)
+    return header_symbols, so_symbols
+
+
+def main():
+    src_dir = os.path.join(os.getcwd(), "..")
+    header_symbols, so_symbols = analyze(src_dir)
+    print("Shared object symbols: %d" % len(so_symbols))
+    print("Public header symbols: %d" % len(header_symbols))
+    print()
+    so_only = so_symbols.difference(header_symbols)
+    print("The following %d symbols are exported in binary, but are not present in public header files:" % len(so_only))
+    for s in sorted(so_only):
+        print(s)
+    header_only = header_symbols.difference(so_symbols)
+    print()
+    print("The following %d symbols are present in public header files, but are not exported in binary:" % len(header_only))
+    for s in sorted(header_only):
+        print(s)
+
+
+if __name__ == "__main__":
+    main()
--- a/tests/check_public_api.py
+++ b/tests/check_public_api.py
@@ -23,7 +23,8 @@ def get_public_symbols_from_shared_object(prefix):
 
 def get_public_headers(prefix):
     """
-    It is useful to check if all the public headers are actually installed under $PREFIX/include/openscap.
+    It is useful to check if all the public headers are actually installed
+    under $PREFIX/include/openscap.
     :param prefix: CMAKE_INSTALL_PREFIX
     :return: a set of public headers filepaths
     """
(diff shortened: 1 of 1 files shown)


tests/check_public_api.py
```diff
@@ -0,0 +1,113 @@
...
+    :param headers: a list of public headers file paths
+    :return: a set of public symbols of public headers file paths
+    """
+    symbols = []
+    for h in headers:
+        symbols.extend(_get_public_symbols_from_header(h))
+    return set(symbols)
+
+
+def analyze(src_dir):
```
cipherboy: Any chance we could get a docstring and perhaps a better function name? I realize that this is internal to the script but perhaps something like `analyze_project_artifacts`?

tests/check_public_api.py
```diff
@@ -0,0 +1,131 @@
...
+    library. This is done on a real build and a real installation.
+    The artifacts are installed in a temporary directory which is
+    automatically removed when this function returns.
+    :param src_dir: Path to project sources
+    :return: tuple (header_symbols, so_symbols)
+    """
+    cwd = os.getcwd()
+    with tempfile.TemporaryDirectory() as prefix:
+        build_dir = os.path.join(src_dir, "build")
+        os.chdir(build_dir)
```
cipherboy: I think it would be good to add a sanity check here for when `build` is non-empty or perhaps when the build was configured with `ninja` instead of `make`.

That's a non-trivial gotcha that some people will run into that is easily detected by this script (`build.ninja`) and causes an opaque error:

```
-- OpenSCAP 1.3.0_alpha3
-- (see /home/cipherboy/GitHub/jan-cerny/openscap/docs/developer/developer.adoc for build instructions)
--  
-- Checking for module 'acl'
--   Package 'acl', required by 'virtual:world', not found
-- Found ACL 
-- Found BLKID 
-- Found GLib 
-- Found GObject 
-- Found GCONF 
-- Checking for module 'opendbx'
--   Package 'opendbx', required by 'virtual:world', not found
-- Found POPT 
-- Checking for module 'libprocps'
--   Package 'libprocps', required by 'virtual:world', not found
-- Found RPM 4.14.1
-- Found SELINUX 
--  
-- CMake:
-- generator: Ninja
-- source directory: /home/cipherboy/GitHub/jan-cerny/openscap
-- build directory: /home/cipherboy/GitHub/jan-cerny/openscap/build
--  
-- Core features:
-- SCE: OFF
--  
-- OVAL:
-- base probe support: ON
-- SEAP msgid bit-size: 32
-- independent probes: ON
-- unix probes: ON
-- linux probes: ON
-- solaris probes: OFF
--  
-- Language bindings:
-- python3 bindings: FALSE
-- perl bindings: FALSE
--  
-- Utilities:
-- oscap: ON
-- oscap-docker: ON
-- scap-as-rpm: ON
-- oscap-ssh: ON
-- oscap-vm: ON
-- oscap-chroot: ON
--  
-- Configuring done
-- Generating done
-- Build files have been written to: /home/cipherboy/GitHub/jan-cerny/openscap/build
make: *** No targets specified and no makefile found.  Stop.
make: Nothing to be done for 'docs'.
make: *** No rule to make target 'install'.  Stop.
nm: '/tmp/tmp4udeo22_/lib64/libopenscap.so': No such file
  File "../tests/check_public_api.py", line 130, in <module>
    main()
  File "../tests/check_public_api.py", line 110, in main
    header_symbols, so_symbols = analyze_project_artifacts(src_dir)
  File "../tests/check_public_api.py", line 102, in analyze_project_artifacts
    so_symbols = get_public_symbols_from_shared_object(prefix)
  File "../tests/check_public_api.py", line 18, in get_public_symbols_from_shared_object
    nm_output_bytes = subprocess.check_output(nm_command)
  File "/usr/lib64/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib64/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['nm', '--demangle', '--dynamic', '--defined-only', '/tmp/tmp4udeo22_/lib64/libopenscap.so']' returned non-zero exit status 1.
```

In addition, you might want to check the return codes of the `make` commands so they don't all get called if they are failing.

cipherboy:
@jan-cerny I'm also confused about the output of this script. See this:

When I build your branch (admittedly after rebasing on top of dbae6b2294b85e5d78bf571e91cea3a55ec2f8e4), I get the following output (truncated):

```
The following 1491 symbols are exported in binary, but are not present in public header files:
CVRF_BRANCH_TYPE_MAP

~snip~

SEXP_list_pop
SEXP_list_push
```

(They are listed as not having OSCAP_API prefix in the header files).

However:

https://github.com/OpenSCAP/openscap/blob/dbae6b2294b85e5d78bf571e91cea3a55ec2f8e4/src/OVAL/probes/SEAP/public/sexp-manip.h#L407

and 

https://github.com/OpenSCAP/openscap/blob/dbae6b2294b85e5d78bf571e91cea3a55ec2f8e4/src/OVAL/probes/SEAP/public/sexp-manip.h#L400

Both show them as having `OSCAP_API`. 

Further:

```python
Python 3.6.6 (default, Jul 19 2018, 14:25:17) 
[GCC 8.1.1 20180712 (Red Hat 8.1.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import check_public_api
>>> 'SEXP_list_pop' in check_public_api.get_public_symbols_from_header("../src/OVAL/probes/SEAP/public/sexp-manip.h")
True
>>> 
```


So the error seems to be in `get_public_headers(...)`: 

```
public_headers = {'/tmp/tmpr38yr6m5/include/openscap/check_engine_plugin.h',
 '/tmp/tmpr38yr6m5/include/openscap/cpe_dict.h',
 '/tmp/tmpr38yr6m5/include/openscap/cpe_lang.h',
 '/tmp/tmpr38yr6m5/include/openscap/cpe_name.h',
 '/tmp/tmpr38yr6m5/include/openscap/cve_nvd.h',
 '/tmp/tmpr38yr6m5/include/openscap/cvrf.h',
 '/tmp/tmpr38yr6m5/include/openscap/cvss_score.h',
 '/tmp/tmpr38yr6m5/include/openscap/ds_rds_session.h',
 '/tmp/tmpr38yr6m5/include/openscap/ds_sds_session.h',
 '/tmp/tmpr38yr6m5/include/openscap/oscap.h',
 '/tmp/tmpr38yr6m5/include/openscap/oscap_debug.h',
 '/tmp/tmpr38yr6m5/include/openscap/oscap_download_cb.h',
 '/tmp/tmpr38yr6m5/include/openscap/oscap_error.h',
 '/tmp/tmpr38yr6m5/include/openscap/oscap_export.h',
 '/tmp/tmpr38yr6m5/include/openscap/oscap_reference.h',
 '/tmp/tmpr38yr6m5/include/openscap/oscap_source.h',
 '/tmp/tmpr38yr6m5/include/openscap/oscap_text.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_adt.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_agent_api.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_agent_xccdf_api.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_definitions.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_directives.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_probe.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_probe_session.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_results.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_schema_version.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_session.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_system_characteristics.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_types.h',
 '/tmp/tmpr38yr6m5/include/openscap/oval_variables.h',
 '/tmp/tmpr38yr6m5/include/openscap/scap_ds.h',
 '/tmp/tmpr38yr6m5/include/openscap/xccdf_benchmark.h',
 '/tmp/tmpr38yr6m5/include/openscap/xccdf_policy.h',
 '/tmp/tmpr38yr6m5/include/openscap/xccdf_session.h'}
```

But do tell me if I'm doing something wrong. :)

jan-cerny:
It looks like the problem is that "SEXP" headers are "public" but they are not installed by CMake.
But the SEXP functions will be moved to private by #1150 .

cipherboy:
@jan-cerny Sure, but I think this is bigger than just the SEXP functions. 

https://github.com/OpenSCAP/openscap/blob/dbae6b2294b85e5d78bf571e91cea3a55ec2f8e4/src/OVAL/probes/public/probe-api.h#L431
^ is also flagged, and it is part of `public/*.h`. 

I think we also need to consider if we want to support things in more internal headers as well:

https://github.com/OpenSCAP/openscap/blob/dbae6b2294b85e5d78bf571e91cea3a55ec2f8e4/src/OVAL/probes/probe/option.h#L22

(The XCCDF-related functions I've not done a thorough review of, but most of them seem to be in `unused.h` (which are, I think, unimplemented) and/or otherwise private).

jan-cerny:
The public/probe-api.h is another example of a public file that is not installed, but it is also moved away by #1150.

The second example is also a wrong use of OSCAP_API, and the macro OSCAP_API is moved away in #1150 as well.

jan-cerny:
I'll fix the PEP8 problems.

tests/check_public_api.py
```diff
@@ -0,0 +1,131 @@
...
+        cmake_command = ["cmake",
+                         "-DENABLE_PYTHON3=FALSE",
+                         "-DENABLE_PERL=FALSE",
+                         "-DCMAKE_INSTALL_PREFIX=" + prefix,
+                         "-DENABLE_TESTS=FALSE",
+                         ".."]
+        subprocess.run(cmake_command)
```
cipherboy: One last feature request similar to my other comment. Can we get this output not-outputted by default? I'd probably redirect stdout+stderr to the same variable, then only output it if the command failed (and then exit this script too with the same return code).

That way you can pipe the output of this script to a file and easily diff the output between different runs without having to edit the output of this script to remove build output.

jan-cerny: that would be great

tests/check_public_api.py
```diff
@@ -0,0 +1,154 @@
...
+    return symbols
+
+
+def _run_command(command):
+    try:
+        subprocess.run(command, check=True,
+                       stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+                       encoding="utf-8")
+    except subprocess.CalledProcessError as e:
+        print(e.output)
```
cipherboy: I'd prefer if this merged `stdout` and `stderr` so that they were interleaved as during build `stderr` gets written to and having context of `stdout` helps (by setting `stderr=subprocess.STDOUT`), but I won't hold up merging this on that.

cipherboy:
LGTM. Thanks @jan-cerny! :)

cipherboy:
@jan-cerny :

Heh. Looks like there's a bug in this script. See #1167 for details.

What happens is that variables aren't correctly processed by the regex, so they show up as `unknown` when processed from headers. Because _every_ unknown item takes the same form, they are condensed by the set. I'd at minimum suggest making the `unknown` represent the line (`p`) it was taken from so they don't get condensed.

PR #1163, opened 2018-08-14
Remove suspicious occurences of OSCAP_API in oscap_error.h

jan-cerny:
These occurences were inside a comment. Again, we see that
it was not a good idea to perform this change automatically.


cipherboy:
ACK.

cipherboy:
Thanks @jan-cerny! :)

Issue #936, opened 2017-12-13
oscap: unrecognized option '--rule'


ghost:
According to the openscap 1.2 documentation here : https://static.open-scap.org/openscap-1.2/oscap_user_manual.html
`
You can additionaly add --rule option to the above command to evaluate a specific rule:
$ oscap xccdf eval --profile Desktop --rule ensure_gpgcheck_globally_activated  --results xccdf-results.xml --cpe cpe-dictionary.xml scap-xccdf.xml
`
However, attempting to actually use the --rule flag fails:
oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_nist-800-171-cui --rule content_rule_aide_build_database --results results.xml --report report.html /usr/share/xml/scap/ssg/content/ssg-centos7-ds.xml
oscap: unrecognized option '--rule'
Usage:    oscap [options] xccdf eval [options] INPUT_FILE [oval-definitions-files]
Help:     oscap xccdf eval -h

yum list installed openscap-scanner
Installed Packages
openscap-scanner.x86_64  1.2.14-2.el7

The help does not show this option as being valid either.
oscap xccdf eval --help
oscap -> xccdf -> eval

Perform evaluation driven by XCCDF file and use OVAL as checking engine

Usage: oscap [options] xccdf eval [options] INPUT_FILE [oval-definitions-files]

INPUT_FILE - XCCDF file or a source data stream file

Options:
   --profile <name>     	 - The name of Profile to be evaluated.
   --tailoring-file <file>       - Use given XCCDF Tailoring file.
   --tailoring-id <component-id> - Use given DS component as XCCDF Tailoring file.
   --cpe <name> 		 - Use given CPE dictionary or language (autodetected)
                		   for applicability checks.
   --oval-results       	 - Save OVAL results as well.
   --sce-results		 - Save SCE results as well. (DEPRECATED! use --check-engine-results)
   --check-engine-results        - Save results from check engines loaded from plugins as well.
   --export-variables   	 - Export OVAL external variables provided by XCCDF.
   --results <file>     	 - Write XCCDF Results into file.
   --results-arf <file> 	 - Write ARF (result data stream) into file.
   --thin-results       	 - Thin Results provides only minimal amount of information in OVAL/ARF results.
                        	   The option --without-syschar is automatically enabled when you use Thin Results.
   --without-syschar    	 - Don't provide system characteristic in OVAL/ARF result files.
   --report <file>      	 - Write HTML report into file.
[...]

PR #1165, opened 2018-08-14
Fix broken compiler visibility presets

### LINKED ISSUE #963: C++ is not C
In https://github.com/OpenSCAP/openscap/blob/master/src/CMakeLists.txt#L22
we set  `CXX_VISIBILITY_PRESET hidden`, which is then used on all our objects.

`CXX` means C++, but our code is written in C. Why is it there? Should it be removed?

jan-cerny:
The main problem is that C is not C++, therefore the symbol visibility
setting for C++ is not applicable to C. This caused that many symbols
which must be private were actually exported in libopenscap.so and they
were publicly visible. I have discovered this implication of C++
not being C by using `nm` command and analysing `nm` output carefully.
Fixes #963.


CMakeLists.txt
```diff
@@ -441,9 +441,13 @@ include_directories(
...
 )
 
+# Honor visibility properties for all target types
+# Run "cmake --help-policy CMP0063" for policy details
+cmake_policy(SET CMP0063 NEW)
+
 function(set_oscap_generic_properties TARGET_OBJECT)
 	set_target_properties(${TARGET_OBJECT} PROPERTIES
-		CXX_VISIBILITY_PRESET hidden    # Make global variables and functions HIDDEN by default.
+		C_VISIBILITY_PRESET hidden    # Make global variables and functions HIDDEN by default.
```
cipherboy: Might as well add in those two spaces...

mpreisler: we want both IMO because dpkg probe is in C++

cipherboy:
@jan-cerny In addition to @mpreisler's comment, it looks like the Jenkins failure is legitimate:

```python
32/87 Test #33: bindings/test_python.sh .............................................***Failed    0.27 sec
+ python_import                                               
TEST: python_import
  File "<string>", line 1, in <module>
  File "/home/jenkins/workspace/openscap-master-pull-requests/build/swig/python3/openscap_api.py", line 49, in <module>
    OSCAP = _import_helper()
  File "/home/jenkins/workspace/openscap-master-pull-requests/build/swig/python3/openscap_api.py", line 45, in _import_helper
    _mod = imp.load_module('_openscap_py', fp, pathname, description)
  File "/usr/lib64/python3.6/imp.py", line 243, in load_module
    return load_dynamic(name, filename, file)
  File "/usr/lib64/python3.6/imp.py", line 343, in load_dynamic
    return _load(spec)
ImportError: /home/jenkins/workspace/openscap-master-pull-requests/build/swig/python3/_openscap_py.so: undefined symbol: OSCAP_LANG_DEFAULT
RESULT: FAILED
```

jan-cerny:
So that's pretty unpleasant. I'll try to resolve it.

CMakeLists.txt
jan-cerny: OK

cipherboy:
@jan-cerny It looks like it might actually be fairly easy:

```
src/common/text.c:44:const char * const OSCAP_LANG_DEFAULT    = "en-US";
```

I'm guessing it isn't exported / marked public :)

jan-cerny:
I have set both C++ and C visibility due to dpkg probe and I have exported remaining symbols from public API.

The second commit (438d977) fixes problem of Python symbols. These symbols are supposed to be a part of pubilc API, but they have been forgotten to be labeled by OSCAP_API macro. Due to coincidens with Issue #963 this problem was not visible. After Issue #936 has been fixed this issue poped by by breaking our upstream tests of SWIG Python 3 Bindings.

cipherboy:
Thanks @jan-cerny! :)

PR #1170, opened 2018-08-16
Made the rpm-related pkg-config configuration more robust.

### LINKED ISSUE #453: PKG_CHECK_MODULES macros ignore package version on RHEL7
On RHEL7.2 we have seen that autoconf macros like
`PKG_CHECK_MODULES([rpm], [rpm >= 4.18],[],[])`
gets evaluated to true everytime, regardless to the version of respective package installed.

Currently I am not able to say whether it is an issue in our build system or an issue in the toolset.

matejak:
The same target of the check (`rpm`) screwed up other checks.
Checks for different `librpm` versions have `rpm4`, `rpm5`, ... targets
that are then transferred to the `rpm_...` env variables.

Fixes #453

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -375,22 +375,26 @@ AC_SUBST(sigwaitinfo_LIBS)
 # libopenscap links against librpm if found. Otherwise we carry own implementation of rpmvercmp.
 echo
 echo '* Checking for rpm library  (optional dependency of libopenscap) '
-PKG_CHECK_MODULES([rpm], [rpm >= 4.4],[
-	SAVE_LIBS=$LIBS
+AC_SUBST([rpm_CFLAGS])
+AC_SUBST([rpm_LIBS])
+PKG_CHECK_MODULES([rpm4], [rpm >= 4.4],[
 	AC_DEFINE([HAVE_RPMVERCMP], [1], [Define to 1 if there is rpmvercmp available.])
-	AC_SUBST([rpm_CFLAGS])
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -375,22 +375,26 @@ AC_SUBST(sigwaitinfo_LIBS)
...
-PKG_CHECK_MODULES([rpm], [rpm >= 4.6],[
+PKG_CHECK_MODULES([rpm6], [rpm >= 4.6],[
 	AC_DEFINE([HAVE_RPM46], [1], [Define to 1 if rpm is newer than 4.6.])
+	rpm_CFLAGS="$rpm6_CFLAGS"
+	rpm_LIBS="$rpm6_LIBS"
 ],[
 	AC_MSG_NOTICE([librpm is older than 4.6])
 ])
-PKG_CHECK_MODULES([rpm], [rpm >= 4.7],[
+PKG_CHECK_MODULES([rpm7], [rpm >= 4.7],[
```
jan-cerny: It doesn't work for me at all on my RHEL6 virtual machine. If I change this line to >= 4.10 it still defines the respective variable in config.h, which shouldn't happen. I have rpm-4.8.0-59.el6.x86_64.

jan-cerny: I have forgotten to run `autogen.sh` before that. If I run it, your patch works.

jan-cerny:
Thank you!

PR #1171, opened 2018-08-16
Public headers

cipherboy:
This brings the headers and exported symbols further into alignment.

 - Removes several unimplemented functions. 

I believe the SEXP/SEAP and probes related issues will disappear with @jan-cerny's #1150.

This leaves:
```
oscap_basename
oscap_dirname
oscap_enum_to_string
oscap_realpath
oscap_split
oscap_sprintf
oscap_strcasecmp
oscap_strerror_r
oscap_string_to_enum
oscap_strncasecmp
oscap_strtok_r
```

These are all defined in `src/common/util.h` -- I am fine moving these to `src/common/public/openscap.h`, however, several of these have comments along the lines of `FIXME: make private` (e.g., `oscap_sprintf`) or a comment about their usage in probes.

Thoughts?

### DIFF
--- a/src/OVAL/probes/public/probe-api.h
+++ b/src/OVAL/probes/public/probe-api.h
@@ -450,8 +450,6 @@ typedef struct probe_ctx probe_ctx;
 
 OSCAP_API bool probe_item_filtered(const SEXP_t *item, const SEXP_t *filters);
 
-OSCAP_API int probe_result_additem(SEXP_t *result, SEXP_t *item);
-
 /**
  * Collect generated item (i.e. add it to the collected object)
  * The function takes ownership of the item reference and takes


jan-cerny:
I think you shouldn't install  "probes/public/*.h" and "probes/SEAP/public/*.h" because these directories are going to be removed and the symbols will get to private headers.

cipherboy:
Ah I missed that the probes changes were part of #1150 -- I thought it was the SEXP stuff only. 

I'll rescope this so it only includes removing functions you haven't already removed as part of that.

jan-cerny:
> These are all defined in src/common/util.h -- I am fine moving these to src/common/public/openscap.h, however, several of these have comments along the lines of FIXME: make private (e.g., oscap_sprintf) or a comment about their usage in probes.
> Thoughts?

If they are used in probes, they can be made private. Previously the probes used to be separated binaries, so all the functions used in probes had to be made public. However, this is no longer true, so these comments are not relevant anymore and they should be removed.

On the other hand, if some of the functions is used in `oscap` utility, they need to be kept private. Grep the `utils` directory and if you find it there, then probably they have to be public.

I really like the idea of moving the public functions out of `util.h`. I agree we should keep only the private symbols in `util.h`. Maybe we could create something like `src/common/public/openscap_helpers.h`.

cipherboy:
@jan-cerny Okay! I'll look into doing those once #1150 merges. :)

cipherboy:
I'm guessing #1150 won't get done before I leave / in time for me to rebase this PR, so I have made sure that this PR has maintainer-edits enabled.

cipherboy:
(Otherwise, feel free to pick the changes and update as necessary).

cipherboy:
@jan-cerny Okay! Then I think the rest of this is independent of #1150 so I'm removing the WIP flag.

PR #1176, opened 2018-08-21
Move public symbols to a public header

jan-cerny:
Our header file src/common/util.h deifned both public and private
symbols. If they are used in probes, they can be made private.
Previously the probes used to be separated binaries, so all the
functions used in probes had to be made public. However, this is
no longer true, so these comments are not relevant anymore and they
should be removed. On the other hand, if some of the functions is
used in oscap utility, they need to be kept public. This commit
moves the symbols that need to be kept public to a new header file
src/common/public/oscap_helpers.h Moreover, the symbols that doesn't have
to be public won't be public anymore.

### DIFF
--- a/src/CPE/cpe_session.c
+++ b/src/CPE/cpe_session.c
@@ -34,6 +34,7 @@
 #include "OVAL/public/oval_agent_api.h"
 #include "source/public/oscap_source.h"
 #include "source/oscap_source_priv.h"
+#include "oscap_helpers.h"
 
 static inline bool cpe_session_add_default_cpe(struct cpe_session *session)
 {
--- a/src/CPE/cpename.c
+++ b/src/CPE/cpename.c
@@ -42,6 +42,7 @@
 
 #include "cpe_name.h"
 #include "common/util.h"
+#include "oscap_helpers.h"
 
 #define CPE_URI_SUPPORTED "2.3"
 
--- a/src/CVRF/cvrf_priv.c
+++ b/src/CVRF/cvrf_priv.c
@@ -49,6 +49,7 @@
 
 #include "source/oscap_source_priv.h"
 #include "source/public/oscap_source.h"
+#include "oscap_helpers.h"
 
 
 /***************************************************************************
--- a/src/CVSS/cvss.c
+++ b/src/CVSS/cvss.c
@@ -41,6 +41,7 @@
 #include "public/cvss_score.h"
 #include "cvss_priv.h"
 #include "common/elements.h"
+#include "oscap_helpers.h"
 
 #define CVSS_SUPPORTED "2.0"
 #define NS_VULN_STR BAD_CAST "vuln"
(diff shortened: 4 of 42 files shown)


cipherboy:
I agree that these changes do what is advertised on the label:
```diff
1,2c1,2
< Shared object symbols: 2881
< Public header symbols: 2630
---
> Shared object symbols: 2875
> Public header symbols: 2635
5c5
< The following 251 symbols are exported in binary, but are not present in public header files:
---
> The following 240 symbols are exported in binary, but are not present in public header files:
159,169d158
< oscap_basename
< oscap_dirname
< oscap_enum_to_string
< oscap_realpath
< oscap_split
< oscap_sprintf
< oscap_strcasecmp
< oscap_strerror_r
< oscap_string_to_enum
< oscap_strncasecmp
< oscap_strtok_r
```

However, it introduces several new warnings:

```diff
287a288,293
> /home/cipherboy/GitHub/jan-cerny/openscap/src/OVAL/oval_session.c: In function ‘oval_session_setup_agent’:
> /home/cipherboy/GitHub/jan-cerny/openscap/src/OVAL/oval_session.c:323:20: warning: implicit declaration of function ‘oscap_basename’; did you mean ‘oscap_strcmp’? [-Wimplicit-function-declaration]
>   char *base_name = oscap_basename(path_clone);
>                     ^~~~~~~~~~~~~~
>                     oscap_strcmp

368a375,382
> /home/cipherboy/GitHub/jan-cerny/openscap/src/OVAL/probes/oval_fts.c: In function ‘oval_fts_read_recurse_path’:
> /home/cipherboy/GitHub/jan-cerny/openscap/src/OVAL/probes/oval_fts.c:1301:37: warning: implicit declaration of function ‘oscap_dirname’; did you mean ‘oscap_ptr_cmp’? [-Wimplicit-function-declaration]
>     ofts->ofts_recurse_path_curpth = oscap_dirname(ofts->ofts_recurse_path_curpth);
>                                      ^~~~~~~~~~~~~
>                                      oscap_ptr_cmp

1002a1017,1024
> /home/cipherboy/GitHub/jan-cerny/openscap/src/OVAL/probes/oval_fts.c: In function ‘oval_fts_read_recurse_path’:
> /home/cipherboy/GitHub/jan-cerny/openscap/src/OVAL/probes/oval_fts.c:1301:37: warning: implicit declaration of function ‘oscap_dirname’; did you mean ‘oscap_ptr_cmp’? [-Wimplicit-function-declaration]
>     ofts->ofts_recurse_path_curpth = oscap_dirname(ofts->ofts_recurse_path_curpth);
>                                      ^~~~~~~~~~~~~
>                                      oscap_ptr_cmp
```

I think it'd be good not to introduce any new warnings. :)

jan-cerny:
I have added 2 missing includes.

cipherboy:
This looks good to me. Thanks @jan-cerny for the updates! :)

Can confirm this fixes the warnings.

PR #1177, opened 2018-08-21
Allow oscap-ssh to use custom SSH options

matejak:
This PR allows the caller to pass SSH connection options.
The use case for this is the SSG test suite - as containers or VMs have the same IP address but different SSH server certificates, one would be bothered by the `known_hosts` records that would cause false alarms.
This PR consists of some refactoring and backwards-compatible usage of the `SSH_ADDITIONAL_OPTIONS` environment variables for a list of SSH options.

### DIFF
--- a/utils/oscap-ssh
+++ b/utils/oscap-ssh
@@ -92,9 +92,14 @@ OSCAP_SUDO=""
 # SSH_ADDITIONAL_OPTIONS may be defined in the calling shell
 SSH_TTY_ALLOCATION_OPTION=""
 
-# $1: The SSH command.
-# $2: More of additional options (optional, space-separated string)
+# $1, $2, ... SSH options (pass them as separate arguments)
 function ssh_execute_with_options {
+    ssh -o ControlPath="$MASTER_SOCKET" $SSH_ADDITIONAL_OPTIONS "$@" -p "$SSH_PORT" "$SSH_HOST"
+}
+
+# $1: The SSH command.
+# $2: More of additional options (optional, pass one space-separated string)
+function ssh_execute_with_command_and_options {
     ssh -o ControlPath="$MASTER_SOCKET" $SSH_ADDITIONAL_OPTIONS $2 -p "$SSH_PORT" "$SSH_HOST" "$1"
 }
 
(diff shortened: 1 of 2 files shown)


utils/oscap-ssh
```diff
@@ -80,61 +80,94 @@ function usage()
...
 shift 2
 
+check_oscap_arguments "$@"
+
 MASTER_SOCKET_DIR=$(mktemp -d)
 MASTER_SOCKET="$MASTER_SOCKET_DIR/ssh_socket"
 
 echo "Connecting to '$SSH_HOST' on port '$SSH_PORT'..."
-ssh -M -f -N -o ServerAliveInterval=60 -o ControlPath="$MASTER_SOCKET" -p "$SSH_PORT" "$SSH_HOST" || die "Failed to connect!"
+ssh -M -f -N -o ServerAliveInterval=60 -o ControlPath="$MASTER_SOCKET" -p "$SSH_PORT" $SSH_ADDITIONAL_OPTIONS "$SSH_HOST" || die "Failed to connect!"
```
cipherboy: Could we perhaps add a wrapper for this command as well? I realize that it isn't reused and it adds more lines than necessary, but I think it'd be nice to have.

utils/oscap-ssh
```diff
@@ -80,61 +80,94 @@ function usage()
...
+    elif [ "$1" == "info" ]; then
+        true
+    elif [ "$1 $2" == "xccdf eval" ]; then
+        true
+    elif [ "$1 $2" == "oval eval" ]; then
+        true
+    elif [ "$1 $2" == "oval collect" ]; then
+        true
+    else
+        die "This script only supports '-h', '--help', '--v', '--version', 'info', 'xccdf eval', 'oval eval' and 'oval collect'."
```
cipherboy: I realize that the old version is wrong, but doesn't this technically support `--sudo` as well?

matejak:
Thanks for review @cipherboy , I have updated the PR.

matejak:
@matusmarhefka Good point, I have rebased and inserted the man page section in.

matusmarhefka:
@matejak Great stuff, thank you, I am merging this.

PR #1178, opened 2018-08-22
Resolve testing of functions that are not a part of public API

jan-cerny:
This PR consists of 5 changes that block #1150 where we had to face problems with linking executable tests written in C.

1. Define symbols that are exported in DSO using an export map
The export map will be generated automatically from the header
files by a Python scripts. This way prevents us from having
to maintain an explicit list. All the decalrations labeled by
OSCAP_API macros in *public* header files will be exported.
All the remaining symbols will be hidden.
For more details, please see especially Section 2.2.5.
https://www.akkadia.org/drepper/dsohowto.pdf

2. Refactor: Save the variables into a list
In order to be able to reuse the list of relevant libraries to link
multiple libraries in future, we can save all the libraries into
a list.

3. Build a testing library that does not hide any symbol
This library will export completely everything. That can be
used in future to link tests against. This commit reuses the
previous refactoring of library list.

4. Link tests written in C to libopenscap_testing.so
This enables us to use all the functions in our tests, including
functions that are not in public API. Therefore, unit testing
and quality revamping activities will be easier. Also we simplify
some of CMakeLists.txt. In future we are going to move SEXP functions
out of public API. This change will allow us to still run the tests
that use SEXP functions without having to create serious lists of
files.

5. Do not define OSCAP_API
We have started to use an export map to define public API.
Therefore, we don't have to use the GCC modifiers. We can
make the OSCAP_API macro to an empty string in Linux.

6. Enable Clang to use export map as well

This PR is blocked by #1176 . The reason of being blocked by #1176 is that this PR has a side effect that only symbols in public header files will be considered public. Symbols in private header files will be always considered private even if they would be labeled by OSCAP_API macro.


cipherboy:
@jan-cerny I just pushed #1176 to `master`; do you need to rebase this?

jan-cerny:
Yes, I'll rebase that.

jan-cerny:
I have rebased it on the top of master which means it is unblocked.

cipherboy:
@jan-cerny It looks like Apple doesn't ship a python3 executable in the location you expect it:

```
/bin/sh: ../../src/generate_export_map.py: /usr/bin/python3: bad interpreter: No such file or directory
make[2]: *** [src/libopenscap.map] Error 126
make[1]: *** [src/CMakeFiles/export_map.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....```

**Comment by jan-cerny** [comment]
I think we should use Python in CMakeLists.txt .

**Comment by jan-cerny** [comment]
It looks like it doesn't know the --version-script option.
```
ld: unknown option: --version-script=/Users/travis/build/OpenSCAP/openscap/build/src/libopenscap.map
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [src/libopenscap.24.0.0.dylib] Error 1
make[1]: *** [src/CMakeFiles/openscap.dir/all] Error 2
```

**Comment by cipherboy** [comment]
Yeah, I think this is expected. macOS doesn't seem to support version scripts.

**Comment by jan-cerny** [comment]
@cipherboy  This is really nice, Thank you very much. I have tried to do it. However, it looks to me we need to resolve the public API mess before. Travis CI reported a bunch of unresolved symbols then. But they are already fixed in #1150 . Therefore I have created a patch with the OSX flag and opened it as a separate PR, which will be blocked by #1150.

File: src/CMakeLists.txt [hunk 1]
```diff
@@ -69,55 +69,73 @@ if (HAVE_MMAN_H AND (GCRYPT_FOUND OR NSS_FOUND))
...
+		add_custom_command(
+			OUTPUT "${VERSION_SCRIPT}"
+			COMMAND "${PYTHON_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}/generate_export_map.py" --linux --output "${VERSION_SCRIPT}" "${CMAKE_CURRENT_SOURCE_DIR}"
+			COMMENT "Generating export map"
+		)
+		set_target_properties(openscap PROPERTIES
+			LINK_FLAGS "-Wl,--version-script=\"${VERSION_SCRIPT}\""
+		)
+	endif()
+	add_custom_target(export_map DEPENDS "${VERSION_SCRIPT}")
```
jan-cerny: I think that this should be moved into the if now and moved back after it is merged in a separate PR.

jan-cerny:
@mpreisler What do you suggest instead of this?

mpreisler:
We discussed this several time in meetings, a quick summary of my opinion of this:
1) This is a very heavy solution to a pretty small and limited problem we have
2) It's odd and very uncommon, will be difficult for newcomers to pick up
3) Adds python as a required dependency to openscap build, this is problematic on Windows where python is not installed by default

An alternative is to add the C files we are testing to the tests, this is what we were doing before.

jan-cerny:
That was suggested by the team.

I see another alternative to close this PR, keep the current behavior regarding the symbols, then refactor the SEXP module to be able to link the test. @mpreisler Will that be OK?

mpreisler:
> I see another alternative to close this PR, keep the current behavior regarding the symbols, then refactor > the SEXP module to be able to link the test. @mpreisler Will that be OK?

yeah, I think that's reasonable

jan-cerny:
OK, thank you. I think it will be better to do it in a separate PR.

PR #1182, opened 2018-08-27
 Use dynamic symbols on OS X and Windows

jan-cerny:
OS X uses exported symbols list instead.
Windows uses "definition files" instead.
This is blocked by #1150

For more information about Windows and their definition files, please see https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-def-files


jan-cerny:
It looks like we need to move compat.h to some public directory. The Windows build is going to be broken by this change, because there will be unresolved symbols `_optind` ,`_opterr`, `_optarg`, etc. I think that should be solved in a separate PR, because it is a huge change.

jan-cerny:
Blocked by #1184 and #1150

jan-cerny:
This will need to be rebased on the top of master's head once #1150 is merged.

jan-cerny:
Closing, because this strongly depends on #1178 but #1178 will be totally reworked.

PR #1183, opened 2018-08-27
Add some missing free calls to prevent possible memory leaks

mildass:
Fix some defects, which could lead to memory leaks.

There are some variables which have allocated memory but they aren't free before 'return' from function.


cipherboy:
@openscap-ci add to whitelist.

jan-cerny:
Thanks!

PR #1184, opened 2018-08-27
Move fallback implementation of getopt to utils

jan-cerny:
The utils module is the consumer of the getopt API. Instead of
including the getopt symbols in the shared library, we can
build them into the oscap binary. This will usually happen
on Windows and similar platforms hthat do not provide getopt.h.

This blocks #1182


utils/getopt.c
```diff
@@ -29,7 +29,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#include "compat.h"
+#include "oscap-tool.h"
```
jan-cerny: I think that we should put `#ifndef HAVE_GETOPT_H` to make it more reliable.

jan-cerny: Done.

PR #1189, opened 2018-08-29
Add Windows CPEs

### LINKED ISSUE #1181: oscap reports all items as not applicable (Win7, MS Win7SP1 xccf from SCM 4.0)
#### Description of Problem:
Exporting a SCAP file for Windows 7 SP1 from Microsoft SCM 3.0, running scap with it returns a report saying every single item is "not applicable".  This is not useful, and probably not intended.

#### OpenSCAP Version:
1.3.0_alpha2 downloaded from
https://github.com/OpenSCAP/openscap/releases/tag/1.3.0_alpha2 , specifically the .msi installer at https://github.com/OpenSCAP/openscap/releases/download/1.3.0_alpha2/OpenSCAP-1.3.0-win32.msi

Windows 7 Professional Service Pack 1
Microsoft Windows Version 6.1 (Build 7601: Service Pack 1)

#### Steps to Reproduce:

1. Install openscap 1.3.0_alpha2
2. Download Microsoft Security Compliance Manager 4.0 from https://www.microsoft.com/en-us/download/confirmation.aspx?id=53353
3. Run the downloaded `Security_Compliance_Manager_Setup.exe` file to install it.  Use all defaults.
4. Run Security Compliance Manager, let it update and install all the included Baselines
5. Click on the + sign next to "Windows 7 SP1" to expand it in the left panel of SCM
6. In the right panel of SCM, under Export, click "SCAP v1.0 (.cab) and save the cab file using its default name of "Win7SP1 Computer Security Compliance_SCAP.CAB". [ You will get a message about some records being dropped from the export and logged, which can be ignored.]
7. Extract the files from that .cab file, uzing 7zip or any other .cab archive extractor.  This should result in the creation of files named "Win7SP1 Computer Security Compliance-oval.xml" and "Win7SP1 Computer Security Compliance-xccdf.xml" (among others).
8. Run
```
    oscap.exe info "Win7SP1 Computer Security Compliance-xccdf.xml"
```
which should reveal a profile id of Win7SP1_Computer_Security_Compliance
9. Run a scan with oscap.exe as follows (command is all one line) in an elevated Command prompt window, from the directory containing the extracted .xml file:
```
    oscap xccdf eval --profile Win7SP1_Computer_Security_Compliance --results-arf
 win7-comp-arf.xml --report win7-comp-report.html "Win7SP1 Computer Security Compliance-xccdf.xml"
```
10. Open the results file win7-comp-report.html in a web browser.

#### Actual Results:
The Score is reported as 0.0, and every individual result is shows as "notapplicable".

#### Expected Results:
A non-zero score and some results passing, others failing, others showing as "unknown" since oscap.exe is still alpha on Windows.

Tried using the SCAP file from SCM 3.0, same result, files seem identical.
Tried using a different SCAP file, and got more expected results (mix of fail, unknown, and pass).
I do not know how to check whether the issue is with the Microsoft SCAP data, or the oscap.exe port to Windows.  
Are there free/open SCAP or XCCF validation tools that could help with this?  
Are there debug options to oscap that would provide useful further insight?

jan-cerny:
This commit adds CPEs for Windows systems (Windows 7, 8, 8.1, 10,
Server 2008, Server 2012) to OpenSCAP CPE dictionary.

The PR doesn't fix #1181 though, because according to NVD the CPE for Windows 7 is `cpe:/o:microsoft:windows_7`, not `cpe:/o:microsoft:windows7`

### DIFF
--- a/src/XCCDF/item.c
+++ b/src/XCCDF/item.c
@@ -41,6 +41,7 @@
 #include "helpers.h"
 #include "xccdf_impl.h"
 #include "common/util.h"
+#include "oscap_helpers.h"
 
 /* According to `man 3 pcreapi`, the number passed in ovecsize should always
  * be a multiple of three.
@@ -767,23 +768,21 @@ void xccdf_item_add_applicable_platform(struct xccdf_item *item, xmlTextReaderPt
 	int rc = pcre_exec(regex, NULL, platform_idref, strlen(platform_idref), 0, 0, ovector, OVECTOR_LEN);
 	/* 1 pattern + 2 groups = 3 */
 	if (rc == 3) {
-		size_t match_len = ovector[1] - ovector[0];
-		/* match_len + 1 underscore + 1 zero byte */
-		char *alternate_platform_idref = malloc(match_len + 1 + 1);
 		const int first_group_start = ovector[2];
 		const int first_group_end = ovector[3];
 		size_t first_group_len = first_group_end - first_group_start;
+		char *first_group = malloc(first_group_len + 1); // + 1 for '\0'
+		strncpy(first_group, platform_idref + first_group_start, first_group_len);
+		first_group[first_group_len] = '\0';
 		const int second_group_start = ovector[4];
 		const int second_group_end = ovector[5];
 		size_t second_group_len = second_group_end - second_group_start;
-		char *aptr = alternate_platform_idref;
-		strncpy(aptr, platform_idref + first_group_start, first_group_len);
-		aptr += first_group_len;
-		*aptr = '_';
-		aptr++;
-		strncpy(aptr, platform_idref + second_group_start, second_group_len);
-		aptr += second_group_len;
-		*aptr = '\0';
+		char *second_group = malloc(second_group_len + 1); // + 1 for '\0'
+		strncpy(second_group, platform_idref + second_group_start, second_group_len);
+		second_group[second_group_len] = '\0';
+		char *alternate_platform_idref = oscap_sprintf("%s_%s", first_group, second_group);
+		free(first_group);
+		free(second_group);
 		oscap_list_add(item->item.platforms, alternate_platform_idref);
 	}
 
--- a/src/XCCDF/item.c
+++ b/src/XCCDF/item.c
@@ -770,11 +770,11 @@ void xccdf_item_add_applicable_platform(struct xccdf_item *item, xmlTextReaderPt
 		size_t match_len = ovector[1] - ovector[0];
 		/* match_len + 1 underscore + 1 zero byte */
 		char *alternate_platform_idref = malloc(match_len + 1 + 1);
-		int first_group_start = ovector[2];
-		int first_group_end = ovector[3];
+		const int first_group_start = ovector[2];
+		const int first_group_end = ovector[3];
 		size_t first_group_len = first_group_end - first_group_start;
-		int second_group_start = ovector[4];
-		int second_group_end = ovector[5];
+		const int second_group_start = ovector[4];
+		const int second_group_end = ovector[5];
 		size_t second_group_len = second_group_end - second_group_start;
 		char *aptr = alternate_platform_idref;
 		strncpy(aptr, platform_idref + first_group_start, first_group_len);
--- a/src/XCCDF/item.c
+++ b/src/XCCDF/item.c
@@ -838,7 +838,7 @@ bool xccdf_item_process_element(struct xccdf_item * item, xmlTextReaderPtr reade
         oscap_list_add(item->item.rationale, oscap_text_new_parse(XCCDF_TEXT_HTMLSUB, reader));
 		return true;
         case XCCDFE_PLATFORM:
-        oscap_list_add(item->item.platforms, xccdf_attribute_copy(reader, XCCDFA_IDREF));
+		xccdf_item_add_applicable_platform(item, reader);
                 return true;
 	case XCCDFE_QUESTION:
         oscap_list_add(item->item.question, oscap_text_new_parse(XCCDF_TEXT_PLAIN, reader));
--- a/src/XCCDF/item.c
+++ b/src/XCCDF/item.c
@@ -30,6 +30,7 @@
 #include <string.h>
 #include <time.h>
 #include <math.h>
+#include <pcre.h>
 
 #include <libxml/tree.h>
 #include <libxml/xpath.h>
@@ -40,6 +41,11 @@
 #include "helpers.h"
 #include "xccdf_impl.h"
 #include "common/util.h"
+
+/* According to `man 3 pcreapi`, the number passed in ovecsize should always
+ * be a multiple of three.
+ */
+#define OVECTOR_LEN 30
 
 const struct oscap_string_map XCCDF_OPERATOR_MAP[] = {
 	{XCCDF_OPERATOR_EQUALS, "equals"},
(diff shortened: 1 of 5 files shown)


redhatrises:
Do we need to handle things like service packs and R2 e.g. `cpe:/o:microsoft:windows_7:-:sp2:*:*:*:*:*:* `, etc?

I really, really wish that we could just include the official NIST CPE dictionary list so that we can eliminate some of this type of work.

jan-cerny:
But the official NIST CPE dictionary does not contain OVAL, so we would need to develop OVALs anyway, is that correct?

mpreisler:
Yeah, including the whole CPE dict would be great but it's just not practical because of performance reasons. Plus it doesn't have OVALs as @jan-cerny said.

jan-cerny:
On the other hand, @redhatrises is true that there are a larger amount of CPEs on Windows. To what extent should we include them? I started with just the most basic CPEs for only major Windows versions in this PR.

redhatrises:
> On the other hand, @redhatrises is true that there are a larger amount of CPEs on Windows. To what extent should we include them? I started with just the most basic CPEs for only major Windows versions in this PR.

Great question. IMO should be all. We could cut the problem in half where we use the NIST CPE dictionary, then if you want oscap to scan an OS, require a PR to add the OVAL.

jmarsden:
Is there a reason Windows Server 2016 was omitted?

Regarding NIST using `cpe:/o:microsoft:windows_7, and Microsoft apparently using `cpe:/o:microsoft:windows7` : is creating a script to fix up the MS-generated XCCDF files a useful way to handle that without burdening the codebase with knowing about both names?

Are there other commonly used sources of SCAP / XCCDF files used in the Windows community, and which CPE naming convention do *they* use?

redhatrises:
@jmarsden more than 2016 needs to be selected.

The correct CPE definitions are defined in the NIST database. Anything else is considered invalid and a bug. In other words if MS didn't use the NIST CPE, it is invalid and a bug.

mpreisler:
I don't think it's fair to ask @jan-cerny of more CPEs than he has implemented here. This PR definitely makes the situation better. If you want more CPEs for more flavors of Windows please open separate PRs.

Not sure what to do about inconsistent CPE IDs. In general we use NIST CPE IDs if they exist and we reject any other ID in the openscap CPE dict.

jan-cerny:
@jmarsden The official source of CPE naming is NIST National Vulnerability Database, https://nvd.nist.gov/products/cpe , where it is possible to download an official CPE dictionary.

jan-cerny:
@mpreisler @redhatrises 

When I  was reading this thread again, I have realized it would be useful to check which of the CPEs of Linux products from the NIST official CPE dicitionary we implement in OpenSCAP "inbuilt" CPE dictionary.

And I have found something that can be seen as inconsistency:

In OpenSCAP CPE dict we have: ```cpe:/o:redhat:enterprise_linux:7```. But   in official NIST CPE dict,
there is no "cpe:/o:redhat:enterprise_linux:7". There are only CPEs for minor versions.  In the NIST official CPE dictionary there is:

```
grep "cpe:/o:redhat:enterprise_linux:7" official-cpe-dictionary_v2.2.xml
  <cpe-item name="cpe:/o:redhat:enterprise_linux:7.0">
  <cpe-item name="cpe:/o:redhat:enterprise_linux:7.1">
  <cpe-item name="cpe:/o:redhat:enterprise_linux:7.2">
  <cpe-item name="cpe:/o:redhat:enterprise_linux:7.3">
  <cpe-item name="cpe:/o:redhat:enterprise_linux:7.4">
  <cpe-item name="cpe:/o:redhat:enterprise_linux:7.5">
```

mpreisler:
@jan-cerny I'd say that is for convenience and since it's already there we will keep it. Especially since we are using it for content development.

redhatrises:
Slightly concerned that we are going to have to handle all windows cpes here.

src/XCCDF/item.c
```diff
@@ -742,6 +748,48 @@ bool xccdf_item_process_attributes(struct xccdf_item *item, xmlTextReaderPtr rea
...
+		aptr += first_group_len;
+		*aptr = '_';
+		aptr++;
+		strncpy(aptr, platform_idref + second_group_start, second_group_len);
+		aptr += second_group_len;
+		*aptr = '\0';
+		oscap_list_add(item->item.platforms, alternate_platform_idref);
+	}
+
+	oscap_list_add(item->item.platforms, platform_idref);
```
mpreisler: I am not 100% sure but I think we might want to add both the unaltered and altered. Just in case the user supplies a CPE dict.

src/XCCDF/item.c
```diff
@@ -742,6 +748,48 @@ bool xccdf_item_process_attributes(struct xccdf_item *item, xmlTextReaderPtr rea
...
+		size_t second_group_len = second_group_end - second_group_start;
+		char *aptr = alternate_platform_idref;
+		strncpy(aptr, platform_idref + first_group_start, first_group_len);
+		aptr += first_group_len;
+		*aptr = '_';
+		aptr++;
+		strncpy(aptr, platform_idref + second_group_start, second_group_len);
+		aptr += second_group_len;
+		*aptr = '\0';
+		oscap_list_add(item->item.platforms, alternate_platform_idref);
```
mpreisler: couple suggestions:
1) don't mix signed and unsigned, the compiler warns about this...
2) be consistent about types, you are using size_t for some lengths but int for others
3) use "const" to make it easier to read the code
4) the pointer artihmetic code can (probably) be replaced with something that's way easier to verify

src/XCCDF/item.c
jan-cerny: I think it adds both. The altered is added inside the if (on line 787) and unaltered is added always (on line 790).

jan-cerny:
@redhatrises If you want to handle these CPEs feel free to open a pull request. I don't plan to add every CPE for minor versions.

redhatrises:
> @redhatrises If you want to handle these CPEs feel free to open a pull request. I don't plan to add every CPE for minor versions.

Let me ask this another way.... Does it work for all minor versions as is in this PR? Kinda like how we work and deal with minor RHEL versions?

yuumasato:
> Slightly concerned that we are going to have to handle all windows cpes here.

@redhatrises Is it the case that there are some Windows minor versions that are more relevant from security standpoint? We could start with them.
For example, versions that notoriously fixed some vulnerability, or introduced security features that we would likely configure and use for hardening.

> I don't think it's fair to ask @jan-cerny of more CPEs than he has implemented here.

I agree with Martin here.

jan-cerny:
@redhatrises I think it should work the same way as for minor RHEL versions. It is basically a prefix match.

@yuumasato That makes sense for me, but let's not do it in this PR.

jan-cerny:
The CI fail is probably due to #1199

jan-cerny:
@openscap-ci test this please

mpreisler:
Let's merge this. To answer the various questions:
1) it's super confusing but cpe dict and cpe2 language model are different. they look similar but one is basically just an ID and one is a query language. openscap supports both so you can use cpe2 language model style CPE platforms
2) additional Windows versions can be added later

Kudos for the compatibility hack for the non-NIST Windows IDs!

ACK

mpreisler:
jenkins decided to rerun this after my comment so I will wait for it to finish

jan-cerny:
@jan-cerny test this please

PR #1193, opened 2018-09-17
Introduce a "virtual" "(all)" profile that will select all groups and…

mpreisler:
… all rules

This is useful for testing and debugging. It will use default values but
will select everything in the benchmark.

This is a WIP, there are still opportunities for refactoring and some minor code deduplication but I wanted your feedback as soon as possible. Thanks!

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy.c
+++ b/src/XCCDF_POLICY/xccdf_policy.c
@@ -2283,8 +2283,10 @@ void xccdf_policy_free(struct xccdf_policy * policy) {
 	/* A policy which is set to use default profile has its profile member set to NULL,
 	 * check it so we don't try to get the ID from a NULL profile.
 	 * */
-	if (policy->profile && xccdf_profile_get_id(policy->profile) == NULL)
-		/* If ID of policy's profile is NULL then this
+	if (policy->profile && (
+			(xccdf_profile_get_id(policy->profile) == NULL) ||
+			(strcmp(xccdf_profile_get_id(policy->profile), "(all)") == 0)))
+		/* If ID of policy's profile is NULL or "(all)" then this
 		 * profile is created by Policy layer and need
 		 * to be freed
 		 */
--- a/src/XCCDF_POLICY/xccdf_policy_model.c
+++ b/src/XCCDF_POLICY/xccdf_policy_model.c
@@ -31,6 +31,7 @@
 #include "xccdf_policy_model_priv.h"
 #include "xccdf_policy_priv.h"
 #include "XCCDF/item.h"
+#include "XCCDF/helpers.h"
 
 struct xccdf_policy *xccdf_policy_model_get_existing_policy_by_id(struct xccdf_policy_model *policy_model, const char *profile_id)
 {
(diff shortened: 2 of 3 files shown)


mpreisler:
I am not sure where to add it but the obvious trick for testing specific rules is to use `--profile "(all)" --rule "some_id"`.

redhatrises:
Seems a little weird to do "(all)". Any reason why "all" won't work? Or am I reading it wrong?

redhatrises:
> The main reason is to avoid squating on a name that is a legal XCCDF profile name. Plus it's similar to `(default)` which is a virtual profile name from the XCCDF spec.

Ah. Got it. Thanks for the explanation @mpreisler!

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -2283,8 +2283,10 @@ void xccdf_policy_free(struct xccdf_policy * policy) {
 	/* A policy which is set to use default profile has its profile member set to NULL,
 	 * check it so we don't try to get the ID from a NULL profile.
 	 * */
-	if (policy->profile && xccdf_profile_get_id(policy->profile) == NULL)
-		/* If ID of policy's profile is NULL then this
+	if (policy->profile && (
+			(xccdf_profile_get_id(policy->profile) == NULL) ||
+			(strcmp(xccdf_profile_get_id(policy->profile), "(all)") == 0)))
```
jan-cerny: I would put the id to a variable.

src/XCCDF_POLICY/xccdf_policy_model.c
```diff
@@ -46,6 +47,34 @@ struct xccdf_policy *xccdf_policy_model_get_existing_policy_by_id(struct xccdf_p
 	return NULL;
 }
 
+static void _add_selectors_for_all_xccdf_items(struct xccdf_profile *profile, struct xccdf_item *item)
+{
+	struct xccdf_item_iterator *children = NULL;
+	if (xccdf_item_get_type(item) == XCCDF_BENCHMARK) {
+		children = xccdf_benchmark_get_content(XBENCHMARK(item));
+	}
+	else if (xccdf_item_get_type(item) == XCCDF_GROUP) {
```
jan-cerny: The coding style https://www.kernel.org/doc/html/v4.10/process/coding-style.html prefers
```
if (x == y) {
        ..
} else if (x > y) {
        ...
} else {
        ....
}
```

src/XCCDF_POLICY/xccdf_policy_model.c
```diff
@@ -56,6 +85,9 @@ struct xccdf_policy *xccdf_policy_model_create_policy_by_id(struct xccdf_policy_
 		profile = xccdf_tailoring_get_profile_by_id(tailoring, id);
 	}
 
+	// The (default) and (all) profiles are de-facto owned by the xccdf_policy
+	// and will be freed by it when it's freed. See xccdf_policy_free.
```
jan-cerny: This is great!

src/XCCDF_POLICY/xccdf_policy.c
mpreisler: I considered that but notice the policy-profile additional conditional. To handle that the resulting code would be pretty much the same size.

src/XCCDF_POLICY/xccdf_policy_model.c
mpreisler: Fixed

mpreisler:
@openscap-ci test this please

matejak:
jenkins test this please

matejak:
test this please

matejak:
test this please

jan-cerny:
It works perfectly for me. Thank you.

PR #1200, opened 2018-10-01
Check if CMake policy CMP0063 is defined

### LINKED ISSUE #1188: CMake fails on unknown policy
#### Description of Problem:
CMake fails on unknown policy.
This happened on our Jenkins when building #1093 
https://jenkins.open-scap.org/job/openscap-master-pull-requests/267/console

#### OpenSCAP Version:
master

RHEL 7 on our Jenkins

#### Steps to Reproduce:

1.  `cd build`
2.  `cmake ..`
3. 

#### Actual Results:
```
CMake Error at CMakeLists.txt:467 (cmake_policy):
  Policy "CMP0063" is not known to this version of CMake.
```

#### Expected Results:
no fail

jan-cerny:
This policy was introduced in CMake 3.3.
Fixes #1188.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -464,7 +464,9 @@ include_directories(
 
 # Honor visibility properties for all target types
 # Run "cmake --help-policy CMP0063" for policy details
-cmake_policy(SET CMP0063 NEW)
+if (POLICY CMP0063)
+	cmake_policy(SET CMP0063 NEW)
+endif()
 
 function(set_oscap_generic_properties TARGET_OBJECT)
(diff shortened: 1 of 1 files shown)


mpreisler:
I am not sure how I feel about this, we change behavior if cmake supports it but if it doesn't we silently keep the old (broken? wrong?) behavior. We should either have some compat support for when cmake doesn't support the policy or we should output a warning.

jan-cerny:
I wil add a warning because I haven't found how to do that on old CMake.

jan-cerny:
@openscap-ci test it please

mpreisler:
ACK, the jenkins fail is unrelated.

PR #1206, opened 2018-10-02
Fix leak in textfilecontent_probe

yuumasato:
The leak was found after assume_d macro was replace by if code.

- Move the free for line_ent and ent_val before check for error condition,
they should be freed anyway.
- Add free for other leaking SEXP variables.


yuumasato:
@openscap-ci test this please

PR #1209, opened 2018-10-02
Add missing 'free' calls to src/OVAL/probes/probe/worker.c (master branch)

mildass:
Add missing `SEXP_free` calls.


src/OVAL/probes/probe/worker.c
```diff
@@ -396,6 +396,7 @@ static SEXP_t *probe_obj_eval(probe_t *probe, SEXP_t *id)
 
 	rid = SEXP_list_first(res);
 	if (SEXP_string_cmp(id, rid) != 0) {
+		SEXP_free(rid);
```
jan-cerny: Add SEXP_free(res);

mildass:
Updated. It should be ok now.

PR #1215, opened 2018-10-02
Plug memory leaks

jan-cerny:
The MODEP returned a SEXP_t structure which needs to be freed, but
it wasn't freed, which leads to memory leaks. Instead, we can just
pass a boolean value to probe_item_create and use a different data
type, which means we will avoid working with heap completely.


mpreisler:
For reference I checked that in maint-1.2 this is not an issue, it was introduced in master.

ACK

PR #1216, opened 2018-10-02
Fixed some Coverity issues

matejak:
* Refactored cleanup section of `_xccdf_fix_execute`:
   `fix_text` is allocated in `_xccdf_fix_decode_xml` and was possibly freed in `_write_text_to_fd_and_free`, but in meantime could have been be jumped over free, causing a memory leak.

* Improved freeing process of the env_values array: 
  Extracted the process into a function and called it where it was missing.
* Freed a se_ruid sexp that is deep-copied by probe_item_create.


mpreisler:
ACK

PR #1220, opened 2018-10-02
OVAL/probes/SEAP/sexp-manip.c: fixed false positives reported by coverity

matusmarhefka:
Coverity reports 2 false positives:
```
openscap-1.3.0_alpha1/src/OVAL/probes/SEAP/sexp-manip.c:934: overwrite_var: Overwriting "list" in "list = SEXP_list_new_rv(list, memb, ap)" leaks the storage that "list" points to.
```
and
```
openscap-1.3.0_alpha1/src/OVAL/probes/SEAP/sexp-manip.c:1037: overwrite_var: Overwriting "sexp" in "sexp = SEXP_list_rest_r(sexp, list)" leaks the storage that "sexp" points to.
```

In both cases pointer variable is passed as an argument to a function which only accesses that pointer but does not modify it.

### DIFF
--- a/src/OVAL/probes/SEAP/sexp-manip.c
+++ b/src/OVAL/probes/SEAP/sexp-manip.c
@@ -932,7 +932,10 @@ SEXP_t *SEXP_list_new (SEXP_t *memb, ...)
 
         va_start(ap, memb);
         list = SEXP_new ();
-        list = SEXP_list_new_rv(list, memb, ap);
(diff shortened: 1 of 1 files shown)


jan-cerny:
@openscap-ci test this please

jan-cerny:
@matusmarhefka it breaks test_api_probes_smoke

PR #1222, opened 2018-10-03
Change category of message

jan-cerny:
This message is annoying because it pollutes the output.
It is an usual situation.


matejak:
LGTM.

Issue #1199, opened 2018-10-01
crashes in pthread_mutex_unlock


jan-cerny:
#### Description of Problem:
oscap sometimes crashes randomly due to an abort  in `pthread_mutex_unlock`.
It happens in multiple XCCDF "unit tests" when they run command `oscap xccdf eval --remediate`.

#### OpenSCAP Version:
current master, as of 1st October 2018

#### Operating System & Version:
Fedora 28 Server

#### Steps to Reproduce:

1. cmake -DCMAKE_BUILD_TYPE=Debug .. && make
2. ctest --verbose -j4 -R  API/XCCDF/unittests/all.sh
3. run again and again until it crashes
4. coredumpctl 

#### Actual Results:
crash with a coredump
```
           PID: 30187 (oscap)
           UID: 0 (root)
           GID: 0 (root)
        Signal: 6 (ABRT)
     Timestamp: Mon 2018-10-01 12:47:57 CEST (10min ago)
  Command Line: /home/user/openscap/build/utils/oscap xccdf eval --remediate --results /tmp/test_remediation_bad_fix.out.CJvy44 /home/user/openscap/tests/API/XCCDF/unittests/test_remediation_bad_fix.xccdf.xml
    Executable: /home/user/openscap/build/utils/oscap
 Control Group: /user.slice/user-0.slice/session-3.scope
          Unit: session-3.scope
         Slice: user-0.slice
       Session: 3
     Owner UID: 0 (root)
       Boot ID: 72e2da705e284977bbdd33f6741adbe3
    Machine ID: a4187eabc7a340a1853e60d5ba212501
      Hostname: localhost.localdomain
       Storage: /var/lib/systemd/coredump/core.oscap.0.72e2da705e284977bbdd33f6741adbe3.30187.1538390877000000.lz4
       Message: Process 30187 (oscap) of user 0 dumped core.
                
                Stack trace of thread 30197:
                #0  0x00007f7414528eab raise (libc.so.6)
                #1  0x00007f74145135b9 abort (libc.so.6)
                #2  0x00007f7414513491 __assert_fail_base.cold.0 (libc.so.6)
                #3  0x00007f7414521612 __assert_fail (libc.so.6)
                #4  0x00007f741630acbe __pthread_tpp_change_priority (libpthread.so.0)
                #5  0x00007f74172be278 n/a (/home/user/openscap/build/src/libopenscap.so.24.0.0)
[...]

PR #1223, opened 2018-10-03
Check lock return values

jan-cerny:
This will inform us when some of the locks fails.
It can help us with debugging #1199.


matusmarhefka:
LGTM

PR #1226, opened 2018-10-04
Move SEAP out of public API

jan-cerny:
SEAP API is our internal detail. That means it shouldn't be exposed as public interface. In past it was necessary because probe processes used it, but now when probes are integrated into our library we don't have to expose these symbols.

This PR is a part of the activity that splits out non-aggressive patches  from PR #1150 .


jan-cerny:
@openscap-ci test this please

mpreisler:
Great stuff, ACK!

PR #1229, opened 2018-10-04
Remove useless functions

jan-cerny:
These functions are either unused or perform no action.
They were needed in the old system of separated processes,
but now each probe has its own functions.


mpreisler:
The fail is kind of mysterious. It doesn't seem to be related to this PR but it also doesn't seem random. Could you please take a look @jan-cerny ?

jan-cerny:
@mpreisler I beilieve that it is related to #1199. I cannot reproduce it on my machine, I tried to run the tests multiple times. I agree that it is mysterious because #1199 used to shout out only on test `API/XCCDF/unittests/all.sh`, but this time it's multiple tests.
Recently @matejak got a similar fail in #1230 , but only in test `API/XCCDF/unittests/all_python.sh`

matejak:
test this please

PR #1232, opened 2018-10-05
Fix SCE

jan-cerny:
Fix SCE runtime crash

47878e9 has fixed symbol visibilty.
However, it has completely broken SCE. SCE uses some private API
functions. When the symbol visibility wasn't set correctly, it was able
to find the private functions in libopenscap.so. After we have fixed
visibility, SCE engine cannot resolve the symbols it needs to run.
Instead of making these functions public, we can add the source files
containing the problematic symbols to the source file list for SCE
library.

Build SCE by default

Downstream distributions as RHEL or Fedora build it anyway.
SCE library doesn't require any special dependencies.
By building SCE by default we will increase the probability
of running SCE tests and discovering potential issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ find_program(GIT_EXECUTABLE git)
 
 # ---------- CORE FEATURE SWITCHES
 
-option(ENABLE_SCE "enables Script Check Engine - an alternative checking engine that lets you use executables instead of OVAL for checks" FALSE)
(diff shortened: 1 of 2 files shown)


matejak:
test this please

jan-cerny:
I think that CTest on RHEL7 does not understand SKIP_RETURN_CODE property.

jan-cerny:
@openscap-ci test this please

jan-cerny:
I think that SCE shouldn't be built on OSX.

PR #1234, opened 2018-10-08
Plug memory leaks

jan-cerny:
This PR fixes 3 potential leaks that have been found by the Coverity scan. The datives can be found in commit messages.


matejak:
LGTM, the Fedora test failiure is unrelated. Thanks!

PR #1235, opened 2018-10-08
Fix 2 Coverity problems

jan-cerny:
* Plug a memory leak in CVRF

If there is 0 iterations of the outer while loop, the 'filtered_ids'
variable isn't freed. Fortunately, we can allocate it inside
the while loop. If it's not free'd in this function, it is free'd
in cvrf_product_status_free.

* Specify size of the buffer in fscanf

The terminating null byte is not included in the length.


matejak:
Great PR, good catch with the buffer size!

PR #1239, opened 2018-10-09
Fixed memory errors in tests.

matejak:


jan-cerny:
Hi @matejak . I like your PR very much, but it would be more exciting if it had a description.

PR #1244, opened 2018-10-11
Fixed a bug with sudo as the first argument.

matejak:
A call to `shift` is necessarry outside of any function, because one can't use `shift` to shift script's arguments from a function.


mpreisler:
> because one can't use shift to shift script's arguments from a function.

would be great to remove the shift call in the function then and comment why.

Issue #1152, opened 2018-08-09
Test of sysctl probe fails with procps-ng >= 3.3.13


jan-cerny:
#### Description of Problem:
Linux 4.1.0 introduced a per-NIC IPv6 stable_secret file. The stable_secret file cannot be read until it is set.

To workaround this problem, we have created a fix in 7b8ca9c7829969232cc11c4363790c09d8e41806
to skip the stable_secret files when it is not readable. Otherwise we collect them.

That behavior was consistent with behavior of `sysctl` command. `sysctl` command also skipped them.

However, at procps-ng 3.3.13 the sysctl command changed its behavior concerning the stable_secret files. The change has been introduced in https://gitlab.com/procps-ng/procps/commit/854e2c5528e85420af253da51e8ec4703c679bb3#2ac4d3e6ddf3306cb8fcbae7b632520feb2ff20e
Since procps-ng 3.3.13, `sysctl` doesn't skip stable_secret files anymore.

This change starts to break test `probes/sysctl/all.sh`. In this test, we compare a list of items obtained by `sysctl -aN --deprecated` with OVAL results produce by OpenSCAP scan of ` tests/probes/sysctl/test_sysctl_probe_all.oval.xml`.
In OVAL results, the `stable_secret`items are not present, whereas in sysctl output they are present.

We need to find a way of being consistent with `sysctl` behavoir and at the same time not breaking the OpenSCAP test suite if using procps-ng older than 3.3.13. 

#### OpenSCAP Version:
current upstream master as of d8dd104b962451a35cc19b8153a5bf49d43e43ad

#### Operating System & Version:
Fedora Rawhide

#### Steps to Reproduce:

1. ctest --verbose -R probes/sysctl/all.sh

or
1. oscap oval eval --results results.xml tests/probes/sysctl/test_sysctl_probe_all.oval.xml
2. sysctl -aN --deprecated 
3. compare OVAL results with sysctl output

#### Actual Results:
Output of sysctl probe is not consistent with output of `sysctl` tool from procps-ng >= 3.3.13 regarding `stable_secret` files.

#### Expected Results:
Output of sysctl probe is consistent with output of `sysctl` tool from any procps-ng regarding `stable_secret` files.

matusmarhefka:
[...]

PR #1246, opened 2018-10-11
tests/probes/sysctl/test_sysctl_probe_all.sh: updated for procps-ng > 3.3.12

matusmarhefka:
Some sysctl options cannot be read until they are set. Sysctl from
procps-ng > 3.3.12 reports an error when reading these options but still
prints them to stdout. On the other hand openscap does not include these
options into results when they are not set. This was introduced in the commit
7b8ca9c and it is causing different results between openscap and sysctl from
procps-ng > 3.3.12. This commit updates the sysctl test to count with these
differences in case that procps-ng > 3.3.12 is installed on the system.

Fixes https://github.com/OpenSCAP/openscap/issues/1152


matejak:
LGTM, thank you!

PR #1249, opened 2018-10-15
Ensure there are no unresolved symbols in libopenscap_sce.so

jan-cerny:
After installing the library to /usr/lib64/ oscap wasn't able to find
SCE plugin and SCE was not shown in list of plugins in `oscap -V`. That
was because the `libopenscap_sce.so` library could not be loaded because
it contained unresolved symbol `__oscap_seterr`. We need to add
`error.c` and its dependencies to make sure `ldd` doesn't report any
unresolved symbols.


matejak:
test this please

Issue #902, opened 2017-11-06
ARF doesn't contain tailoring


jan-cerny:
I customized Fedora Standard profile using SCAP Workbench. I saved tailoring file to some folder. I evaluated the customized profile using `oscap xccdf eval` command-line and saved ARF results into ARF file. Then I wanted to generate profile-based bash remediation fix using oscap. However I got this error:
 
```
[jcerny@thinkpad wb-tailoring]$ myoscap xccdf generate fix --result-id xccdf_org.open-scap_testresult_xccdf_org.ssgproject.content_profile_standard_customized arf.xml 
OpenSCAP Error: Could not find Profile/@id="xccdf_org.ssgproject.content_profile_standard_customized" to build policy from TestResult/@id="xccdf_org.open-scap_testresult_xccdf_org.ssgproject.content_profile_standard_customized" [xccdf_session.c:1719]
```

I use OpenSCAP from current upstream GIT (head  05d3f20d8d1710e4497a3783db581a820eaaa4c4).

I think the problem is that ARF doesn't contain the customized profile.

EDIT: It could be solved by adding `--tailoring-file` option.

However I think the ARF should contain the tailored profile.

jan-cerny:
Fixed in #1253

PR #1253, opened 2018-10-24
Tailoring in arf

mpreisler:
Created as part of a collaboration with the team yesterday. Needs a lot more work. This is just a code dump for further development.
Fixes https://github.com/OpenSCAP/openscap/issues/902

### DIFF
--- a/src/DS/rds.c
+++ b/src/DS/rds.c
@@ -733,7 +733,7 @@ static int ds_rds_create_from_dom(xmlDocPtr* ret, xmlDocPtr sds_doc, xmlDocPtr t
 	xmlDOMWrapReconcileNamespaces(sds_wrap_ctxt, sds_res_node, 0);
 	xmlDOMWrapFreeCtxt(sds_wrap_ctxt);
 
-	if (tailoring_doc) {
+	if (tailoring_doc && strcmp(tailoring_filepath, "NONEXISTENT")) {
 		char *mangled_tailoring_filepath = ds_sds_mangle_filepath(tailoring_filepath);
 		char *tailoring_component_id = oscap_sprintf("scap_org.open-scap_comp_%s_tailoring", mangled_tailoring_filepath);
 		char *tailoring_component_ref_id = oscap_sprintf("scap_org.open-scap_cref_%s_tailoring", mangled_tailoring_filepath);
@@ -849,8 +849,6 @@ struct oscap_source *ds_rds_create_source(struct oscap_source *sds_source, struc
 			const size_t max_timestamp_len = 32;
 			tailoring_doc_timestamp = malloc(max_timestamp_len);
 			strftime(tailoring_doc_timestamp, max_timestamp_len, "%Y-%m-%dT%H:%M:%S", localtime(&file_stat.st_mtime));
-		} else {
-			return NULL;
 		}
 	}
 


src/DS/rds.c
```diff
@@ -730,6 +732,36 @@ static int ds_rds_create_from_dom(xmlDocPtr* ret, xmlDocPtr sds_doc, xmlDocPtr x
 	xmlDOMWrapReconcileNamespaces(sds_wrap_ctxt, sds_res_node, 0);
 	xmlDOMWrapFreeCtxt(sds_wrap_ctxt);
 
+	if (tailoring_doc) {
+		const char *tailoring_component_id = "scap_org.open-scap_comp_user_file_tailoring";
+		const char *tailoring_component_ref_id = "scap_org.open-scap_cref_user_file_tailoring";
```
matejak: For reference, this needs to be added somewhere here:
- ensure that the tailoring component IDs are unique.
- Make a function `get_unique_id(DOM, starting_id)`

src/DS/rds.c
```diff
@@ -730,6 +732,36 @@ static int ds_rds_create_from_dom(xmlDocPtr* ret, xmlDocPtr sds_doc, xmlDocPtr x
...
+
+		xmlDOMWrapCtxtPtr tailoring_wrap_ctxt = xmlDOMWrapNewCtxt();
+		xmlNodePtr tailoring_res_node = NULL;
+		xmlDOMWrapCloneNode(tailoring_wrap_ctxt, tailoring_doc, xmlDocGetRootElement(tailoring_doc),
+				&tailoring_res_node, doc, NULL, 1, 0);
+		xmlNsPtr sds_ns = sds_res_node->ns;
+		xmlNodePtr tailoring_component = xmlNewNode(sds_ns, BAD_CAST "component");
+		xmlSetProp(tailoring_component, BAD_CAST "id", BAD_CAST tailoring_component_id);
+		xmlAddChild(tailoring_component, tailoring_res_node);
+		xmlAddChild(sds_res_node, tailoring_component);
```
jan-cerny: This code aims to add a new component to a SCAP source datastream. I suggest investigating if we can reuse `ds_sds_compose_add_component` or `ds_sds_compose_add_component_with_ref` here.

src/DS/rds.c
jan-cerny: @matejak We already have it at other place. Take a look here:
https://github.com/OpenSCAP/openscap/blob/166a0a32fc9f72dc5aa9bddba164af8066fce724/src/DS/sds.c#L1059-L1079

src/DS/rds.c
```diff
@@ -730,6 +733,51 @@ static int ds_rds_create_from_dom(xmlDocPtr* ret, xmlDocPtr sds_doc, xmlDocPtr x
 	xmlDOMWrapReconcileNamespaces(sds_wrap_ctxt, sds_res_node, 0);
 	xmlDOMWrapFreeCtxt(sds_wrap_ctxt);
 
+	if (tailoring_doc) {
+		char* mangled_tailoring_filepath = ds_sds_mangle_filepath(tailoring_filepath);
```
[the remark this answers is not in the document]
jan-cerny: Oooops :smile_cat: :1st_place_medal: :nepal: 
Yes, I mean that.

src/DS/rds.c
```diff
@@ -730,6 +733,68 @@ static int ds_rds_create_from_dom(xmlDocPtr* ret, xmlDocPtr sds_doc, xmlDocPtr x
...
+		else {
+			checklists_element = node_get_child_element(datastream_element, "checklists");
+		}
+
+		xmlNodePtr tailoring_component_ref = xmlNewNode(sds_ns, BAD_CAST "component-ref");
+		xmlSetProp(tailoring_component_ref, BAD_CAST "id", BAD_CAST tailoring_component_ref_id);
+		char *tailoring_cref_href = oscap_sprintf("#%s", tailoring_component_id);
+		xmlNsPtr xlink_ns = xmlSearchNsByHref(doc, sds_res_node, BAD_CAST xlink_ns_uri);
+		if (!xlink_ns) {
+			oscap_seterr(OSCAP_EFAMILY_GLIBC,
```
jan-cerny: I guess `OSCAP_EFAMILY_XML` will be better here.

tests/API/XCCDF/tailoring/all.sh
```diff
@@ -84,6 +84,55 @@ function test_api_xccdf_tailoring_autonegotiation {
...
+    assert_exists 1 $component_xpath'/xccdf:Tailoring/xccdf:Profile[@id="xccdf_com.example.www_profile_customized"]'
+    assert_exists 1 $component_xpath'/xccdf:Tailoring/xccdf:Profile[@extends="xccdf_com.example.www_profile_baseline1"]'
+    rm -f $result
+}
+
+function test_api_xccdf_tailoring_profile_generate_fix {
+    local INPUT=$srcdir/$1
+    local TAILORING=$srcdir/$2
+
+    tailoring_result='tailoring_res'
```
jan-cerny: Please create a temporary file using `mktemp`.

tests/API/XCCDF/tailoring/all.sh
```diff
@@ -84,6 +84,55 @@ function test_api_xccdf_tailoring_autonegotiation {
...
+    fix_result='fix_res'
+    # tailoring profile only with "always fail" rule and generate bash fix
+    $OSCAP xccdf eval --tailoring-file $TAILORING --profile "xccdf_com.example.www_profile_customized" --results-arf $tailoring_result $INPUT || [ "$?" == "2" ]
+    $OSCAP xccdf generate fix --tailoring-id "scap_org.open-scap_cref_tests--API--XCCDF--tailoring--baseline.tailoring.xml_tailoring" --result-id xccdf_org.open-scap_test-result_xccdf-com.example.www_profile_customized --results $fix_result $tailoring_result
```
jan-cerny: The tailoring ID here is dependent on current working directory, which means this won't work if  you do `cd tests/API/XCCDF/tailoring` and you run `make check` in that directory. You'll get `"scap_org.open-scap_cref_.--baseline.tailoring.xml_tailoring"` as an component reference ID. 

I was thinking how to get the component reference ID.

A first idea that comes to my mind is to use absolute paths for files, eg. `local INPUT=$(realpath $srcdir/$1)`. 
However, then you'll have to build the tailoring ID from the absolute path. That would be a PITA. So instead you can hopefully get the compontent reference using an XPath query, eg `xpath -q -e 'string(//ds:component-ref[contains(@id, "_tailoring")]/@id)' tailoring_res`

tests/API/XCCDF/tailoring/all.sh
```diff
@@ -101,5 +150,8 @@ test_run "test_api_xccdf_tailoring_ds_hybrid_override" test_api_xccdf_tailoring_
 test_run "test_api_xccdf_tailoring_oscap_info_11" test_api_xccdf_tailoring_oscap_info simple-tailoring11.xml 1
 test_run "test_api_xccdf_tailoring_oscap_info_12" test_api_xccdf_tailoring_oscap_info simple-tailoring.xml 1
 test_run "test_api_xccdf_tailoring_autonegotiation" test_api_xccdf_tailoring_autonegotiation simple-tailoring-autonegotiation.xml xccdf_org.open-scap_profile_default 1
+test_run "test_api_xccdf_tailoring_simple_include_in_arf" test_api_xccdf_tailoring_simple_include_in_arf simple-xccdf.xml simple-tailoring.xml
+test_run "test_api_xccdf_tailoring_profile_include_in_arf" test_api_xccdf_tailoring_profile_include_in_arf baseline.xccdf.xml baseline.tailoring.xml
+
```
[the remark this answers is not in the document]
mildass: It should be ok now :)

jan-cerny:
We were able to make the tests for tailoring passing. :smile: 

However, the NIST test suite started to fail. (R3005). :sob: 
I was able to reproduce: 
```
cd tests/nist/R3005
oscap xccdf eval --datastream-id scap_gov.nist_datastream_r3005-tailoring_02 --tailoring-id scap_gov.nist_cref_r3005-xccdf_tailoring_02 --profile xccdf_gov.nist.validation_profile_r3005_tailoring_02 --results-arf arf.xml r3005-datastream-02.xml
```
Notice no `arf.xml` is created.

src/DS/rds.c
```diff
@@ -758,22 +823,45 @@ static int ds_rds_create_from_dom(xmlDocPtr* ret, xmlDocPtr sds_doc, xmlDocPtr x
...
 
+	xmlDoc *tailoring_doc = NULL;
+	char *tailoring_doc_timestamp = NULL;
+	char *tailoring_filepath = NULL;
+	if (tailoring_source) {
+		tailoring_doc = oscap_source_get_xmlDoc(tailoring_source);
+		if (tailoring_doc == NULL) {
+			return NULL;
+		}
+		tailoring_filepath = oscap_source_get_filepath(tailoring_source);
```
jan-cerny: We need to solve the situation when tailoring isn't a separate file, but it is a part of a source datastream. In this situation, the `tailoring_filepath` is unknown and the subsequent `stat()` returns error. As a result, this function returns NULL, and finally ARF isn't created at all.

jan-cerny:
@mildass Thanks!

PR #1254, opened 2018-10-24
Make sure that the manual HTML is built and distributed.

matejak:
Fixes the upsteam part of https://bugzilla.redhat.com/show_bug.cgi?id=1465661 point No.1


jan-cerny:
I have checked that it builds the HTML manual correctly during `make dist` and the resulting HTML manual is a part of distributed tarball. Thank you!

PR #1255, opened 2018-10-25
Check if system has enough memory

jan-cerny:
As a team effort we tried to mitigate `oscap` memory consumption.
We found out that there is already code that checks for memory
limits, but unfortunately it was never run, because a wrong SEXP
was measured. We will instead count items collected by each thread
and every 100 collections we will check if the system has enough
free memory.

### DIFF
--- a/src/OVAL/probes/probe/icache.c
+++ b/src/OVAL/probes/probe/icache.c
@@ -457,39 +457,26 @@ int probe_icache_nop(probe_icache_t *cache)
         return (0);
 }
 
-#define PROBE_RESULT_MEMCHECK_CTRESHOLD  32768  /* item count */
-#define PROBE_RESULT_MEMCHECK_MINFREEMEM 512    /* MiB */
-#define PROBE_RESULT_MEMCHECK_MAXRATIO   0.8   /* max. memory usage ratio - used/total */
+/* Every n-th call of probe_cobj_memcheck actually checks memory. This is due to performance reasons */
+#define PROBE_RESULT_MEMCHECK_FREQUENCY  100
+#define PROBE_RESULT_MEMCHECK_MINFREEMEM 256    /* MiB */
 
 /**
  * Returns 0 if the memory constraints are not reached. Otherwise, 1 is returned.
  * In case of an error, -1 is returned.
  */
-static int probe_cobj_memcheck(size_t item_cnt)
+static int probe_cobj_memcheck()
 {
-	if (item_cnt > PROBE_RESULT_MEMCHECK_CTRESHOLD) {
-		struct proc_memusage mu_proc;
+	static __thread unsigned long cnt = PROBE_RESULT_MEMCHECK_FREQUENCY;
+	if (++cnt > PROBE_RESULT_MEMCHECK_FREQUENCY) {
+		cnt = 0;
 		struct sys_memusage  mu_sys;
-		double c_ratio;
-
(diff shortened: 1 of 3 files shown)


jan-cerny:
I can reproduce the CI fail on my RHEL 7.5 VM. I'm going to investigate.

jan-cerny:
Steps to reproduce:
```
[root@localhost process58]# pwd
/root/openscap/tests/probes/process58
[root@localhost process58]# /root/openscap/run /root/openscap/utils/.libs/oscap oval eval  --results results.xml ./selinux_domain_label.oval.xml
Definition oval:x:def:1: unknown
Evaluation done.
```
results.xml contains:
```
........
          </interfaces>
        </system_info>
        <collected_objects>
          <object id="oval:x:obj:1" version="1" flag="incomplete">
            <message level="warning">Object is incomplete due to memory constraints.</message>
          </object>
        </collected_objects>
      </oval_system_characteristics>
```

src/OVAL/probes/probe/icache.c
```diff
@@ -457,39 +457,26 @@ int probe_icache_nop(probe_icache_t *cache)
...
 /**
  * Returns 0 if the memory constraints are not reached. Otherwise, 1 is returned.
  * In case of an error, -1 is returned.
  */
-static int probe_cobj_memcheck(size_t item_cnt)
+static int probe_cobj_memcheck()
 {
-	if (item_cnt > PROBE_RESULT_MEMCHECK_CTRESHOLD) {
-		struct proc_memusage mu_proc;
+	static __thread unsigned long cnt = PROBE_RESULT_MEMCHECK_FREQUENCY;
```
jan-cerny: It seems to me that we have to use a different way to declare thread-local storage in MSVC. Please see https://msdn.microsoft.com/en-us/library/6yh4a9k1.aspx

dahaic:
@jan-cerny are you waiting for review?

jan-cerny:
@dahaic A review would be very nice.

Also we need to check if the workaround works on Windows as expected. Note that this PR is opened against maint-1.2 but we run Windows CI only on maint-1.3. So currently we don't know what will happen when this patch will be forward-ported to maint-1.3 if it even compiles.

evgenyz:
Do we want it to be a part of 1.3? Do we have any use-cases or outstanding bugs upstream or downstream related to this?

jan-cerny:
@evgenyz If I recall it well, it's a workaround for problems with memory consumption s.a. https://bugzilla.redhat.com/show_bug.cgi?id=1548949

matejak:
We have realized that in 1.2 process-per-probe oscap, this fix would have a value - a probe that consumes a lot of memory would be killed, the particular rule result would be an error, but the scan could continue. However in 1.3, the fix is meaningless, as there is only one process, and as Evgeny independently realized, killing oscap because of lack of memory is pretty much the same as letting it to crash.
Therefore, I am closing this PR - we very likely won't make a 1.2 upstream release, and we can't use this in 1.3.

jan-cerny:
@matejak There is an open bugzilla against RHEL 7: https://bugzilla.redhat.com/show_bug.cgi?id=1548949 In RHEL7, we have OpenSCAP 1.2.17 from maint-1.2 branch, so this PR can help at least on RHEL 7. Also, the benefit of having the memory limit inside scanner would be that the process would quit gracefully with an error message, instead of killing boxes.

Issue #1264, opened 2018-11-09
unittest: test_probes_rpmverifypackage fails on systems without epoch


rfrohl:
#### Description of Problem:
Suse is not using epoch for rpms, which is causing test_probes_rpmverifypackage to fail

#### OpenSCAP Version:
1.3

#### Operating System & Version:
Suse

#### Steps to Reproduce:

1.  run ctest

#### Additional Information / Debugging Steps:
Will send a proposal change.

Issue #1258, opened 2018-10-26
unittest:  rpmverifyfile and rpmverify fail, because /etc/passwd not owned by a package (suse)


rfrohl:
#### Description of Problem:
passwd is not owned by a package in SUSE derived products. This is causing problems with the rpmverifyfile and rpmverify unit tests. 

What would be an alternative that would be acceptable? My patches use /etc/securetty instead (owned by pam).

Another issue is '/' which is also not owned by any package either. Here I used '/etc' for now.

Please let me know if any additional information could be useful. I am not sure if this is easy to communicate without looking at the patch.

#### OpenSCAP Version:
1.3.0

#### Operating System & Version:
SUSE

PR #1266, opened 2018-11-09
Disable test_probes_rpmverifypackage_epoch for Suse based systems

rfrohl:
Suse does not use epoch for rpms, therefor the test will always fail.

fix for #1264

### DIFF
--- a/tests/probes/rpmverifypackage/test_probes_rpmverifypackage.sh
+++ b/tests/probes/rpmverifypackage/test_probes_rpmverifypackage.sh
@@ -14,6 +14,8 @@
 set -e -o pipefail
 set -x
 
+[ -f /etc/os-release ] && . /etc/os-release
+
 function test_probes_rpmverifypackage {
 
     probecheck "rpmverifypackage" || return 255
@@ -79,7 +81,9 @@ function test_probes_rpmverifypackage_noepoch {
 
 test_init
 
+if [[ $ID_LIKE != *"suse"* ]]; then
 test_run "test_probes_rpmverifypackage_epoch" test_probes_rpmverifypackage_epoch
+fi
 test_run "test_probes_rpmverifypackage_noepoch" test_probes_rpmverifypackage_noepoch
 
 test_exit
--- a/tests/probes/rpmverifypackage/test_probes_rpmverifypackage.sh
+++ b/tests/probes/rpmverifypackage/test_probes_rpmverifypackage.sh
@@ -81,7 +81,7 @@ function test_probes_rpmverifypackage_noepoch {
 
 test_init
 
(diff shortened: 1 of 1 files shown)


tests/probes/rpmverifypackage/test_probes_rpmverifypackage.sh
```diff
@@ -79,7 +81,9 @@ function test_probes_rpmverifypackage_noepoch {
 
 test_init
 
+if [[ $ID_LIKE != *"suse"* ]]; then
```
matusmarhefka: Can we rather use `$ID` here? On my Fedora 29, there is no entry with `$ID_LIKE` in the `/etc/os-release` file. Also on RHEL-7.6 this is how the file looks like:
```
NAME="Red Hat Enterprise Linux Server"
VERSION="7.6 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
...
```

rfrohl: Would you like me to check that it is a redhat based distro? Meaning something along this line:
` if [[ $ID = *"rhel"* ]]; then`

I wanted to be conservative with my change originally. So that the unit test is only disabled it on suse based systems. But I have not problem with making the change, just want to be sure I understand you correctly.

One concern I would have with the change mentioned above is that the unit tests would be disabled on rhel7 with that change. My understanding is that rhel7 does not have a `/etc/os-release`, at least that was my interpretation of the CI results to my last change for #1258. I haven't found the time to investigate the issue further yet. Meaning I am nut a 100% sure.

rfrohl: @matusmarhefka, @jan-cerny sorry for the slow response I am quite busy atm.

I am okay to make the general change, but I have to investigate if this would work for openSUSE and SLE. I know that openSUSE would be okay. Still have to look into SLE.

rfrohl: change needs to look like this if `$ID` is used:
```
if [[ $ID != *"sle"* && $ID != *"suse"* ]]; then
```

matusmarhefka:
@openscap-ci test this please

matusmarhefka:
@openscap-ci okay to test

matusmarhefka:
@openscap-ci ok to test

jan-cerny:
@openscap-ci ok to test

rfrohl:
@jan-cerny sorry, should have caught that

rfrohl:
I think this should be good, I did not touch the test that failed for fedora. Can someone confirm that this is the correct assumption ?

Issue #1262, opened 2018-11-07
RFE: Include error message in HTML report by default


shawndwells:
When viewing the OpenSCAP HTML report for an ``error`` result, no information is given about what the error is. Could details be included in the results?

Current:
![image](https://user-images.githubusercontent.com/5713754/48144888-ae6e7a80-e27f-11e8-8c7b-b19493469203.png)

Could new row could be added, or perhaps a "details" link to expand the ``error`` row, to show the error?

redhatrises:
@shawndwells what happens when you add `--oval-results` to the scan?

shawndwells:
``--oval-results`` isn't in documentation, labs, etc. 

Why not include detailed results by default?

redhatrises:
@shawndwells what if you are not using OVAL?

shawndwells:
Even more reason to include scan results by default.

jan-cerny:
@shawndwells @redhatrises 

I like the idea that `--oval-results` and `--check-engine-results` could be turned on by default. That will be easy to change.

There is other problem that some errors don't show any information in the HTML report even if you add `--oval-results`. That would be more difficult to fix.

redhatrises:
@shawndwells btw adding `--oval-results` shows why there is an error.

redhatrises:
To be able to close this, error details should be given by default without having to specify `--results-arf` or `--oval-results`. See https://github.com/OpenSCAP/openscap/pull/1267#issuecomment-452024470

PR #1267, opened 2018-11-09
Show details in HTML report by default if --results-arf is given

jan-cerny:
This commit changes the `oscap` behavior in the following way:
If `oscap xccdf eval` is called with both `--results-arf` and
`--report` the HTML report will contain `OVAL details` section
even if `--oval-results` is not given by the user.
I think we can do it because the ARF file already contains OVAL
results anyway. I think it's not convenient to type `--oval-results`
explicitly.

### DIFF
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -1327,7 +1327,7 @@ int xccdf_session_export_xccdf(struct xccdf_session *session)
 
 	struct oscap_source* results = session->xccdf.result_source;
 	struct oscap_source* arf = NULL;
-	if (session->export.oval_results) {
+	if (session->export.oval_results || session->export.arf_file) {
 		arf = xccdf_session_create_arf_source(session);
 		if (arf == NULL) {
 			return 1;


jan-cerny:
And if this would be OK the next step could be to turn on --oval-results and --check-engine-results by default as is discussed in #1262 .

redhatrises:
> And if this would be OK the next step could be to turn on --oval-results and --check-engine-results by default as is discussed in #1262 .

I like it. What happens if there is no ARF file? Does it just continue without an error?

shawndwells:
Aweomse! thank you!

jan-cerny:
`--results` still generates plain XCCDF results, that isn't changed by this PR. 

If in future we would make --oval-results and --check-engine-results implicit (turned on by default), then I think it should work the same way as when you type those options today.

jan-cerny:
@openscap-ci test this please

jan-cerny:
Hi @matusmarhefka @redhatrises @shawndwells @mildass @yuumasato @matejak @dahaic @ggbecker , are there any further changes needed?

matusmarhefka:
Original issue was resolved so I am merging this. Thanks @jan-cerny

matusmarhefka:
@jan-cerny Maybe one more question, do we want to add this also into `maint-1.2`?

shawndwells:
On 1/7/19 3:37 AM, Jan Černý wrote:
> Hi @matusmarhefka <https://github.com/matusmarhefka> @redhatrises 
> <https://github.com/redhatrises> @shawndwells 
> <https://github.com/shawndwells> @mildass <https://github.com/mildass> 
> @yuumasato <https://github.com/yuumasato> @matejak 
> <https://github.com/matejak> @dahaic <https://github.com/dahaic> 
> @ggbecker <https://github.com/ggbecker> , are there any further 
> changes needed?



Seems like an overall improvement - thanks!

Could the HTML report included details by *default*, without needing the 
various --oval-results and --arf-results flags? Most people don't know 
what those flags are as they're undocumented.

redhatrises:
> On 1/7/19 3:37 AM, Jan Černý wrote: Hi @matusmarhefka <https://github.com/matusmarhefka> @redhatrises <https://github.com/redhatrises> @shawndwells <https://github.com/shawndwells> @mildass <https://github.com/mildass> @yuumasato <https://github.com/yuumasato> @matejak <https://github.com/matejak> @dahaic <https://github.com/dahaic> @ggbecker <https://github.com/ggbecker> , are there any further changes needed?
> Seems like an overall improvement - thanks! Could the HTML report included details by *default*, without needing the various --oval-results and --arf-results flags? Most people don't know what those flags are as they're undocumented.

@shawndwells where are you looking for docummentation? CLI or elsewhere?

jan-cerny:
@shawndwells 

> Could the HTML report included details by *default*, without needing the various --oval-results and --arf-results flags?
Yes, they could. I think that will be a great UX improvement.

>  Most people don't know what those flags are as they're undocumented.
They aren't undocumented, they're described in man pages etc., but they're often missing in slides or tutorials.

PR #1268, opened 2018-11-13
Fix trailing whitespace in Ansible Playbooks

jan-cerny:
Code which generates Ansible remediations updated to:
* remove trailing whitespaces from the Ansible remediation lines,
* not indent empty Ansible remediation lines.

These issues were found by `ansible-lint`. Also trailing whitespaces
got removed from the generated remediation headers.

There is still one issue reported by the `yamllint` remaining not fixed in openscap code: `hosts`, `vars` and `tasks` sections are indented by an additional space which is reported as an error:
```
error    wrong indentation: expected 0 but found 1  (indentation)
```
We can fix it in `openscap`, but it also needs an update in `ComplianceAsCode/content` (generation scripts).


jan-cerny:
@matusmarhefka Let's fix that indentation in a separate PR. If you'll do that we need to make sure ComplianceAsContent content can build with both new and old OpenSCAP.

jan-cerny:
This PR has been completely reworked by @matusmarhefka so I think it's OK if I review that.

jan-cerny:
@matusmarhefka Tremendous patch! Thank you.

PR #1270, opened 2018-11-16
Fix dpkginfo_probe

teselkin:
dpkginfo_init doesn't check if dpkg cache initialization has succeeded.

It may happen that 'Dir::Cache::pkgcahce' is set to an empty string
to save some space (e.g. Ubuntu Xenial docker image does that), and in
such case calling dpkgingo_probe_main will end up in a SEGFAULT.

This commit checks if dpkg initialization has succeded before performing
any check.

This issue affects 1.2 and 1.3 versions of OpenSCAP.

### DIFF
--- a/src/OVAL/probes/unix/linux/dpkginfo_probe.c
+++ b/src/OVAL/probes/unix/linux/dpkginfo_probe.c
@@ -62,16 +62,23 @@
 #include "dpkginfo_probe.h"
 
 struct dpkginfo_global {
+        int init_done;
         pthread_mutex_t mutex;
 };
 
-static struct dpkginfo_global g_dpkg;
+static struct dpkginfo_global g_dpkg = {
+        .init_done = -1,
+};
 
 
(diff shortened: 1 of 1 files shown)


jan-cerny:
@openscap-ci ok to test

jan-cerny:
Hi @teselkin 

Thank you very much for your patch.

You told us the issue affects 1.2 version as well. Could you please switch the PR to be based on `maint-1.2` branch? The maint-1.2 branch will be merged into maint-1.3 so your patch will eventually get into maint-1.3, but it doesn't work the opposite way. Thanks.

The CI fail should be unrelated.

Otherwise LGTM.

teselkin:
Hi @jan-cerny ,
Here is the patch adopted for 1.2 - https://github.com/OpenSCAP/openscap/pull/1272

matejak:
@openscap-ci add to whitelist

jan-cerny:
@teselkin I'll do that, thanks.

PR #1272, opened 2018-11-20
dpkginfo_init doesn't check if dpkg cache initialization has succeeded.

teselkin:
It may happen that 'Dir::Cache::pkgcache' is set to an empty string
to save some space (e.g. Ubuntu Xenial docker image does that), and in
such case calling dpkginfo_probe_main will end up in a SEGFAULT.

This commit checks if dpkg initialization has succeded before performing
any check.

This issue affects 1.2 and 1.3 versions of OpenSCAP.

Fix for 1.3 - https://github.com/OpenSCAP/openscap/pull/1270

### DIFF
--- a/src/OVAL/probes/unix/linux/dpkginfo.c
+++ b/src/OVAL/probes/unix/linux/dpkginfo.c
@@ -62,16 +62,23 @@
 
 
 struct dpkginfo_global {
+        int init_done;
         pthread_mutex_t mutex;
 };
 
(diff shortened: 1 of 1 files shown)


jan-cerny:
@openscap-ci ok to test

jan-cerny:
@openscap-ci test this please

jan-cerny:
Thank you!

PR #1273, opened 2018-11-20
Fix possible invalid read

jan-cerny:
If "current" is empty string, then end = current - 1 which is
invalid address.

Unfortunately discovered after I merged #1268


matusmarhefka:
@jan-cerny Good catch, thanks for fixing this :smile:

PR #1274, opened 2018-11-20
Prevent duplicate variables in Ansible Playbooks

jan-cerny:
The same XCCDF value can appear in multiple rules, which means
there can be same variable in multiple Ansible task. But when
we generate a Playbook we should put the variable only once.
Ansible produces a warning if there are multiple definitions
of the same variable.

See https://github.com/OpenSCAP/openscap/pull/896 for further context if you are interested.

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -134,11 +134,11 @@ static int _write_remediation_to_fd_and_free(int output_fd, const char* template
 					free(text);
 					return 1;
 				}
-			}
 
-			if (_write_text_to_fd(output_fd, "\n") != 0) {
-				free(text);
-				return 1;
+				if (_write_text_to_fd(output_fd, "\n") != 0) {
+					free(text);
+					return 1;
+				}
 			}
 
 			if (next_delim != NULL) {
@@ -610,7 +610,7 @@ static int _write_fix_missing_warning_to_fd(const char *sys, int output_fd, stru
 	}
 }
 
-static inline int _xccdf_policy_rule_generate_fix_ansible(const char *template, int output_fd, const char *fix_text, bool ansible_variable_mode)
+static inline int _parse_ansible_fix(const char *fix_text, struct oscap_list *variables, struct oscap_list *tasks)
 {
 #if defined USE_REGEX_PCRE
 	// TODO: Tolerate different indentation styles in this regex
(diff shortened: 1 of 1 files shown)


src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -733,37 +710,55 @@ static inline int _xccdf_policy_rule_generate_fix(struct xccdf_policy *policy, s
...
 				xccdf_rule_get_id(rule));
 		xccdf_fix_free(cfix);
 		return 1;
 	}
 	xccdf_fix_free(cfix);
+	return 0;
+}
 
-	int ret = _write_fix_header_to_fd(template, output_fd, rule, current, total);
+static int _xccdf_policy_rule_generate_fix(struct xccdf_policy *policy, struct xccdf_rule *rule, const char *template, int output_fd, unsigned int current, unsigned int total, bool ansible_variable_mode)
```
jan-cerny: ansible_variable_mode is unused in this function, please remove it

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -977,30 +972,45 @@ int xccdf_policy_generate_fix(struct xccdf_policy *policy, struct xccdf_result *
 
 	// In Ansible we have to generate variables first and then tasks
 	if (strcmp(sys, "urn:xccdf:fix:script:ansible") == 0) {
+		struct oscap_list *variables = oscap_list_new();
+		struct oscap_list *tasks = oscap_list_new();
+		struct oscap_iterator *rules_to_fix_it = oscap_iterator_new(rules_to_fix);
```
matejak: I propose to do some function extraction here. What about introducing `get_vars_and_tasks_from_rules(rules_to_fix, variables, tasks)` here?

jan-cerny:
@matejak I have tried to do a function refactoring.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -733,37 +710,55 @@ static inline int _xccdf_policy_rule_generate_fix(struct xccdf_policy *policy, s
...
 				xccdf_rule_get_id(rule));
 		xccdf_fix_free(cfix);
 		return 1;
 	}
 	xccdf_fix_free(cfix);
+	return 0;
+}
 
-	int ret = _write_fix_header_to_fd(template, output_fd, rule, current, total);
+static int _xccdf_policy_rule_generate_fix(struct xccdf_policy *policy, struct xccdf_rule *rule, const char *template, int output_fd, unsigned int current, unsigned int total)
```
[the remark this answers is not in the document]
jan-cerny: Nice, I'll look into that.

matejak:
@jan-cerny And please don't forget to rebase, tests should pass now :-)

matejak:
Looks great, thank you!

PR #1265, opened 2018-11-09
Change test_sysctl_probe_all to succeed for non-root users

rfrohl:
- add blacklist for confidential kernel params
- set PATH explicitly to make sure sysctl is found

fix for #1263

PR #1283, opened 2018-11-27
Expand the sysctl test blacklist so it passes.

### LINKED ISSUE #1263: unittest: test_sysctl_probe fails for non-root users
#### Description of Problem:
With the test_sysctl_probe I have two problems on Suse based systems:
- The main problem is that not all sysctl params are accessible to non-root users. 
- Another minor problem is that the sysctl binary lives in sbin, which is not part of the user path.

#### OpenSCAP Version:
1.3

Suse

#### Steps to Reproduce:

1. run ctest as non-root user

Will send a proposal patch

matejak:
Complements #1265, fixes #1263.


matusmarhefka:
@matejak Thanks for extending the test.

PR #1288, opened 2019-01-16
Fix Ansible mapping keys space indentation

redhatrises:


jan-cerny:
Thanks @redhatrises .

I wonder why it was indented that way. It looks like it was introduced here https://github.com/OpenSCAP/openscap/pull/522 but with no reason.

PR #1289, opened 2019-01-17
Improve error message for pcre_exec fail

jan-cerny:
This can help us to resolve problems similar to RHBZ#1655943.


jan-cerny:
@matejak Is this what you meant? Would you like to have an human readable translation of error codes? They can be found in `man pcreapi`.

jan-cerny:
@openscap-ci test this please

matejak:
@jan-cerny Would it make sense to add a similar message few lines above in case the pattern can't be compiled? That could happen when people editing a datastream and introduce some typo.

jan-cerny:
@matejak  Yes, it would, I'll do that.

jan-cerny:
@matejak I have updated the PR with another commit.

PR #1290, opened 2019-01-18
Always show OVAL details in HTML report

jan-cerny:
Users will no longer need to specify `--oval-results` to get the OVAL details in HTML report. The OVAL details will be shown always.

I know this change is highly controversial, but it has been requested multiple times, for example in #1262.

Moreover, I have discovered that it isn't true that you have to specify `--results` together with report. Probably this statement has been forgotten in the man page during the HTML report redesign.

### DIFF
--- a/tests/oval_details/test_oval_details.sh
+++ b/tests/oval_details/test_oval_details.sh
@@ -14,13 +14,15 @@ function test_oval_details_implicit {
   # Tests if OVAL Details are present in HTML report by default
   # without specifing --oval-results --results (new behavior)
   xccdffile=$srcdir/$1.xccdf.xml
-  resultfile=$output_dir/$1.result.xml
   reportfile=$output_dir/$1.report.html
+  resultfile="$1.result.xml"
   oval_resultfile="$1.oval.xml.result.xml"
   $OSCAP xccdf eval --report $reportfile $xccdffile
   [ -f $reportfile ]
   [ ! -f $resultfile ]
+  [ ! -f $output_dir/$resultfile ]
   [ ! -f $oval_resultfile ]
+  [ ! -f $output_dir/$oval_resultfile ]
   grep -i $2 $reportfile >/dev/null && grep -i $3 $reportfile >/dev/null
   rm $reportfile
 }
--- a/tests/oval_details/test_oval_details.sh
+++ b/tests/oval_details/test_oval_details.sh
@@ -10,13 +10,37 @@ output_dir=`mktemp -d -t oval_details_XXXXXX`
 
 # Test cases.
 
-function test_oval_details {
+function test_oval_details_implicit {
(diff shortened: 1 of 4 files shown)


jan-cerny:
Hi @ggbecker @matejak @matusmarhefka @yuumasato @dahaic . Is there any volunteer to review?

matusmarhefka:
@jan-cerny Great PR, thank you. Have you also considered to add a test for this new functionality?

jan-cerny:
@matusmarhefka That's a great idea. I'll try to add a new test.

jan-cerny:
@matusmarhefka I have modified the existing test to test that OVAL details are present in HTML report by default and at the same time the old way still works.

jan-cerny:
Hi, do you think that we need to add an option to explicitly remove OVAL details from HTML report, eg. `--no-oval-details`? I doubt it would be useful.

tests/oval_details/test_oval_details.sh
```diff
@@ -10,13 +10,37 @@ output_dir=`mktemp -d -t oval_details_XXXXXX`
...
+function test_oval_details_implicit {
+  # Tests if OVAL Details are present in HTML report by default
+  # without specifing --oval-results --results (new behavior)
   xccdffile=$srcdir/$1.xccdf.xml
   resultfile=$output_dir/$1.result.xml
   reportfile=$output_dir/$1.report.html
+  oval_resultfile="$1.oval.xml.result.xml"
+  $OSCAP xccdf eval --report $reportfile $xccdffile
+  [ -f $reportfile ]
+  [ ! -f $resultfile ]
```
matusmarhefka: I would add one more test here to also check if the results file is not generated in the CWD, just to be sure: `[ ! -f $(basename $resultfile) ]`

matusmarhefka:
@jan-cerny I don't think that option like `--no-oval-details` would be useful, I would just fill OVAL details into the HTML report all the time.

jan-cerny:
> I don't think that option like --no-oval-details would be useful, I would just fill OVAL details into the HTML report all the time.

I totally agree.

tests/oval_details/test_oval_details.sh
jan-cerny: OK, I'll do that.

jan-cerny:
I have updated the test to check result files in current working directory.

dahaic:
> @jan-cerny I don't think that option like `--no-oval-details` would be useful, I would just fill OVAL details into the HTML report all the time.

Depends on the size difference. Remember, the HTML is already quite a whale. :)

matusmarhefka:
@dahaic It is not so tragic, report with OVAL results for Fedora standard profile is 1.8 MB compared to the report without OVAL results which is 1.6 MB.

matusmarhefka:
@jan-cerny Great work, also thanks for implementing the test.

PR #1291, opened 2019-01-18
 Merge branch 'maint-1.2' into maint-1.3 

jan-cerny:


matejak:
The Fedora tests that fail are executed under RHEL, where they pass. Therefore LGTM, merging.

PR #1292, opened 2019-01-18
Removed coverage tracking and Travis tests

matejak:
Coverage output is relevant only for unittests, where it can be assumed that the covered code is tested sufficiently. As our tests don't work on Travis anyway, I have disabled them.
Conversely, our acceptance testing scripts don't provide any reasonable indication that the code they execute is correct.


jan-cerny:
I think it would be better to fix the tests to be able to run on Ubuntu.

Could you elaborate more on "Conversely, our acceptance testing scripts don't provide any reasonable indication that the code they execute is correct.", please?

matejak:
Let's not forget that it will stay with us in the git repo.
Moreover, the codecov service was fine without it, it was the Sonar that may need it. If we start adding proper unit tests to the project at some point, we certainly won't need it.

PR #1295, opened 2019-01-30
Small fixes in "oscap xccdf generate fix", mostly for Ansible

### LINKED ISSUE #1112: Fix profile titles in generated ansible roles
#### Description of Problem:
When there are newlines in profile title there are no commentary marks for these newlines in the generated ansible role headers. This results in an invalid generated ansible roles.

#### Steps to Reproduce:
To reproduce you need to build older version of SSG (tested with `0.1.36`) which has one such a profile title. The role with the issue is `ssg-rhel6-role-nist-cl-il-al.yml`.

This issue is also reported in [RHBZ#1548499](https://bugzilla.redhat.com/show_bug.cgi?id=1548499) and can be fixed in a simillar way as the profile description was fixed in the PR #838.

jan-cerny:
* Fixes variables indentation. As we use 2 spaces to indent our content, the variables
which are on second level of indentation, should use 4 spaces.
* Handles multiline profile tittles and descriptions. Multiline titles were handled at all (See issue #1112). The code is improved to strip the whitespaces.
* Replaces `--template` by `--fix-type` in Playbook header. Using `--fix-type` is more understandable and it's easier for users than using the URN namespace.

 Fixes #1112.


jan-cerny:
Hi @ggbecker @matejak @matusmarhefka @yuumasato @dahaic . Is there any volunteer to review?

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -802,6 +802,57 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
 	return ret;
 }
 
+static char *_comment_multiline_text(char *text)
```
matejak: Please split this function into smaller functions - it is somehow mind-bending to guess the intent of the code.

Unfortunately, we don't have easy unit-test support, so please include a comment indicating what the functions is supposed to do - it may not be obvious outside of the context of this PR.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -802,6 +802,57 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+		return oscap_strdup("Not available");
+	}
+	const char *filler = "\n# ";
+	size_t buffer_size = strlen(text) + 1; // +1 for terminating '\0'
+	char *buffer = malloc(buffer_size);
+	char *saveptr, *str;
+	size_t filler_len = strlen(filler);
+	size_t result_len = 0;
+	bool first = true;
+	for (str = text; ; str = NULL) {
```
matejak: The `for` condition is somehow mind-bending.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -802,6 +802,57 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+		char *token = oscap_strtok_r(str, "\n", &saveptr);
+		if (token == NULL) {
+			break;
+		}
+		while (isspace(*token)) {
+			token++;
+		}
+		size_t token_len = strlen(token);
+		if (token_len > 0) {
```
matejak: This is a good candidate for a function for extraction.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -802,6 +802,57 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+			char *token_last_char = token + token_len - 1;
+			while(isspace(*token_last_char)) {
+				token_last_char--;
+			}
+			*(token_last_char + 1) = '\0';
+			token_len = strlen(token);
+		}
+		if (token_len > 0) {
```
matejak: This is also a good candidate for a function for extraction.

src/XCCDF_POLICY/xccdf_policy_remediate.c
jan-cerny: This function would be a perfect candidate for an unit test!

src/XCCDF_POLICY/xccdf_policy_remediate.c
jan-cerny: It doesn't seem resuable to me because there is multiple assumptions (size of the buffer, size of the filler, size of the token, index of the token).

matejak: It may not be reusable, but it would decrease the cognitive complexity of the code. If you are against an extraction, then put a comment there.

src/XCCDF_POLICY/xccdf_policy_remediate.c
matejak: Yes. and I think that you can rename it to `trim_trailing_whitespace` or `rstrip`, as  multiline comments are out of scope of this function.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -802,6 +802,72 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
 	return ret;
 }
 
+static void _comment_multiline_text_strip_trailing_whitespace(char *token, size_t token_len)
```
matejak: I think that you can rename this one to `trim_trailing_whitespace` or `rstrip`, as multiline comments are out of scope of this function.

src/XCCDF_POLICY/xccdf_policy_remediate.c
jan-cerny: I think we can survive coding normally it this case. We don't need to create lot of pointers if we are not going to use the code again.

src/XCCDF_POLICY/xccdf_policy_remediate.c
jan-cerny: OK

jan-cerny:
@matejak I have renamed the static function which removes the trailing whitespace to just `_trim_trailing_whitespace`.

However, I haven't found any nice function to extract from the `_comment_multiline_text`.

src/XCCDF_POLICY/xccdf_policy_remediate.c
matejak: OK, may be.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -802,6 +802,72 @@ static int _xccdf_item_recursive_gather_selected_rules(struct xccdf_policy *poli
...
+		if (token == NULL) {
+			break;
+		}
+		/* Strip leading whitespace */
+		while (isspace(*token)) {
+			token++;
+		}
+		size_t token_len = strlen(token);
+		if (token_len > 0) {
+			/* Strip trailing whitespace */
```
matejak: Pleare remove that comment - it is not needed, as you have introduced the `_trim_trailing_whitespace` function, so it is now obvious what is going on.

matejak:
@jan-cerny The PR is basically good to merge. If you have time, please remove a redundant comment.

matejak:
Thank you for the PR!

PR #1296, opened 2019-01-30
Split tests/API/XCCDF/unittests/all.sh to separate test cases

jan-cerny:
The `all.sh` script combines roughly 80 test cases in a single test case. When running unit tests, it's difficult to debug fails.  It's also difficult to orient in output. It isn't possible to easily run selected
single test case. If we split the test cases, it will be easier to work with the tests. For example, you can run `ctest -R <test_name>`.  The `all.sh` script can be considered an unnecessary abstraction layer.


matejak:
LGTM, but what about the omission of `test_run` call, have you examined related potential side-effects?

jan-cerny:
@matejak Thanks for the question, I haven't examined it before.

But I have looked at it and I think that we shouldn't use test_run anywhere, because the functionality is more or less replaced by the ctest framework. Take a look at its source: 

https://github.com/OpenSCAP/openscap/blob/be146691167900e920d14cbd118bfe2b95c601c4/tests/test_common.sh.in#L76-L98

It runs the script and based on the results it prints PASS/FAIL/SKIP which is basically what ctest does.

What do you think? Is it OK to remove it or do we need to somehow wrap the tests in functions that will be passed as arguments to test_run?

jan-cerny:
@matejak Do you remember the potential side effects that you had in mind? It seems to me that the change should be harmless because the return codes are recognized by `ctest` itself so we don't need to handle it by a wrapper function. What's your opinion?

matejak:
Agreed. Therefore, LGTM, thank you very much for your PR!

Issue #1173, opened 2018-08-17
Bug: double chdir/chroot in probe rpmverifypackage


matusmarhefka:
There is a problem in both branches that probe rpmverifypackage calls chdir/chroot second time:

* `maint-1.2`:
  * first called from src/OVAL/probes/probe/main.c
  * second call from src/OVAL/probes/unix/linux/rpmverifypackage.c
* `master`:
  * first called from src/OVAL/probes/probe/probe_main.c
  * second call from src/OVAL/probes/unix/linux/rpmverifypackage_probe.c

This causes an issue for offline scan (tested on maint-1.2):
```
E: probe_rpmverifypackage: RPM: Failed to initialize NSS library

error: Failed to initialize NSS library
E: probe_rpmverifypackage: Unable to chdir to '/mnt/root', errno: 2 "No such file or directory"
W: oscap:     Can't receive message: 125, Operation canceled.
E: oscap:     Recv: retry limit (0) reached.
OpenSCAP Error: Probe at sd=1 (rpmverifypackage) reported an error: Initialization failed [oval_probe_ext.c:393]
Unable to receive a message from probe [oval_probe_ext.c:579]
Invalid oval result type: -1. [oval_resultTest.c:179]
```

Also reported in [RHBZ#1603347](https://bugzilla.redhat.com/show_bug.cgi?id=1603347)

matusmarhefka:
We also need to check other rpm* probes as we set them to work in chroot, we should not call the `rpmtsSetRootDir` function inside these probes.

* https://github.com/rpm-software-management/rpm/blob/master/lib/rpmts.c#L853
* http://ftp.rpm.org/api/4.12.0.1/group__rpmts.html#ga2cc54aa978a49d0b5cefa64095c8ee41

PR #1306, opened 2019-02-28
Added utils/oscap-podman for SCAP evaluation of podman images and containers

matusmarhefka:
The `oscap-podman` script should make it possible to perform the SCAP evaluation of podman images and containers. The script complements `oscap-docker` script which does the same for docker images and containers.

Currently, there are multiple issues with offline scanning in RPM probes (for example https://github.com/OpenSCAP/openscap/issues/1173) which prevent this script to work without issues (the same applies also for the `oscap-docker` script).

### DIFF
--- a/utils/oscap-podman
+++ b/utils/oscap-podman
@@ -1,0 +1,90 @@
+#!/bin/bash
+
+# Copyright 2019 Red Hat Inc., Durham, North Carolina.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


utils/oscap-podman
```diff
@@ -0,0 +1,80 @@
...
+    echo "oscap-podman -- Tool for SCAP evaluation of podman images and containers."
+    echo
+    echo "Compliance scan of podman image:"
+    echo "$ sudo oscap-podman IMAGE_NAME OSCAP_ARGUMENT [OSCAP_ARGUMENT...]"
+    echo
+    echo "Compliance scan of podman container:"
+    echo "$ sudo oscap-podman CONTAINER_NAME OSCAP_ARGUMENT [OSCAP_ARGUMENT...]"
+    echo
+    echo "See \`man oscap\` to learn more about semantics of OSCAP_ARGUMENT options."
+}
```
jan-cerny: We should add a man page as well.

matusmarhefka:
@jan-cerny Thanks for the review, I have updated the PR based on your comments. Man page is still missing.

jan-cerny:
@matusmarhefka you can do it in a separate PR

PR #1307, opened 2019-03-06
add block_size support for Linux partition_state

maltek:
it got specified in OVAL 5.11.2: https://github.com/OVALProject/Language/blob/5.11.2/docs/linux-definitions-schema.md#-partition_state-

### DIFF
--- a/src/OVAL/probes/unix/linux/partition_probe.c
+++ b/src/OVAL/probes/unix/linux/partition_probe.c
@@ -219,6 +219,7 @@ static int collect_item(probe_ctx *ctx, oval_schema_version_t over, struct mnten
                                  "total_space",   OVAL_DATATYPE_INTEGER, (int64_t)stvfs.f_blocks,
                                  "space_used",    OVAL_DATATYPE_INTEGER, (int64_t)(stvfs.f_blocks - stvfs.f_bfree),
                                  "space_left",    OVAL_DATATYPE_INTEGER, (int64_t)stvfs.f_bfree,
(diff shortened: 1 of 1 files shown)


redhatrises:
@openscap-ci test this please

jan-cerny:
Hi! Thanks for your patch. If this attribute got added in 5.11.2 and wasn't present in 5.11.1 it would mean that OpenSCAP would produce invalid results according to previous schemas. This needs to be handled by a condition that compares the schema version used using `oval_schema_version_cmp`. The attribute should be populated only if the schema of input content is in version that defines this attribute.
You can get inspired by 
https://github.com/OpenSCAP/openscap/blob/4ca85d11d35211d942478aa698edc7cfc726f63f/src/OVAL/probes/unix/shadow_probe.c#L161-L165

redhatrises:
@openscap-ci test this please

jan-cerny:
Hi @maltek . Thanks for your update. The patch looks good to me. However, I have noticed that it's created against the master branch. That will take a long time to get to users. The patch can be included into maint-1.3 branch and it will get forward-ported. I'll try if it can be merged in maint-1.3.

jan-cerny:
It seems that the patch cannot be merged, so I will merge it manually.

jan-cerny:
Merged manually into maint-1.3 as 38a4dd6432066e6773543c6aacff957a7f6e046b and 683ed8c6ed9e79f6a50c4c5ff8582f6954b22c01

jan-cerny:
Thank you!

PR #1315, opened 2019-03-18
Don't ignore file behaviors if filepath is specified

jan-cerny:
The effect of the code removed by this patch is that the max_depth, recurse, recurse_direction and recurse_file_system attributes of the behaviors element are completely ignored and reset to their default values if filepath element is specified in the textfilecontent54_object. This is against OVAL specification
and it is inconsistent with file probe, from where a similar code has been removed in 93d5f1416f232d6fa21fe8f2ad771d003749ea7b.


matejak:
LGTM, thank you for fixing this omission!

PR #1316, opened 2019-03-19
Add missing OVAL license

jan-cerny:
The new OVAL 5.11.2 schemas contain terms_of_use element which
says that it is required to include OVAL license header when distributing
copies of the OVAL Schema. To satisfy their requirement, we will
add a copy of the OVAL license obtained from:
https://github.com/OVAL-Community/OVAL/blob/master/terms-of-use.rst

See eg.:
https://github.com/OpenSCAP/openscap/blob/ef16b42e5327dea1e6708584326b3230a42a913c/schemas/oval/5.11.2/oval-definitions-schema.xsd#L39

### DIFF
--- a/schemas/oval/terms-of-use.rst
+++ b/schemas/oval/terms-of-use.rst
@@ -1,0 +1,45 @@
+.. _terms-of-use:
+
+Terms of Use
+============
+
+Introduction
+------------
+OVAL is an open standard developed by the information security community as represented by the OVAL Board and the OVAL discussion lists, and maintained by the Center for Internet Security, Inc. under license from the United States Department of Homeland Security (U.S. DHS) on this public OVAL Website.
+
+The OVAL Language and any resulting OVAL content based upon the language that is stored in the OVAL Repository are free to use by any organization or individual for any research, development, and/or commercial purposes, per below.
+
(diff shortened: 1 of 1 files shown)


yuumasato:
> When distributing copies of the OVAL Schema, this license header must be included.

Does "this header" means this `<terms_of_use>` element or the whole terms of use?

matusmarhefka:
IMO, by "this header" they mean only the `<terms_of_use>` element which links to the full license.

jan-cerny:
@matusmarhefka If that would be true, why would we create this patch at all?

matusmarhefka:
@jan-cerny Exactly that was my point. Do we really need to copy and maintain the whole file with license? I think that only OVAL license header (`<terms_of_use>`) is needed to be in our OVAL schema and that header is referencing the whole license which IMO doesn't need to be copy pasted in our project.

jan-cerny:
@matusmarhefka I have the same question as you.

dahaic:
@matusmarhefka @jan-cerny can we finish (somehow) this PR?

jan-cerny:
@dahaic Oh dear, this is still opened. It was a request by @isimluk . 

Hi @isimluk , could you please shed some light whether this change is what is expected?

isimluk:
I think You are right that just inclusion of the header is required by the header.

Although, I would be slightly more inclined to include the full text. The text online can change or become unavailable. Nevertheless, IANAL.

matejak:
Closing this PR, as the `terms_of_use` text doesn't indicate that a separate file has to be shipped (... When distributing copies of the OVAL Schema, **this** license header must be included.).
Keeping any shipped license in sync with upstream would just impose a burden on project maintainers.

PR #1319, opened 2019-03-29
improve --stig-viewer output when there is no 1:1 connection between rules

maltek:
There can both be rules covering multiple STIG IDs (easy to handle: just
output the same result multiple times), and STIG IDs covered by more
than one rule. The DISA Stig Viewer seems to only consider the last
result for a given rule when it imports results, so when that last rule
happens to pass, that means the whole STIG will be marked as 'not a
finding'. To counteract that, this commit attempts to merge the results
in a sensible way - only if all tests passed, the STIG should be
marked as 'not a finding' in the way.

I'm not attempting to merge anything besides the result state, because the STIG Viewer seems to ignore all that information anyways.

(Note that currently, [ComplianceAsCode/content](https://github.com/ComplianceAsCode/content) doesn't handle these N:N connections very well either - although it does allow them so some extent. I have some patches for that I'll try to upstream there, soon.)

### DIFF
--- a/src/XCCDF/result.c
+++ b/src/XCCDF/result.c
@@ -1065,12 +1065,67 @@ void xccdf_result_to_dom(struct xccdf_result *result, xmlNode *result_node, xmlD
 	}
 	xccdf_setvalue_iterator_free(setvalues);
 
+	struct oscap_htable *nodes_by_rule_id = oscap_htable_new();
+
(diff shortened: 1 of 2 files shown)


jan-cerny:
@openscap-ci test this please

jan-cerny:
Thank you very much for your patch!

PR #1323, opened 2019-04-26
[WIP] Compose SCAP 1.3 DataStreams

yuumasato:
This is a very early and rough patch to update openscap to compose SCAP 1.3 compliant Data Streams.

What this PR does for now:
- When a `check-content-ref`'s `href` attribute is a supported URL (i.e.: `http://` or `https://`)
  - Add a `uri` entry in the `Benchmark`s component-ref catalog, but don't include its contents in DS
  - Add a `component-ref` to datastream's `<ds:checks>`

What it needs to do:
- Get the path from the URL and generate the  component `name` and `uri`.
- `oscap_source_new_from_url` not sure if makes sense to implement it.

Related to https://github.com/ComplianceAsCode/content/pull/4302

### DIFF
--- a/src/DS/sds.c
+++ b/src/DS/sds.c
@@ -828,6 +828,7 @@ static int ds_sds_compose_add_component_dependencies(xmlDocPtr doc, xmlNodePtr d
 		struct oscap_htable *exported = oscap_htable_new();
 		char* filepath_cpy = oscap_strdup(oscap_source_readable_origin(component_source));
 		char *dir = oscap_dirname(filepath_cpy);
+		bool href_is_remote = false;
 
 		for (int i = 0; i < nodeset->nodeNr; i++)
 		{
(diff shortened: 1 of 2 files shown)


src/DS/sds.c
```diff
@@ -988,8 +997,16 @@ static int ds_sds_compose_add_component_source_with_ref(xmlDocPtr doc, xmlNodePt
 	xmlNodePtr cref_parent;
 
 	bool extended_component = false;
+    bool source_is_remote;
 
-	oscap_document_type_t doc_type = oscap_source_get_scap_type(component_source);
+    if (oscap_str_startswith(filepath, "com.")) {
```
jan-cerny: This is suspicious for me. Is it remote when the file path starts with "com."? 

Also, ds_sds_compose_add_component_source_with_ref is called from ds_sds_compose_add_component_with_ref which is called from ds_sds_compose_add_component which is used when composing datastreams.
What if I have a local file `com.example.xml` in current working directory  and I run this:
`oscap ds sds-compose com.example.xml ds.xml`

src/DS/sds.c
```diff
@@ -848,17 +849,25 @@ static int ds_sds_compose_add_component_dependencies(xmlDocPtr doc, xmlNodePtr d
 
 				if (oscap_acquire_url_is_supported(href)) {
 					/* If the referenced component is remote one, do not include
-					 * it within the DataStream. Such component shall only be
-					 * downloaded once the scan is run. */
-					xmlFree(href);
-					continue;
+					* it within the DataStream. Such component shall only be
```
jan-cerny: Does this mean after this change it will get included?

src/DS/sds.c
```diff
@@ -848,17 +849,25 @@ static int ds_sds_compose_add_component_dependencies(xmlDocPtr doc, xmlNodePtr d
...
 				}
 
-				// skip over file:// if it's used in the file href
-				const char *altered_href = oscap_str_startswith(href, "file://") ? href + 7 : href;
+				char *real_path;
+				if (href_is_remote) {
+					// fix real_path for remote source
+					// Update href name to be only remote file (com.redhat.rhsa-RHEL7.xml)
+					// char *real_path = oscap_strdup(href)
+					real_path = oscap_strdup("com.redhat.rhsa-RHEL7.xml");
```
jan-cerny: I don't understand. If content references any remote content from anywhere, will it always get replaced by RHEL 7 RHSA? Hard coding anything here seems suspicious.

jan-cerny:
I think we will need a test for this.

src/DS/sds.c
yuumasato: >  Is it remote when the file path starts with "com."?

No. I just used a hardcoded string, and a dumb match to it just to see how the output would lookl ike. This PR is very rough, take it with a grain of salt, :)

src/DS/sds.c
yuumasato: I was just lazy to implement a URL parser before I could test that this is a good approach.

src/DS/sds.c
yuumasato: Not the contents, just a `component-ref` in the `<ds:checks>` that will reference the remote content, and a `<uri>` in each Benchmark component that uses the remote resource.

evgenyz:
We are not using OpenSCAP to compose DS anymore. Closing.

PR #1326, opened 2019-05-09
Add SCAP 1.3 schema and enable XML validation

jan-cerny:
This pull request adds SCAP 1.3 source datastream XML schema. OASIS XML Catalog schema and WC3 XLink schema, that are shared between SCAP 1.3 and 1.2 are moved to `/schemas/common` directory to avoid duplication. 

So far version 1.2 was hard-coded, therefore I have added a function that detects the version from the datastream using `data-stream-collection@schematron-version` attribute.

Some of the 1.2 datastreams in our test suite have set the version to 1.0, so I fixed this.


src/source/oscap_source.c
```diff
@@ -363,7 +363,7 @@ const char *oscap_source_get_schema_version(struct oscap_source *source)
 		}
 		switch (oscap_source_get_scap_type(source)) {
 			case OSCAP_DOCUMENT_SDS:
-				source->origin.version = oscap_strdup("1.2");
+				source->origin.version = ds_sds_detect_version(reader);
```
matusmarhefka: Can we add a simple test for this? It should be sufficient to use `oscap info` which uses the `oscap_source_get_schema_version` function and call it against prepared content, one with `schematron-version` `1.2 and the other with `1.3` and check if the versions are reported correctly.

matusmarhefka:
LGTM except missing test. SCAP 1.3 schema is aligned with the https://csrc.nist.gov/Projects/Security-Content-Automation-Protocol/SCAP-Releases/SCAP-1-3

matusmarhefka:
@openscap-ci test this please

jan-cerny:
@matusmarhefka I have added a simple test.

jan-cerny:
@matusmarhefka I have added another simple test for validating SCAP 1.3 datastreams.

matusmarhefka:
@jan-cerny Great work, thank you.

PR #1329, opened 2019-05-17
Handle autofs entries in /etc/mtab

jan-cerny:
Some file systems can be mounted using autofs, which should be
considered during analysis of /etc/mtab. For more details,
please see the comment introduced in this patch.


matejak:
Looks great, thank you for the test!

PR #1330, opened 2019-05-17
Check local filesystems when using 'filepath' element

jan-cerny:
The recurse_file_system="local" doesn’t work with file_object and
textfilecontent54_object where <filepath> element is used to define the
file path and operation=”equals” is used on filepath.  As of OVAL
specification of FileBehaviors element:
https://oval.mitre.org/language/version5.10/ovaldefinition/documentation/unix-definitions-schema.html#FileBehaviors
"'recurse_file_system' defines the file system limitation of any
searching and applies to all operations as specified on the path or
filepath entity." Notice the words “any searching” and notice the words
“all operations”. We can solve this bug by performing a similar check
as we do in oval_fts_read_recurse_path.

### DIFF
--- a/src/OVAL/probes/oval_fts.c
+++ b/src/OVAL/probes/oval_fts.c
@@ -1346,6 +1346,12 @@ OVAL_FTSENT *oval_fts_read(OVAL_FTS *ofts)
 		if (ofts->ofts_sfilepath) {
 			fts_ent = ofts->ofts_match_path_fts_ent;
 			ofts->ofts_match_path_fts_ent = NULL;
+			if (ofts->filesystem == OVAL_RECURSE_FS_LOCAL
+				&& (!OVAL_FTS_localp(ofts, fts_ent->fts_path,
+					(fts_ent->fts_statp != NULL) ?
+					&fts_ent->fts_statp->st_dev : NULL))) {
+				continue;
+			}
 			break;
 		} else {
 			fts_ent = oval_fts_read_recurse_path(ofts);


matejak:
test this pls

matejak:
test this please

jan-cerny:
@matejak What's the status of this?

jan-cerny:
@matejak Do you have any update?

jan-cerny:
@matejak Do you have any update?

jan-cerny:
@matejak If you have any justification please close this PR and please put the justification in a comment.

jan-cerny:
@matejak BUMP

jan-cerny:
@matejak Do you have nay updates on this problem?

matejak:
What I don't understand now is that as the PR addresses recursion on `filepath` with `equals`, there is IMO not much recursion to do here.
The referenced OVAL documentation states that recursion applies when `path` is used, and this PR addresses `filepath` specification together with `equals` operation, which has little to do with what I imagine as recursion.
Is the point of this PR to exclude objects that correspond to files existing on remote filesystems?

jan-cerny:
@matejak

> The referenced OVAL documentation states that recursion applies when path is used,

That statement is about the 'max_depth' and 'recurse_direction'  attributes, it doesn't mention 'recurse_file_system' attribute.

> and this PR addresses filepath specification together with equals operation, 

The definition of  recurse_file_system attribute says that it can be used with either path or filepath elements. 

It also says it applies to any operation, which I think also include `equals` operation. For example, it also says that "The value of 'defined' only applies when an equality operation is used", which suggests equals is a valid operation in this context.

> which has little to do with what I imagine as recursion

The spec of the recurse_file_system says it "defines the file system limitation of any searching". My understanding of "any searching" is that it also includes finding files based on specified filepath. I admit it's ambiguous definition and people might have different understanding. You can object that you don't need to "search" anything if you already know the path to it. Understanding this statement properly is the first thing that we need to have.

> Is the point of this PR to exclude objects that correspond to files existing on remote filesystems?

Yes, I think that goal is to not access remote filesystems at all when recurse_file_system is set to "local", even if an exact real path is specified. The same path can be resolved to a local file on one machine and to remote file on other machine. It would give us a way to ensure we never scan remote FSs regardless specific filesystem layout.

matejak:
I was not against, I just looked back after a prolonged period of time, and realized that it doesn't make sense to me.
Thanks for the explanation, I am fine with that now.

evgenyz:
@jan-cerny Why `1.2.18` and `maint-1.2` by the way, do we need it to fix some BZ for 1.2?

jan-cerny:
@evgenyz I am not aware of any BZ that this would fix.  I think this is related to https://bugzilla.redhat.com/show_bug.cgi?id=1655943. It was created as a follow up on https://github.com/OpenSCAP/openscap/pull/1315 and https://github.com/OpenSCAP/openscap/pull/1329. These are in maint-1.2 so I think the reason was to be consistent with them.

jan-cerny:
@openscap-ci test this please

evgenyz:
The BZ is fixed. All related PRs are merged. Do we need this at all?

jan-cerny:
@evgenyz In general, we should comply with the OVAL specification. If there is no BZ it doesn't mean that there isn't a problem.

evgenyz:
Can we have it rebased on maint-1.3? I think that we all agreed on the correctness of the approach and usefulness for the current release.

jan-cerny:
rebased and changed the target to maint-1.3

evgenyz:
Okay, I checked the change locally - everything is OK. Disregard the CI results.

PR #1333, opened 2019-05-21
Fix invalid OVAL in test ds_continue_without_remote_resources

jan-cerny:
Addressing:
```
$ oscap ds sds-split remote_content_1.2.ds.xml /tmp/xxx
$ oscap oval validate --schematron /tmp/xxx/test_single_rule.oval.xml
<?xml version="1.0"?>
oval:x:obj:1 - referenced variable oval:x:var:1 not found. The var_ref entity must hold a variable id that exists in the document.
```


matusmarhefka:
@jan-cerny Thanks for fixing this.

PR #1338, opened 2019-05-23
Extend unit test for is_local_fs from fsdev.h

jan-cerny:
The test uses a fake `mtab` file which contains 4 entries for a local filesystem, 1 entry for a direct autofs map, 1 entry for a NFS system mounted using autofs and 1 entry for a mounted cifs system. By parsing the `mtab` file only 4 local filesystems should be found. It will help us to test https://github.com/OpenSCAP/openscap/pull/1329


jan-cerny:
@openscap-ci test this please

tests/API/probes/fake_mtab
```diff
@@ -0,0 +1,3 @@
+/dev/mapper/fedora-root / ext4 rw,seclabel,relatime 0 0
```
matusmarhefka: @jan-cerny Can we add more entries into this fake mtab? At least 2 more local filesystems (for example proc, sysfs or tmpfs) and one more remote filesystem?

jan-cerny: @matusmarhefka done

matusmarhefka: @jan-cerny Great, thank you.

PR #1341, opened 2019-05-28
Improve SDS detect version test to cover future releases.

ggbecker:
Taking advantage of `oscap info`, we can generate the datastream during runtime as it only needs to have the version set. This way we can extend the test to test future versions of SCAP we add in `schemas/sds`.


ggbecker:
@openscap-ci test this please

tests/DS/sds_detect_version/test_detect_version.sh
```diff
@@ -16,12 +16,21 @@ function test_oscap_info {
...
 	[ ! -s $stderr ]
 	grep -q "Version: $version" $stdout
 	rm $stdout
 	rm $stderr
+	rm $ds
 }
 
-test_oscap_info "1.2"
-test_oscap_info "1.3"
+SDS=$(find $builddir/schemas/sds -maxdepth 1 -mindepth 1 -type d -printf '%f\n')
```
jan-cerny: I think it should be `$top_srcdir` instead of `$builddir`.

ggbecker: I needed to add this variable to the `Makefile.am`, this was the only way to make it work. Is it correct?

jan-cerny: It seems the right way. But please don't rename it, keep `$top_srcdir` to be consistent across the whole project.

tests/DS/sds_detect_version/test_detect_version.sh
```diff
@@ -16,12 +16,21 @@ function test_oscap_info {
 	version="$1"
 	stdout="$(mktemp)"
 	stderr="$(mktemp)"
-	$OSCAP info $srcdir/scap-$version-ds.xml > $stdout 2> $stderr
+	ds="$(mktemp)"
+	cp scap-ds.xml $ds
```
jan-cerny: missing `$srcdir`

tests/DS/sds_detect_version/test_detect_version.sh
ggbecker: Done

tests/DS/sds_detect_version/test_detect_version.sh
ggbecker: Done

jan-cerny:
@openscap-ci test this please

ggbecker:
rhel6 node failed in process58 test ;(

@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
The CI fail is unrelated and has been fixed in https://github.com/OpenSCAP/openscap/pull/1344

PR #1342, opened 2019-05-29
Do not skip 'fs.binfmt_misc.status' by sysctl probe

jan-cerny:
Directory /proc/sys/fs/binfmt_misc is a separate file system,
mounted by systemd-automout and is represented by an autofs
entry in /etc/mtab. /proc/sys/fs/binfmt_misc is mounted
on demand when accessed. If accessed the first time, we don't
have a way to determine the if the autofs is remote or local.
See 309f8230d67f229b6091876c3ace62370fb3d451.
However, we don't see a reason why not to traverse the whole
/proc/sys regardless the type of file systems mounted there.


matejak:
I totally agree. Great catch!

PR #1344, opened 2019-05-30
Update blacklist of kernel params with net.core.bpf_jit_limit

matusmarhefka:
The `net.core.bpf_jit_limit` kernel parameter cannot be set by
non root users and so add it to blacklist in the test.


jan-cerny:
I vote for removing the test `ensure command line is collected`.

jan-cerny:
@openscap-ci test this please

jan-cerny:
@matusmarhefka I like your commit very much. Thank you. I wait for CI if it will not trigger zombies this time.

PR #1347, opened 2019-06-06
Make is_local_fs static again

jan-cerny:
It isn't necessary to expose this function in public API.
The function has been accidentaly introduced to public API
in fff58197d9747a08d0fc23914a31fefbe44f07ea which hasn't
been released yet, so it can be safe to remove it.


matusmarhefka:
LGTM

PR #1350, opened 2019-06-10
OVAL details in HTML report improvements

jan-cerny:
The scope of this PR is smaller than #964 so hopefully it won't get stalled.

This PR adds the following XCCDF HTML report improvements:
* adds support for `extend_definition` element in  "OVAL details" section 
* Fixes misleading `pass` or `fail` in OVAL details section. Instead, it shows real OVAL results.
* Doesn't show OVAL state when no OVAL items were collected.

For more details and explanation, please read the commit messages of every commit.

### DIFF
--- a/xsl/xccdf-report-oval-details.xsl
+++ b/xsl/xccdf-report-oval-details.xsl
@@ -50,7 +50,14 @@ Authors:
     </xsl:apply-templates>
 </xsl:template>
 
-<xsl:template mode='brief' match='ovalres:definition|ovalres:criteria|ovalres:criterion|ovalres:extend_definition'>
+<xsl:template mode='brief' match='ovalres:extend_definition'>
+    <xsl:param name='result'/>
+    <xsl:apply-templates select='key("oval-definition", @definition_ref)' mode='brief'>
+        <xsl:with-param name='result' select='$result'/>
+    </xsl:apply-templates>
+</xsl:template>
+
+<xsl:template mode='brief' match='ovalres:definition|ovalres:criteria|ovalres:criterion'>
     <xsl:param name='result'/>
     <xsl:apply-templates select='key("oval-test", @test_ref)' mode='brief'>
         <xsl:with-param name='title' select='key("oval-testdef", @test_ref)/@comment'/>
--- a/xsl/xccdf-report-oval-details.xsl
+++ b/xsl/xccdf-report-oval-details.xsl
@@ -123,7 +123,6 @@ Authors:
             <xsl:variable name='object_id' select='key("oval-testdef", @test_id)/*[local-name()="object"]/@object_ref'/>
             <xsl:variable name='object_info' select='key("oval-objectdef",$object_id)'/>
             <xsl:variable name='state_id' select='key("oval-testdef", @test_id)/*[local-name()="state"]/@state_ref'/>
-            <xsl:variable name='state_info' select='key("oval-statedef",$state_id)'/>
             <xsl:variable name='comment' select='$object_info[1]/@comment'/>
             <xsl:if test="$object_info">
                 <h4>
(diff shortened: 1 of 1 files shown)


matejak:
This is a great improvement - however, I suggest to make it more explicit that what we see in OVAL details section are all test results - i.e. AND/OR/negate operations are not taken into account.
This came to my mind:

- Include test IDs next to those comments, so it is obvious that those are test results, not criteria.
- Tell the reader that what they see are linearized test results.

shawndwells:
Could a screen shot be included to show what the new changes look like?

jan-cerny:
@shawndwells Yes, it could. This is an example with @matejak 's feedback already incorporated.

Screenshot_2019-06-11 xccdf_org open-scap_testresult_xccdf_org ssgproject content_profile_ospp OpenSCAP Evaluation Report(1)

matejak:
Before:
image
After:
image

matejak:
LGTM, thank you!

shawndwells:
On 6/11/19 4:26 AM, Jan Černý wrote:
>
> @shawndwells <https://github.com/shawndwells> Yes, it could. This is 
> an example with @matejak <https://github.com/matejak> 's feedback 
> already incorporated.
>
> Screenshot_2019-06-11 xccdf_org open-scap_testresult_xccdf_org 
> ssgproject content_profile_ospp OpenSCAP Evaluation Report(1) 
> <https://user-images.githubusercontent.com/9050916/59256030-5cd20800-8c33-11e9-86b8-6d27065fd383.png>
>


The true/false is extremely confusing.

Could this be replaced by pass/fail?

People won't know what true/false means.... how do they know when 
something should be true versus false? Which one means that a check 
"passed" or "failed"?

matejak:
This is supposed to be like a traceback - it is useful, but not so user-friendly. However, it is a step ahead in terms of user-friendliness, so this is why it got merged.

What you see are results of tests. Whether a rule passes or not is determined by the criteria tree, which is much more difficult to present in the report. You are not able to tell why exactly a rule has failed, unless you map test results to criteria with all of those AND/OR + negate logic.
We have some progress on that on multiple fronts, but the idea of this PR was to introduce the work gradually.

jan-cerny:
@shawndwells 

The pass/fail in OVAL results was even more confusing. Let me explain it.

First, in this part of HTML report results of individual OVAL tests are presented (and always were). As of [OVAL Results Schema](https://oval.mitre.org/language/version5.11/ovaldir/documentation/oval-results-schema.html#ResultEnumeration), result of an OVAL test can be: "true", "false", "unknown", "error", "not evaluated", "not applicable". All of these values have an exactly defined meaning. There is no "pass" or "fail". Theoretically, we could say that we will rename "true" to "pass" for the purposes of the HTML report, but then it would have to explain what we actually mean by that. It is not clear to me if  "fail" would be then defined as a negation of "pass" or if it would correspond to "false" in OVAL.

Second, the way the HTML report worked before this pull request was logically broken. It worked this way: It took the result of the XCCDF rule (the final rule result).  Then it propagated this value to the test results and displayed it as a result of each of the OVAL tests. If the XCCDF result was "pass" it used "passed" everywhere and if the result was anything else it used "failed" everywhere, regardless of the actual result of the specific OVAL test. That lead to weird situation like in this example obtained using OpenSCAP 1.2.17 on RHEL 7. Notice it says that the test "TMOUT in /etc/profile" passed because there is no TMOUT in /etc/profile. If user takes a look at the definition of this OVAL test, he will be confused because that test requires that TMOUT must be present in /etc/profile, so it shouldn't pass when the string isn't present in the file. And if he checks the XML results he founds out that this test was evaluated as "false", yet the report shows that it passed.
Screenshot_2019-06-12 xccdf_org open-scap_testresult_xccdf_org ssgproject content_profile_ospp42 OpenSCAP Evaluation Report


In my opinion, the main drawback of this section of HTML report, is that it doesn't show the relationship between the results of individual OVAL tests and the final result of the XCCDF rule. It isn't visible how the final results have been computed by OpenSCAP. However, it never was, except the most simple rules which had only one OVAL test with a single OVAL object and no state. I didn't attempt to fix this problem in this PR because it's more complex. Probably some sort of tree visualization would be a solution.

I agree with you that "people won't know what true/false means", but now we can point them to OVAL specification where it's clearly stated. Before this PR we had to explain them quirks of our XSLT template implementation, which was hard to understand.

PR #1354, opened 2019-06-12
Update versions.sh in maint-1.2

jan-cerny:
Future version of OpenSCAP is 1.2.18. Latest released version is 1.2.17.
Latest supported Fedora is 31 (current Rawhide). Latest RHEL is 8.


yuumasato:
@openscap-ci test this please

PR #1355, opened 2019-06-12
Disable fallthrough warnings in switch

mildas:
Build produces a lot of warnings about fallthrough, e.g.
```
MurmurHash3.c:132:14: warning: this statement may fall through [-Wimplicit-fallthrough=]
   case 3: k1 ^= tail[2] << 16;
           ~~~^~~~~~~~~~~~~~~~
```
but it is intended. `#pragma GCC diagnostic` disables these warning messages.

### DIFF
--- a/src/OVAL/probes/SEAP/MurmurHash3.c
+++ b/src/OVAL/probes/SEAP/MurmurHash3.c
@@ -127,6 +127,7 @@ void MurmurHash3_x86_32 ( const void * key, int len,
 
   uint32_t k1 = 0;
 
+#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
   switch(len & 3)
@@ -137,6 +138,7 @@ void MurmurHash3_x86_32 ( const void * key, int len,
           k1 *= c1; k1 = ROTL32(k1,15); k1 *= c2; h1 ^= k1;
   };
 #pragma GCC diagnostic pop
+#endif
 
   //----------
   // finalization
@@ -205,6 +207,7 @@ void MurmurHash3_x86_128 ( const void * key, const int len,
   uint32_t k3 = 0;
   uint32_t k4 = 0;
 
+#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
   switch(len & 15)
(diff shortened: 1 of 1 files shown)


evgenyz:
Hmm, maybe it'd be better to use `/* fall through */` or ` __attribute__ ((fallthrough));` to mark exact points where falling through is intended?

mildas:
Yes, first I was thinking about `__attribute__((fallthrough));`, but we'd need it for every `case` and I'd say it's more readable to have whole block in `#pragma`

jan-cerny:
@mildas on RHEL 6  node: https://jenkins.complianceascode.io/job/openscap-maint-1.2-pull-requests-multi/151/label=rhel6/consoleText
```

../../../../../src/OVAL/probes/SEAP/MurmurHash3.c: In function ‘MurmurHash3_x86_32’:
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c:130: error: #pragma GCC diagnostic not allowed inside functions
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c:131: error: #pragma GCC diagnostic not allowed inside functions
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c:139: error: #pragma GCC diagnostic not allowed inside functions
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c: In function ‘MurmurHash3_x86_128’:
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c:208: error: #pragma GCC diagnostic not allowed inside functions
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c:209: error: #pragma GCC diagnostic not allowed inside functions
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c:235: error: #pragma GCC diagnostic not allowed inside functions
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c: In function ‘MurmurHash3_x64_128’:
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c:300: error: #pragma GCC diagnostic not allowed inside functions
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c:301: error: #pragma GCC diagnostic not allowed inside functions
../../../../../src/OVAL/probes/SEAP/MurmurHash3.c:323: error: #pragma GCC diagnostic not allowed inside functions
make[6]: *** [libseap_la-MurmurHash3.lo] Error 1
```
It seems that RHEL 6 has an older version of GCC which doesn't support them.
I think you need to guard your pragmas by preprocessor directives.
eg.
```
#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
#endif
```

mildas:
@jan-cerny thank you for pointing out the GCC version.
Updated it with >= GCC 4.6 version, where the `#pragma GCC diagnostic` was added. (https://gcc.gnu.org/gcc-4.6/changes.html)

jan-cerny:
@mildas You're great!

PR #1361, opened 2019-06-17
Change creating of zombie process and waiting for processes at process58 test

mildas:
The old creating zombie process did not work properly (seems like there were some changes in planning/stopping processes) and it caused problems with running tests (it was randomly passing/failing/waiting because of the test).

The commit also changes waiting for process, because the name of created zombie process isn't `stopped_process.sh`, but `command_line.sh` (line [98](https://github.com/mildas/openscap/blob/a923e01a29df6e051d9b8a0dfa2336084adb1829/tests/probes/process58/command_line.sh#L98)), so it was waiting 30 seconds for `stopped_process.sh` zombie, but it never appears.

### DIFF
--- a/tests/probes/process58/command_line.sh
+++ b/tests/probes/process58/command_line.sh
@@ -24,10 +24,14 @@ function get_process_cmdline() {
 # We can scan processes before process's exec() and get wrong command_line
 function wait_for_process() {
 	local PID="$1"
+	local PROC_NAME="$2"
+	if [[ -z "$PROC_NAME" ]]; then
+		PROC_NAME="$PROC"
+	fi
 	for i in $(seq 1 100); # wait max 100 * 300ms
 		do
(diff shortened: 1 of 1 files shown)


mildas:
Looks like sometimes subshell dies before killing parent shell, then it won't create zombie process. I added `sleep 1` to subshell so there's time for dying parent shell and then subshell becomes zombie.

jan-cerny:
@mildas Now it passed on all 4 nodes.

jan-cerny:
@openscap-ci test this please

jan-cerny:
@mildas Now it passed on all 4 nodes.

jan-cerny:
@openscap-ci test this please

jan-cerny:
@mildas  It passed again. Let's try to merge it.

PR #1363, opened 2019-06-18
Include additional headers in dkpginfo-helper for apt 1.9

julian-klode:
apt 1.9 cleans up the header includes a bit, causing the build
to fail, so add the additional headers we are missing.


ggbecker:
@openscap-ci test this please

jan-cerny:
Hi @julian-klode , are these header files present also in previous versions of the apt library? In other words, will it build with the older versions of library?

julian-klode:
Sure, I guess they're probably a decade old or something

PR #1364, opened 2019-06-18
Fix (null) string when XCCDF value not set

teselkin:
Similar to 793475c2dfddf4d122722f097b99b5e0d0ca48ce assign an
emtpy string to XCCDF value if xccdf_value_binding_get_value
returns NULL.


jan-cerny:
Thanks!

PR #1370, opened 2019-06-26
Fix #1173, RHBZ#1603347: Double chdir/chroot in probe rpmverifypackage

### LINKED ISSUE #1173: Bug: double chdir/chroot in probe rpmverifypackage

evgenyz:
The probe is using chroot in the init function, where all the packages
information is being collected. So, the comment is actually misleading
everybody and the probe should be switched to the custom offline mode
right away.

This will fix #1173, [RHBZ#1603347](https://bugzilla.redhat.com/show_bug.cgi?id=1603347).

I used the content attached to the RHBZ ticket to check this patch. I ran all the test from this file (not just 6, as in example) and all where successful (except for 8, but that is expected I suppose).

### DIFF
--- a/src/OVAL/probes/unix/linux/rpmverifypackage_probe.c
+++ b/src/OVAL/probes/unix/linux/rpmverifypackage_probe.c
@@ -314,8 +314,7 @@ ret:
 
 int rpmverifypackage_probe_offline_mode_supported()
 {
-	// TODO: Switch this to OFFLINE_MODE_OWN once rpmtsSetRootDir is fully supported by librpm
-	return PROBE_OFFLINE_CHROOT;
+	return PROBE_OFFLINE_OWN;
 }
 
 void *rpmverifypackage_probe_init(void)


evgenyz:
@openscap-ci test this please

jan-cerny:
@evgenyz I think that this will confuse future developers, because based on PROBE_OFFLINE_OWN they will think that the probe uses some sort of prefix prepending like other probes that use PROBE_OFFLINE_OWN do.

I think the correct fix is to not use CHROOT_.* macros and  probe_chroot_.* functions in the probe code. The CHROOT_.* macros and  probe_chroot_.* functions duplicate the code from the generic probe code, which is executed in every probe that declares PROBE_OFFLINE_CHROOT. See `src/OVAL/probes/probe/probe_main.c` around line 279 and `src/OVAL/probes/probe/worker.c` around line 1095, you will see that `chroot()` is called in `probe_common_main()` and later the `chroot()` is called in `probe_chroot_enter()` again, so the `chroot()` is called twice.

I suggest completely removing `src/OVAL/probes/unix/linux/probe-chroot.c` `src/OVAL/probes/unix/linux/probe-chroot.h` and all their usages. They're used only in the rpmverifypackage probe anyway. There is no reason why the rpmverifypackage  should use a custom code. By removing these you will make it consistent with the other probes. 

Regarding the comment you touched, please see https://bugzilla.redhat.com/show_bug.cgi?id=1566985. After you remove all the code you can start implementing the OWN offline mode and you can investigate if this bug doesn't affect the rpmverifypackage probe. That can be done in a separate PR.

evgenyz:
Yes, chroot handling is inconsistent across the probes. And there are a lot of ad-hoc style code pieces that ensures library preloading and stuff. This should be refactored, of course. And I actually started this process, but then I realized that it's 2AM and this refactoring process is not a part of the bug.

This patch is exactly the fix for the RHBZ ticket. And semantically it is correct because the probe is implementing custom solution for jailing itself, it does not require global chroot call.

Regarding the removal of the chroot call from inside the probe I have completely opposite opinion. And here is why: if you use chroot call outside the thread you won't be able to use unshare mechanism to evade global impact. I actually thought about removing all chroot calls from probe init (as it is called from the main thread) and let first worker to fill the RPM info database (with help of a mutex). Or even change the chroot approach to rpmtsSetRootDir. And then do this for all chroot probes.

evgenyz:
But anyway, this refactoring effort would be big for sure. And I predict a lot of discussions about it. So, I just wanted the bug to be fixed right away, in a simplest manner.

matusmarhefka:
I agree with @evgenyz here, IMO it is better to fix the probe with this one line change for now and only go for the proper fix of the offline chroot mode when we agree on the approach which will work and will be consistent for all the probes.

@jan-cerny If you are not strictly against it I would merge this.

jan-cerny:
@matusmarhefka I'm not strictly against it

jan-cerny:
I only consider this a very temporary solution.

matusmarhefka:
@jan-cerny Yes it is, but it is still better than the original behavior when `oscap` exited with error as `chroot` was called twice.

PR #1376, opened 2019-07-10
Prevent broken output in "--progress" mode

jan-cerny:
In "--progress" mode we print out the rule id and ': ' before the rule
evaluation starts and the rest of the line after the rule evaluation
finishes. If any warning occurs during the evaluation, it's printed in
the middle, which looks ugly. We can fix this by printing the
whole line at the same moment, after the rule evaluation finishes.
Also, the warnings should use dW macros in order to be able to turn them
on/off using `--verbose` option.
Related to https://bugzilla.redhat.com/show_bug.cgi?id=1640889

### DIFF
--- a/tests/API/XCCDF/unittests/test_xccdf_check_processing_selector_empty.sh
+++ b/tests/API/XCCDF/unittests/test_xccdf_check_processing_selector_empty.sh
@@ -12,7 +12,7 @@ $OSCAP xccdf eval --profile xccdf_moc.elpmaxe.www_profile_1 --results $result $s
 
 echo "Stderr file = $stderr"
 echo "Result file = $result"
-[ -f $stderr ]; [ ! -s $stderr ]; rm $stderr
+rm $stderr
 
 $OSCAP xccdf validate-xml $result
 
--- a/tests/API/XCCDF/unittests/test_xccdf_check_processing_selector_empty.sh
+++ b/tests/API/XCCDF/unittests/test_xccdf_check_processing_selector_empty.sh
@@ -12,6 +12,7 @@ $OSCAP xccdf eval --profile xccdf_moc.elpmaxe.www_profile_1 --results $result $s
 
 echo "Stderr file = $stderr"
 echo "Result file = $result"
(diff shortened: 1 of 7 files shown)


tests/API/XCCDF/unittests/test_xccdf_check_processing_selector_empty.sh
```diff
@@ -12,7 +12,7 @@ $OSCAP xccdf eval --profile xccdf_moc.elpmaxe.www_profile_1 --results $result $s
 
 echo "Stderr file = $stderr"
 echo "Result file = $result"
-[ -f $stderr ]; [ ! -s $stderr ]; rm $stderr
+rm $stderr
```
matusmarhefka: So now the `stderr` should contain the warning message right? What about greping it similarly as it is done in the `tests/API/XCCDF/unittests/test_xccdf_multiple_testresults.sh`?

tests/API/XCCDF/unittests/test_xccdf_check_unsupported_check_system.sh
```diff
@@ -8,7 +8,7 @@ stderr=`mktemp`
 
 $OSCAP xccdf eval --results $result $srcdir/test_xccdf_check_unsupported_check_system.xml 2> $stderr
 echo "Stderr file = $stderr"
-[ -f $stderr ]; [ ! -s $stderr ]; rm $stderr
+rm $stderr
```
matusmarhefka: The same, we can grep the warning message here.

tests/API/XCCDF/unittests/test_xccdf_check_without_content_refs.sh
```diff
@@ -10,7 +10,8 @@ $OSCAP xccdf eval --results $result $srcdir/test_xccdf_check_without_content_ref
 
 echo "Stderr file = $stderr"
 echo "Result file = $result"
-[ -f $stderr ]; [ ! -s $stderr ]; rm $stderr
+[ -f $stderr ]
```
matusmarhefka: And here too.

jan-cerny:
@matusmarhefka updated

PR #1377, opened 2019-07-10
Show which OVAL object isn't supported

jan-cerny:
When a probe that can evaluate the given OVAL object isn't present
it will tell user which specific type of OVAL object isn't supported
Related to:
https://bugzilla.redhat.com/show_bug.cgi?id=1640889


yuumasato:
LGTM.

PR #1383, opened 2019-07-19
fixing accessibility of html report

vojtapolasek:
When I opened the detail of a certain rule in the report, the popup dialog was not visible for the screenreader (Orca). Therefore I could not read it. This fixes it.
replaced aria-hidden="true" with "false" in openscap.js
regenerated the xsl template


yuumasato:
@openscap-ci add to whitelist

yuumasato:
Nice!

PR #1384, opened 2019-07-23
Fix missing filepaths in rpminfo items

jan-cerny:
When the queried RPM package has null epoch (or "(none)" as returned by "rpm -q --qf '%{EPOCH}\n'" command), the rpminfo_item produced by the probe didn't contain any filepath element. To use the null epoch as a parameter of rpmdbSetIteratorRE, the "(none)" string should be replaced by "0".

I have a reproducer on Fedora 30: [rpminfo_reproducer.zip](https://github.com/OpenSCAP/openscap/files/3421607/rpminfo_reproducer.zip) The rpminfo_item in OVAL results doesn't contain any filepath.


matejak:
Great catch, thank you for the PR!

PR #1387, opened 2019-07-25
Dpkg info offline mode, fixes

### LINKED ISSUE #1367: dpkginfo_free_reply should return NULL
- Is this an issue with SCAP Workbench?
    - don't know
- Is this an issue with SCAP Security Guide (i.e., related to the content of scans, not the scanner proper)?
    - don't know
- Is this an issue during the OS installation process?
    - don't know

Thanks!

#### Description of Problem:
When run `oscap oval collect --id ID OVAL.XML`, prog will crash. By debugging, I found this function messed the caller's stack because it doesn't return value as its function signature declared.

#### OpenSCAP Version:
git clone in 6/20/2019

#### Steps to Reproduce:

1. oscap oval collect --id ID OVAL.XML
   where id is oval:com.ubuntu.disco:obj:2017111090000000
2.

#### Actual Results:
crash

#### Expected Results:
not crash

n/a

evgenyz:
Adding PROBE_OFFLINE_OWN support was very easy. But still there is a problem here: I'm unable to check any pkgcache.bin, be it a mounted file from VM or local file without any chroot shenanigans. The `new pkgCache (dpkg_mmap);` call always returns error 'The package cache file is corrupted'.

I'll just leave it here in case someone could shed some light on the matters.

This PR will also fix #1367

### DIFF
--- a/src/OVAL/probes/probe-table.c
+++ b/src/OVAL/probes/probe-table.c
@@ -246,7 +246,7 @@ static const probe_table_entry_t probe_table[] = {
 	{OVAL_INDEPENDENT_XML_FILE_CONTENT, xmlfilecontent_probe_init, xmlfilecontent_probe_main, xmlfilecontent_probe_fini, xmlfilecontent_probe_offline_mode_supported},
 #endif
 #ifdef OPENSCAP_PROBE_LINUX_DPKGINFO
-	{OVAL_LINUX_DPKG_INFO, dpkginfo_probe_init, dpkginfo_probe_main, NULL, NULL},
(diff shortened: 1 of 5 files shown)


evgenyz:
BTW, @pschneiders can you look at this PR as well? You might be interested in the fix commit at least.

pschneiders:
Currently this PR branch is the only one that builds and runs cleanly for me on Debian.  It fixes the errors in my ComplianceAsCode PR and compiles without changes due to the CAP_LIBRARIES fix.  Let me know if I can do anything specific to verify the specific intentions of this patch.

pschneiders:
@sbjornaas777 can you assist with this

evgenyz:
OK. We will iron out offline mode later, with specific test.

PR #1388, opened 2019-07-30
Fix URL link mechanism for Red Hat Errata

isimluk:
Previously, public description of each erratum was available under

    https://rhn.redhat.com/errata/

at some point in time this has changed to

    https://access.redhat.com/errata/

and since not all links redirect properly from old to new, we have to fix
OpenSCAP end as well.

This enables people reviewing XCCDF policy guide or XCCDF result report to click
on RH[SBE]A-YYYY-ABCD acronym and be redirected to public description of the
given erratum.

# Before
scrot-before-oscap-fix

# After
scrot-after-oscap-fix

### DIFF
--- a/xsl/oval-to-xccdf.xsl
+++ b/xsl/oval-to-xccdf.xsl
@@ -86,9 +86,9 @@ Authors:
     </xsl:template>
 
     <xsl:template match="oval-def:reference[@source='RHSA']">
-        <xsl:if test="starts-with(@ref_url, 'https://rhn.redhat.com/errata')">
-            <xccdf:ident system="https://rhn.redhat.com/errata">
-                <xsl:copy-of select="substring-before(substring-after(@ref_url, 'https://rhn.redhat.com/errata/'), '.html')"/>
+        <xsl:if test="starts-with(@ref_url, 'https://access.redhat.com/errata')">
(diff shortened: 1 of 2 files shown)


jan-cerny:
@isimluk 

I don't know if I use the templates correctly. But I have run the following command:
```
wget https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL8.xml

xsltproc --stringparam ovalfile "./com.redhat.rhsa-RHEL8.xml"  -o xccdf.xml ~/openscap/xsl/oval-to-xccdf.xsl com.redhat.rhsa-RHEL8.xml

~/openscap/build/oscap_wrapper xccdf eval --report report.html xccdf.xml
```
The result report contain broken links to RHSAs. There is a missing slash. For example I have https://access.redhat.com/errataRHSA-2019:0966.html which leads to an error 404 instead of https://access.redhat.com/errata/RHSA-2019:0966.html for `RHSA-2019:0966: firefox security update (Critical)`.

Could you please take a look at it?

isimluk:
Nice catch! Let me take a look.

jan-cerny:
I have re-executed the workflow from my previous comment. I get correct links in the HTML report now. Thank you very much for your patch.

PR #1389, opened 2019-07-30
Remove duplicate case in a switch (xslt:choose)

isimluk:
Interestingly, I overlooked this years ago and nobody noticed since then.

Please also review following funny animation, it is supposed to make you smile.


yuumasato:
@isimluk revisiting code?

Thanks for the fix.

PR #1390, opened 2019-08-02
Fix --fix-type anaconda

dahaic:
`oscap xccdf generate fix --fix-type anaconda` was failing, because of wrong template associated with the fix type.


matusmarhefka:
LGTM

PR #1392, opened 2019-08-09
Fix typo in manpage and add missing option help usage documentation.

ggbecker:
- Fix typo in oscap manpage.
- Add missing option from oval eval help usage documentation (--fetch-remote-resources)


jan-cerny:
Nice catch! Thank you for this PR.

PR #1393, opened 2019-08-12
Offline mode support for environmentvariable58_probe

evgenyz:
I have some doubts about importing variables from containers as they are, without prefixing, but I can't put my finger on any case in particular.

### DIFF
--- a/utils/oscap-podman
+++ b/utils/oscap-podman
@@ -79,6 +79,10 @@ else
 fi
 DIR=$(podman mount $ID) || die
 
+for VAR in `podman inspect $ID --format '{{join .Config.Env " "}}'`; do
+    eval "export $VAR"
+done
+
 export OSCAP_PROBE_ROOT="$(cd "$DIR"; pwd)"
 export OSCAP_PROBE_OS_NAME="Linux"
 export OSCAP_PROBE_OS_VERSION="$(uname --kernel-release)"
--- a/utils/oscap-podman
+++ b/utils/oscap-podman
@@ -80,7 +80,7 @@ fi
 DIR=$(podman mount $ID) || die
 
 for VAR in `podman inspect $ID --format '{{join .Config.Env " "}}'`; do
-    eval "export $VAR"
+    eval "export OSCAP_OFFLINE_$VAR"
 done
 
 export OSCAP_PROBE_ROOT="$(cd "$DIR"; pwd)"
--- a/utils/oscap-podman
+++ b/utils/oscap-podman
@@ -28,13 +28,15 @@ function usage()
     echo "oscap-podman -- Tool for SCAP evaluation of Podman images and containers."
     echo
     echo "Compliance scan of Podman image:"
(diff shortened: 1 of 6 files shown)


utils/oscap-podman
```diff
@@ -74,14 +79,18 @@ else
 fi
 DIR=$(podman mount $ID) || die
 
+for VAR in `podman inspect $ID --format '{{join .Config.Env " "}}'`; do
+    eval "export $VAR"
```
jan-cerny: Can a container contain an environment variable that affects the `oscap` execution?

evgenyz: It was my initial concern.

In theory — it could. But I don't think it's different from environment affecting on-line scan. We are talking about situations where customer can't check JBOSS container because there is a check against JAVA_HOME variable (or something like that, can't find the bug right now).

Anyway, I can add all these VARs with some prefix (like OFFLINE_) and use this prefix in the probe. But I'm not sure if 1) this is really needed, and 2) it won't affect some content functionality (as names would be prefixed).

evgenyz: The bug: https://bugzilla.redhat.com/show_bug.cgi?id=1493614

jan-cerny: @evgenyz Probably it is OK then, let's not do anything now.

jan-cerny:
@evgenyz I have tried to run it and I have a problem that I haven't realized before:  If I write OVAL to collect every variable such as [environmentvariable58.oval.xml.zip](https://github.com/OpenSCAP/openscap/files/3496661/environmentvariable58.oval.xml.zip)  the collected results contain also the variables from the host, including the variables like SUDO_USER. I think it isn't acceptable to include the variables from the host in the collected results, because we scan only the image/container as a separate unit.

If there would be an SCAP rule that a variable must bet set, you could fake the results by setting the variable on the host, which probably isn't what we want.

evgenyz:
@jan-cerny Good catch! Then I definitely have to use some prefix. I'll change the patch.

jan-cerny:
@evgenyz OK, makes sense, sorry for that.

jan-cerny:
@evgenyz 
There is an interesting inconsistency when scanning podman images:
When I run `podman inspect fedora` I get the following variables: PATH, DISTTAG, FGC, FBR.
But if I run `oscap-podman` scan using the OVAL file attached in previous comment, the OVAL results file contains the following variables: TERM, PATH, FGC, FBR, DISTTAG, HOSTNAME. I would expect it to be consistent. It's because we don't scan the image directly, but we create a container using `podman create` and then mount it. If you run `podman create` and then `podman inspect` on the created image, you will get the same list as OpenSCAP produces.

In general, it's us who decide what will be the behavior of offline mode. But I don't know if it's OK or not. What is your opinion on this problem?

evgenyz:
I think that as we are trying to verify not just the configuration, but the state of to-be-running container (as it will be used by the end user) we should stick with current behaviour.
This behaviour is consistent with `podman inspect <container>` and should be logically expected.

We could have a problem in the future when we will be able to scan the image without instantiating a container, but I think we should cross that bridge when (and if) we will be there.

jan-cerny:
@evgenyz makes sense

PR #1396, opened 2019-08-15
fetch-remote-resources can be used only together with datastreams

isimluk:
Similarly to oval-id and datastream-id.

### DIFF
--- a/utils/oscap-oval.c
+++ b/utils/oscap-oval.c
@@ -113,7 +113,8 @@ static struct oscap_module OVAL_EVAL = {
 	"                                   (only applicable for source datastreams)\n"
 	"   --oval-id <id>                - ID of the OVAL component ref in the datastream to use.\n"
 	"                                   (only applicable for source datastreams)\n"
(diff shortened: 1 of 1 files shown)


utils/oscap-oval.c
```diff
@@ -113,7 +113,8 @@ static struct oscap_module OVAL_EVAL = {
 	"                                   (only applicable for source datastreams)\n"
 	"   --oval-id <id>                - ID of the OVAL component ref in the datastream to use.\n"
 	"                                   (only applicable for source datastreams)\n"
-	"   --fetch-remote-resources      - Download remote content referenced by OVAL Definitions.\n",
+	"   --fetch-remote-resources      - Download remote content referenced by OVAL Definitions.\n"
+  "                                   (only applicable for source datastreams)\n",
```
evgenyz: Please, stick to the original indentation style. This is even more important in blocks like this.

isimluk: Ha, too much time spent in ruby, rust, go makes you forget the tab key. Sorry. will fix.

evgenyz:
Great! Thank you!

PR #1397, opened 2019-08-20
Remove attributes that are not needed from our embedded OVAL

isimluk:
Introduced by 90e701800afbc4b255ddaf6c33c035883d0714c0, but was always overabundant.

¯\\_(ツ)_/¯


jan-cerny:
@isimluk Are they really superfluous? From the output of Jenkins fail it seems it doesn't allow prefix matches anymore:

```
TEST: test_api_xccdf_default_cpe_rhel
System CPE: cpe:/o:redhat:enterprise_linux:8.0:ga
Tested CPE: cpe:/o:redhat:enterprise_linux:8
Title
	
Rule
	xccdf_moc.elpmaxe.www_rule_1
Result
	notapplicable

XCCDF Results are exported correctly.
RESULT: FAILED
```
Could you please take a look into this?

isimluk:
You are right. Luckily, we have those tests. Sorry for the noise.

PR #1401, opened 2019-09-19
Add new STIG Viewer URI public.cyber.mil and be backward compatible

abergmann:
As described over at ComplianceAsCode/content [issue#4846](https://github.com/ComplianceAsCode/content/issues/4846) the STIG Viewer URI change is causing problems with newly created security benchmarks and the ```--stig-viewer``` output.

This PR adds the new URI ```https://public.cyber.mil/stigs/srg-stig-tools/``` as an additional identifier. The olf URI is still in place to be backward compatible with existing security benchmarks.

### DIFF
--- a/src/XCCDF/result.c
+++ b/src/XCCDF/result.c
@@ -70,7 +70,8 @@
 #define XCCDF_NUMERIC_SIZE 32
 
 // References containing STIG Rule IDs can be found by their href attribute, it must match the following url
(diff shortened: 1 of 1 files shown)


yuumasato:
Thanks for the fix!

I think this should go to maint-1.2 branch as well.
@abergmann Would you be willing to do back port the patch into maint-1.2?
Due to #1319 the changes will not apply smoothly.

PR #1402, opened 2019-09-19
SWIG openscap_api: Use __repr__() to get consistent string output

yuumasato:
Lets use `__repr__()` to infer object structure, instead of `__str__()`.

`__str__()` returns different strings on different python versions

On python2, it returns something like
`'_e016160100000000_p_xccdf_result_iterator'`
On python3, it returns something like
`'<Swig Object of type 'struct xccdf_result_iterator *' at 0x7fc81474adb0>'`

`__repr__()` returns the second format on both python versions.

Fixes traceback where it tried to use a strange function:
```
  File "get_results.py", line 17, in <module>
    results = benchmark.get_results()
  File "/usr/lib64/python3.6/site-packages/openscap_api.py", line 196, in __getter_wrapper
    list.generate(OSCAP_Object.new(retobj))
  File "/usr/lib64/python3.6/site-packages/openscap_api.py", line 89, in generate
    while iterator.has_more():
  File "/usr/lib64/python3.6/site-packages/openscap_api.py", line 277, in __call__
    raise NameError("name '"+self.object+"' is not defined")
NameError: name 'wig Object of type 'struct xccdf_result_iterator *' at 0x7f7f7dc20090>_has_more' is not defined
Exception ignored in: <bound method OSCAP_List.__del__ of []>
  File "/usr/lib64/python3.6/site-packages/openscap_api.py", line 81, in __del__
    self.iterator.free()
  File "/usr/lib64/python3.6/site-packages/openscap_api.py", line 314, in free
    raise Exception("Can't free %s" % (self.object,))
Exception: Can't free wig Object of type 'struct xccdf_result_iterator *' at 0x7f7f7dc20090>
```

### DIFF
--- a/swig/openscap_api.py
+++ b/swig/openscap_api.py
@@ -27,6 +27,7 @@ __author__ = 'Maros Barabas'
 __version__ = '1.0'
 
 import logging                  # Logger for debug/info/error messages
+import re
 logger = logging.getLogger("openscap")
 
 from sys import version_info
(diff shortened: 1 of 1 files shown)


yuumasato:
Does it make sense for it to fail on `filehash58` on fedora?
@openscap-ci test this please

jan-cerny:
@yuumasato no fails please :smiley:

yuumasato:
@jan-cerny There was some infrastructure problem.

PR #1404, opened 2019-09-23
Add new STIG Viewer URI public.cyber.mil and be backward compatible in maint-1.2

abergmann:
As described in [PR#1401](https://github.com/OpenSCAP/openscap/pull/1401) the STIG Viewer URI change is causing problems with newly created security benchmarks and the ```--stig-viewer``` output.

This PR is addressing the issue for the 1.2 branch.


yuumasato:
@abergmann Thank you very much!

PR #1408, opened 2019-10-16
oscap-docker without atomic dependency

DominiqueDevinci:
Here is my proposal of new oscap-docker utility.
The existing one rely on atomic which works well only on redhat based systems and which is furthermore abandoned. This new tool is able to perform exactly same task, but uniquely rely on the native docker api (also used by current oscap-docker). The only disadvantage is you need to run this utility with root privileges (because the mounted docker filesystem is located in /proc/container_pid/root ). If it's really a problem, it's possible to figure that by copying the container first, but it's a lot longer and disk/cpu consumming.

demo : https://pastebin.com/AynvWfPC (error handling)
https://pastebin.com/vFsrrTzx (image-cve scanning). also support scan args for --report etc.
https://pastebin.com/cXAFeNKA (container scan of a debian8 img).

Quote from the updated doc : 
When `oscap-docker` scans an image, it runs it before in a temporary container
which will be cleaned at the end. If you scan a running container, it's directly
scanned, and if it's not running, `oscap-docker` will run it temporarily 
for you, before stop it at the end. In some case, if a container doesn't 
keep running (for instance because its command is `echo "hello world"` instead 
of "/bin/bash", it's not possible to scan it, but you still can perform 
your scan on the initial container image.


DominiqueDevinci:
@jan-cerny  I was planning to support cve scanning for other platforms, but i fear that oscap-docker isn't the right place for this ... do you not think that it's should be better to do that in a separate util ? 

For instance, docker-scapcontent (or a name like this) could retrieve scap content (which can be cve-oval files, and why no content from ssg), and other scap-utility should be able to use it; oscap-docker, but also oscap-daemon for instance. An utility available through command line AND a python lib should be adequate, i think.

Of course in order to eval the right cve-oval file on a container, i plan to test a whole CPE file on it to determine which distrib is used.

utils/oscap_docker_python/oscap_docker_util.py
```diff
@@ -28,291 +29,248 @@
...
-        )
-        sys.exit(1)
-
-except ImportError:
-    sys.stderr.write(
-        "Failed to import \"Atomic.mount.DockerMount\". It seems Atomic has "
-        "not been installed.\n"
-    )
-    sys.exit(1)
+def randomString(stringLength=10):
```
matejak: Python standard library has the [UUID module](https://docs.python.org/2/library/uuid.html) exactly for this purpose.

DominiqueDevinci:
@matejak Hi, yes i really understand the problem, the current request is bad for the legacy support, and the new code have nothing common with the precedent one.

I guess that's better to close this PR and create another one based on the precedent code ? Because the only other possibility i see is reverting most of commits in order to start on a proper base (or rebase dropping all commits) ... and it's dirty :D So feel free to close this PR and i will create a proper one (or tell me what process you prefer).

jan-cerny:
@DominiqueDevinci 

I second @matejak in his review.

We're fine with opening a new PR.

matejak:
Yes, feel free to close it after you open the successor PR.

Issue #1403, opened 2019-09-19
Remediation is not working when referencing a whole OVAL file


abergmann:
####  Description of Problem:
I've noticed a remediation problem when referencing a whole OVAL file without selecting a definition via a ```name``` value inside a XCCDF rule stanza.
```
<Rule id="xccdf_org.ssgproject.content_rule_test_file_presence_all" selected="false" severity="medium">
  <title>Test file presence all</title>
  <fix system="urn:xccdf:fix:script:sh">touch /tmp/file1; touch /tmp/file2</fix>
  <check multi-check="true" system="http://oval.mitre.org/XMLSchema/oval-definitions-5">
    <check-content-ref href="oval.xml"/>
  </check>
</Rule>
```
When a rule fails openscap is starting the remediation phase, but no script gets executed.
```
#> oscap xccdf eval --remediate --profile standard xccdf.xml 
Title   Test file presence all
Rule    xccdf_org.ssgproject.content_rule_test_file_presence_all
Result  fail

Title   Test file presence all
Rule    xccdf_org.ssgproject.content_rule_test_file_presence_all
Result  fail


 --- Starting Remediation ---
```

#### OpenSCAP Version:
```
#> oscap -V
OpenSCAP command line tool (oscap) 1.3.1
Copyright 2009--2018 Red Hat Inc., Durham, North Carolina.

==== Supported specifications ====
XCCDF Version: 1.2
OVAL Version: 5.11.1
CPE Version: 2.3
CVSS Version: 2.0
CVE Version: 2.0
Asset Identification Version: 1.1
Asset Reporting Format Version: 1.1
CVRF Version: 1.1
```

#### Operating System & Version:
Running on SUSE Linux Enterprise Server 12 SP4.

#### Steps to Reproduce:

Simply follow the steps I've outlined [here](https://gist.github.com/abergmann/5e48b2c0eb7888bc9b0082fed8fae986)

#### Actual Results:
The remediation commands do not get executed.

#### Expected Results:
All commands listed in the ```<fix>...</fix>``` section should be executed when the rule failed before.

DominiqueDevinci:
@abergmann I think it's not a but but a desired behavior. You can find the problem origin in my linked PR (which currently only adds verbose details concerning this case)

evgenyz:
@jan-cerny So, was this bug fixed by #1410 or not?

jan-cerny:
Yes, I tested it and the problem doesn't occur with OpenSCAP 1.3.2.

Issue #712, opened 2017-03-24
Warn when user asks to remediate rules that don't have a remediation


iankko:
`Disclaimer:` Didn't check latest version, so feel free to close this, if already implemented, or decided as wontfix.

`Use case:`
User scans the system (baremetal, vm, container, whatever) and requests online remediation. Since not all remediations might be implemented in the benchmark, user might expect the system will be corrected / remediated as a whole, though this obviously won't happen.

Thus rather to confuse users, and make them to later find out, why the scan is still failing, it's better right at the scan issue a NOTE / WARNING in the case, rule is present in the original (modified) profile, and online remediation was requested, but the corresponding remediation for that rule is missing (thus oscap won't correct this system property).

`Current user experience (see above disclaimer though, didn't check latest version):`
User requests remediation. After completion, some rules are still failing.

`Expected user experience:`
User requests remediation. Rules not having remediation implemented yet, would be printed by oscap either beforehand -- something like:

  The following rules are missing remediation:
  * Rule_1
  * Rule_2
  Remediation won't be performed for these rules.

or during the scan (if oscap knows if remediation is available only at the moment rule is "touched") -- something like:

Rule: Audit rules privileged commands (or whatever rule title here):
Result: Pass / Fail
Note: The rule is missing remediation. The corrective operation won't be performed.

Thank you for consideration!
Jan

mpreisler:
Makes sense to me. We should issue the same warning for `generate fix` as well. This would clearly communicate that the resulting fixes aren't complete for those test results.

shawndwells:
Love this. Would really be great for UX.

PR #1410, opened 2019-10-17
Remediation verbosity : don't apply fix for multicheck ? issue #1403

DominiqueDevinci:
- Description
I investigate the issue #1403 and found that it's not really a bug but a desired behavior which forbids to apply fix for multichecks. I guess that it's to force to use one fix by oval check. 

However, the is actually no information displayed, and the user doesn't undertsand why it's not working, of course. So i've add more details about remediations cases.

- Example
Result with the example provided in the issue (just look at last line):

```
oscap xccdf eval --verbose INFO --remediate --profile standard xccdf.xml 
I: oscap: Identified document type: Benchmark
I: oscap: Created a new XCCDF session from a XCCDF Checklist 'xccdf.xml'.
I: oscap: Identified document type: oval_definitions
I: oscap: Started new OVAL agent oval.xml.
I: oscap: Querying system information.
I: oscap: Starting probe on URI 'queue://system_info'.
I: oscap: probe_common_main started
I: oscap: SEAP_add_probe
I: oscap: I will run system_info_probe_main:
I: oscap: Extracting item from the cache queue: cnt=2, beg=0
I: oscap: cache MISS
I: oscap: Extracting item from the cache queue: cnt=1, beg=1
I: oscap: Evaluating a XCCDF policy with selected 'xccdf_org.ssgproject.content_profile_standard' profile.
I: oscap: Evaluating XCCDF group 'xccdf_org.ssgproject.content_group_intro'.
I: oscap: Evaluating XCCDF rule 'xccdf_org.ssgproject.content_rule_test_file_presence_1'.
I: oscap: Rule 'xccdf_org.ssgproject.content_rule_test_file_presence_1' is not selected.
I: oscap: Evaluating XCCDF rule 'xccdf_org.ssgproject.content_rule_test_file_presence_2'.
I: oscap: Rule 'xccdf_org.ssgproject.content_rule_test_file_presence_2' is not selected.
I: oscap: Evaluating XCCDF rule 'xccdf_org.ssgproject.content_rule_test_file_presence_all'.
Title   Test file presence all
Rule    xccdf_org.ssgproject.content_rule_test_file_presence_all
I: oscap: Evaluating definition 'oval:test_file_presence_2:def:1': Test file presence 2.
I: oscap:   Evaluating textfilecontent54 test 'oval:test_file_presence_2:tst:1': test file presence 2.
I: oscap:     Querying textfilecontent54 object 'oval:test_file_presence_2:obj:1', flags: 0.
I: oscap:     Creating new syschar for textfilecontent54_object 'oval:test_file_presence_2:obj:1'.
I: oscap:     Starting probe on URI 'queue://textfilecontent54'.
I: oscap:     probe_common_main started
I: oscap:     SEAP_add_probe
I: oscap:     I will run textfilecontent54_probe_main:
I: oscap:     Opening file '/tmp/file2'.
I: oscap:     Extracting item from the cache queue: cnt=1, beg=0
I: oscap:     Test 'oval:test_file_presence_2:tst:1' requires that every object defined by 'oval:test_file_presence_2:obj:1' exists on the system.
I: oscap:     0 objects defined by 'oval:test_file_presence_2:obj:1' exist on the system.
I: oscap:     Test 'oval:test_file_presence_2:tst:1' does not contain any state to compare object with.
I: oscap:     No item matching object 'oval:test_file_presence_2:obj:1' was found on the system. (flag=does not exist)
I: oscap:   Test 'oval:test_file_presence_2:tst:1' evaluated as false.
I: oscap: Definition 'oval:test_file_presence_2:def:1' evaluated as false.
Result  fail


 --- Starting Remediation ---
I: oscap: Identified document type: oval_definitions
I: oscap: Started new OVAL agent oval.xml.
I: oscap: Querying system information.
I: oscap: Starting probe on URI 'queue://system_info'.
I: oscap: probe_common_main started
I: oscap: SEAP_add_probe
I: oscap: I will run system_info_probe_main:
I: oscap: Extracting item from the cache queue: cnt=2, beg=0
I: oscap: cache MISS
I: oscap: Extracting item from the cache queue: cnt=1, beg=1
I: oscap: xccdf_org.ssgproject.content_rule_test_file_presence_all: cannot apply fix for multicheck.
```


DominiqueDevinci:
@jan-cerny I suggest (following issue #712 ) to add a warning in the result/report . For instance, if we run a bugged bash script as remediation, it displays an error in the report (and it's very well). And i think it should be great to have the same behavior if the remediation isn't applied for any other reason (online remediation not found, bad/missing interpreter, or not applied because of multicheck rule etc.)

If you agree this proposal i can do that :)

DominiqueDevinci:
Furthermore, I just see that some message info saved in the ARF file aren't displayed by the verbose mode ... example :  `<message severity="info">None of the check-content-ref elements was resolvable.</message>`
I guess it's because it's come from _rule_add_info_message and not dI ... do you not think that a proper restructuring of the log system is needed  ?

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -495,19 +500,22 @@ int xccdf_policy_rule_result_remediate(struct xccdf_policy *policy, struct xccdf
 
 	if (fix == NULL) {
 		fix = _find_suitable_fix(policy, rr);
-		if (fix == NULL)
-			// We may want to append xccdf:message about missing fix.
```
jan-cerny: The comment should be kept because it hasn't been addressed. The comment is about adding a message to the XCCDF results which is the big XML file generated using `--results` option.

DominiqueDevinci: yes my bad it's because at this time I had not understand the diff between DI and xccdf:message :(

DominiqueDevinci:
>  The fix element contains code which brings the system to compliance with the rule, not with individual OVAL checks. So it might be a bug.

No i've not found too but i was thinking it was a desired behavior because it seems a deliberate choice. I will test to see what's the result when trying to apply fix on a multichecks rule.

> But the other direction could be possible, though
Yes i also think it's the best solution. I will look at the feasibility of this :)

src/XCCDF_POLICY/xccdf_policy_remediate.c
jan-cerny: That's fine.

jan-cerny:
> Yes i also think it's the best solution. I will look at the feasibility of this :)

It doesn't have to block this PR. I expect that it won't be simple. Please only shorten the aforementioned message, add the comment you removed and then this PR can be merged, and you can propose another PR later.

DominiqueDevinci:
@jan-cerny I've just seen your message, too late i was working on that this morning :D 
Hum so you were right, the dI / dD / dW functions are very different that xccdf_messages. In fact, xccdf_message are much better because they are applied to an exact rule, and it's better, but the only problem is they are not displayed in some case.

So i've modify the error handling, and with this new patch the remediation is displayed in the result in any case, example with multicheck : 
```
 --- Starting Remediation ---
I: oscap: Identified document type: oval_definitions
I: oscap: Started new OVAL agent oval.xml.
I: oscap: Querying system information.
I: oscap: Starting probe on URI 'queue://system_info'.
I: oscap: probe_common_main started
I: oscap: SEAP_add_probe
I: oscap: I will run system_info_probe_main:
I: oscap: Extracting item from the cache queue: cnt=2, beg=0
I: oscap: cache MISS
I: oscap: Extracting item from the cache queue: cnt=1, beg=1
I: oscap: [xccdf_org.ssgproject.content_rule_test_file_presence_all]->msg: cannot apply fix for multicheck.
Title   Test file presence all
Rule    xccdf_org.ssgproject.content_rule_test_file_presence_all
Result  error

I: oscap: Identified document type: oval_definitions
```
And it's also in the HTML report. It's similar for "no suitable fix found etc.", before this patch they were several case where the rule was just "skipped" without any report nor logs.

Finally, i've also manually redefine the function xccdf_message_set_content in order it print the new message only in --verbose DEVEL mode. It can be usefull for display "lost" messages (that was the case here before the patch).

src/XCCDF_POLICY/xccdf_policy_remediate.c
DominiqueDevinci: In the last PR i've a little modify the comment since it's now reported ... i think now it can be even removed but I've prefer to keep too much comments than not enough.

DominiqueDevinci:
Here is an example of "skipping" when triggering a the error "Fix execution was aborted: Text substitution failed" before that I modify this part. You can see here why printing the message in DEVEL verbose mode is usefull, whithout that the message could have been lost.
```
D: oscap: packet size: 7749 [oscap(23786):oscap(7f910d5c9b00):seap-packet.c:794:SEAP_packet_recv]
D: oscap: Message received. [oscap(23786):oscap(7f910d5c9b00):oval_probe_ext.c:579:oval_probe_comm]
D: oscap: name=(null), value=0x55ecb01a74b0 [oscap(23786):oscap(7f910d5c9b00):seap-message.c:74:SEAP_msg_free]
D: oscap: XCCDF_SET_MESSAGE: Fix execution was aborted: Text substitution failed. [oscap(23786):oscap(7f910d5c9b00):result.c:537:xccdf_message_set_content]
I: oscap: Identified document type: oval_definitions [oscap(23786):oscap(7f910d5c9b00):doc_type.c:96:oscap_determine_document_type_reader]
D: oscap: probe_common_main_cleanup started [oscap(23786):common_main(7f90fe918700):probe_main.c:151:probe_common_main_cleanup]
D: oscap: probe_input_handler thread has joined with status -1 [oscap(23786):common_main(7f90fe918700):probe_main.c:165:probe_common_main_cleanup]
D: oscap: probe_common_main_cleanup finished [oscap(23786):common_main(7f90fe918700):probe_main.c:179:probe_common_main_cleanup]
```

DominiqueDevinci:
- in **test_remediation_subs_unresolved, the unittest expect to have the status "fail"** in the xml report. After my patch, it's "error" and not "fail", and in my mind it should be "error" since "fail" is when an evluation have been evaluated as failed (potentially when the fix have ben ran but haven't fix the rule compliance).

So ... i am wrong or should I change the unit test in order it expect "error" and not fail ? The rest of the xml result is exactly the same than before (but not the html report, the new html report include the error message, it was not the case before).

DominiqueDevinci:
Allow remediation for multicheck seems to work fine : 
https://pastebin.com/K99E32F5
That fixes issue #1403

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -493,39 +499,58 @@ int xccdf_policy_rule_result_remediate(struct xccdf_policy *policy, struct xccdf
...
+
 	if (fix == NULL) {
 		fix = _find_suitable_fix(policy, rr);
-		if (fix == NULL)
-			// We may want to append xccdf:message about missing fix.
-			return 0;
+		if (fix == NULL) {
+			// We want to append xccdf:message about missing fix.
+			_rule_add_info_message(rr, "No suitable fix found.");
+			xccdf_rule_result_set_result(rr, XCCDF_RESULT_ERROR);
```
jan-cerny: This is controversial.

The first point of view is that if you have no remediation nothing changed so the rule result is still fail.

The second point of view is that if users run "oscap xccdf remediate" users expect that the rule will start passing after the command finishes and if it doesn't it should be an error.

If you have found anything about that in the XCCDF specification, please point me to that, but I haven't found anything. To be on the safe side I prefer to keep the current behavior, which is to keep the XCCDF_RESULT_FAIL.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -493,39 +499,58 @@ int xccdf_policy_rule_result_remediate(struct xccdf_policy *policy, struct xccdf
...
-
-	/* Execute the fix. */
-	res = _xccdf_fix_execute(rr, cfix);
-	if (res != 0) {
-		_rule_add_info_message(rr, "Fix was not executed. Execution was aborted.");
-		return res;
+		_rule_add_info_message(rr, "cannot apply fix for multicheck.");
+		xccdf_rule_result_set_result(rr, XCCDF_RESULT_ERROR);
+		misc_error=1;
+	}*/
```
jan-cerny: No need to leave the commented code here, we have git history. Please remove this including the comment.

src/XCCDF_POLICY/xccdf_policy_remediate.c
DominiqueDevinci: I understand this point of view too :) that's done for both change request.

DominiqueDevinci:
Hello, i've adapt the last blocking test according to the new result ('error' instead of 'fail' for messages "Fix element contains unresolved child elements. Fix was not executed. Execution was aborted."). I understand that can be controversial (even it's more an error than when no suitable fix are found), so just tell me if you want to change this or if it's ok and the PR is finished :+1:

jan-cerny:
@DominiqueDevinci 

I like the code very much now. Thanks for your update.

However, the Jenkins CI tests still fails. This time `API/XCCDF/applicability/all.sh` fails in 

https://github.com/OpenSCAP/openscap/blob/490983b2adbc730fdbfc8e83df97901cc073d5e4/tests/API/XCCDF/applicability/test_remediate_fix_notapplicable.sh#L30

because it doesn't expect `<message severity="info">No suitable fix found.</message>` in XCCDF results.

The tests need to be updated. I suspect there will be multiple occurrences of that assertion, maybe in multiple files. Please try to find them all.

I admit it's unfortunate that the tests terminate on the first failed assertion.

DominiqueDevinci:
@jan-cerny My bad, this time i think it's ok.
In fact some tests are failing on my machine even they pass on the CI ... i will probably open an issue about this, because it's a real problem to not have same results in dev env and on CI :D

jan-cerny:
Great contribution! Thanks.

PR #1412, opened 2019-10-25
Use and return canonical paths in rpmverifyfile probe

jan-cerny:
This is done to prevent false positives in implementation of rules file_permissions_unauthorized_suid and file_permissions_unauthorized_sgid proposed in ComplianceAsCode in https://github.com/ComplianceAsCode/content/pull/4648 caused by inconsistent symlink usage in rpmdb.

Also improves logic when an exact file path is defined in rpmverifyfile_object we don't have to iterate over all RPMs in the rpmdb but we query the rpmdb directly.


src/OVAL/probes/unix/linux/rpmverifyfile_probe.c
```diff
@@ -229,51 +244,75 @@ static int rpmverify_collect(probe_ctx *ctx,
...
-		      break;
+					match1 = pcre_exec(re, NULL, current_file, strlen(current_file), 0, 0, NULL, 0);
+					if (match1 < -1) {
+						ret = -1;
+						dE("pcre_exec() failed!");
+						free(current_file_realpath);
+						goto ret;
+					}
+					if (current_file_realpath) {
+						match2 = pcre_exec(re, NULL, current_file_realpath, strlen(current_file_realpath), 0, 0, NULL, 0);
```
jan-cerny: Maybe we don't have to change the `case OVAL_OPERATION_PATTERN_MATCH` at this moment because we don't need it now and we haven't created any OVAL content on which we would test it.

matusmarhefka:
@jan-cerny Do we have some unit test which covers this change?

ggbecker:
Thanks for the patch.

ggbecker:
> @jan-cerny Do we have some unit test which covers this change?

as far as I know the current test `test_probes_rpmverifyfile.sh` does not cover files which are symbolic links. @jan-cerny is it feasible to extend this test so it checks also for symlinks?

jan-cerny:
@matusmarhefka @ggbecker I think the test depends on a specific RPM package which  mean that it isn't suitable for the upstream test suite.

PR #1413, opened 2019-10-31
added a comment about not removing attributes improving a11y

vojtapolasek:
Added a comment into openscap.js not to remove certain attributes because it breaks accessibility of reports and guides

### DIFF
--- a/xsl/xccdf-resources/openscap.js
+++ b/xsl/xccdf-resources/openscap.js
@@ -2,6 +2,8 @@ function openRuleDetailsDialog(rule_result_id)
 {
     $("#detail-modal").remove();
 
+    /*DO NOT remove the aria-hidden="false" attribute from the following lines!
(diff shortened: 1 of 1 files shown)


xsl/xccdf-resources/openscap.js
```diff
@@ -2,6 +2,7 @@ function openRuleDetailsDialog(rule_result_id)
 {
     $("#detail-modal").remove();
 
+// DO NOT remove the aria-hidden="false" attribute from the following lines! This attribute ensures that rule detail can be accessed by accessibility technologies such as screenreaders.
```
yuumasato: The indentation is borked.
And nitpick, I'd break the comment into two lines.

dahaic:
@vojtapolasek cool, the comment makes sense, no objections there. Can you squish those two commits into one, though? It will improve readability of the git log.

With that changed, it's good to be merged.

PR #1418, opened 2019-11-06
Fix #528 - Eval SCE script when /tmp is in mode noexec

### LINKED ISSUE #528: SCE & run scripts without permissions
According to open-scap mailing list,

we are not able to execute SCE scripts without exec permissions (e.g. from **/tmp** with **noexec** mount option.)

In my opinion, we can parse shebang from script and use interpret to run it (`/usr/bin/bash /tmp/sce.sh`) - in similar way we run remediations. Can we run binary files in such way?

DominiqueDevinci:
Hello,
I experienced recently an error already signalled in issue #528 . As suggested by Martin Presiler, i've used a wrapper (so it's easy to fix a problem linked to the wrapper for everyone). My wrapper parse the shebang (or use bash by default if there is no shebang).

I use the wrapper only if there is no posssibility to acquire +x rights on the extracted .sh file, so the legacy code is not affected (and the wrapper is less reliable than a direct execution, so it's better to only use in when needed).

Result:
With a /tmp partition with noexec mode: 
```oscap xccdf eval --verbose INFO test_sce_in_ds.xml
I: oscap: Identified document type: data-stream-collection
I: oscap: Created a new XCCDF session from a SCAP Source Datastream 'test_sce_in_ds.xml'.
I: oscap: Identified document type: Benchmark
I: oscap: Evaluating a XCCDF policy with selected 'default-profile' profile.
I: oscap: Evaluating XCCDF rule 'xccdf_org.open-scap.sce-community-content_rule_services_obsolete-disable_nfs_exports'.
Title
Rule    xccdf_org.open-scap.sce-community-content_rule_services_obsolete-disable_nfs_exports
I: oscap: /tmp/oscap.UmezuG/services/obsolete/disable_nfs_exports.sh isn't executable, oscap-run-sce-script will be use.
Result  pass
```
This test was not passing before this fix.

### DIFF
--- a/src/SCE/sce_engine.c
+++ b/src/SCE/sce_engine.c
@@ -384,7 +384,7 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
 	{
 		// use the sce wrapper if it's not possible to acquire +x rights
 		use_sce_wrapper = true;
-		dI("%s isn't executable, oscap-run-sce-script will be use.", tmp_href);
+		dI("%s isn't executable, oscap-run-sce-script will be used.", tmp_href);
 	}
 
 	// all the result codes are shifted by 100, because otherwise syntax errors in scripts
@@ -515,7 +515,6 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
 
 		if (fork_result == 0)
 		{
-
 			// we won't read from the pipes, so close the reading fd
 			close(stdout_pipefd[0]);
 			close(stderr_pipefd[0]);
@@ -544,7 +543,7 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
 			if(use_sce_wrapper)
 				execvpe("oscap-run-sce-script", argvp, env_values);
 			else
-                execve(tmp_href, argvp, env_values);
+				execve(tmp_href, argvp, env_values);
 
 			free_env_values(env_values, index_of_first_env_value_not_compiled_in, env_value_count);
 
--- a/src/SCE/sce_engine.c
+++ b/src/SCE/sce_engine.c
@@ -362,6 +362,7 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
 {
 	struct sce_parameters* parameters = (struct sce_parameters*)usr;
 	const char* xccdf_directory = parameters->xccdf_directory;
+	bool use_sce_wrapper = false; // use osca-run-sce-script ?
 
 	char* tmp_href = oscap_sprintf("%s/%s", xccdf_directory, href);
 
@@ -381,27 +382,26 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
 
 	if (access(tmp_href, F_OK | X_OK))
 	{
-		// again, only to provide helpful error message
-		oscap_seterr(OSCAP_EFAMILY_SCE, "SCE has found script file '%s' at '%s' "
-				"but it isn't executable!", href, tmp_href);
-		free(tmp_href);
-		return XCCDF_RESULT_ERROR;
+		// use the sce wrapper if it's not possible to acquire +x rights
+		use_sce_wrapper = true;
+		dI("%s isn't executable, oscap-run-sce-script will be use.", tmp_href);
 	}
 
 	// all the result codes are shifted by 100, because otherwise syntax errors in scripts
 	// or even their nonexistence would cause XCCDF_RESULT_PASS to be the result
 
-	char* argvp[1 + 1] = {
+	char* argvp[3] = {
+		tmp_href,
 		tmp_href,
 		NULL
 	};
-
+	dI("arvp[0] = %s", argvp[0]);
 	// bound values in KEY=VALUE form, ready to be passed as environment variables
 	char ** env_values = malloc(10 * sizeof(char * ));
 	size_t env_value_count = 10;
 	const size_t index_of_first_env_value_not_compiled_in = 10;
 
-	env_values[0] = "PATH=/bin:/sbin:/usr/bin:/usr/sbin";
+	env_values[0] = "PATH=/bin:/sbin:/usr/bin:/usr/local/bin:/usr/sbin";
 
 	env_values[1] = "XCCDF_RESULT_PASS=101";
 	env_values[2] = "XCCDF_RESULT_FAIL=102";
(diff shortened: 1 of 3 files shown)


utils/oscap-run-sce-script
```diff
@@ -0,0 +1,24 @@
...
+if [ ! -z $1 ] && [ -f $1 ] 
+then
+	# file exists. first check if shebang is here 
+	
+	firstline=$(head -n1 $1)
+	if [ ${firstline:0:2} = "#!" ]
+	then  # it's a shebang
+		cmd=${firstline:2}  # remove the begin (#!)
+		cmd=${cmd##*( )}  # trim whitespaces
+		eval $cmd $1 > /dev/stdout
```
jan-cerny: How do we know that we don't run arbitrary binary?

evgenyz: Isn't it the point to run some arbitrary binary? Or you expect that this wrapper could be abused?

src/SCE/sce_engine.c
```diff
@@ -381,27 +382,26 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
...
 	if (access(tmp_href, F_OK | X_OK))
 	{
-		// again, only to provide helpful error message
-		oscap_seterr(OSCAP_EFAMILY_SCE, "SCE has found script file '%s' at '%s' "
-				"but it isn't executable!", href, tmp_href);
-		free(tmp_href);
-		return XCCDF_RESULT_ERROR;
+		// use the sce wrapper if it's not possible to acquire +x rights
+		use_sce_wrapper = true;
+		dI("%s isn't executable, oscap-run-sce-script will be use.", tmp_href);
```
evgenyz: Will be *used*.

src/SCE/sce_engine.c
```diff
@@ -539,7 +540,11 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
 #endif
 
 			// we are the child process
-			execve(tmp_href, argvp, env_values);
+
+			if(use_sce_wrapper)
+				execvpe("oscap-run-sce-script", argvp, env_values);
+			else
+                execve(tmp_href, argvp, env_values);
```
evgenyz: Indentation seems to be wrong here (tabs/spaces mix?)

src/SCE/sce_engine.c
```diff
@@ -515,6 +515,7 @@ xccdf_test_result_type_t sce_engine_eval_rule(struct xccdf_policy *policy, const
 
 		if (fork_result == 0)
 		{
+
```
evgenyz: Wasteful spacing

utils/oscap-run-sce-script
DominiqueDevinci: Hello,
In the case were oscap-run-sce-script is used directly by the user, the privileges are exactly the same, there is no special permissions accorded here. And if the cmd contains an arbitrary "binary", it will not be executed if it have not +x right (if if it have it, the user can run it directly).

Example, chmod -x on /usr/bin/lsub, create a .sh file in tmp with the shebang #! /usr/bin/lsusb, run oscap-run-sce-script /tmp/tmp.sh, it will try to eval /usr/bin/lsusb but this binary will not be executed.

The only case where an "privilege escalation" could be observed is if the user have no bash access, but is able to run arbitrarily oscap-run-sce-script, and not that if he can do that he can probably also run oscap on a custom datastream containing bash commands. So yes, I agree that there is a possibility of "bash injection", but in our case i don't see the risk since there is no privilege escalation (when using directly oscap-run-sce-script), and moreover this "bash injection" can also be done by evaluating a custom datastream, so for me there is no more risks with or without oscap-run-sce-script.

The only potential security issue that I see is if oscap is running with root privileges, we need to be absolutely sure that the file evaluated (previously extracted from the DS) is safe, i.e. the DS should be signed and we should ensure that the content of the tmp file haven't been modified between its extraction and it's launch; I've not analyse more this potential issue, but the issue is exactly the same with or without oscap-run-sce-script (because oscap-run-sce-script can have root permissions only if launched from oscap running as root). But this subject should be in the scope on another PR (and i think it should be a good idea to discuss about "should we run sce script as root if oscap is running as root ? should we introduce a specific attribute specifying if the sce script should be ran as root ? Perhaps this possibility should be allowed only with DSIG enabled on the DS").

I don't see any other way of "abusing" the oscap-run-sce-script, but feel free to correct me if i'm going wrong and to suggest a mitigation (note that originally this wrapper script if an idea of martin preisler, it's not mandatory to use it but easy to use and/or customize, and currently I don't see in what it adds a security issue).

evgenyz: To bring a bit more context to the topic: https://serverfault.com/questions/72356/how-useful-is-mounting-tmp-noexec

So, in general I agree with @DominiqueDevinci about it not increasing the surface of attack in any way, because there is already a possibility to run scripts from data streams.

On the other hand, if you would prefer to not to execute arbitrary binaries as interpreters, the only way to go it is to create a "white list" right in the wrapper. Does it look like feasible and/or sensible, @jan-cerny?

jan-cerny: Do you think that attackers could write their payload to `/tmp` and then call our wrapper to execute the payload if they find that there is `noexec` set on `/tmp`?

DominiqueDevinci: @evgenyz about the context I agree that the gain of /tmp with noexec mode isn't huge.
However, it's a specification (in my case it's required by ANSSI NT28.R12), and even if we can  disagree with that, the goal here is to be compliant, and it's not possible currently with openscap using SCE (and SCE is incredibly usefull for custom advanced checks). I think that's the most important point of the context.

A white list seems a correct mitigation for me (the real question is to agree with SCE "specifications", if exists ... users should be informed of this white list), and so we should use only this wrapper (even if the tmp file is executable). I will investigate later about the risks of running oscap as root with SCE, because it's an importante question.

evgenyz: It won't work as /tmp is noexec, and it won't allow to have anything executable there. And eval won't execute it.

DominiqueDevinci:
Change requests fixed :) If you cannot decide about whitelist or not (for me the problem is that there is no SCE specifications about that, and it can be considered as a regression to restrict the supported scripts - even if practically it's not a big deal to add an interpreter into the whitelist). So, tell me ... anyway, this PR don't extend the vulnerable area, so i think we can close it and make an issue about this interpreter restriction.

jan-cerny:
@openscap-ci test this please

jan-cerny:
So far it looks good to me, I have executed the CI again, because the fail was an infrastructure problem, and wasn't related to the pull request.

PR #1424, opened 2019-11-25
Small verbose mode improvements

jan-cerny:
I have wished to do this for a long time, I finally decided to do a small refresh of verbose mode.

I have changed the category of many messages from INFO to DEVEL. These are mostly the messages that remained in the code from the legacy system. However, in the legacy system the messages were printed only if complied with debug symbols, and therefore they had a different purpose. I believe the purpose of INFO level is to provide description of the evaluation process on a logical level, and not to overwhelm users by implementation details.

I also stopped showing messages about evaluation of rules and groups when they aren't evaluated. From the scanner point of view, the XCCDF group doesn't affect rule evaluation. We don't need to report them, because it makes the verbose output longer. And, the rules that are not selected by the profile don't need to be reported in the verbose output as well.

I believe this can also improve the user experience of SSGTS if SSGTS decides to switch to INFO level logs.

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy.c
+++ b/src/XCCDF_POLICY/xccdf_policy.c
@@ -986,7 +986,6 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
 	 * evaluated is not equal to the selected rule, do not evaluate it and
 	 * mark it as notselected. */
 	if (policy->rule != NULL) {
-		dI("Policy selects only 1 rule");
 		if (strcmp(policy->rule, rule_id) != 0) {
 			return _xccdf_policy_report_rule_result(policy, result, rule, NULL, XCCDF_RESULT_NOT_SELECTED, NULL);
 		}
@@ -1001,9 +1000,9 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
 	xccdf_role_t role = xccdf_get_final_role(rule, r_rule);
 
 	if (!is_selected) {
-		dI("Rule '%s' is not selected.", rule_id);
 		return _xccdf_policy_report_rule_result(policy, result, rule, NULL, XCCDF_RESULT_NOT_SELECTED, NULL);
 	}
+	dI("Evaluating XCCDF rule '%s'.", rule_id);
 
 	if (role == XCCDF_ROLE_UNCHECKED)
 		return _xccdf_policy_report_rule_result(policy, result, rule, NULL, XCCDF_RESULT_NOT_CHECKED, NULL);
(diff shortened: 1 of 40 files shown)


evgenyz:
I perceive this PR as pretty harmless, and it would for sure clean up the logs a great deal. But I have a concern about the Workbench. Wouldn't it break something there?

jan-cerny:
@evgenyz Great concern! I believe it will not break anything there. SCAP Workbench calls `oscap` without `--verbose`. That means `oscap` produces only "W" and "E" messages on stderr. SCAP workbench displays everything that appears on stderr as an "error" in diagnostic window. That means this PR will reduce the amount of errors displayed in SCAP Workbench, because this PR moves everything to D category and D messages aren't produced by default. Also, the change of "I" messages about rules and groups shouldn't affect SCAP Workbench because "I" messages aren't produced by default by `oscap`.

evgenyz:
Aha, makes sense.

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -1000,9 +1000,9 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
 	xccdf_role_t role = xccdf_get_final_role(rule, r_rule);
 
 	if (!is_selected) {
-		dI("Rule '%s' is not selected.", rule_id);
 		return _xccdf_policy_report_rule_result(policy, result, rule, NULL, XCCDF_RESULT_NOT_SELECTED, NULL);
 	}
+	dI("Evaluating XCCDF rule '%s'.", rule_id);
```
matusmarhefka: This is being printed for all rules and groups even when I use `--rule`. Can we do something about it?

evgenyz: Let me check

evgenyz: Hmm. `./run utils/oscap --verbose=DEVEL xccdf eval --profile xccdf_org.ssgproject.content_profile_ospp --rule=xccdf_org.ssgproject.content_rule_no_empty_passwords ../../content/build/ssg-fedora-ds.xml` gives me exactly one line `I: oscap: Evaluating XCCDF rule 'xccdf_org.ssgproject.content_rule_no_empty_passwords'...`.

evgenyz: @matusmarhefka Can you please give me an example of a command to use to reproduce behaviour observed by you?

matusmarhefka: @evgenyz Okay, this was probably something wrong on my side as now I have tried to re-test it and the message is not printed anymore. Before, I have used `make install` and used `oscap` instead of the `./run utils/oscap`, but now it works correctly for both.

Issue #1320, opened 2019-04-16
Console output of check with multi-check="true" is confusing


yuumasato:
#### Description of Problem:

When OpenSCAP is evaluating a rule whose check is `multi-check="true"`, the console output is confusing.

#### OpenSCAP Version:
OpenSCAP 1.2.17
Very likely reproducible in OpenSCAP 1.3 too

#### Steps to Reproduce:

1. Use Datastream linked below
2. oscap xccdf eval --profile pci-dss --fetch-remote-resources --results results.xml --report report.html ./ssg-rhel7-ds-1.2-catalog-component.xml
3. There will be multiple reports for rule `security_patches_up_to_date`.
4. In `results.xml` file it is possible to see that each XCCDF result entry is for a different OVAL definition.

#### Actual Results:
```
Title   Disable Prelinking
Rule    xccdf_org.ssgproject.content_rule_disable_prelink
Ident   CCE-27078-5
Result  pass

Title   Ensure Software Patches Installed
Rule    xccdf_org.ssgproject.content_rule_security_patches_up_to_date
Ident   CCE-26895-3
Result  fail

Title   Ensure Software Patches Installed
Rule    xccdf_org.ssgproject.content_rule_security_patches_up_to_date
Ident   CCE-26895-3
Result  pass

Title   Ensure Software Patches Installed
Rule    xccdf_org.ssgproject.content_rule_security_patches_up_to_date
Ident   CCE-26895-3
Result  pass

Title   Ensure Software Patches Installed
Rule    xccdf_org.ssgproject.content_rule_security_patches_up_to_date
Ident   CCE-26895-3
Result  fail
```

#### Expected Results:
It is somehow clear from console output that each result for `security_patches_up_to_date` is about a different CVE.

#### Additional Information / Debugging Steps:
RHEL7 DataStream SCAP1.2 with `multi-check="true"` attribute in Rule `security_patches_up_to_date`: https://fedorapeople.org/~wsato/DS/ssg-rhel7-ds-1.2-multi-check.xml

jan-cerny:
Also reported in https://bugzilla.redhat.com/show_bug.cgi?id=1771438

PR #1426, opened 2019-11-26
Show check identifiers at multi-check rules

jan-cerny:
Shows OVAL Definition ID and Title when an XCCDF rule uses
multi-check='true'. The multi-check is used in rule
security_patches_up_to_date in SCAP 1.3 datastreams. It wasn't possible
to see which vulnerabilities have been found because all the checks have
the same XCCDF rule ID.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1771438

Fixes: #1320

### DIFF
--- a/src/OVAL/oval_agent.c
+++ b/src/OVAL/oval_agent.c
@@ -630,19 +630,34 @@ _oval_agent_list_definitions(void *usr, xccdf_policy_engine_query_t query_type, 
 {
 	__attribute__nonnull__(usr);
 	struct oval_agent_session *sess = (struct oval_agent_session *) usr;
-	if (query_type != POLICY_ENGINE_QUERY_NAMES_FOR_HREF || (query_data != NULL && strcmp(sess->filename, (const char *) query_data)))
+	if (query_data != NULL && strcmp(sess->filename, (const char *) query_data)) {
 		return NULL;
-	struct oval_definition_iterator *iterator = oval_definition_model_get_definitions(sess->def_model);
-	struct oscap_stringlist *result = oscap_stringlist_new();
-	struct oval_definition *oval_def;
-
-	while (oval_definition_iterator_has_more(iterator)) {
-		oval_def = oval_definition_iterator_next(iterator);
-		oscap_stringlist_add_string(result, oval_definition_get_id(oval_def));
 	}
+	if (query_type == POLICY_ENGINE_QUERY_NAMES_FOR_HREF) {
+		struct oval_definition_iterator *iterator = oval_definition_model_get_definitions(sess->def_model);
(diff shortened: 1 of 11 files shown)


jan-cerny:
We need to update the HTML report as well, see https://github.com/ComplianceAsCode/content/issues/5037

matejak:
test this please

xsl/xccdf-report-oval-details.xsl
```diff
@@ -63,6 +65,30 @@ Authors:
 </xsl:template>
 
 <xsl:template mode='brief' match='ovalres:definition'>
+    <xsl:variable name="definition" select="key('ovaldef-definition', ./@definition_id)" />
```
jan-cerny: @jan-cerny Please submit this change as a separate PR, it isn't mandatory part of the fix.

jan-cerny:
This is now ready for a review. You can check the example stdout and HTML report from evaluation of PCI-DSS profile on a RHEL 8 machine:
[pci-dss-results.zip](https://github.com/OpenSCAP/openscap/files/3943778/pci-dss-results.zip)
Focus on rule "Ensure Software Patches Installed".

matejak:
test this please

evgenyz:
@jan-cerny Would it affect Workbench? Does it have to be patched as well?

jan-cerny:
@evgenyz You're great! Thanks for pointing this out. I played with this Unfortunately, we need to add support for the multi-check feature in SCAP Workbench as well. Currently, SCAP Workbench shows only 1 of the N results of the checks (probably the latest evaluated check result).  I tried to compile and run Workbench with the patched OpenSCAP and I have found that this patch neither improves it nor breaks it, as it's already broken. I assume that to integrate this feature in SCAP Workbench we also need to add additional column to the `--progress` output of `oscap xccdf eval` or some other way to distinguish between the checks within a single XCCDF rule, which means changes on OpenSCAP side.

jan-cerny:
I checked SCAP Workbench code. Unfortunately, I haven't found any easy way to bring this feature to SCAP Workbench. I have reported an issue against Workbench about that in https://github.com/OpenSCAP/scap-workbench/issues/242.

I suggest not blocking this PR by SCAP Workbench. As I said in my previous comment, this patch neither fixes nor breaks the Workbench. That's mainly because SCAP Workbench uses `oscap` in `--progress` mode, but this patch modifies the normal output mode.

Please let me know your thoughts about the improvements in the HTML report and check also if the stdout format is OK for you.

evgenyz:
The patch set looks good to me. We will deal with the Workbench in a separate PR. HTML report is definitely makes more sense now, and as stdout is not supposed to be used for inter-op we can allow a a bit of swag in any direction.

PR #1428, opened 2019-11-29
Tests for symlink probe in offline mode

vojtapolasek:
The symlink probe in offline mode lacks test coverage.
Tests are created based on tests for online mode and extended with several tests cases specific for offline mode.
Tests could not be ran yet because of a bug in Openscap.


tests/probes/symlink/test_offline_mode_symlink.sh
```diff
@@ -0,0 +1,83 @@
...
+
+. $builddir/tests/test_common.sh
+
+set -e -o pipefail
+
+function test_offline_mode_symlink {
+    probecheck "symlink" || return 255
+
+    DF="test_offline_mode_symlink.xml"
+    RF="results.xml"
```
matejak: I recommend to go for `offline_mode_results.xml` for consistency.

tests/probes/symlink/test_offline_mode_symlink.sh
```diff
@@ -0,0 +1,83 @@
...
+
+    rm -f $RF
+
+    tmpdir=$(mktemp -t -d "test_symlink.XXXXXX")
```
matejak: ditto - why not to go for `test_offline_mode_symlink`?

tests/probes/symlink/test_offline_mode_symlink.sh
```diff
@@ -0,0 +1,83 @@
...
+    ln -s /other_file $tmpdir/middle_symlink
+    ln -s /middle_symlink $tmpdir/chained_symlink
+    mkdir $tmpdir/etc
+    touch $tmpdir/etc/shadow
+    ln -s /etc/shadow $tmpdir/etc/passwd
+    touch $tmpdir/file_in_root
+    ln -s ../../../../../file_in_root $tmpdir/path_traversal_symlink
+    ln -s ./some_file $tmpdir/symlink_with_period_in_path
+    touch /tmp/symlinktarget
+    ln -s /tmp/symlinktarget /tmp/symlinktest
```
matejak: These two cause problems if they exist from the previous test run - and if a test run fails, those links won't be deleted. I recommend using `mktemp` without the `-d` here.

tests/probes/symlink/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 if(ENABLE_PROBES_UNIX)
 	add_oscap_test("all.sh")
+	add_oscap_test("test_offline_mode_symlink.sh")
```
matejak: I recommend to add `set_tests_properties("probes/symlink/test_offline_mode_symlink.sh" PROPERTIES WILL_FAIL true)` for the time being so the test doesn't block gating.

tests/probes/symlink/test_offline_mode_symlink.sh
```diff
@@ -0,0 +1,83 @@
...
+    assert_exists 1 $p'system_data/unix-sys:symlink_item/unix-sys:canonical_path[text()="/other_file"]'
+    assert_exists 1 $p'system_data/unix-sys:symlink_item/unix-sys:filepath[text()="'/path_traversal_symlink'"]'
+    assert_exists 1 $p'system_data/unix-sys:symlink_item/unix-sys:canonical_path[text()="/file_in_root"]'
+    assert_exists 1 $p'system_data/unix-sys:symlink_item/unix-sys:filepath[text()="'/symlink_with_dot_in_path'"]'
+    assert_exists 1 $p'system_data/unix-sys:symlink_item/unix-sys:canonical_path[text()="/some_file"]'
+    assert_exists 3 $p'collected_objects/object[@flag="error"]'
+    assert_exists 3 $p'collected_objects/object[@flag="does not exist"]'
+    assert_exists 6 $p'collected_objects/object/message'
+
+    rm -rf $tmpdir
```
matejak: In theory, the test data can be deleted before the phase with assertions.

matejak:
test this please

matejak:
test this please

tests/probes/symlink/test_offline_mode_symlink.sh
vojtapolasek: I would leave this for future improvements. In current state it is not possible to use mktemp here, because the file name is hardcoded in the generated oval. And if the file exists from the previous run, that actually doees not matter. The probe should not be able to see files if working correctly.

matejak: We actually need to address this, as there is `set -e` effective, os if there is already `/tmp/symlinktarget`, the test will abort on this line. I wouldn't worry about hardcoding either - you can parametrize those names in the same manner as `tmpdir` is parametrized.

matejak: In your last commit, you have moved the cleanup commands. However, you need to make sure that `/tmp/symlinktest` and `/tmp/symlinktarget` are not there *before* you try to create them. And I am pretty sure that if you remove results before actually checking them, the test will fail.

matejak:
test this please

jan-cerny:
Blocked by #1431

vojtapolasek:
test this please

matejak:
test this please

matejak:
@jan-cerny This is not blocked, as `set_tests_properties("probes/symlink/test_offline_mode_symlink.sh" PROPERTIES WILL_FAIL true)` ensures that the failing test won't cause any trouble.

matejak:
test this please

matejak:
test this please

matejak:
Thank you for adjustments, the PR adds test coverage for the symlink probe in offline mode.

PR #1431, opened 2019-12-02
Move the chroot call to probe_worker

jan-cerny:
We have discovered a problem when testing offline mode in the symlink
probe. When the probe main function needs to be executed multiple times,
function `probe_common_main`, where the chroot is entered, is called
only once, but function `probe_worker`, which leaves the chroot, is called
many times. This leads to a situation that the first OVAL test is
evaluated in chroot, but all further tests have retrieved data from the
host. Therefore the collected data were wrong. We try to address this
problem by moving the code.

EDIT:
* I have checked that the symlink probe test proposed without that #1428 fails without my patch and passes with the patch applied. Actually it's vice-versa due to  `PROPERTIES WILL_FAIL true` in CMakeLists.txt
* tests of RPM probes from #1430 started to pass when I merged this PR in.

### DIFF
--- a/src/OVAL/probes/probe/input_handler.c
+++ b/src/OVAL/probes/probe/input_handler.c
@@ -124,7 +124,7 @@ void *probe_input_handler(void *arg)
 			SEXP_VALIDATE(oid);
 
 			if (probe->offline_mode && probe->supported_offline_mode == PROBE_OFFLINE_NONE) {
-				dW("Requested offline mode is not supported by %s.", probe->name);
+				dW("Requested offline mode is not supported by %s probe.", oval_subtype_get_text(probe->subtype));
 				/* Return a dummy. */
 				probe_out = probe_cobj_new(SYSCHAR_FLAG_NOT_APPLICABLE, NULL, NULL, NULL);
 				probe_ret = 0;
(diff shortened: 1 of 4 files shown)


jan-cerny:
@openscap-ci test this please

jan-cerny:
Rebased on maint-1.3 after #1428 was merged.

src/OVAL/probes/probe/input_handler.c
```diff
@@ -122,99 +122,86 @@ void *probe_input_handler(void *arg)
...
+			probe_out = probe_rcache_sexp_get(probe->rcache, oid);
 
-			if (probe->offline_mode && probe->supported_offline_mode == PROBE_OFFLINE_NONE) {
-				dW("Requested offline mode is not supported by %s.", probe->name);
-				/* Return a dummy. */
-				probe_out = probe_cobj_new(SYSCHAR_FLAG_NOT_APPLICABLE, NULL, NULL, NULL);
-				probe_ret = 0;
-				SEXP_free(oid);
+			if (probe_out == NULL) { /* cache miss */
+				SEXP_t *skip_flag, *obj_mask;
```
jan-cerny: note for reviewers: this block is only indented 1 level less

src/OVAL/probes/probe/worker.c
```diff
@@ -41,6 +43,29 @@
 extern bool  OSCAP_GSYM(varref_handling);
 extern void *OSCAP_GSYM(probe_arg);
 
+
+static int fail(int err, const char *who, int line)
```
jan-cerny: @jan-cerny Please remove it!

evgenyz:
Looks good. Thanks for the fix!

PR #1432, opened 2019-12-06
Modified test suite to make use of setcap-blessed oscap binary.

matejak:
Some offline-mode tests require chroot syscalls during execution.

The build system and test suite are now aware of this and support build/usage of oscap "buddy binary" that is oscap with setcap chroot blessing.
The person who builds may either decide to run `sudo make oscap-chrootable`, or `make oscap-chrootable-nocap` followed by manual `setcap` call.

### DIFF
--- a/tests/probes/symlink/test_offline_mode_symlink.sh
+++ b/tests/probes/symlink/test_offline_mode_symlink.sh
@@ -45,7 +45,12 @@ function test_offline_mode_symlink {
 
     bash ${srcdir}/test_offline_mode_symlink.xml.sh "" > "$DF"
     export OSCAP_PROBE_ROOT="$tmpdir"
-    $OSCAP oval eval --results $RF $DF
+    if test -x "$OSCAP_CHROOTABLE_EXEC"; then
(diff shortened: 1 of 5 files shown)


tests/probes/symlink/test_offline_mode_symlink.sh
```diff
@@ -45,7 +45,12 @@ function test_offline_mode_symlink {
 
     bash ${srcdir}/test_offline_mode_symlink.xml.sh "" > "$DF"
     export OSCAP_PROBE_ROOT="$tmpdir"
-    $OSCAP oval eval --results $RF $DF
+    if test -x "$OSCAP_CHROOTABLE_EXEC"; then
```
evgenyz: Can we reuse `require "$OSCAP_CHROOTABLE_EXEC" || return 255` or create another special function `require_offline`? And then also we could just replace $OSCAP with OSCAP=$OSCAP_CHROOTABLE in all related tests.

jan-cerny:
@matejak I expect you will remove PROPERTIES WILL_FAIL from tests/probes/symlink/CMakeLists.txt

evgenyz:
LGTM

PR #1434, opened 2019-12-10
[WIP] OVAL details in HTML report layout improvement

jan-cerny:
This patch was originally a part of #1426 but it was not a mandatory part of the fix so I decided to split it and submit it as a separate PR.

It will show OVAL definition ID and title in the OVAL details section
of the HTML report. These values will be displayed similarly to
OVAL test ID and title.

However, that created confusing results (see screen shots in comments below). I improved that by using nested unordered lists, which basically mirror the tree structure of an OVAL definition.

We still need to improve the graphics layout, though.

### DIFF
--- a/xsl/xccdf-report-oval-details.xsl
+++ b/xsl/xccdf-report-oval-details.xsl
@@ -26,13 +26,15 @@ Authors:
 <xsl:stylesheet version="1.1"
     xmlns="http://www.w3.org/1999/xhtml"
     xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+    xmlns:ovaldef="http://oval.mitre.org/XMLSchema/oval-definitions-5"
     xmlns:ovalres="http://oval.mitre.org/XMLSchema/oval-results-5"
     xmlns:ovalsys="http://oval.mitre.org/XMLSchema/oval-system-characteristics-5"
     xmlns:ovalunixsc="http://oval.mitre.org/XMLSchema/oval-system-characteristics-5#unix"
     xmlns:ovalindsc="http://oval.mitre.org/XMLSchema/oval-system-characteristics-5#independent"
     exclude-result-prefixes="xsl ovalres ovalsys ovalunixsc ovalindsc">
 
-<xsl:key name='oval-definition' match='ovalres:definition'    use='@definition_id' />
+<xsl:key name='ovaldef-definition' match='ovaldef:definition' use='@id' />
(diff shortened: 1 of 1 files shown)


evgenyz:
OK, so what are your doubts? Or you just want this PR to be merged as it is (why still draft then)?

jan-cerny:
I think that the way the information is presented is ugly and can create confusion.

Especially when a definition is extended by another definition using extend_definition.

Before:
Screenshot_2020-06-04 xccdf_org open-scap_testresult_xccdf_org ssgproject content_profile_ospp OpenSCAP Evaluation Report

After:
Screenshot_2020-06-04 xccdf_org open-scap_testresult_xccdf_org ssgproject content_profile_ospp OpenSCAP Evaluation Report(1)

Notice that you can't see the hierarchy and it feels like some results are missing but in fact they aren't missing.

I think it would help to transform the whole section to a bullet list to reflect the tree structure of OVAL definitions.

evgenyz:
Aha, got it. I'll fiddle around and check what UI options we could have here.

jan-cerny:
If we use the lists, we get better results, and we can group criteria and add their operator and their results. See this example:

Screenshot_2020-06-08 xccdf_org open-scap_testresult_xccdf_org ssgproject content_profile_ospp OpenSCAP Evaluation Report

What do you think?

evgenyz:
I started exactly with this. Obvious problem is (as in any other tree-view with deep branching) – lot of wasted space on the left (would be even more painful when printed). 

You went beyond and introduced another problem: placement of operators. It's hard to understand that they are actually combining upper-level entries, not the checks immediately under them.

evgenyz:
But I like operators, can you please update PR with the latest code? We will iron out the layout.

jan-cerny:
Yes, it definitely needs more work.

jan-cerny:
I think it isn't possible to add just the original idea of "showing the OVAL definition ID" without any PR scope creep or making the report confusing.

I see following options:
1. Use the currently proposed code and on top of it improve the layout by using better CSS - colours, borders, paddings.
2. Apply the "clickable table tree" library, in a similar way that we use it in the report for the list of rules and groups.
3. Close the PR because it makes the report more complex and we have OVAL graph project that visualises the OVAL results better.

I would like to pursue option (1) or (2) because the report is something that users use the most and they start their investigation of failed rules based on the results present there, so having the report very detailed and accurate should help them.

PR #1436, opened 2019-12-10
Fixed oscap-chrootable dependency.

matejak:
so now one can run `sudo make oscap-chrootable` right away without calling `make oscap-chrootable-nocap` first.


jan-cerny:
@matejak This breaks the Windows build. Can it be guarded by `if(NOT WIN32)`?

evgenyz:
Great! Thanks!

PR #1439, opened 2019-12-19
Suppress SWIG-related warnings

evgenyz:
The suppression is already there, but it is not really effective for all of the warnings


jan-cerny:
@openscap-ci test this please

jan-cerny:
@evgenyz Thanks!

PR #1441, opened 2019-12-19
Add test that ensures there is no remediation attempt in the "offline" mode

evgenyz:
It would check for a message in *stderr* of the scanner, stating that `...remediation in offline mode: not implemented...`, when it is being executed with `--remediate` option in "offline" mode.


tests/API/XCCDF/unittests/test_remediation_offline.sh
```diff
@@ -0,0 +1,21 @@
...
+stderr=$(mktemp -t ${name}.err.XXXXXX)
+stdout=$(mktemp -t ${name}.out.XXXXXX)
+root=$(mktemp -d -t ${name}.root.XXXXXX)
+
+set_chroot_offline_test_mode "$root"
+
+$OSCAP xccdf eval --remediate --results $result $srcdir/${name}.xccdf.xml 2> $stderr 1> $stdout || echo "Scanner returned non-zero code (OK)"
+rm $result
+
+grep "remediation in offline mode: not implemented" $stderr > /dev/null
```
evgenyz: Maybe even 
```suggestion
grep "remediation in offline mode: not implemented" $stderr
```
for the line `remediation in offline mode: not implemented` to be present in the log.

jan-cerny: We usually suppress it (you can also use `grep -q`), but I'm fine either way.

zimmertr:
Why is remediation unavailable in offline mode? Is there any way to perform a remediation using `oscap-chroot`? Every time I try it successfully performs a scan and then fails to remediate with the aforementioned error message.

PR #1447, opened 2020-01-03
Added packit service configuration.

matejak:


redhatrises:
Please add the .spec file back so that we comply with the upstream first policy. In addition, most the upstream RH projects have the spec file in them.

dhodovsk:
packit failed with: 
```
ParserError: while parsing a block mapping   in "<unicode string>", line 23, column 3:       post-upstream-clone: 'bash -c "w ...        ^ expected <block end>, but found '<scalar>'   in "<unicode string>", line 23, column 139:      ... .spec && sed -i \"/^Patch/d\""' openscap.spec                                          ^
--
```

evgenyz:
Where are we with this? @matejak Do you need any help?

matejak:
> Please add the .spec file back so that we comply with the upstream first policy. In addition, most the upstream RH projects have the spec file in them.

1. Upstream first has nothing to do with the location of a spec file.
2. Modification to the spec file assume adherence to Fedora packaging guidelines, and there is no way how to ensure that upstream won't merge a PR that modifies the spec file in a way that would break those guidelines.

Therefore, I think that an aliasing the Fedora spec file is a good approach that combines zero duplication with being in touch with the downstream development.

matejak:
I have removed outdated comments that referred to past status of this PR and that completely lost their relevance.

.packit.yaml
```diff
@@ -0,0 +1,32 @@
...
+synced_files:
+- .packit.yaml
+upstream_package_name: openscap
+actions:
+  post-upstream-clone:
+    - "wget -N https://src.fedoraproject.org/rpms/openscap/raw/rawhide/f/openscap.spec"
+    - "sed -i '/^Patch/d' openscap.spec"
+    - "sed -i '23 a BuildRequires: xmlsec1-devel xmlsec1-openssl-devel' openscap.spec"
+    - "sed -i '24 a Requires: xmlsec1 xmlsec1-openssl' openscap.spec"
+    # use git_am instead of patch since packit patches are generated with git-format-patch
```
evgenyz, line 27:
```suggestion
    - "sed -E -i 's/%cmake (.+)$/%cmake -S \"%{_vpath_srcdir}\" -B \"%{__cmake_builddir}\" \\1/'"
    # use git_am instead of patch since packit patches are generated with git-format-patch
```

.packit.yaml
```diff
@@ -0,0 +1,33 @@
...
+synced_files:
+- .packit.yaml
+upstream_package_name: openscap
+actions:
+  post-upstream-clone:
+    - "wget -N https://src.fedoraproject.org/rpms/openscap/raw/rawhide/f/openscap.spec"
+    - "sed -i '/^Patch/d' openscap.spec"
+    - "sed -i '23 a BuildRequires: xmlsec1-devel xmlsec1-openssl-devel' openscap.spec"
+    - "sed -i '24 a Requires: xmlsec1 xmlsec1-openssl' openscap.spec"
+    - "sed -E -i 's/%cmake (.+)$/%cmake -S \"%{_vpath_srcdir}\" -B \"%{__cmake_builddir}\" \\1/'"
```
evgenyz, line 27:
```suggestion
    - "sed -E -i 's/%cmake (.+)$/%cmake -S \"%{_vpath_srcdir}\" -B \"%{__cmake_builddir}\" \\1/' openscap.spec"
```

.packit.yaml
```diff
@@ -0,0 +1,33 @@
...
+synced_files:
+- .packit.yaml
+upstream_package_name: openscap
+actions:
+  post-upstream-clone:
+    - "wget -N https://src.fedoraproject.org/rpms/openscap/raw/rawhide/f/openscap.spec"
+    - "sed -i '/^Patch/d' openscap.spec"
+    - "sed -i '23 a BuildRequires: xmlsec1-devel xmlsec1-openssl-devel' openscap.spec"
+    - "sed -i '24 a Requires: xmlsec1 xmlsec1-openssl' openscap.spec"
+    - "sed -E -i 's/%cmake (.+)$/%cmake -S \"%{_vpath_srcdir}\" -B \"%{__cmake_builddir}\" \\1/' openscap.spec"
```
evgenyz, line 27:
```suggestion
    - "sed -i '127 a %undefine __cmake_in_source_build' openscap.spec"
```

.packit.yaml
matejak: This won't work. I have tried to come up with a spec file without "if Fedora 32 then ... else" blocks downstream, as features like this shouldn't be delivered by means of monkey-patching.
It is possible to make `cmake` aware of the actual build directory (by passing it `-B %_vpath_builddir` that would be defined before, `__cmake_builddir` is not set correctly), but that would cause problems with the `cmake_build` macro.

Therefore, I propose to abandon efforts to please F32, as it will be irrelevant next month due to F34 being released.

evgenyz: Hmm, well the right thing to do for Fedora 32 is to `sed -i '127 a %undefine __cmake_in_source_build' openscap.spec` as I did in the latest commit.

evgenyz: You have to undefine __cmake_in_source_build in order to make out-of-source cmake-related macros to work. In F33+ is is done by default.

.openscap.spec
```diff
@@ -0,0 +1,702 @@
...
+%{_libdir}/libopenscap_sce.so.*
+
+%files containers
+%{_bindir}/oscap-docker
+%{_mandir}/man8/oscap-docker.8*
+%{python3_sitelib}/oscap_docker_python/*
+%{_bindir}/oscap-podman
+%{_mandir}/man8/oscap-podman.8*
+
+%changelog
```
evgenyz, line 205: Are you sure we want to have the changelog in the repo?

matejak:
I have rebased the PR and squashed commits that were exploring the dead end of monkey-patching the retrieved downstream spec file, as we have come to the conclusion that having a spec file in the repository is the only feasible option.
This spec file has a role of being the Fedora CI configuration rather than a mirror of the downstream Fedora spec file.

openscap.spec
```diff
@@ -0,0 +1,205 @@
+# This spec file is not synchronized to the Fedora downstream.
+# It serves as Fedora CI configuration and as support for downstream updates.
+Name:           openscap
+Version:        1.3.4
+Release:        4%{?dist}
```
evgenyz, line 5:
```suggestion
Release:        0%{?dist}
```

matejak: I wonder whether it's not worth to monkey-patch the version number according to the last tag in that branch or something like that.

openscap.spec
```diff
@@ -0,0 +1,205 @@
...
+BuildRequires:  libselinux-devel
+BuildRequires:  libcap-devel
+BuildRequires:  libblkid-devel
+BuildRequires:  bzip2-devel
+BuildRequires:  asciidoc
+BuildRequires:  openldap-devel
+BuildRequires:  glib2-devel
+BuildRequires:  dbus-devel
+BuildRequires:  libyaml-devel
+BuildRequires:	xmlsec1-devel xmlsec1-openssl-devel
```
evgenyz, line 29:
```suggestion
BuildRequires:  xmlsec1-devel xmlsec1-openssl-devel
```

openscap.spec
```diff
@@ -0,0 +1,205 @@
...
+# It serves as Fedora CI configuration and as support for downstream updates.
+Name:           openscap
+Version:        1.3.4
+Release:        0%{?dist}
+Epoch:          1
+Summary:        Set of open source libraries enabling integration of the SCAP line of standards
+License:        LGPLv2+
+URL:            http://www.open-scap.org/
+Source0:        https://github.com/OpenSCAP/%{name}/releases/download/%{version}/%{name}-%{version}.tar.gz
+BuildRequires: make
```
evgenyz, line 11:
```suggestion
BuildRequires:  make
```

openscap.spec
```diff
@@ -0,0 +1,205 @@
...
+Requires:       glib2
+Requires:       libacl
+Requires:       libblkid
+Requires:       libcap
+Requires:       libselinux
+Requires:       openldap
+Requires:       popt
+# Fedora has procps-ng, which provides procps
+Requires:       procps
+Requires:	xmlsec1 xmlsec1-openssl
```
evgenyz, line 46:
```suggestion
Requires:       xmlsec1 xmlsec1-openssl
```

evgenyz:
Okay, I'm merging it. Let's see how it would behave.

PR #1448, opened 2020-01-06
Do not emit warning when a rule has only OCIL or SCE

jan-cerny:
OCIL is a valid SCAP check, we only don't support it in OpenSCAP.  SCE
is supported by OpenSCAP, but users can use OpenSCAP without SCE plugin.
Instead we will display an info message (in verbose mode). We will still
keep the warning for really invalid check systems. Problem with warning
messages is that they're understood as errors by SCAP Workbench and the
diagnostic window pops up when they're produced.


evgenyz:
Great patch and test. Thank you!

PR #1451, opened 2020-01-08
make links showing remediation snippets in guides accessible

vojtapolasek:
Links showing and hiding remediation snippets in HTML guides generated by Openscap were not accessible to screenreader users. This is now fixed by adding appropriate parameters to the <a> tag in the XSLT template.

### DIFF
--- a/xsl/xccdf-share.xsl
+++ b/xsl/xccdf-share.xsl
@@ -284,7 +284,7 @@ Authors:
         </xsl:choose>
     </xsl:variable>
 
-    <span class="label label-success">Remediation <xsl:value-of select="$fix_type"/>:</span>&#160;&#160;&#160;<a data-toggle="collapse" data-target="#{generate-id($fix)}">(show)</a><br />
+    <span class="label label-success">Remediation <xsl:value-of select="$fix_type"/>:</span>&#160;&#160;&#160;<a data-toggle="collapse" data-target="#{generate-id($fix)}" tabindex="0" role="link" aria-expanded="false" href="#!">(show)</a><br />
     <div class="panel-collapse collapse" id="{generate-id($fix)}">
         <xsl:if test="$fix/@complexity or $fix/@disruption or $fix/@reboot or $fix/@strategy">
             <table class="table table-striped table-bordered table-condensed">
--- a/xsl/xccdf-share.xsl
+++ b/xsl/xccdf-share.xsl
@@ -284,7 +284,7 @@ Authors:
         </xsl:choose>
     </xsl:variable>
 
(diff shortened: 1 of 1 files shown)


evgenyz:
I have a suggestion. The control element kinda looks ugly. And now it's further marked as a link by the role attribute. Maybe it'd be better to make it a button?

xsl/xccdf-share.xsl
```diff
@@ -284,7 +284,7 @@ Authors:
         </xsl:choose>
     </xsl:variable>
 
-    <span class="label label-success">Remediation <xsl:value-of select="$fix_type"/>:</span>&#160;&#160;&#160;<a data-toggle="collapse" data-target="#{generate-id($fix)}">(show)</a><br />
+    <span class="label label-success">Remediation <xsl:value-of select="$fix_type"/>:</span>&#160;&#160;&#160;<a data-toggle="collapse" data-target="#{generate-id($fix)}" tabindex="0" role="link" aria-expanded="false" href="#!">(show)</a><br />
```
evgenyz: Even maybe:

```suggestion
    <a class="label label-success" data-toggle="collapse" data-target="#{generate-id($fix)}" tabindex="0" role="button" aria-expanded="false" title="Activate to reveal" href="#!">Remediation <xsl:value-of select="$fix_type"/>:</a><br />
```

evgenyz: @vojtapolasek Actually I don't know which would be better in this case `aria-label` or `title`.

evgenyz: This should look like
Screenshot from 2020-01-08 12-15-16

evgenyz:
@jan-cerny Can you please check this as well, while you after report's look and feel?

jan-cerny:
@evgenyz yes, I will

jan-cerny:
It looks great and works perfectly

But I have a problem. When I do the following:
1. click on rule in the rule overview table to show the pop-up window with the rule details
2. in the pop-up window click on the green button to show the remediation
3. close the pop-up window
4. scroll down
5. click on blue button "Show all result details"
6. scroll down find the same rule that you viewed in step 1
7. with that rule click on the green button to show the remediation
8. nothing is displayed, the button doesn't work

Maybe it's a very corner case and we can ignore that.

evgenyz:
Huh. You know your ways to break things, don't you? This is actually not related to this PR, it was brken before, I just checked. I'll try to find the exact problem.

evgenyz:
@jan-cerny Fixed the problem in #1454

jan-cerny:
@vojtapolasek you are great!

PR #1452, opened 2020-01-08
Improve printed HTML report

jan-cerny:
Improvements in the printed report:
* Do not show value of title attribute of abbr elements
* Do not crop count of rule results and severity
* Do not show "Scroll Back to top" and "(show)" links

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1640839


evgenyz:
Looks good to me in FF regarding the changes listed.
I see another problem, though:
Screenshot from 2020-01-08 15-18-42
Is it known? Is it ours? Should I file an issue?

evgenyz:
Old version looks somehow better:
Screenshot from 2020-01-08 15-31-01

jan-cerny:
@evgenyz Thanks, nice finding! Do you know what content is expected to be there instead of the cut line?

Regarding the grey lines merged with red lines, maybe I was very eager in reducing the empty space. I can set the padding to 1px or something. Another solution is to convert the red frame to black, because the rest of the printed report is black, and this is the only color thing in there.

evgenyz:
Colour printers are not a rarity nowadays; I think that failed rules with a red outline is a nice touch. Just give it a bit more space. And, maybe, some more thickness.

Content is there (in PDF) but it is partially hidden:
Screenshot from 2020-01-08 15-52-09

jan-cerny:
@evgenyz Check now! The crop happened because the whole OVAL details section was inside a table row (inside <tr> element) and table rows are not allowed by Bootstrap to have a page break inside. Also, it had set display:inline-block which also seems to prevent line breaks. I have solved it by moving the OVAL details section outside the big table and then remove the display:inline-block property from CSS class. But it means that the OVAL details section changed it's position, it's now at the end of the rule result panel.

evgenyz:
Cool, nice and tidy.

> But it means that the OVAL details section changed it's position, it's now at the end of the rule result panel.

It's even better now. Great! Thanks!

PR #1455, opened 2020-01-09
oscap-podman: add podman init to every scanned image/container.

ggbecker:
It makes sure that /run/.containerenv gets created.

Tested with ubi8-init from redhat and the output is:

```
$ oscap-podman registry.access.redhat.com/ubi8-init oval eval --id oval:ssg-installed_env_is_a_container:def:1 --oval-id scap_org.open-scap_cref_ssg-rhel8-oval.xml /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml 
Definition oval:ssg-installed_env_is_a_container:def:1: true
Evaluation done.
```

### DIFF
--- a/utils/oscap-podman
+++ b/utils/oscap-podman
@@ -77,7 +77,16 @@ else
     ID=$1
     TARGET="podman-container://$1"
 fi
+
+# podman init creates required files such as: /run/.containerenv - we don't care about output and exit code
+podman init $ID &> /dev/null || true
+
 DIR=$(podman mount $ID) || die
+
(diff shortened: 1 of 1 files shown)


matusmarhefka:
LGTM, but I will do some testing before merging.

matusmarhefka:
I have tested all the variations of image/container for ubi7 and ubi8 base images and CPE evaluation of `oval:ssg-installed_env_is_a_container:def:1` and `oval:ssg-installed_env_is_a_machine:def:1` works correctly. Good job @ggbecker

PR #1456, opened 2020-01-09
Do not run probes if they don't support offline mode

jan-cerny:
When the probe doesn't implement offline mode function it means
the probe doesn't support offline mode, therefore it should
set the supported_offline_mode flag to PROBE_OFFLINE_NONE.
This omission caused that the probe was run and scanned the host.

### DIFF
--- a/src/OVAL/probes/probe/worker.c
+++ b/src/OVAL/probes/probe/worker.c
@@ -969,6 +969,8 @@ SEXP_t *probe_worker(probe_t *probe, SEAP_msg_t *msg_in, int *ret)
 	probe_offline_mode_function_t offline_mode_function = probe_table_get_offline_mode_function(probe->subtype);
 	if (offline_mode_function != NULL) {
 		probe->supported_offline_mode = offline_mode_function();
+	} else {
+		probe->supported_offline_mode = PROBE_OFFLINE_NONE;
 	}
 
 	/*


src/OVAL/probes/probe/worker.c
```diff
@@ -969,6 +969,8 @@ SEXP_t *probe_worker(probe_t *probe, SEAP_msg_t *msg_in, int *ret)
 	probe_offline_mode_function_t offline_mode_function = probe_table_get_offline_mode_function(probe->subtype);
 	if (offline_mode_function != NULL) {
 		probe->supported_offline_mode = offline_mode_function();
+	} else {
+		probe->supported_offline_mode = PROBE_OFFLINE_NONE;
```
redhatrises: Should add a debug message here if possible as well.

evgenyz: In the worker thread function there is a warning already:
```c
        if (probe->supported_offline_mode == PROBE_OFFLINE_NONE) {
            dW("Requested offline mode is not supported by %s probe.", oval_subtype_get_text(probe->subtype));
            *ret = 0;
            return probe_cobj_new(SYSCHAR_FLAG_NOT_APPLICABLE, NULL, NULL, NULL);
```

evgenyz: Should it be a warning, thought? Maybe info?

evgenyz:
The purpose of this PR is to reach feature parity with 1.2 version in terms of the "offline mode" scanning. Back then probes were initialized with PROBE_OFFLINE_NONE flag and it was lost at some point.

We ought to investigate the answer we should return from the probe that is not applicable at the moment of execution. And in accordance with our findings we will change the log message and response object in the future, but right now it is out of the scope of current PR.

PR #1457, opened 2020-01-10
Add "offline" mode test for the password OVAL probe

evgenyz:
The probe does not support any of offline modes,
thus it should return "not applicable" response.


evgenyz:
I did not check for the warning message as it is not a part of the implementation.

jan-cerny:
It passes for me and it fails when the build doesn't contain #1456 as expected. Thanks.

PR #1459, opened 2020-01-13
Add a test to check --fix-type anaconda

jan-cerny:
There should be 2 equal ways of generating Anaconda remediations:
"oscap xccdf generate fix --fix-type anaconda" and
"oscap xccdf generate fix --template urn:redhat:anaconda:pre"
Both commands should give the same output.
This tests a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1736850
introduced by b1448ec95a957a76eb8be6d439531c532d97ff3c


evgenyz:
LGTM

PR #1460, opened 2020-01-14
1.3.2 release

evgenyz:


NEWS
```diff
@@ -1,3 +1,23 @@
+openscap-1.3.2                                                  13-01-2020
+  - New features
+    - Offline mode support for environmentvariable58 probe
+  - Maintenance, bug fixes
+    - Improved support of multi-check rules (report, remediations, console output)
+    - Improved HTML report look and feel, including printed version
+    - Less clutter in verbose mode output; some warnings and errors demoted to verbose mode levels
+    - Probe rpmverifyfile use and return canonical paths
```
jan-cerny: uses and returns

CMakeLists.txt
```diff
@@ -23,15 +23,15 @@ endif()
 # See http://sources.redhat.com/autobook/autobook/autobook_91.html#SEC91 for details
 
 ## increment if the interface has additions, changes, removals.
-set(LT_CURRENT 25)
+set(LT_CURRENT 26)
 
 ## increment any time the source changes; set 0 to if you increment CURRENT
```
jan-cerny: the comment says "set 0 to if you increment CURRENT"

NEWS
```diff
@@ -1,3 +1,23 @@
...
+    - The oscap-docker wrapper is available without Atomic
```
jan-cerny: Maybe this could be promoted to a new feature?

NEWS
```diff
@@ -1,3 +1,23 @@
+openscap-1.3.2                                                  13-01-2020
```
jan-cerny: I recall that we discussed with the team long time ago that we want to make the release notes more exciting, but I don't know how to do that. If you have any idea, feel free to be creative.

NEWS
evgenyz: Isn't it an implementation detail? Is there any new feature coming with this change?

NEWS
evgenyz: Mmm, I don't know. This update really feels like a maintenance release to me.

NEWS
jan-cerny: Yes and no, due to this change, Ubuntu users can now scan the Docker containers, which they could not do before, because Atomic isn't available on Ubuntu.

PR #1465, opened 2020-01-15
Fix case where CMake couldn't find libacl or xattr.h

redhatrises:
This might be a Fedora 31ism or a later version of CMake but found some issues where CMake wasn't finding some of the installed libraries.


jan-cerny:
Great discovery. I checked the latest F30 and F31 builds of OpenSCAP 1.3.2 are affected by this. And the older version `openscap-1.3.1-2.fc31.x86_64` which I use on my laptop doesn't contain fileextendedattribute probe in `oscap --version` as well.

redhatrises:
> Great discovery. I checked the latest F30 and F31 builds of OpenSCAP 1.3.2 are affected by this. And the older version `openscap-1.3.1-2.fc31.x86_64` which I use on my laptop doesn't contain fileextendedattribute probe in `oscap --version` as well.

@jan-cerny thanks. I was doing some experimentation and ran across this in passing.

PR #1466, opened 2020-01-16
utils/oscap-podman: Detect ambiguous scan target

matusmarhefka:
In case that a container image and a running container have the same
name, `oscap-podman` scans container image and a running container is
skipped. This might be unexpected and might cause a confusion for user.
Therefore, this commit adds a code which detects such situation and
rather informs user about ambiguous scan target and terminates.
In such cases the unique container image/container ID should be used
for specifying the target of the scan.

### DIFF
--- a/utils/oscap-podman
+++ b/utils/oscap-podman
@@ -65,17 +65,30 @@ if grep -q "\-\-remediate" <<< "$@"; then
     die
 fi
 
+IMAGE_NAME=$(podman image exists "$1" \
+    && podman image inspect --format "{{.Id}} {{.RepoTags}}" "$1")
+CONTAINER_NAME=$(podman container exists "$1" \
+    && podman container inspect --format "{{.Id}} {{.Name}}" "$1")
+
+if [ -n "$IMAGE_NAME" ] && [ -n "$CONTAINER_NAME" ]; then
+    echo "Ambiguous target, container image and container with the same name detected: '$1'." >&2
+    echo "Please rather use an unique ID to specify the target of the scan." >&2
+    die
+fi
+
 # Check if the target of scan is image or container.
 CLEANUP=0
-if podman images | grep -q $1; then
+if [ -n "$IMAGE_NAME" ]; then
     ID=$(podman create $1) || die
-    IMG_NAME=$(podman images --format "{{.ID}} ({{.Repository}}:{{.Tag}})" | grep -m1 $1)
-    TARGET="podman-image://$IMG_NAME"
+    TARGET="podman-image://$IMAGE_NAME"
     CLEANUP=1
-else
+elif [ -n "$CONTAINER_NAME" ]; then
     # If the target was not found in images we suppose it is a container.
     ID=$1
-    TARGET="podman-container://$1"
+    TARGET="podman-container://$CONTAINER_NAME"
+else
+    echo "Target of the scan not found: '$1'." >&2
+    die
 fi
 
 # podman init creates required files such as: /run/.containerenv - we don't care about output and exit code


utils/oscap-podman
```diff
@@ -65,6 +65,12 @@ if grep -q "\-\-remediate" <<< "$@"; then
     die
 fi
 
+if (podman images | grep -q $1) && (podman ps -a --format "{{.ID}} {{.Names}}" | grep -q $1); then
```
jan-cerny: The list of images could be reused later, but it's a nitpick.

utils/oscap-podman
```diff
@@ -65,6 +65,12 @@ if grep -q "\-\-remediate" <<< "$@"; then
...
+    echo "Ambiguous targer, container image and container with the same name detected: '$1'." >&2
```
jan-cerny: targer

matusmarhefka:
@jan-cerny Thx for review. I have reworked the code because I didn't like that it was not possible to scan for example target `ubi8` when you had `ubi8` and `ubi8-minimal` in a local container registry.

Please review and test once more. Scan target should look like this:
Screenshot from 2020-01-17 10-37-19

utils/oscap-podman
```diff
@@ -65,17 +65,53 @@ if grep -q "\-\-remediate" <<< "$@"; then
...
 
+# When searching the scan target in a local container registry, we use internal
+# representation of container images in the following format:
+#     ID:REPOSITORY:TAG:
+# This is to prevent as many collisions as possible, for example, in case there
+# would be 'ubi8' and 'ubi8-minimal' images in the local container registry, it
+# is possible to provide just 'ubi8' as a scan target. We would be looking for
+# the 'ubi8:' and therefore 'ubi8-minimal' would not be matched.
+# This approach prevents most collisions, but there still might be some. For
+# example, having container images '1fedora' and '2fedora' and specifying
```
jan-cerny: I think we should support only exact matches. Why do we have support for substrings?

matusmarhefka: Exact matches will not work in this case, only in case you would use the hash ID of the image (which is unique). For example, if you would have images like `registry.access.redhat.com/ubi8` and `registry.access.redhat.com/ubi8-minimal` and you would like to scan `ubi8` (not minimal), it would not help if you would type `registry.access.redhat.com/ubi8` because it would match also the `ubi8-minimal` so `oscap-podman` would need to produce an error. I find this annoying and the change in this commit tries to make it work in such cases.

On the other hand, I have no problem to change it so the error is printed, but IMO this is more user friendly.

matusmarhefka: Another reason why exact matches can be a problem is that user can also specify a tag, for example, `ubi8:latest`, `ubi8:8.2.0`, etc.

matusmarhefka: @jan-cerny Actually, I don't like my code, I will make it simpler and I would like to use `podman-image-exists`/`podman-container-exists`.

jan-cerny: I think that in your situation when you have images registry.access.redhat.com/ubi8 and registry.access.redhat.com/ubi8-minimal  and you request scan of registry.access.redhat.com/ubi8 the tool should scan only registry.access.redhat.com/ubi8 and not registry.access.redhat.com/ubi8-minimal. I understand registry.access.redhat.com/ubi8 as an unambiguous identifier, unless there are more of them with different tags.

If you want to add support for substring matches, then introduce some sort of glob or regular expressions. My opinion is that we don't need the substring matches, but we can have it if it works and is documented.

Let me know once you're done.

PR #1467, opened 2020-01-17
Covscan fixes

evgenyz:
Some fixes for covscan of 1.3.2

### DIFF
--- a/utils/oscap-chroot
+++ b/utils/oscap-chroot
@@ -25,6 +25,13 @@ function die()
     exit 1
 }
 
+function invalid()
+{
+    echo -e "$*\n" >&2
+    usage
+    exit 1
+}
+
 function usage()
 {
     echo "oscap-chroot -- Tool for offline SCAP evaluation of filesystems mounted in arbitrary paths."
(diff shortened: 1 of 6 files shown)


jan-cerny:
@evgenyz Looks great, but please rebase it and remove the "Merge" commit.

utils/oscap-chroot
```diff
@@ -74,26 +81,23 @@ function usage()
...
-    echo "Invalid arguments provided."
-    usage
-    die
+    invalid "Invalid arguments provided."
 fi
 
 # Learn more at https://www.redhat.com/archives/open-scap-list/2013-July/msg00000.html
 export OSCAP_PROBE_ROOT
-OSCAP_PROBE_ROOT="$(cd "$1"; pwd)"
+OSCAP_PROBE_ROOT="$(cd "$1" && pwd)"
```
matejak: This doesn't really handle the case when one can't cd into that directory - in that case, the `OSCAP_PROBE_ROOT` will be empty, which is not handled. I would go for calling `die` right afterwards if `OSCAP_PROBE_ROOT` is empty.

utils/oscap-podman
```diff
@@ -92,14 +93,16 @@ for VAR in `podman inspect $ID --format '{{join .Config.Env " "}}'`; do
     eval "export OSCAP_OFFLINE_$VAR"
 done
 
-export OSCAP_PROBE_ROOT="$(cd "$DIR"; pwd)"
+export OSCAP_PROBE_ROOT
+OSCAP_PROBE_ROOT="$(cd "$DIR" && pwd)"
```
matejak: ditto

utils/oscap-vm
```diff
@@ -128,7 +131,7 @@ fi
 
 # Learn more at https://www.redhat.com/archives/open-scap-list/2013-July/msg00000.html
 export OSCAP_PROBE_ROOT
-OSCAP_PROBE_ROOT="$(cd "$MOUNTPOINT"; pwd)"
+OSCAP_PROBE_ROOT="$(cd "$MOUNTPOINT" && pwd)"
```
matejak: ditto

utils/oscap-chroot
evgenyz: Added fall-back `die`.

utils/oscap-podman
evgenyz: Situation here is not really like it is in `oscap-chroot`, but added as well, for sake of completeness.

utils/oscap-vm
evgenyz: Situation here is not really like it is in `oscap-chroot`, but added as well, for sake of completeness.

evgenyz:
Rebased and updated.

matejak:
Thanks, merging!

PR #1468, opened 2020-01-21
Fix fileextendedattribute covscan errors

evgenyz:


evgenyz:
@openscap-ci test this, please

matejak:
Nice fixes, appreciate splitting the PR to two commits. Thanks you for the PR!

jan-cerny:
@matejak Next time please don't do mass indentation changes. People that apply patches on our code will have to rework them.

evgenyz:
@jan-cerny First of all, this probe was not used for a long time, there is a little chance of patches in the wild.

And if there are some patches in the wild for this, used to be broken, probe (without a PR against open-source, welcoming and all in all nice upstream) I would say that we should not be blamed for fixing really awful indentation mess.

PR #1469, opened 2020-01-22
Fix duplicate global variables

jan-cerny:
This should fix failed builds with GCC 10 on Rawhide.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1793914


evgenyz:
LGTM, thanks!

PR #1471, opened 2020-01-22
[DO NOT MERGE] Fixed ACL cmake refactoring.

matejak:
Completes #1465 - the change of pkg-config module has implications that need to be integrated into the build system setup.


matejak:
So this PR that swaps `ACL` with `LIBACL` works, but this is not the correct approach - there are at least two ways how to check for `libacl` in the project, probably each with its own defects, and they are somehow competing with each other. Proper solution for this is to have one (possibly compound) check that works.

PR #1472, opened 2020-01-23
Fix FindACL.cmake

jan-cerny:
find_path parameter `NAMES` values should be separated. According to
https://cmake.org/cmake/help/latest/command/find_path.html it should be:
`NAMES name1 [name2 ...]`

find_library parameter `NAMES` either should not contain `lib` or should
contain both `lib` and `.so.` The documentation at
https://cmake.org/cmake/help/latest/command/find_library.html says: Each
library name given to the `NAMES` option is first considered as a
library file name and then considered with platform-specific prefixes
(e.g. `lib`) and suffixes (e.g. `.so`).

This bug caused that even if cmake reported that libacl was found, the
library wasn't linked to the built `libopenscap.so`. Also,
`HAVE_ACL_EXTENDED_FILE`, `HAVE_ACL_LIBACL_H` and `HAVE_SYS_ACL_H` were
undefined in `config.h`, which caused some guarded pieces of code to not
compile, which means features missing.

### DIFF
--- a/cmake/FindACL.cmake
+++ b/cmake/FindACL.cmake
@@ -12,13 +12,13 @@ libfind_pkg_check_modules(ACL_PKGCONF libacl)
 
 # Include dir
 find_path(ACL_INCLUDE_DIR
-	NAMES "acl/libacl.h sys/libacl.h"
+	NAMES "acl/libacl.h" "sys/libacl.h"
 	PATHS ${ACL_PKGCONF_INCLUDE_DIRS}
 )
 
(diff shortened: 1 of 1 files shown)


matejak:
The output of `cmake` at configure-time as well as the respective file (incl. `CMakeCache.txt`) indicate that ACL checks are all positive when the library is present on the system.
The compiled library contains the `libacl` entry in the `NEEDED` section as well, which validates the assumption.
Great work!

PR #1473, opened 2020-01-26
Fix tests on Travis to not ignore their results

jan-cerny:
This should fix tests on Travis.

### DIFF
--- a/tests/probes/rpm/rpminfo/test_probes_rpminfo.sh
+++ b/tests/probes/rpm/rpminfo/test_probes_rpminfo.sh
@@ -17,7 +17,7 @@
 
 set -e -o pipefail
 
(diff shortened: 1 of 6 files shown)


jan-cerny:
@openscap-ci test this please

tests/probes/rpm/rpminfo/test_probes_rpminfo.sh
```diff
@@ -17,10 +17,11 @@
 
 set -e -o pipefail
 
-require "rpm" || return 255
+require "rpm" || exit 255
```
evgenyz:
```suggestion
require rpm || exit 255
```

evgenyz:
LGTM! Thanks!

PR #1474, opened 2020-01-27
Prevent infinite recursion in systemdunitdependency probe

jan-cerny:
Systemd units trees can contain cycles in them, which causes infinite recursion in our probe.

We will save processed units in a hash table and if we meet an unit that is already processed we won't process it again.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1478285
See comment 11 in this BZ for a reproducer.

This PR also contains some refactoring to make the bug fix easier. For more details about specific changes please read the commit messages of respective commits.


evgenyz:
I've tried to break it with different types of broken dependencies and it looks like we are good now. LGTM, merging. Thank you.

PR #1476, opened 2020-01-28
Do not drop empty lines from Ansible remediations

jan-cerny:
We are removing empty lines from generated Ansible Playbook.  It looks
nice, but there is a catch.

Ansible tasks are a YAML snippets. So they contain YAML block scalars,
indicated by a pipe (|). Within these block scalars, the newlines should
be preserved.  By removing the empty lines from the scalars we are
actually changing the meaning of the scalars.

Unfortunately, OpenSCAP doesn't understand YAML, but only reads the
Ansible fix line-by-line as s string. That means it doesn't know when it
reads a block scalar. Therefore we need to stop removing the empty lines
anywhere.

This commit also adds a small test.

Resolves: RHBZ#1795563

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -139,11 +139,10 @@ static int _write_remediation_to_fd_and_free(int output_fd, const char* template
 					free(text);
 					return 1;
 				}
-
-				if (_write_text_to_fd(output_fd, "\n") != 0) {
-					free(text);
-					return 1;
-				}
+			}
+			if (_write_text_to_fd(output_fd, "\n") != 0) {
+				free(text);
+				return 1;
 			}
 
 			if (next_delim != NULL) {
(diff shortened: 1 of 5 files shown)


jan-cerny:
The problem with this simple solution is that it adds a lot of newlines to the playbooks. For example. the RHEL7 OSPP playbook grows from 4228 lines by 305 lines to 4533 lines.

We could start removing all the newline characters from the end of the contents of the <fix> element until first non-newline character. What do you think?

jan-cerny:
Bump

evgenyz:
@jan-cerny Makes sense, can't see any drawbacks, but it won't get us much I suppose.

vojtapolasek:
By newlines in playbooks you mean trailing newlines at the end of the file?

jan-cerny:
@vojtapolasek I meant empty lines anywhere in the file. And I propose reducing the trailing newlines at each rule block, i.e. at the end of tasks.

vojtapolasek:
I understand. Although I am not sure why excessive empty lines are a problem.

jan-cerny:
@vojtapolasek They shouldn't be a problem, they only look ugly and increase the file size.

jan-cerny:
OK, let's not do it.

evgenyz:
PR fixes the problem in BZ and we decided not to proceed with micro-optimizations. Merging.

Issue #1449, opened 2020-01-06
Timezone missing in XCCDF TestResult start-time and end-time


akofink:
#### Description of Problem:
XCCDF scan results outputs a test result file with a TestResult node like the following:

```
<TestResult id="xccdf_org.open-scap_testresult_xccdf_org.ssgproject.content_profile_standard" start-time="2020-01-03T18:39:24" end-time="2020-01-03T18:40:35" version="0.1.46" test-system="cpe:/a:redhat:openscap:1.2.17">
```

The start-time and end-time do not have any timezone information, and the times printed are in the local scanned system's time zone. This can be worked around by passing `TZ=UTC` to oscap, but the time zone information really ought to be there.

#### OpenSCAP Version:
openscap-1.2.17-4.el7.x86_64

#### Operating System & Version:
RHEL 7.7

#### Steps to Reproduce:

1. Evaluate a system
```sh
$ sudo oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_standard --results ~/rhel7-xccdf_org.ssgproject.content_profile_standard.xml /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml
```
2. Check the TestResult start-time and end-time

#### Actual Results:
Time zone information is missing
```sh
$ grep end-time ~/rhel7-xccdf_org.ssgproject.content_profile_standard.xml
  <TestResult id="xccdf_org.open-scap_testresult_xccdf_org.ssgproject.content_profile_standard" start-time="2020-01-03T18:39:24" end-time="2020-01-03T18:40:35" version="0.1.46" test-system="cpe:/a:redhat:openscap:1.2.17">
```

#### Expected Results:
The TestResult node output by oscap should match the spec and include time zone information:

> <xccdf:TestResult id="xccdf_org.example_testresult_ios-test5" end-time="2007-09-25T7:45:02-04:00" xmlns:xccdf="http://checklists.nist.gov/xccdf/1.2" version=”1.0”>

#### Additional Information / Debugging Steps:
It seems the dateTime type in the XCCDF 1.2 spec is not explicitly defined in that document but rather, is defined in the XML spec: https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#dateTime as ISO 8601. This is referenced in the XCCDF 1.2 spec, Chapter 2: Normative References.

evgenyz:
[...]

PR #1477, opened 2020-02-05
Add timezone to XCCDF TestResult start/end time

jan-cerny:
We add timezone information to the evaluation start date
and time using the format defined in XML specification
https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#dateTime

We can't use strftime as strftime's %z modifier doesn't
put ':' (colon) between hours and minutes in timeshift
which is required by XML specification.

Fixes: #1449


evgenyz:
I wonder if glibc will finally be fully compatible with ISO 8601 at some day... Anyhow, LGTM. Nice improvement!

PR #1478, opened 2020-02-06
Fix #1453 pointer lost in Swig API

### LINKED ISSUE #1453: [SWIG] Argument pointer destroyed too early in xccdf_session_set_rule
#### Description of Problem:
Hi, I've encountered recently a serious bug in python bindings.
Example:

```
import openscap_api as oscap
sess = oscap.xccdf.session_new("./samples/xccdf_sample.xml")
rule="R-SHOULD_PASS"
sess.set_rule(rule)
sess.load() # should return 0 but return 1
sess.evaluate()
```
After an horrible debugging i've found that when xccdf_session_set_rule is called, the rule is well stored, but when xccdf_session_load is called, the pointer is lost (random/inconsistant value). I've modify xccdf_session_set_rule by adding a malloc/strcpy before store the value in session->rule, and the problem doesn't exists anymore.

Of course it's not a good solution. The best compromise that i've found is to add a wrap function in file openscap.i like

```
void xccdf_session_set_rule_py(struct xccdf_session  *sess, const char *rule) {
    char *n_rule=(char*)malloc((strlen(rule)+1)*sizeof(char));
    xccdf_session_set_rule(sess, rule);
}
```
I've tested that an it works (to avoid memory leak we should also wrapp the xccdf_session_destroy function in orde to release the pointer). I write this as an issue because it's not a trivial problem, and it can probably happens with other functions (i guess that the char* agrument is released when the call to set_rule is terminated ... ).

So, what o you suggest in order to mitigate these problems ? I will propose some wrapping functions in openscap.i in order to have a working API but it's not the better mitigation ever.

What do you think about these kinds of problems concerning SWIG bindings ?

#### OpenSCAP Version:
maint-1.3

Ubuntu 18.04

DominiqueDevinci:
Hello,
Some time ago i've signalled a serious issue #1453  in the python binding, the xccdf_session_set_rule was just not working.

Here is a proposal of correction, including two more unit tests which ensure it's working now.
I can understand some technical choices are debatable in this solution but i've to be as clear as possible. If you have a better proposal i'm all ears !

Regards,
Dominique.

### DIFF
--- a/swig/openscap.i
+++ b/swig/openscap.i
@@ -549,10 +549,24 @@ int oval_agent_eval_system_py(oval_agent_session_t * asess, PyObject * func, PyO
     return oval_agent_eval_system(asess, agent_reporter_callback_wrapper, (void *) new_usrdata);
 }
 
-void xccdf_session_set_rule_py(struct xccdf_session  *sess, const char *rule) {
+/* we only recopy the first part of the struct, in order to free the rule property
+* since it's defined in xccdf_session.c and not xccdf_session.h, this struct is unknown
+* and so it's not possible to access to the rule property without defining it
+*/
+struct xccdf_session {
+        char *filename;
+        char *rule;
+};
+
+void xccdf_session_set_rule_py(struct xccdf_session  *sess, char *rule) {
     char *n_rule=(char*)malloc((strlen(rule)+1)*sizeof(char));
     strcpy(n_rule, rule);
     xccdf_session_set_rule(sess, n_rule);
+}
+
+void xccdf_session_free_py(struct xccdf_session *sess){
+    free(sess->rule);
+    xccdf_session_free(sess);
 }
 
 %}
--- a/swig/openscap.i
+++ b/swig/openscap.i
@@ -559,8 +559,7 @@ struct xccdf_session {
 };
 
 void xccdf_session_set_rule_py(struct xccdf_session  *sess, char *rule) {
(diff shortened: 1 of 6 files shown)


swig/openscap.i
```diff
@@ -549,5 +549,25 @@ int oval_agent_eval_system_py(oval_agent_session_t * asess, PyObject * func, PyO
...
+* and so it's not possible to access to the rule property without defining it
+*/
+struct xccdf_session {
+        char *filename;
+        char *rule;
+};
+
+void xccdf_session_set_rule_py(struct xccdf_session  *sess, char *rule) {
+    char *n_rule=(char*)malloc((strlen(rule)+1)*sizeof(char));
+    strcpy(n_rule, rule);
```
jan-cerny: Use `strdup` instead.

swig/openscap_api.py
```diff
@@ -273,13 +273,24 @@ def __getattr__(self, name):
         if name in self.__dict__:
             return self.__dict__[name]
 
+        # cache self.object + "_" + name for perf / clearness purpose
+        obj_name = self.object + "_" + name
+
+        """ Catch  potential C function overriden in OSCAP_Object
+        like xccdf_session_set_rule ir xccdf_session_free for instance. """
```
jan-cerny: ir should be or?

DominiqueDevinci:
Yes you are right, fixed :)

jan-cerny:
Thanks for your contribution.

PR #1479, opened 2020-02-13
Add simple tests for yamlprobe

yuumasato:
#### Description
- Add a few simple tests using `yamlfilecontent_test` to exercise not-yet-existing `yamlfileprobe` (time to TDD).


tests/probes/yamlfilecontent/test_probes_yamlfilecontent.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+. $builddir/tests/test_common.sh
+
+function test_probes_yamlfilecontent {
+
+    probecheck "yamlfilecontent" || return 255
```
yuumasato: This line causes the test to be skipped.

Removing it will run the test, and cause fail due to invalid schema:
```
File '/home/wsato/git/openscap/tests/probes/yamlfilecontent/test_probes_yamlfilecontent.xml' line 40: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#independent}yamlfilecontent_test': This element 
is not expected. Expected is one of ( {http://oval.mitre.org/XMLSchema/oval-definitions-5}test, {http://oval.mitre.org/XMLSchema/oval-definitions-5#aix}interim_fix_test, {http://oval.mitre.org/XMLSchema/oval-def
initions-5#aix}fileset_test, {http://oval.mitre.org/XMLSchema/oval-definitions-5#aix}fix_test, {http://oval.mitre.org/XMLSchema/oval-definitions-5#aix}no_test, {http://oval.mitre.org/XMLSchema/oval-definitions-5
#aix}oslevel_test, {http://oval.mitre.org/XMLSchema/oval-definitions-5#apache}httpd_test, {http://oval.mitre.org/XMLSchema/oval-definitions-5#catos}line_test, {http://oval.mitre.org/XMLSchema/oval-definitions-5#
catos}module_test, {http://oval.mitre.org/XMLSchema/oval-definitions-5#catos}version55_test ).
```

yuumasato:
@evgenyz @matejak I'd appreciate any feedback on the tests and the direction it is going.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent.sh
evgenyz: When we will have the probe test will start to fail and we will be fixing it, that line is OK. But, what we could do to fix the schema *before* we officially commit and publish it in OVAL?

tests/probes/yamlfilecontent/test_probes_yamlfilecontent.xml
```diff
@@ -0,0 +1,136 @@
...
+    </ind-def:yamlfilecontent_test>
+
+  </tests>
+
+  <objects>
+
+    <ind-def:yamlfilecontent_object version="1" id="oval:0:obj:1">
+      <ind-def:path>/tmp</ind-def:path>
+      <ind-def:filename>openshift-logging.yaml</ind-def:filename>
+      <ind-def:yamlpath>/kind</ind-def:yamlpath>
```
evgenyz: Current YAML Path definition is here: https://github.com/evgenyz/libyaml-yamlpath-filter/wiki/YAML-Path-(OVAL-subset). It became even smaller; for now there are only *keys*, *indexes* and *slices* (for lists, not for maps). And only dot notation (to be more compatible with JSONPath).

You can get the idea of what it is capable of right now here: https://github.com/evgenyz/libyaml-yamlpath-filter/blob/afb1fec90a5eb3974d1115bab1d1bfff4956b91a/test-paths.c#L170.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent.sh
jan-cerny: I think we can add OVAL 5.11.3 schemas (as a future version) and then modify them. We should add proper comments there.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent.xml
```diff
@@ -0,0 +1,136 @@
...
+      <ind-def:filename>openshift-logging.yaml</ind-def:filename>
+      <ind-def:yamlpath>/spec/outputs[type=forward]/name</ind-def:yamlpath>
+      <ind-def:instance operation="greater than or equal" datatype="int">0</ind-def:instance>
+    </ind-def:yamlfilecontent_object>
+
+    <ind-def:yamlfilecontent_object version="1" id="oval:0:obj:4">
+      <ind-def:path>/tmp</ind-def:path>
+      <ind-def:filename>openshift-logging.yaml</ind-def:filename>
+      <ind-def:yamlpath>/spec/outputs[type=elasticsearch]/name</ind-def:yamlpath>
+      <ind-def:instance operation="greater than or equal" datatype="int">0</ind-def:instance>
```
jan-cerny: I can't see `instance` element in the schema in https://github.com/OVAL-Community/OVAL/pull/90.

evgenyz: I designed it after `xmlfilecontent` probe. There is no instance as well. I guess we don't need it because we can distinctively enumerate all the nodes in the YAML document, unlike `filecontent` probe.

yuumasato: `instance` is also about selecting how many or which collected objects should be evaluated in the test.
If a yamlpath yields two values, like this test would, should it evaluate to error?

evgenyz: How `xmlfilecontent` is able to return *a list of text strings* without it?

yuumasato: I misinterpreted the `instance` field., it can return multiple values, it just won't let you choose the objects.

yuumasato: I'll remove the `instance`.

jan-cerny: great!

tests/probes/yamlfilecontent/test_probes_yamlfilecontent.xml
yuumasato: > for now there are only _keys_, _indexes_ and _slices_ (for lists, not for maps).

I'm not sure if array slicing will be that useful. Can we be sure of the order of the items in a list?

So hash key searching is not supported? :cry:
I think key searching will be the most useful feature.

evgenyz: There is a case already for slices (which could be used as *any* filter for sequences), in the example we have we are looking for at least one record of certain type. The path is `.spec.pipelines[:].inputSource` and result would be `['logs.app', 'logs.infra', 'logs.audit']`. We are after logs.audit.

evgenyz: Key searching is another thing. You can address anything by an exact key. But there is no any kind of fuzzy key search right now, until we prove it to be useful. I try to keep it as dump as possible to actually have all the logic in OVAL territory.

evgenyz: > Can we be sure of the order of the items in a list?

Yes, it'd be the same as in the document.

yuumasato: > > Can we be sure of the order of the items in a list?
> 
> Yes, it'd be the same as in the document.

I'm more concerned about the order in the yaml file.
Is it the case that a field we are looking for, will always be the first entry in list?

yuumasato: > There is a case already for slices (which could be used as _any_ filter for sequences), in the example we have we are looking for at least one record of certain type. The path is `.spec.pipelines[:].inputSource` and result would be `['logs.app', 'logs.infra', 'logs.audit']`. We are after logs.audit.

In this example you want to know if any pipeline has `logs.audit` as inputSource.
Can there be an use case in which you want to retrieve a certain field from a pipeline that has `inputSource: logs.audit`?

evgenyz: No, of course it won't. But when we are dealing with arrays, we, in general, have to check all the items for that reason. I understand that `[type='logs.app']` feature is attractive. But, in addition to another round of filter library development (this check can't be done in a single pass), we would face the need for some type coercion in the probe, which we are trying to avoid.

Lets just stick to what we have right now, and evolve with the content, should the need arise.

evgenyz: > Can there be an use case in which you want to retrieve a certain field from a pipeline that has inputSource: logs.audit?

*There are more things in heaven and Earth, Horatio, 
Than are dreamt of in your philosophy.* 

There *might* be, for sure. Even the aforementioned case could be rewritten as `.spec.pipelines[inputSource=logs.audit]` and then just have to check if results count >1. But I want to be sure that workaround is ugly enough or impossible before starting to deal with all the problems related to [XXX=YYY].

yuumasato: Playing around with `yaml-get`:

```bash
$ yaml-get -p '.spec.pipelines[:].inputSource' openshift-logging.yaml
CRITICAL:  : is not an integer array slice at ':' in 'spec.pipelines[:].inputSource'.                                                                                                                              
$ yaml-get -p '.spec.pipelines[0:2].inputSource' openshift-logging.yaml                                                                                         
logs.app                                                                                                 
logs.infra                                                                                               
$ yaml-get -p '.spec.pipelines[0:5].inputSource' openshift-logging.yaml                                                                                         
logs.app                                                                                                 
logs.infra                                                                                               
logs.audit                                          
$ yaml-get -p '.spec.pipelines.inputSource' openshift-logging.yaml                                                                                             
logs.app                                                                                                 
logs.infra                                                                                                                                                                                                         
logs.audit          
```

Is the [python style slicing syntax](https://github.com/evgenyz/libyaml-yamlpath-filter/wiki/YAML-Path-(OVAL-subset)#slices-arrays) specific to [libyaml-yamlpath-fitler](https://github.com/evgenyz/libyaml-yamlpath-filter)?
`yaml-get` doesn't like it.

@evgeni Would behavior of the last command work with the filter?

evgenyz: My (and Matej's) beloved example: https://docs.ansible.com/ansible/latest/modules/file_module.html#parameter-mode.

From some Ansible file you want to collect paths of files with access mode is `0777`. What will you write in `[mode=XXXXXX?]`. What will you do in the probe when you get this XXXXX (as a char*, with the only notion about type that it was explicitly quoted or wasn't)?

evgeni: I am not the Evgeni you're looking for ;)

yuumasato: > Lets just stick to what we have right now, and evolve with the content, should the need arise.

I understand, we need to start writing content as soon as possible then, to collect use cases, even when the probe is not ready.

> But I want to be sure that workaround is ugly enough or impossible before starting to deal with all the problems related to [XXX=YYY].

At the same time, one of the goals of the probe is not to have to write workarounds or complex OVAL contents.

yuumasato: @evgeni Apologies, :open_mouth:

evgeni: No worries :)

evgenyz: > At the same time, one of the goals of the probe is to avoid need for workarounds or complex OVAL contents.

Balance. That is what we usually looking for :)

>  I understand, we need to start writing content as soon as possible then, to collect use cases, even when the probe is not ready.

Exactly. So I'm trying to create something to give our OCP colleagues tools to start working and pinpointing weak spots.

> Is the python style slicing syntax specific to libyaml-yamlpath-fitler?
yaml-get doesn't like it.

Yes, it is. And it is how things are done in JSONPath. I don't like YAML Paths' version because you can't collect the whole list with it.

YAML Paths' `'.spec.pipelines.inputSource'` is too ambiguous. It completely skips the list in the middle, inputSource may as well be just a key in pipelines dictionary, which is not acceptable at all. It is so because YAML Path uses so called *arrays pass-through*, which is not good from a security-related tool point of view, IMHO.

evgenyz: BTW, if you want to play with *our* (he-he) YAML Path, go ahead: https://github.com/OpenSCAP/yaml-filter. We even have a console tool now :)

evgenyz:
I feel like you overdoing this right now. We have unit tests for YAML Path behaviour in the library. We are after the probe integration tests: no results, 1 result, multiple results, errors (no file, invalid path, path pointing to a non-scalar value).

tests/probes/yamlfilecontent/test_probes_yamlfilecontent.sh
jan-cerny: @evgenyz @yuumasato We need an agreement on the schemas, because it's a blocker for both test and implementation (you can temporarily do --skip-valid but I don't want to see anything like that in our repo). Do you agree with my proposal from the previous comment? If yes, I can create a separate PR for schemas.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+. $builddir/tests/test_common.sh
+
+function test_probes_yamlfilecontent_array {
+
+    probecheck "yamlfilecontent" || return 255
+
+    local ret_val=0;
```
jan-cerny: remove the semicolon

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+. $builddir/tests/test_common.sh
+
+function test_probes_yamlfilecontent_key {
+
+    probecheck "yamlfilecontent" || return 255
+
+    local ret_val=0;
```
jan-cerny: remove semicolon

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.sh
```diff
@@ -0,0 +1,33 @@
...
+    local YAML_FILE="/tmp/openshift-logging.yaml"
+
+    $OSCAP oval eval --results $RF $DF
+
+    if [ -f $RF ]; then
+	verify_results "def" $DF $RF 2 && verify_results "tst" $DF $RF 3
+	ret_val=$?
+    else
+	ret_val=1
+    fi
```
jan-cerny: borked indents, please use spaces or tabs consistently

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0"?>
+<oval_definitions xmlns:oval-def="http://oval.mitre.org/XMLSchema/oval-definitions-5" xmlns:oval="http://oval.mitre.org/XMLSchema/oval-common-5" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:ind-def="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent" xmlns:unix-def="http://oval.mitre.org/XMLSchema/oval-definitions-5#unix" xmlns:lin-def="http://oval.mitre.org/XMLSchema/oval-definitions-5#linux" xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5" xsi:schemaLocation="http://oval.mitre.org/XMLSchema/oval-definitions-5#unix unix-definitions-schema.xsd http://oval.mitre.org/XMLSchema/oval-definitions-5#independent independent-definitions-schema.xsd http://oval.mitre.org/XMLSchema/oval-definitions-5#linux linux-definitions-schema.xsd http://oval.mitre.org/XMLSchema/oval-definitions-5 oval-definitions-schema.xsd http://oval.mitre.org/XMLSchema/oval-common-5 oval-common-schema.xsd">
+
+  <generator>
+    <oval:product_name>yamlfilecontent</oval:product_name>
+    <oval:product_version>1.0</oval:product_version>
+    <oval:schema_version>5.10</oval:schema_version>
```
jan-cerny: The schema version is questionable, I propose 5.11.3, see previous discussion.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.sh
```diff
@@ -0,0 +1,33 @@
...
+    local YAML_FILE="/tmp/openshift-logging.yaml"
+
+    $OSCAP oval eval --results $RF $DF
+
+    if [ -f $RF ]; then
+	verify_results "def" $DF $RF 2 && verify_results "tst" $DF $RF 3
+	ret_val=$?
+    else
+	ret_val=1
+    fi
```
jan-cerny: indents

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
```diff
@@ -0,0 +1,88 @@
...
+      <ind-def:object object_ref="oval:0:obj:1"/>
+      <ind-def:state state_ref="oval:0:ste:1"/>
+    </ind-def:yamlfilecontent_test>
+
+    <ind-def:yamlfilecontent_test version="1" id="oval:0:tst:2" check="all" comment="true">
+      <ind-def:object object_ref="oval:0:obj:2"/>
+      <ind-def:state state_ref="oval:0:ste:2"/>
+    </ind-def:yamlfilecontent_test>
+
+    <ind-def:yamlfilecontent_test version="1" id="oval:0:tst:3" check="all" comment="error">
```
jan-cerny: Why is error expected?

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.sh
```diff
@@ -0,0 +1,33 @@
...
+
+    rm -f $YAML_FILE
+
+    return $ret_val
```
jan-cerny: Where do you test that the correct data was extracted from the YAML? Please add assertions on collected items.

yuumasato: The tests have `states`, if they pass doesn't it mean we have collected the correct data?

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
```diff
@@ -0,0 +1,88 @@
...
+      <ind-def:object object_ref="oval:0:obj:5"/>
```
yuumasato: The object here should be `oval:0:obj:3`.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
yuumasato: Because the path `.metadata` on the test yaml file will return an non-scalar: https://github.com/OpenSCAP/openscap/pull/1479/files/68289f537fbb8607212c56e6300d413af24b24be#diff-5ca741cc9861804e1ff83bfe153329cbR3

tests/probes/yamlfilecontent/test_probes_yamlfilecontent.sh
yuumasato: @jan-cerny Yes, your suggestion makes sense to me.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
jan-cerny: OK

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.sh
jan-cerny: Probably yes

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
```diff
@@ -0,0 +1,99 @@
...
+      <ind-def:filename>openshift-logging.yaml</ind-def:filename>
+      <ind-def:yamlpath>.spec.outputs[0]</ind-def:yamlpath>
+    </ind-def:yamlfilecontent_object>
+
+  </objects>
+
+  <states>
+
+    <ind-def:yamlfilecontent_state version="1" id="oval:0:ste:1">
+      <ind-def:value-of datatype="string">secureforward-offcluster</ind-def:value-of>
```
jan-cerny: According to schema in https://github.com/OVAL-Community/OVAL/pull/90/files the element should be `value_of`, with an underscore.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
```diff
@@ -0,0 +1,99 @@
...
+    </ind-def:yamlfilecontent_state>
+
+    <ind-def:yamlfilecontent_state version="1" id="oval:0:ste:2">
+      <ind-def:value-of datatype="string" var_ref="oval:0:var:2"/>
```
jan-cerny: dtto

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
```diff
@@ -0,0 +1,99 @@
...
+    </ind-def:yamlfilecontent_state>
+
+  </states>
+
+  <variables>
+
+    <local_variable comment="oval:0:var:2" datatype="string" version="1" id="">
```
jan-cerny: id can't be empty

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
```diff
@@ -0,0 +1,88 @@
...
+  </objects>
+
+  <states>
+
+    <ind-def:yamlfilecontent_state version="1" id="oval:0:ste:1">
+      <ind-def:value-of datatype="string">LogForwarding</ind-def:value-of>
+    </ind-def:yamlfilecontent_state>
+
+    <ind-def:yamlfilecontent_state version="1" id="oval:0:ste:2">
+      <ind-def:value-of datatype="string">openshift-logging</ind-def:value-of>
```
jan-cerny: dtto

tests/probes/yamlfilecontent/test_probes_yamlfilecontent.sh
jan-cerny: See https://github.com/OpenSCAP/openscap/pull/1482

yuumasato:
> I feel like you overdoing this right now.

> I think we have a perfect testing OVAL content now,

:smiley: 

> We have unit tests for YAML Path behaviour in the library.  We are after the probe integration tests: no results, 1 result, multiple results, errors (no file, invalid path, path pointing to a non-scalar value).

@evgenyz I don't understand what you are missing.
I believe these work as integration tests, and most of what you mention is already covered.

matejak:
> @evgenyz I don't understand what you are missing.
> I believe these work as integration tests, and most of what you mention is already covered.

I think that Evgeny is not happy with the library being tested twice - once in respective unit-tests and basically once by analysing OVAL results produced by the probe that uses the library.
I am not against this "double testing", as I find it as an appropriate "black box" testing of the probe.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
```diff
@@ -0,0 +1,124 @@
...
+    <ind-def:yamlfilecontent_object version="1" id="oval:0:obj:4">
+      <ind-def:path>/tmp</ind-def:path>
+      <ind-def:filename>openshift-logging.yaml</ind-def:filename>
+      <ind-def:yamlpath>.doesnt.exist</ind-def:yamlpath>
+    </ind-def:yamlfilecontent_object>
+
+    <ind-def:yamlfilecontent_object version="1" id="oval:0:obj:5">
+      <ind-def:path>/tmp</ind-def:path>
+      <ind-def:filename>yaml-filter.yaml</ind-def:filename>
+      <ind-def:yamlpath>.foo[1].some\.el\/here</ind-def:yamlpath>
```
matejak: The test yaml file modification is needed in order for this to pass, right?

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.sh
matejak: What about the that expected error result? And what about that non-existing scalar? The test could assume that no objects are collected.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
evgenyz: Actually back-slash escaping is not supported so far. This won't work no matter what, ATM.

evgenyz:
>> @evgenyz I don't understand what you are missing.
>> I believe these work as integration tests, and most of what you mention is already covered.

> I think that Evgeny is not happy with the library being tested twice - once in respective unit-tests and basically once by analysing OVAL results produced by the probe that uses the library.

Yeah, I'm grumbling about excessive tests, not because something is missing. While excessive test might be seemed as a neutral or even a positive feature, one should always think that there is a toll of test maintenance (and execution time, BTW) in a developing library (such as libyamlpath) and there is usually no gains from this duplicity.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
yuumasato: @evgenyz Do we expect to support it?

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
yuumasato: Updated to `5.11.3`

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
yuumasato: > The test yaml file modification is needed in order for this to pass, right?

I forgot to add `yaml-filter.yaml` into the commit.
But I'll drop the commit for now.

evgenyz: Yes, we expect. Working on it. But you'd better not include it in this PR.

yuumasato:
> Yeah, I'm grumbling about excessive tests, not because something is missing. While excessive test might be seemed as a neutral or even a positive feature, one should always keep in mind that there is a toll of test maintenance (and execution time, BTW) in a developing library (such as libyamlpath) and usually there are no gains from this duplicity.

I'm happy to remove anything that is excessive and duplicate.
But just like Matěj, I see these as black box tests for the probe. They will define expected behavior of the probe and how content would look like.  They are not focused in finding bugs in yaml-filter.
CC: @matusmarhefka, maybe you have opinions ^

yuumasato:
Per https://github.com/OpenSCAP/openscap/pull/1479#discussion_r380355105 I've dropped escaped key test: b61d5f9f8cb6afa0540ff5a80dd539edc7503a9b

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.sh
yuumasato: Per `yamlpath` description in https://github.com/OVAL-Community/OVAL/pull/90/files#diff-ae56180beddebc6013e4387e9ed9fe0eR2118, `error` is the expected result when the path leads to a non-scalar.

The xpath field in [xmlfilecontent_object](https://github.com/OVALProject/Language/blob/master/docs/independent-definitions-schema.md#-xmlfilecontent_object-) behaves the same way.

yuumasato: Oh, maybe @matejak is asking about how the test verifies that the result was `error`?
In [verify_results()](https://github.com/OpenSCAP/openscap/blob/maint-1.3/tests/test_common.sh.in#L170) the result of evaluation is compared with the results the comments.

yuumasato:
I was going to add tests for [anchor](https://github.com/OpenSCAP/yaml-filter/wiki/YAML-Path-Definition#anchor), but @evgenyz said it is not implemented yet.

So this should be good to merge, if reviewers are ok.

jan-cerny:
I'm fine.

evgenyz:
> I'm fine.

But it's your request for changes still active :)

evgenyz:
I'm fine with this PR, don't take my whining about excessiveness as a blocker

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.sh
matejak: Thanks, I was not able to understand just by looking at [verify_results documentation](https://github.com/OpenSCAP/openscap/blob/maint-1.3/docs/contribute/testing.adoc#verify_results-function). That doc needs a refresh.

PR #1480, opened 2020-02-14
yamlfilecontent probe

jan-cerny:
This PR will add new yamlfilecontent probe based on proposal in https://github.com/OVAL-Community/OVAL/issues/91.

This PR contains:
* basic skeleton
* finding libyaml.so and libyaml-path.so dependencies
* extraction of input data from oval object
* YAML path query
* produces errors when the yamlpath expression doesn't match scalars or list of scalars but a mapping or something else
* handling lists of scalars
* type detection


evgenyz:
Can you please also add the libyaml dependency?

jan-cerny:
@evgenyz Yes, more patches are coming

jan-cerny:
@evgenyz We have discussed this with @matejak and we think that at this moment we will expect that yaml-path is a separate .so library. For this I propose a change in https://github.com/OpenSCAP/yaml-filter/pull/1

Also, I have created a new CMake find module in 9715d1d which detects the yaml-path library. Basically it depends on https://github.com/OpenSCAP/yaml-filter/pull/1

evgenyz:
@jan-cerny Hmm, we have to have a `.pc` file then. And you really think that it has to be a shared library right now?

jan-cerny:
I'm struggling with the problem described in https://github.com/OpenSCAP/yaml-filter/issues/2

evgenyz:
> I'm struggling with the problem described in [OpenSCAP/yaml-filter#2](https://github.com/OpenSCAP/yaml-filter/issues/2)

I'm on it

jan-cerny:
I have rebased it on the top of maint-13 which now contains XML schemas and tests.

It already works with content from test_probes_yamlfilecontent_key.xml. But when evaluating test_probes_yamlfilecontent_array.xml I get unexpected results. Apart from the correct items in `system_data` it also generates incorrect items, it seems to somehow "merge" `outputs` and `pipelines` keys from openshift-logging.yaml. Hopefully it's related to the issue OpenSCAP/yaml-filter#2

jan-cerny:
@openscap-ci test this please

yuumasato:
Apologies for the bugs in the tests, :grimacing:

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
```diff
@@ -80,7 +80,7 @@
     </ind-def:yamlfilecontent_state>
 
     <ind-def:yamlfilecontent_state version="1" id="oval:0:ste:2">
-      <ind-def:value_of datatype="string" var_ref="oval:0:var:2"/>
+      <ind-def:value_of datatype="string" var_ref="oval:0:var:2" var_check="at least one" />
```
yuumasato: So the returning item is a `yamlfilecontent_state` with two entities `value_of`, each with one value?
I thought the returning item would be a `yamlfilecontent_state`, with one entity `value_of` and two values.

We should check behavior of `xmlfilecontent`.

jan-cerny: Great finding, from the specs the value_of element in xmlsfilecontent_item element is 0 to unbounded, and in the probe there is a for cycle that adds value_of elements, so you seem to be right. I will look into that, it will simplify the code.

jan-cerny: @yuumasato Hopefully fixed in 42337d4b35ceb006421cd44c8564b81ed6c75c99

yuumasato: > Hopefully fixed in [42337d4](https://github.com/OpenSCAP/openscap/commit/42337d4b35ceb006421cd44c8564b81ed6c75c99)

Makes sense to me.

I got a bit confused now, for clarity and record, before commit mentioned above, how was the returned item?
2 `item`s with 1 entity `value_of` and 1 `value`?

jan-cerny: Before  42337d4 it was 2 `yamlfilecontent_item` elements, each with 1 `value_of` child element, but now it's 1 `yamlfilecontent_item` element with 2 `value_of` child elements.

yuumasato: Thanks for the clarification.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_types.xml
```diff
@@ -0,0 +1,421 @@
...
+
+    <ind-def:yamlfilecontent_test version="1" id="oval:0:tst:1" check="all" comment="true">
+      <ind-def:object object_ref="oval:0:obj:1"/>
+    </ind-def:yamlfilecontent_test>
+
+    <ind-def:yamlfilecontent_test version="1" id="oval:0:tst:2" check="all" comment="true">
+      <ind-def:object object_ref="oval:0:obj:2"/>
+    </ind-def:yamlfilecontent_test>
+
+    <ind-def:yamlfilecontent_test version="1" id="oval:0:tst:3" check="all" comment="error">
```
yuumasato: Are the comments from the test accurate?
I think all of them should evaluate to `true`, as the object exists and there are no statuses.

tests/probes/yamlfilecontent/types.yaml
```diff
@@ -0,0 +1,20 @@
...
+float: 7.4
+float_neg: -0.3
+float_exp: +12e03
+float_exp_neg: -43e-4
+float_inf: .inf
+float_nan: .NAN
+bool_true_cast: !!bool "true"
+bool_false_cast: !!bool "false"
+int_cast: !!int "369"
+float_cast: !!float "978.65"
```
yuumasato: Should there be tests for the strings?
`string_true: "true"`
`string_int: "369"`

yuumasato: Should there be tests for the strings?
`string_true: "true"`
`string_int: "369"`

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_types.xml
```diff
@@ -0,0 +1,421 @@
...
+
+    <definition class="compliance" version="1" id="oval:0:def:3">
+      <metadata>
+        <title></title>
+        <description></description>
+      </metadata>
+      <criteria operator="AND">
+        <criterion comment="comment" test_ref="oval:0:tst:3"/>
+      </criteria>
+    </definition>
```
yuumasato: I think you can simplify this and have one `definition` and `AND` all tests that should result in `true`.

yuumasato: And then you can add individual definitions for the `tests` that results in error.

yuumasato: I think you can simplify this and have one `definition` and `AND` all tests that should result in `true`.

yuumasato: For the `test`s that evaluate to `error`, you'll need the individual `definition`s, though.

tests/probes/yamlfilecontent/types.yaml
```diff
@@ -0,0 +1,20 @@
...
+int_10_neg: -17
+int_8: 0o33
+int_16: 0xFF
+float: 7.4
+float_neg: -0.3
+float_exp: +12e03
+float_exp_neg: -43e-4
+float_inf: .inf
+float_nan: .NAN
+bool_true_cast: !!bool "true"
```
yuumasato: How about adding tests to exercise the error path? i.e:
```C
	} else if (!question) {
		return NULL;
	}
```
The content would be something like:
```yaml
bool_error_cast: !!bool  "truu"
int_error_cast: !!int "50%"
```

src/OVAL/probes/independent/yamlfilecontent_probe.c
```diff
@@ -35,11 +37,107 @@
...
+			tag = "!";
+		} else {
+			tag = "?";
+		}
+	}
+
+	/* Nodes with "!" tag can be sequences, maps or strings, but we process
+	 * only scalars in this functions, so they can only be strings. */
+	if (!strcmp(tag, "!")) {
+		tag = OSCAP_YAML_STRING_TAG;
```
yuumasato: Do we expect that the tag can be non-NULL and have tag as `!`?
Otherwise, the assignment `tag = OSCAP_YAML_STRING_TAG;` can be done in line 80.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_types.xml
jan-cerny: They aren't accurate, I'll fix that, they all should evaluate to true.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_types.xml
jan-cerny: Yes, that would be simpler. But if you have many definitions and one of the tests gives unexpected results then it's easier to spot which one.

src/OVAL/probes/independent/yamlfilecontent_probe.c
jan-cerny: Yes, according to https://yaml.org/spec/1.2/ section 6.9.1:

It is possible for the tag property to be explicitly set to the “!” non-specific tag. By convention, this “disables” tag resolution, forcing the node to be interpreted as “tag:yaml.org,2002:seq”, “tag:yaml.org,2002:map”, or “tag:yaml.org,2002:str”, according to its kind. 

There is no way to explicitly specify the “?” non-specific tag. This is intentional.

jan-cerny:
@yuumasato Thank you very much for the tests ideas, it helped to discover some issues.

jan-cerny:
@evgenyz @yuumasato If the YAML document is empty or if a key specified by yamlpath expression isn't present, it collects `<object id="OBJ_ID" version="1" flag="does not exist"/>` and from the flag the test result is computed based on the test attributes as usual.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_types.xml
yuumasato: I see, I'm fine keeping the individual `definition`s then.

src/OVAL/probes/independent/yamlfilecontent_probe.c
yuumasato: Thanks for the clarification.

docs/developer/developer.adoc
```diff
@@ -31,7 +31,7 @@ mkdir -p build
 .. Use fresh sources from git repository.
 +
 ----
-$ git clone https://github.com/OpenSCAP/openscap.git
+$ git clone --recurse-submodules https://github.com/OpenSCAP/openscap.git
```
yuumasato: Is it worth adding instructions for those who already have a clone checked out?
I had to do `git submodule update --init --recursive` for git to actually fetch the submodule.

jan-cerny: I would prefer to simply document the fact that we are using git submodules in our project.

yuumasato: Makes sense.

jan-cerny:
@yuumasato  SSG will probably have to start generation (pesudo) OVAL 5.11.3 content.

yuumasato:
> @yuumasato SSG will probably have to start generation (pesudo) OVAL 5.11.3 content.

Well noted. As the `yamlfilecontent_test` specification is not final, and OVAL 5.11.3 version is not out yet, I'm thinking about adding a variable to turn on/off building content with future standards / features.
I'll create an issue in CaC.

PR #1481, opened 2020-02-14
Fixed oscap-ssh failing to retrieve the result files when executing with --sudo

rmetrich:
Depending on the umask configuration of the target system, "sudo oscap"
may create the result files in temporary directory with 600 permissions,
which makes retrieving the log (as the regular user that ssh'ed to the
system) impossible:

~~~
$ oscap-ssh --sudo user@system 22 xccdf eval ...
[...]
oscap exit code: 0
Copying back requested files...
scp: /tmp/tmp.0kfbPWEy6u/report.html: Permission denied
Failed to copy the HTML report back to local machine!
~~~

Scenario to reproduce the failure: set a default umask in /etc/sudoers:

~~~
Defaults	umask = 0077
~~~

The fix consists in changing the result files' ownership from "root" to
user's back, all while in the single sudo (using two sudo commands
wouldn't be nice since the user may get the password prompt twice,
depending on the sudo's configuration).

PR #1485, opened 2020-02-19
Fixed oscap-ssh failing to retrieve the result files when executing with --sudo

rmetrich:
Depending on the umask configuration of the target system, "sudo oscap"
may create the result files in temporary directory with 600 permissions,
which makes retrieving the log (as the regular user that ssh'ed to the
system) impossible:

~~~
$ oscap-ssh --sudo user@system 22 xccdf eval ...
[...]
oscap exit code: 0
Copying back requested files...
scp: /tmp/tmp.0kfbPWEy6u/report.html: Permission denied
Failed to copy the HTML report back to local machine!
~~~

Scenario to reproduce the failure: set a default umask in /etc/sudoers:

~~~
Defaults	umask = 0077
~~~

The fix consists in changing the result files' ownership from "root" to
user's back, all while in the single sudo (using two sudo commands
wouldn't be nice since the user may get the password prompt twice,
depending on the sudo's configuration).

See also #1481

### DIFF
--- a/utils/oscap-ssh
+++ b/utils/oscap-ssh
@@ -280,7 +280,12 @@ echo "Starting the evaluation..."
 # changing directory because of --oval-results support. oval results files are
 # dumped into PWD, and we can't be sure by the file names - we need controlled
 # environment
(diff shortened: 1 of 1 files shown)


rmetrich:
@yuumasato here you go for maint-1.2

yuumasato:
@rmetrich this is against `maint-1.3`. I will try changing base branch of this PR.

yuumasato:
This looks good, merging.
Thanks for the patch.

PR #1487, opened 2020-02-26
Remove useless warnings

jan-cerny:
These warnings give zero value to the user, because they don't specify
what went wrong. Moreover, the severity of the message depends on the
parameters of the macro. For example, in environemntvariable58 probe the
macro is used in a way that "Entity has no value" is a valid and
expected situation, because if the pid element is set to xsi:nil the PID
should be set to the PID of the calling process.

Resolves: RHBZ#1764139


matejak:
The removed messges had no value. As the code in question is a macro, making those messages useful is relatively difficult, it makes sense to first remove them, and go for a proper refactor in a later PR if such need arises.

PR #1489, opened 2020-02-27
Add Fedora 33 CPEs

jan-cerny:
Fedora 32 was branched from Rawhide on 2020-02-12 and Rawhide
became future Fedora 33.


evgenyz:
LGTM, thanks!

PR #1490, opened 2020-03-02
Generate HTML guide using tailoring file

jan-cerny:
If we use a tailoring file we can't apply XSLT directly to source
datastream file or to source XCCDF file, because we need to
resolve tailoring and add the custom profile.

The way that I propose is to load the XCCDF session first, similar
way as we do in `xccdf generate fix module`. That will resolve
resolve the tailoring automatically. Then export XCCDF benchmark
into DOM and apply XSLT on the generated tree in-memory.

The problem is that the current design doesn't add the tailoring
profiles to xccdf_benchmark profiles, but there is a separate hash
table. It would require more refactoring.

I have discovered that --hide-profile-info option doesn't do anything
because the functionality was removed 6 years ago, so I changed
documentation for it.

Resolves: RHBZ#1743835

### DIFF
--- a/src/XCCDF/benchmark.c
+++ b/src/XCCDF/benchmark.c
@@ -866,6 +866,19 @@ xccdf_benchmark_find_target_htable(const struct xccdf_benchmark *benchmark, xccd
 	return XITEM(benchmark)->sub.benchmark.items_dict;
 }
 
+int xccdf_benchmark_include_tailored_profiles(struct xccdf_benchmark *benchmark)
+{
+	struct oscap_list *profiles = XITEM(benchmark)->sub.benchmark.profiles;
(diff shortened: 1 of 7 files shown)


matejak:
I have tested this PR, and it works perfectly.

It also improves the UX, as now it accepts the profile suffix as the `--profile` argument.

matejak:
I recommend to flip the PR to ready to review as soon as possible - I think that it can be merged in the current state.

jan-cerny:
@matejak Why do you think so? Have you also checked the disgusting parts?

matejak:
Could you maybe rebase to get a feedback from our services s.a. LGTM.

jan-cerny:
@matejak rebased

jan-cerny:
@matejak and now?

matejak:
Merging, there is no obvious error in the code, well done!

PR #1493, opened 2020-03-03
Prevent crashes when complicated regexes are executed

jan-cerny:
When a complex unanchored regular expression is provided in OVAL textfilecontent54_object, the pcre_exec might recurse internally many times which can cause a segmenatation fault. A reproducer of this situation is provided as a test. We will prevent the segmentation fault by setting a maximal recursion limit.

Since a similar code existed in 2 probes I decided to unify the code first, and then apply the limit only at 1 place.

Resolves: RHBZ#1686467

### DIFF
--- a/docs/developer/developer.adoc
+++ b/docs/developer/developer.adoc
@@ -317,6 +317,8 @@ behaviour.
 
 * *OSCAP_FULL_VALIDATION=1* - validate all exported documents (slower)
 * *SEXP_VALIDATE_DISABLE=1* - do not validate SEXP expressions (faster)
+* *OSCAP_PCRE_EXEC_RECURSION_LIMIT* - override default recursion limit
+  for match in pcre_exec call in textfilecontent(54) probes.
 
 
 
--- a/src/common/util.c
+++ b/src/common/util.c
@@ -47,6 +47,7 @@
 #endif
 
 #define PATH_SEPARATOR '/'
+#define OSCAP_PCRE_EXEC_RECURSION_LIMIT 1000
 
 int oscap_string_to_enum(const struct oscap_string_map *map, const char *str)
 {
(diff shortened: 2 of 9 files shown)


src/common/util.c
```diff
@@ -45,6 +47,7 @@
 #endif
 
 #define PATH_SEPARATOR '/'
+#define OSCAP_PCRE_EXEC_RECURSION_LIMIT 1000
```
jan-cerny: The question is: How big should the limit be?
On my machine, the segfault happened at ~ 17600.

evgenyz: I wonder if this number is a derivative of the the string or the pattern.

evgenyz: In order to determine where we should draw the line I would start with answering the question of how many recursions is takes to match the content which actually matches (from the set of strings/patterns we have to deal with).

jan-cerny: It probably is, but I don't know how to compute it.

There can be arbitrary string and arbitrary valid regular expression.

evgenyz: Default is: 10 000 000. ModSecurity, for example, has it at 1500. I guess it'd be hard to properly estimate the value.

Maybe we could make it an option for the scanner instead? Then users could adjust it in case it is too small for their needs, without rebuilding the scanner. Just a wild idea, it might be not easy to implement.

jan-cerny: Great idea, we can do that, or we can add possibility to modify this by environment variable. 

However, we need to set a sensible default, because I assume most of the users will use the default.

evgenyz: Oh, I think that it would be your favourite :) Number guessing :)). I'd vote for 5000. Rationale:  ~⅓ of your machine's limit on a set of content we have.

jan-cerny: I'm fine with 5000

docs/developer/developer.adoc
```diff
@@ -317,6 +317,8 @@ behaviour.
 
 * *OSCAP_FULL_VALIDATION=1* - validate all exported documents (slower)
 * *SEXP_VALIDATE_DISABLE=1* - do not validate SEXP expressions (faster)
+* *OSCAP_PCRE_EXEC_RECURSION_LIMIT* - override default recursion limit
+  for match in pcre_exec call in textfilecontent(54) probes.
```
evgenyz: I wonder if other probes could be subjects for this problem. But we probably don't have to address them in this PR.

evgenyz:
LGTM. We will deal with other `pcre_exec` calls in an evolutionary manner.

docs/developer/developer.adoc
jan-cerny: Yes, they are, although I think that the textfilecontent is the most dangerous, because there the user provides the regex and the text file can be arbitrary size.

jan-cerny: Actually, we have many pcre_exec calls across the whole code base, I think it would be better to replace them by some wrapper,  which will always set the limit, and also it can implement printing nice error messages.

PR #1495, opened 2020-03-06
Use pthread required package when building on Windows.

### LINKED ISSUE #1321: Cmake cannot find vcpkg PTHREADS_WIN32 Lib
#### Steps to Reproduce:

1. C:\devel\openscap\build>cmake -D ENABLE_PYTHON3=FALSE -D CMAKE_TOOLCHAIN_FILE=c:
/devel/vcpkg/scripts/buildsystems/vcpkg.cmake ..
2. 
3. 

#### Actual Results: CMake Error at C:/Program Files/CMake/share/cmake-3.14/Modules/FindPackageHandle
StandardArgs.cmake:137 (message):
  Could NOT find Threads (missing: THREADS_PTHREADS_WIN32_LIBRARY)
  C:/Program Files/CMake/share/cmake-3.14/Modules/FindPackageHandleStandardArgs.
cmake:378 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindThreads.cmake:232 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  C:/devel/vcpkg/scripts/buildsystems/vcpkg.cmake:234 (_find_package)
  CMakeLists.txt:145 (find_package)

ggbecker:
When using a fresh instance of Windows on appveyor to build OpenSCAP, it fails to finish with the following:

```
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find Threads (missing: THREADS_PTHREADS_WIN32_LIBRARY)
  C:/Program Files (x86)/CMake/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:393 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindThreads.cmake:232 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  C:/Tools/vcpkg/scripts/buildsystems/vcpkg.cmake:263 (_find_package)
  CMakeLists.txt:149 (find_package)
```

Since the official instance of Windows used by the OpenSCAP project uses a cached version of installed packages since from the beginning (to optimize build time), older versions of dependencies are being used and for some reason they work with the following:

`find_package(Threads REQUIRED)`

This PR possibly fixes #1321 and is related to #1073 

The fix was inspired by: https://github.com/microsoft/vcpkg/pull/8654
Which is a pull request that fixes the Windows build of OpenSCAP in vcpkg repository.

Also ZLIB is required to build on Windows per:
```
  Generating Code...
  xccdf_object.vcxproj -> c:\projects\openscap\build\src\XCCDF\xccdf_object.dir\Release\xccdf_object.lib
  Building Custom Rule C:/projects/openscap/src/CMakeLists.txt
LINK : fatal error LNK1104: cannot open file 'ZLIB::ZLIB.lib' [c:\projects\openscap\build\src\openscap.vcxproj]
```


ggbecker:
To make the build work on AppVeyor, we would have to refresh the cache of installed dependencies because they are too old.

Here is a full log of a working build on a fresh instance of AppVeyor: https://ci.appveyor.com/project/GabrielBecker/openscap/builds/31291312

jan-cerny:
I have realized this: We have a custom CMake module to find threads in cmake/FindThreads.cmake. We have created this module because the stock module shipped in CMake wasn't able to find pthreads on Windows. The changes you made mean that this module won't be used on Windows. That means that we probably don't need the custom module, and we can remove it, and we can use the stock module on Linux. Could you check if cmake/FindThreads.cmake file can be removed?

jan-cerny:
> To make the build work on AppVeyor, we would have to refresh the cache of installed dependencies because they are too old.

Yes, it would be great to refresh them.

I think that you need to trigger a build without using the cache and then re-add cache again. You can do that by commenting `cache` in appveyor.yml See: https://www.appveyor.com/docs/build-cache/#cleaning-up-cache. Then, I guess you will have to add the cache to appveyor.yml again, because without caching it builds all the dependent libraries form sources which takes long time.

ggbecker:
Ok, so I have deleted the `FindThreads.cmake` module and triggered the cache refresh in appveyor. Let's see what the build looks like.

ggbecker:
So far, I am not able to successfully finish the build with cached installed packages. I'm not sure if there is a problem with the cache not being refreshed or if somehow when using the cache `pthread` cannot be found.

I've submitted a ticket in appveyor to check the cache thing, if they can check whether it's updated or not.

Also, I think that we need to add build dependency on:
```
Build-Depends: getopt-win32 (windows)
```
per discussion on: https://github.com/microsoft/vcpkg/pull/8654

ggbecker:
So, I have discovered that refreshing the cache in appveyor through pull requests is disable by default.

From: https://www.appveyor.com/docs/build-cache/
```
Note: By default, saving cache is disabled in Pull Request builds. Use Save build cache in Pull Requests checkbox on the General tab of project settings if you need to save cache during PR builds. This setting is not exposed in YAML to prevent unauthorized cache modifications.
```

So we need to either enable temporarily for this pull request or to handle it after we merge it. I prefer going with the first option.

jan-cerny:
@openscap-ci test this please

jan-cerny:
@ggbecker I don't understand what we should do now. Should I merge it now?

ggbecker:
For posterity:

The cache was updated with the following in the `appveyor.yml`:

```
install:
- cmd: >-  
    vcpkg install curl libxml2 libxslt bzip2 pcre pthreads zlib

    vcpkg upgrade --no-dry-run
```

And to enable saving a new cache of installed packages through pull requests (disable by default) we enabled the option under appveyor OpenSCAP project settings called: `Save build cache in Pull Requests` (This option is only available through the website project settings).

PR #1498, opened 2020-03-18
Use different subject in tests

jan-cerny:
Resolves: rhbz#1814726


matusmarhefka:
@jan-cerny LGTM, thanks for fixing this :)

PR #1482, opened 2020-02-17
Add development versions of OVAL schemas + yamlfilecontent schema

jan-cerny:
This PR adds the OVAL schemas for future version of OVAL language - OVAL 5.11.3.

It will allow us to develop features and test them before the OVAL 5.11.3 is actually released.

The drawback of this solution is that we will have to synchronize or backport changes from the OVAL language as the OVAL community develops the language.

The schemas are taken from `develop` branch from https://github.com/OVAL-Community/OVAL
branch as of 2020-02-17 HEAD: 0fa9c2544433bf7d7840ddf2f58cd0729b067c62.

The schematrons were generated locally using `/schemas/build_schematron.sh`
There are a few minor changes in comparison to the offical schemas. We do these changes each time we import new OVAL version.

On top of that, this PR contains patch that adds **yamlfilecontent_test** schema which was proposed in OVAL-Community/OVAL#90 

For more details please read the commit messages of individual commits.

PR #1500, opened 2020-03-23
Use the latest 5.11.3 OVAL when validating the 1.3 datastreams

matejak:
This PR changes how oscap validates 1.3 datastreams - now it will use the `5.11.3` development OVAL schemas further expanded by us to include the `yamlfilecontent` family.

Related to: https://github.com/OpenSCAP/openscap/pull/1482

### DIFF
--- a/schemas/sds/1.3/scap-source-data-stream_1.3.xsd
+++ b/schemas/sds/1.3/scap-source-data-stream_1.3.xsd
@@ -16,7 +16,7 @@
   <xs:import namespace="http://checklists.nist.gov/xccdf/1.2"
     schemaLocation="../../xccdf/1.2/xccdf_1.2.xsd"/>
   <xs:import namespace="http://oval.mitre.org/XMLSchema/oval-definitions-5"
-    schemaLocation="../../oval/5.11.2/oval-definitions-schema.xsd"/>
+    schemaLocation="../../oval/5.11.3/oval-definitions-schema.xsd"/>
   <xs:import namespace="http://cpe.mitre.org/dictionary/2.0"
     schemaLocation="../../cpe/2.3/cpe-dictionary_2.3.xsd"/>
   <xs:import namespace="http://scap.nist.gov/schema/ocil/2.0"


jan-cerny:
I don't know other ways how to achieve this easily. We did similar things in past, for example in https://github.com/OpenSCAP/openscap/pull/58.

Ideally, the validation engine should recognize versions of datastream components instead of assuming any version. But, it would require some effort.

yuumasato:
Should we be concerned that this may give impression that content with `yamlfilecontent_test` is SCAP-1.3 valid?

jan-cerny:
@yuumasato That's a big problem. It's similar to situation when we say that content is SCAP 1.2 valid even if it contains OVAL systemd tests which are OVAL 5.11 therefore not SCAP 1.2.

Any suggestions?

Obviously, we can add --skip-valid everywhere, but that isn't convenient and doesn't catch errors.

matejak:
The problem is that as there is this support for yaml test in the scanner, you can't use it unless you skip the whole validation. And as validation can reveal issues with the input data, so disabling it just to get the test running is bad UX. I see as important point that other scanners won't have problems with the datastream, unless they try validate it, or to evaluate those rules.
If you check out [OVAL release streams documentation](https://oval-community-guidelines.readthedocs.io/en/latest/proposal-process/release-process.html#how-to-release-streams),  the YAML test officially qualifies as "unofficial extension" :-)

yuumasato:
> It's similar to situation when we say that content is SCAP 1.2 valid even if it contains OVAL systemd tests which are OVAL 5.11 therefore not SCAP 1.2.

That's a great comparison, and this behavior has led to confusion because content will fail SCAPVAL.

> If you check out [OVAL release streams documentation](https://oval-community-guidelines.readthedocs.io/en/latest/proposal-process/release-process.html#how-to-release-streams), the YAML test officially qualifies as "unofficial extension" :-)

It says there that the extension should be expressed as a valid XSD schema, but it doesn't clearly say how they should be used.
I speculate this schema would be loaded in a separate namespace, it wouldn't conflict with the officially released schema and would allow validation to happen.

> Related to: #1482

In the related PR, a future version of the schema was added, the yaml tests won't pass OVAL 5.11.2 validation.

> @yuumasato That's a big problem.

I understand this is a big problem, and that there may not be easier way than changing official schemas and incorporating `yamlfilecontent_test` as valid definition.

Just wanted to check-in that we are OK changing what the scanner considers valid SCAP 1.3.

jan-cerny:
@matejak It seems that nobody is strongly opposed to this patch.

Issue #1394, opened 2019-08-12
"Obtrusive data from probe!" error on every xml with regular expressions in Windows registry


DenisNovac:
#### Description of Problem:
I had package of xml-files written for OVALdi version 5.10. I want to switch to OpenSCAP since it still in development and supports OVAL Language updates. But most of my files does not work. I get the "Obtrusive data from probe" error:

> PS C:\Users\dyablochkin\Documents\dev\oval\wincc> oscap oval eval --report report.html .\wincc_cve_2019_6568.xml
> W: C:\Program Files (x86)\OpenSCAP 1.3.0\oscap.exe:     Obtrusive data from probe!

It is just a warning, but it freezes the process. Seems like it happens when i use regular expressions in registry_object for Windows.

Works great:
`<key>SOFTWARE\WOW6432Node\SIMENS\Automation\_InstalledSW\HMIRTM_V11\EditionMain</key>`

Obtrusive data:
`<key operation="pattern match">^SOFTWARE\\WOW6432Node\\SIMENS\\Automation\\_InstalledSW\\HMIRTM_V\d*\\EditionMain</key>`

From a verbose-log-file i can tell that oscap doing .* check, not my mask. There is every registry branch in output from HKLM hive:

> [C:\Program Files (x86)\OpenSCAP 1.3.0\oscap.exe(6704):unknown(2ba0):c:\projects\openscap\src\oval\probes\windows\registry_probe.c:446:registry_recurse_down]
> D: C:\Program Files (x86)\OpenSCAP 1.3.0\oscap.exe:     Can't open registry key 'DRIVERS\DriverDatabase\DriverPackages\wnetvsc.inf_amd64_86f2f8310179e55e\Configurations\netvsc_Device\Properties': RegOpenKeyExW error: Отказано в доступе.
> 
>  [C:\Program Files (x86)\OpenSCAP 1.3.0\oscap.exe(6704):unknown(2ba0):c:\projects\openscap\src\oval\probes\windows\registry_probe.c:446:registry_recurse_down]
> D: C:\Program Files (x86)\OpenSCAP 1.3.0\oscap.exe:     Can't open registry key 'SAM\SAM': RegOpenKeyExW error: Отказано в доступе.
> 
>  [C:\Program Files (x86)\OpenSCAP 1.3.0\oscap.exe(6704):unknown(2ba0):c:\projects\openscap\src\oval\probes\windows\registry_probe.c:446:registry_recurse_down]
[...]

PR #1502, opened 2020-03-31
Remove a warning

jan-cerny:
This warning has low informational value.

Fixes: rhbz 1722003

### DIFF
--- a/src/OVAL/oval_probe_ext.c
+++ b/src/OVAL/oval_probe_ext.c
@@ -998,7 +998,6 @@ int oval_probe_ext_eval(SEAP_CTX_t *ctx, oval_pd_t *pd, oval_pext_t *pext, struc
                          * The no-reply flag is set and oval_probe_comm returned some
                          * data. This is considered a non-fatal error.
                          */
-                        dW("Obtrusive data from probe!");
                         SEXP_free(s_sys);
 		}
 		return (0);


yuumasato:
Were these warnings always being printed? Or did they become more frequent in lately?
Is it worth investigating why?

evgenyz:
There an issue that mention the warning, #1394. Was this warning related to the root problem of the casse? 

Are we not interested in catching this state at all?

jan-cerny:
This is quite old code, and I recall them seeing them from my beginning. I remember we investigated it couple of times and we never found the purpose of this message.

My personal opinion is that this state can be valid. We don't see any unexpected evaluation results or unexpected items in the OVAL results document.

I think that this warning causes only problems.

I agree that it can be helpful to figure out the root cause, but it's a time investment.

yuumasato:
> There an issue that mention the warning, #1394. Was this warning related to the root problem of the casse?

The BZ and the issue are related, both use `<set>`.

Our content doesn't use much sets, but I think it will become much more common, as it is a nice way to resolve issues with empty variables.

> I think that this warning causes only problems.

How about bringing it to Devel level?

jan-cerny:
We can bring it to devel level. But still, the message says only "obtrusive data from probe". It doesn't specify which data, from which probe, why are they obtrusive. Even a developer needs this information.

yuumasato:
> We can bring it to devel level. But still, the message says only "obtrusive data from probe". It doesn't specify which data, from which probe, why are they obtrusive. Even a developer needs this information.

I'm not familiar with SEXP. But is it the case that we can't convert `s_sys` to anything, or introspect it to check what data the probe sent?

evgenyz:
So, the code checks for presence of `s_sys` when `OVAL_PDFLAG_NOREPLY` flag is set. If `s_sys` exists code frees it (forcefuly cleans probe's response) and then warns us, presumably about a probe, which violates OVAL_PDFLAG_NOREPLY flag protocol.

The question: is it important for the probe to obey OVAL_PDFLAG_NOREPLY or we can consider response clean-up as a satisfying approach.

If it is important, can we add some data that will possibly help to identify rogue probe?

yuumasato:
> The question: is it important for the probe to obey `OVAL_PDFLAG_NOREPLY` or we can consider response clean-up as a satisfying approach.

From my perspective this could be hiding a bug, of which we don't know the impacts.
It could be just this simple message, or there could more issues...

jan-cerny:
The flag `OVAL_PDFLAG_NOREPLY` is defined in oval_probe.h. There is a comment `don't send probe result to library - just an ack`. I'm not sure what this comment exactly means.

The flag `OVAL_PDFLAG_NOREPLY` is masked at multiple places. The first is located in `oval_probe_comm` where when this flag is set it causes setting `no-reply` attribute of a SEAP message using `SEAP_msgattr_set`. The second one is this one in `oval_probe_ext_eval` that prints the warning about obtrusive data. The third occurence in `oval_probe_query_object` in `oval_probe.c` on line 129 is more interesting, because there are conditions that involve variables, OVAL system characteristics and the system characteristics flags. If this flag is set and other conditions are met the existing system characteristics are reused instead of new object evaluation. This looks like an optimalization. The fourth occurence is in the same function (`oval_probe_query_object`) at the end. If the flag isn't present all references for variables are collected.

The only place where the flag `OVAL_PDFLAG_NOREPLY` is set is in `oval_probe_cmd_obj_eval` in oval_probe_ext.c. The flag is passed as an argument  to `oval_probe_query_object`. The function `oval_probe_cmd_obj_eval` is a callback registered in `oval_probe_cmd_init` with command code `PROBECMD_OBJ_EVAL`. We can search that this command is used only in `probe_obj_eval`. This static function has a long comment in `worker.c` starting from line 398. The comment states that the result should not be send to the library, but fetched from the cache and returned from the caller. There might be a problem because I have no knowledge about the various caches that the common probe code uses. The `probe_obj_eval` in called only in `probe_set_eval` which servers to evaluate OVAL `set` elements. Set can be child element of any `_object` element, which allows us to combine multiple different objects to get 1 object. The interesting thing is that `probe_set_eval` is called from `probe_worker` function, which is the same function that calls the main functions of the probes. The difference is that `probe_set_eval` is called if there is a set within the evaluated object whereas the probes main functions are called if set is not present. Therefore, spawning a probe worker thread can lead to spawning 1 or more other probe worker threads if sets are used.

We will continue with investigation.

PR #1506, opened 2020-04-02
Fix filepath pattern matching in offline mode

yuumasato:
Extend offline coverage for `textfilecontent54_test`
- Improve test by adding similar files outside of chroot environment
- Add test to check that a file present only outside chroot environment is not collected by mistake
- Add test for objects defined using pattern matching in filepath and filename
- Updated environment preparation of offline mode
- Fix `filepath` pattern matching in offline mode


tests/offline_mode/textfilecontent54.oval.xml
```diff
@@ -30,29 +30,131 @@
...
+    <ind:textfilecontent54_object id="oval:x:obj:4" version="1" comment="Object representing file">
+      <ind:filepath operation="pattern match">^/tmp/.*\.filepath$</ind:filepath>
+      <ind:pattern operation="pattern match">^.*$</ind:pattern>
+      <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
+    </ind:textfilecontent54_object>
+    <ind:textfilecontent54_object id="oval:x:obj:5" version="1" comment="Object representing file">
+      <ind:filepath operation="pattern match">^/tmp/zzz/.*\.filepath$</ind:filepath>
+      <ind:pattern operation="pattern match">^.*$</ind:pattern>
+      <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
+    </ind:textfilecontent54_object>
```
yuumasato, line 145: For some reason `filepath` with `pattern match` in offline mode doesn't work.

tests/offline_mode/test_offline_mode_textfilecontent54.sh
```diff
@@ -15,38 +15,78 @@ set -e -o pipefail
...
+
+    assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:3" and @result="true"]'
+
+    assert_exists 1 $tfc_item'/ind-sys:filepath[text()="/tmp/bar.filepath"]'
+    assert_exists 1 $tfc_item'/ind-sys:filename[text()="bar.filepath"]'
+    assert_exists 1 $tfc_item'/ind-sys:text[text()="bar filepath pattern"]'
+
+    assert_exists 1 $tfc_item'/ind-sys:filepath[text()="/tmp/zzz/foo.filepath"]'
+    assert_exists 1 $tfc_item'/ind-sys:filename[text()="foo.filepath"]'
+    assert_exists 1 $tfc_item'/ind-sys:text[text()="foo filepath pattern"]'
```
yuumasato, line 70: The objects seem to be collected, though.

yuumasato:
I think we have a bug, `def:5` should be true.
```
128/183 Test #128: offline_mode/test_offline_mode_textfilecontent54.sh ..........................***Failed    1.65 sec
+ test_offline_mode_textfilecontent54                         
TEST: test_offline_mode_textfilecontent54
Definition oval:x:def:7: true
Definition oval:x:def:6: true
Definition oval:x:def:5: false
Definition oval:x:def:4: true
Definition oval:x:def:3: true
Definition oval:x:def:2: true
Definition oval:x:def:1: true
Evaluation done.
Failed: expected count: 1, real count: 0, xpath: '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:5" and @result="true"]'
```

evgenyz:
@yuumasato How about moving `offline_mode/test_offline_mode_textfilecontent54.sh` and stuff to the more appropriate `probes/textfilecontent54`? My PR #1505 will move `system_info` there, so there only this test will remain in `tests/offline_mode`.

jan-cerny:
Nice discovery!

yuumasato:
Tests were moved.

evgenyz:
Great! LGTM! Thanks!

PR #1508, opened 2020-04-03
Added a Python script that can be used for CLI tailoring

matejak:
The script can accept (at the same time) rule IDs to select/deselect, and XCCDF values to set. The result can be saved to a tailoring file that the scanner can use to scann, or to generate a HTML guide (only the git version ATM).

The script doesn't touch the source datastream, so it doesn't check whether profile or rule IDs actually exist.

Example usage:

```
python3 autotailor.py -v var_accounts_maximum_age_login_defs=11 -u display_login_attempts -o tailoring.xml ssg-fedora-ds.xml standard
oscap xccdf generate guide --profile customized --tailoring-file tailoring.xml ssg-fedora-ds.xml  > guide.html
```


evgenyz:
Can we adjust this thing to accept lines up to 120 chars long? I think 100 is a bit too small for contemporary displays... just my IMHO.

utils/autotailor.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def id_is_full(string):
+    return re.match(r"^xccdf_org\.", string) is not None
+
+
+class Tailoring:
+    PROFILES_PREFIX = "xccdf_org.ssgproject.content_profile_"
+    RULES_PREFIX = "xccdf_org.ssgproject.content_rule_"
+    VARS_PREFIX = "xccdf_org.ssgproject.content_value_"
```
jan-cerny, line 40: These prefixes are specific for scap-security-guide. OpenSCAP aims to be a generic SCAP processing tool. I think it would be more beneficial if this is customizable. Then people could use it on any SCAP content.

utils/autotailor.py
```diff
@@ -0,0 +1,193 @@
...
+
+ALL_NS = {
+    "xccdf-1.1": "http://checklists.nist.gov/xccdf/1.1",
+    "xccdf-1.2": "http://checklists.nist.gov/xccdf/1.2",
+}
+DEFAULT_PROFILE_SUFFIX = "_customized"
+
+
+def id_is_full(string):
+    return re.match(r"^xccdf_org\.", string) is not None
```
jan-cerny, line 34: What does this do? I guess it should detect that string is a valid XCCDF 1.2 ID.  For example, the rule ID format is `xccdf_<namespace>_rule_<name>`, where namespace is a valid reversed DNS. IDs of profiles, variables, etc. use a similar layout.

utils/autotailor.py
```diff
@@ -0,0 +1,193 @@
...
+                ET.dump(root)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="This script produces XCCDF 1.2 tailoring files to be used by SCAP scanners and SCAP datastreams.")
+    parser.add_argument(
+        "datastream", metavar="DS-FILENAME",
+        help="The datastream filename is just informational, "
+        "it doesn't have any key role in the tailoring composition.")
```
jan-cerny, line 126: If it's just informational and not key role, then why is it a required argument?

utils/autotailor.py
```diff
@@ -0,0 +1,193 @@
...
+    parser.add_argument(
+        "profile", metavar="BASE_PROFILE_ID",
```
jan-cerny, line 128:
```suggestion
        "--profile", metavar="BASE_PROFILE_ID",
```

utils/autotailor.py
```diff
@@ -0,0 +1,193 @@
...
+        "The argument works the same way as the --select argument.")
+    parser.add_argument(
+        "-p", "--new-profile-id",
+        help="Specify the ID of the tailored profile. "
+        "The ID of the new profile can be either its full ID, or the suffix, in which case "
+        "the '{prefix}' prefix will be prepended internally. "
+        "If left out, the new ID will be obtained "
+        "by appending '{suffix}' to the tailored profile ID."
+        .format(prefix=Tailoring.PROFILES_PREFIX, suffix=DEFAULT_PROFILE_SUFFIX))
+    parser.add_argument("-o", "--output")
```
jan-cerny, line 162: help text missing

utils/autotailor.py
```diff
@@ -0,0 +1,193 @@
...
+            change.text = str(value)
+
+        if location:
+            ET.ElementTree(root).write(location)
+        else:
+            try:
+                # Python >= 3.9
+                ET.indent(root)
+            except AttributeError:
+                ET.dump(root)
```
jan-cerny, line 117: This function should be used for debugging only. I think `write` would work.

redhatrises:
I like this idea! A couple of questions:

- Why isn't this integrated into the existing code? I love python but this means it is now a dependency where it hasn't been before. This affects downstream, container, and cloud usage of OpenSCAP. It would be nice if this was integrated and not having a python dependency and simplified into a single command if possible.
- Namespaces are hardcoded which means it has a very narrow focus towards our projects, and there are so many different namespaces this needs to support.

matejak:
> Can we adjust this thing to accept lines up to 120 chars long? I think 100 is a bit too small for contemporary displays... just my IMHO.

It was quite a fight to get if from 80 to 100, I am still exhausted from it :-) I find 100 bearable, there are exceptions (not necessarily now) when lines have to be longer, and I believe that rules don't have to be exception-proof.
BTW any idea why the LGTM checks are not there?

jan-cerny:
PEP8 says maximum 79 characters, and if the code is maintained exclusively or primarily by the team it allows to extend it to 99. https://www.python.org/dev/peps/pep-0008/#maximum-line-length

evgenyz:
> BTW any idea why the LGTM checks are not there?

No idea :(

matejak:
> I like this idea! A couple of questions:
> 
>     * Why isn't this integrated into the existing code? I love python but this means it is now a dependency where it hasn't been before. This affects downstream, container, and cloud usage of OpenSCAP. It would be nice if this was integrated and not having a python dependency and simplified into a single command if possible.

Having a Python script is a question of less than two hundreds of lines of code including everything and hours of coding max. Doing it in C and integrating into openscap would take ten times the effort (when debugging is considered), which becomes quite sizeable.
Python is not a brand new dependency - `oscap-docker` is also a Python script.

> 
>     * Namespaces are hardcoded which means it has a very narrow focus towards our projects, and there are so many different namespaces this needs to support.

I am not so sure what you mean by namespaces - the script asssumes XCCDF 1.2 XML namespace. It is true that it provides shortcuts for the CAC/SSG way of naming value/profile/rule IDs. However, if anybody uses another ID naming convention, they can use it. The only assumption is that those full IDs start with `xccdf_org` and @jan-cerny related to that as well. Could it be that it is this not generic enough? What would be the way of distinguishing a full ID from our "stripped" IDs?

utils/autotailor.py
matejak: I woudn't do this, as the source of the tailoring is a required input. Therefore, it works well as a positional argument.

utils/autotailor.py
matejak: I have mentioned this in a reply to @redhatrises concern. Your guess is right, and I see that the regex should be changed to match just for `^xccdf_`.

utils/autotailor.py
matejak: People can use it with any SCAP content if they provide valid XCCDF IDs as inputs. If invalid IDs are supplied, then this gets prepended.

utils/autotailor.py
matejak: It is required, because you need to put some kind of datastream href into the tailoring. I know that it is silly, but the workbench does it the same way.

> ... The 1<xccdf:benchmark>1 element holds the associated Benchmark's **location URI**, version string, and id value. ...

utils/autotailor.py
jan-cerny: I have an idea. I think you can create an option `--reverse-dns` that would allow users to provide custom reverse DNSs.

utils/autotailor.py
jan-cerny: Perhaps change it to check for the aforementioned format?

utils/autotailor.py
jan-cerny: That was actually my guess. Then, please change the help text.

utils/autotailor.py
jan-cerny: No, at this moment you have 2 positional arguments.

jan-cerny:
> I am not so sure what you mean by namespaces - the script asssumes XCCDF 1.2 XML namespace. It is true that it provides shortcuts for the CAC/SSG way of naming value/profile/rule IDs. However, if anybody uses another ID naming convention, they can use it. The only assumption is that those full IDs start with xccdf_org and @jan-cerny related to that as well. Could it be that it is this not generic enough? What would be the way of distinguishing a full ID from our "stripped" IDs?

In my opinion, the automated pre-pending of prefixes should be completely optional. The script should work with any valid XCCDF. In XCCDF 1.2 it's required to keep the format that I outlined earlier (we call this "full IDs"). You can find more details in NISTIR 7275 rev. 4, Section 6.2.3, page 10. You can base the distinguishing on that. However, it should always work with the full IDs.

If you want to make it cool and enable the "magic" feature for other people, you can have an command line option for providing the reversed DNS.

utils/autotailor.py
matejak: Yes, the profile and the datastream. There is nothing wrong with having two positional arguments.

matejak:
> In my opinion, the automated pre-pending of prefixes should be completely optional.

Your wish has been granted - check the `_full_id` method that either lets a full ID through without modifications, or promotes a shortened ID to a full valid ID by means of prepending the appropriate prefix.

The reverse DNS trick is nice though. I see that in the future, the utility could support more than one set of default prefixes based on demand, but for now, it's OK as it is.

utils/autotailor
```diff
@@ -0,0 +1,195 @@
...
+        "profile", metavar="BASE_PROFILE_ID",
+        help="Specify ID of the base profile. "
+        "ID of the profile can be either its full ID, or the suffix, in which case "
+        "the 'xccdf_<id-namespace>_profile' prefix will be prepended internally.")
+    parser.add_argument(
+        "--title", default="",
+        help="Title of the new profile.")
+    parser.add_argument(
+        "--id-namespace", default="org.ssgproject.content",
+        help="The reverse-DNS style string that is part of entities IDs in the corresponding datastream.")
```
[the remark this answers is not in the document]
matejak: I have switched on display of defaults in argparse - that way, it reduces duplication.

PR #1511, opened 2020-04-15
Remove OSCAP_EVALUATION_TARGET override

evgenyz:
As the system_info probe now properly support offline mode and it is able to return correct host name of the target system, there is no need in this override.

### DIFF
--- a/src/XCCDF/result.c
+++ b/src/XCCDF/result.c
@@ -224,14 +224,8 @@ void xccdf_result_fill_sysinfo(struct xccdf_result *result)
 
 	_xccdf_result_clear_metadata(XITEM(result));
 
-	/* override target name by environment variable */
-	const char *target_hostname = getenv("OSCAP_EVALUATION_TARGET");
-	if (target_hostname == NULL) {
-		target_hostname = sname.nodename;
-	}
-
 	/* store target name */
-	xccdf_result_add_target(result, target_hostname);
+	xccdf_result_add_target(result, sname.nodename);
 #elif defined(OS_WINDOWS)
 	TCHAR computer_name[MAX_COMPUTERNAME_LENGTH + 1];
 	DWORD computer_name_size = MAX_COMPUTERNAME_LENGTH + 1;
--- a/utils/oscap-chroot
+++ b/utils/oscap-chroot
@@ -94,7 +94,6 @@ fi
 # Learn more at https://www.redhat.com/archives/open-scap-list/2013-July/msg00000.html
 export OSCAP_PROBE_ROOT
 OSCAP_PROBE_ROOT="$(cd "$1" && pwd)" || die "Invalid CHROOT_PATH argument."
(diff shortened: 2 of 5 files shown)


jan-cerny:
@evgenyz 

I prefer to not remove it because it's a useful feature.

For example, I scan ubi8 container image against the OSPP profile using `oscap-podman`:

```
sudo oscap-podman 096cae65a207 xccdf eval \
--report report.html --profile ospp \
/usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml
```

The generated HTML report contains the image hash, name, tage and container engine in Evaluation Characteristics table:
Screenshot_2020-04-15 xccdf_org open-scap_testresult_xccdf_org ssgproject content_profile_ospp OpenSCAP Evaluation Report

But, with this patch, it contains the host hostname. 
Screenshot_2020-04-15 xccdf_org open-scap_testresult_xccdf_org ssgproject content_profile_ospp OpenSCAP Evaluation Report(1)

We lost information about what was scanned. All containers or images that I scan on my system will have this same. I can't identify what was scanned. Moreover, it suggests that the host was scanned which isn't true. Unfortunately, there is no other place in the HTML report that would contain the podman image ID or similar information.

evgenyz:
This PR supposed to be a draft, forgot to mark it like that. I run some tests after creating this PR and found out that there is something wrong with the result build process. I'll continue investigation.

evgenyz:
@jan-cerny The problem actually is in `XCCDF/result.c`, where system information is collected directly via syscalls. So, no matter what we did in sysinfo probe to fix offline mode, it is not used to populate the result. And while evaluation target (hostname) could be overridden, there is no fix for e.g. interfaces at this moment.

My proposal here is to use system information from OVAL probe in `result.c` and remove `OSCAP_EVALUATION_TARGET` from `oscap-chroot` as it is generic enough to leverage the existence of offline mode for sysinfo probe. Is there any reason it was not implemented before?

It's a question if we should remove it from `oscap-docker`, `oscap-podman` and `oscap-vm`. On one hand we have a nice feature to distinguish scanned systems, on the other hand, as @matejak noted, this feature renders the result invalid.

> When I scan using oscap-chroot and save the result as ARF (oscap-chroot /tmp xccdf eval --results-arf r.arf ssg-fedora-ds.xml), it doesn't pass the subsequent validation (oscap ds rds-validate r.arf), because hostname has a chroot://<FS_PATH> format, and that doesn't go well with the ARF schema.

yuumasato:
> We lost information about what was scanned. All containers or images that I scan on my system will have this same. I can't identify what was scanned. Moreover, it suggests that the host was scanned which isn't true.

Shouldn't the system_info probe, in offline mode, actually return `096cae65a207` instead of `thinkpad`?

evgenyz:
It does return exactly that. But the problem is that the probe data is not used to fill Evaluation characteristics.

evgenyz:
Closing it in favour of #1519.

PR #1514, opened 2020-04-17
xccdf: Introduce 'urn:xccdf:fix:script:kubernetes' fix type

JAORMX:
This introduces the `urn:xccdf:fix:script:kubernetes` fix type which
will get added to the `system` entry in the `xccdf:fix` element. Similar
to the `urn:xccdf:fix:script:ignition`, this is initially meant for
OpenShift, however, the intent is to make these fixes more general and
applicable to other Kubernetes distributions.

This is a superset of the `urn:xccdf:fix:script:ignition` type, so it'll
eventually replace it.


jan-cerny:
@openscap-ci add to whitelist

jan-cerny:
@JAORMX I have changed the base branch to maint-1.3 which is more appropriate branch for this patch. Please rebase and remove the unrelated commits from this PR.

JAORMX:
@jan-cerny done, thanks!

PR #1519, opened 2020-04-24
Fixes for Evaluation Characteristics

evgenyz:
OSCAP_EVALUATION_TARGET is now stored in `target-facts` as `xccdf:fact:identifier` and `target` is always a hostname (with offline support).

ARF result is always valid.

Network entities are not collected in offline mode.

Identity entities are not collected in offline mode.

### DIFF
--- a/src/DS/rds.c
+++ b/src/DS/rds.c
@@ -354,6 +354,11 @@ static xmlNodePtr ds_rds_add_ai_from_xccdf_results(xmlDocPtr doc, xmlNodePtr ass
 		}
 	}
 
+	if (xmlGetLastChild(connections) == NULL) {
+		xmlUnlinkNode(connections);
+		xmlFreeNode(connections);
+	}
+
 	return asset;
 }
 
--- a/src/OVAL/probes/independent/system_info_probe.c
+++ b/src/OVAL/probes/independent/system_info_probe.c
@@ -472,25 +472,12 @@ static ssize_t __sysinfo_saneval(const char *s)
 }
 
 #ifndef OS_WINDOWS
-static FILE *_fopen_with_prefix(const char *prefix, const char *path)
-{
-	FILE *fp;
-	if (prefix != NULL) {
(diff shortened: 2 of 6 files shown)


evgenyz:
I have one question here. Do we want to have OSCAP_EVALUATION_TARGET value in ARF?

jan-cerny:
@evgenyz  I like this change very much. I have checked the results and reports and they look fine.


> I have one question here. Do we want to have OSCAP_EVALUATION_TARGET value in ARF? The ai:computing-device does not have any suitable element for this.

Logically, it would make sense to have it there. But, we don't use these elements anywhere. If there is no suitable place let's not add it.


I remember we talked about the fact that the generated ARF is invalid when scanning containers. You have mitigated the cause of it which was the `podman-image://` string in  `<ai:hostname>` element. However, the generated ARF is still invalid due to empty `ai:connections` element. I think it can be easy to fix. Can you check it? Here are the steps to reproduce:
```
[jcerny@thinkpad utils{pr/1519}]$ sudo ./oscap-podman --oscap=/home/jcerny/work/git/openscap/build/oscap_wrapper 096cae65a207 xccdf eval --profile ospp --report report.html --results-arf arf.xml /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml

[jcerny@thinkpad utils{pr/1519}]$ oscap ds rds-validate arf.xml 
File 'arf.xml' line 65535: Element '{http://scap.nist.gov/schema/asset-identification/1.1}connections': Missing child element(s). Expected is ( {http://scap.nist.gov/schema/asset-identification/1.1}connection ).
OpenSCAP Error: Invalid ARF Result Datastream (1.1) content in arf.xml. [/builddir/build/BUILD/openscap-1.3.2/src/source/oscap_source.c:346]
```

evgenyz:
> However, the generated ARF is still invalid due to empty `ai:connections` element. I think it can be easy to fix. Can you check it? Here are the steps to reproduce:

Done.

jan-cerny:
@evgenyz Thanks!

Issue #1512, opened 2020-04-15
Severity refinement lost in generated guide


iokomin:
#### Description of Problem:

Severity refinement available in evaluation scan report while lost in generated guide.

#### OpenSCAP Version:
1.2.17-4

#### Operating System & Version:
OL7.6

#### Steps to Reproduce:

1. create profile with rule refinement e.g. for rule with default _medium_ severity:
`<select idref="grub2_uefi_password" selected="true"/>
<refine-rule idref="grub2_uefi_password" severity="high"/>`
2. generate guide 
`oscap xccdf generate guide --profile stig --output guide.html <xccdf>`
3. review rule severity in guide

#### Actual Results:
severity is default, for our example - _medium_

#### Expected Results:
severity refined, for our example - _high_

#### Additional Information / Debugging Steps:
evaluation scan produces expected severity for this rule - _high_

PR #1520, opened 2020-04-24
Use rule severity refinement

jan-cerny:
This patch will ensure that rule severity changed by refine-rule element
will be correctly displayed in HTML guide and HTML report.  Before this
patch, the refine-rule element was ignored and severity defined in the
Rule element has been shown instead.

This patch still doesn't support refine-rule from parent profile if the
given profile extends another profile. But, the whole XSLT doesn't
support extended profiles and requires fully resolved benchmark as an
input. Therefore I think it doesn't make sense to implement it only
for this particular value.

Fixes: #1512


evgenyz:
LGTM. Great patch, thanks!

PR #1521, opened 2020-04-25
update docker image to fedora 31

ducksecops:
update docker image to fedora 31 as current 29 is EOL as of 30th November 2019


jan-cerny:
Thanks

PR #1524, opened 2020-04-30
Merge maint-1.3

evgenyz:


jan-cerny:
@evgenyz I don't recommend merging the branches via PR because you have to resolve conflicts and you can't do that without force-pushing which is no-go. I recommend merging using command line instead.

evgenyz:
Yeah, figured as much

PR #1525, opened 2020-05-04
Fix oscap-docker when Atomic is not present

jan-cerny:
For more details, please see the commit messages of each commit.


evgenyz:
Great, thanks. It looks like nobody is using Docker any more (or not without Atomic anyway).

PR #1526, opened 2020-05-12
probes/environmentvariable58: Fix regression in container variables

evgenyz:
Force the scanner to account only for variables passed via
OSCAP_CONTAINER_VARIABLES (only if it is not empty).

### DIFF
--- a/src/OVAL/probes/independent/environmentvariable58_probe.c
+++ b/src/OVAL/probes/independent/environmentvariable58_probe.c
@@ -96,32 +96,32 @@ static int read_environment(SEXP_t *pid_ent, SEXP_t *name_ent, probe_ctx *ctx)
 	ssize_t buffer_used;
 	size_t buffer_size;
 
+	const char *extra_vars = getenv("OSCAP_CONTAINER_VARS");
+	if (extra_vars && *extra_vars) {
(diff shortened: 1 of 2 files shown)


jan-cerny:
@evgenyz Why don't we read from /proc of the mounted container?

jan-cerny:
@openscap-ci test this please

evgenyz:
@jan-cerny The way we mount the container in our wrappers the `/proc` is empty. I want to investigate how to do in a way OCP does, but it's another topic.

jan-cerny:
I thought that there was  a random fail, so I restarted the test, but they failed again, It's test `probes/password/test_probes_password.sh`.

jan-cerny:
FTR We have investigated the reason of the CI failure. The failing test is `probes/password/test_probes_password.sh`. This failure is reproducible consistently on any Fedora 32 system. The problem is that `getpwent` that we use in password probe returns multiple records for the `root` and `nobody` users. It's caused by systemd-userdbd.service which is a new thing in systemd 245 and therefore a new thing in Fedora 32. See https://github.com/systemd/systemd/issues/15160 . This problem is unrelated to this PR because it's a problem in systemd.

PR #1529, opened 2020-05-15
Workaround systemd bug in password probe test

jan-cerny:
This test ensures that OpenSCAP password probe returns the same data
as defined in `/etc/passwd`. But, the probe doesn't read the
`/etc/passwd` file. It uses `getpwent` function instead, which employs
also other user sources, including systemd-userdb.service. Systemd
245 is affected by a bug that causes that `getpwent` or `getent passwd`
command returns `root` and `nobody` users twice, each time returns
a different login shell. This bug is already fixed upstream, but
occurs on Fedora 32.

See: https://github.com/systemd/systemd/issues/15160

To prevent CI failures we can workaround the test. If an affected
version is present on the system, it won't test the items returned
for the 2 problematic users. All other user items will be still
tested. Since all the items are produced by the same code, the impact
of workaround will not be drastic.

### DIFF
--- a/tests/probes/password/test_probes_password.xml.sh
+++ b/tests/probes/password/test_probes_password.xml.sh
@@ -1,9 +1,23 @@
 #!/usr/bin/env bash
 
-LINES_COUNT=`cat /etc/passwd | wc -l`
+
+passwd_file=`mktemp`
+
+# getpwent returns duplicate entries for root and nobody users
+# due to a bug in systemd-userdb.service that occurs
+# in systemd 245
+# https://github.com/systemd/systemd/issues/15160
+if which -q rpm &>/dev/null && \
+		[[ "$(rpm -q --qf "%{VERSION}" systemd)" =~ ^245 ]] ; then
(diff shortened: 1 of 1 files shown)


evgenyz:
It is pity that we have to bloat out test suite because of that glitch. I wonder if this fix https://github.com/systemd/systemd/pull/15553 is going to land in F32 any time soon.

jan-cerny:
@evgenyz I also don't like it. I don't know if my proposal is the best thing to do. Other possible solutions are:
* Ignore the failure and merge PRs anyway. That looks unprofessional. But maybe it will get fixed soon and it won't affect many PRs.
* Temporarily disable the test. That introduces a  risk that we will break the probe in meanwhile. Also, somebody will have to remember to enable it again.
* Modify the test in a different way. For example, we can check for specific users, or maybe only for users that we have created. It won't verify that the probe can return data for any user.

evgenyz:
Looks like this problem is not much of a problem for anyone except us. Merging.

PR #1534, opened 2020-05-27
Prevent unwanted recursion in file probe

jan-cerny:
There is a non-optimal behavior of file probe. It happens when file path
is specified using a variable with 2 values with `operation="equals"`
and `var_check="all"`. The probe recurses into a file system tree even
if it's obvious that it won't find any match. If one of values is a big
tree (for example `/`) it eventually runs out of memory and crashes. The
OVAL doesn't make sense because it's impossible that a single file would
have 2 different paths. But despite that it's a valid OVAL document.

We can fix this in oval_fts module. If the path is specified with operation "equals" we know that once we compared the given path with the OVAL object we don't have to continue examining the file system hierachy and we can decide right away.

This PR also adds a test.

Fixes: RHBZ#1686370

### DIFF
--- a/tests/probes/file/CMakeLists.txt
+++ b/tests/probes/file/CMakeLists.txt
@@ -1,3 +1,4 @@
 if(ENABLE_PROBES_UNIX)
 	add_oscap_test("test_probes_file.sh")
+	add_oscap_test("test_probes_file_multiple_file_paths.sh")
 endif()
--- a/tests/probes/file/test_probes_file_multiple_file_paths.sh
+++ b/tests/probes/file/test_probes_file_multiple_file_paths.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+set -e -o pipefail
+
+. $builddir/tests/test_common.sh
+
+probecheck "file" || exit 255
+which strace || exit 255
+
+function check_strace_output {
+	strace_log="$1"
+	grep -q "/tmp/numbers/1" $strace_log && return 1
+	grep -q "/tmp/numbers/1/2" $strace_log && return 1
+	grep -q "/tmp/numbers/1/2/3" $strace_log && return 1
+	grep -q "/tmp/numbers/1/2/3/4" $strace_log && return 1
+	grep -q "/tmp/numbers/1/2/3/4/5" $strace_log && return 1
+	grep -q "/tmp/numbers/1/2/3/4/5/6" $strace_log && return 1
+	grep -q "/tmp/letters/a" $strace_log && return 1
+	grep -q "/tmp/letters/a/b" $strace_log && return 1
+	grep -q "/tmp/letters/a/b/c" $strace_log && return 1
+	grep -q "/tmp/letters/a/b/c/d" $strace_log && return 1
+	grep -q "/tmp/letters/a/b/c/d/e" $strace_log && return 1
+	grep -q "/tmp/letters/a/b/c/d/e/f" $strace_log && return 1
+	return 0
+}
+
+rm -rf /tmp/numbers
+mkdir -p /tmp/numbers/1/2/3/4/5/6
+rm -rf /tmp/letters
+mkdir -p /tmp/letters/a/b/c/d/e/f
+strace_log=$(mktemp)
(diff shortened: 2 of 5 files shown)


jan-cerny:
The tests pass on Jenkins because the test is skipped because strace isn't installed on the machine. I plan to install it to make the test failing.

tests/probes/file/test_probes_file_multiple_file_paths.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -e -o pipefail
+
+. $builddir/tests/test_common.sh
+
+probecheck "file" || exit 255
+which strace || exit 255
+
+function check_strace_output {
```
matusmarhefka, line 10: So basically we expect that only `/tmp/numbers` and `/tmp/letters` should be found in strace and that oscap should not evaluate further inside these directories right?

tests/probes/file/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 if(ENABLE_PROBES_UNIX)
 	add_oscap_test("test_probes_file.sh")
+	add_oscap_test("test_probes_file_multiple_file_paths.sh")
```
matusmarhefka, line 3: As you mentioned, test will start to fail as soon as we add strace dependency. Does it make sense to enable it in CI as it will be always failing? Or do you plan to fix the probe behavior within this PR?

jan-cerny:
@openscap-ci test this please

tests/probes/file/CMakeLists.txt
jan-cerny: I have added strace a couple minutes ago and it started to fail. I think the idea is to have a test available before we start implementing the fix. We plan to push the fix into this PR so then this test will start to pass.

tests/probes/file/test_probes_file_multiple_file_paths.sh
jan-cerny: Your understanding is correct.

evgenyz:
Any particular reason for it not being a draft? So that, you know, no one could get some funny ideas, merge it, and ruin CI.

jan-cerny:
Here is a quick introduction to the implementation details.

In file probe, we use `oval_fts` module. This module helps us to collect files in file system according to the OVAL file_test. The `oval_fts` module is declared in `src/OVAL/probes/oval_fts.h`. It is also used in many other probes that work with the file system, eg. textfilecontent54, sysctl, yamlfilecontent.

The `oval_fts` module is built on standard `fts` functions from glibc that serve for traversing file hierarchies. See `man 3 fts` man page. 

The reason of having a special module for finding files is grounded in the OVAL specification, because the OVAL specification allows many options to describe file objects. A file can be specified either using `filepath` element or by using a combination of `path` and `filename` elements. Each of these elements support other operations than equals, they even can describe the file using a regular expression which means the probe has to search to collect the right file or files. Moreover, the file_object has `behaviors` element, which can instruct the scanner to search recursively in sub-directories and parent directories. There are multiple attributes. Together, there is many possible combinations that the probe has to support. If you’re interested, check the OVAL specification for the file_object.

The file probe operates in the following way: First, it parses the data that come from elements and attributes and normalizes them which includes setting the default values. Second, it opens a file hierarchy stream using `oval_fts_open` which call `fts_open` function inside. Third, it iterates over the files that meet the criteria using `oval_fts_read` which calls `fts_read` inside. For each found file it calls `file_cb` function that processes the found file and adds it to the set of collected OVAL items.

matejak:
The fix addresses the issue by modifying the `oval_fts_read_match_path` function that handles general file searching. That  function doesn't have obvious dedicated code paths for cases when the operation is `equals`, `not equal`, and `pattern match`. Therefore, it behaves inefficiently, as the `equals` operation is relatively trivial. The PR adds a "shortcut" which short-circuits the main function loop with the justification that is recorded in the code comment.
As the fix alters the code flow only for cases when one uses the `equals` operation, possible regressions will be easy to debug - it is easy to see when an equals operation doesn't produce desired results, as opposed to a recursion into a possibly non-local folder etc.
Therefore, I am merging the fix, as it can significantly improve file matching when the `equals` operation is specified.

PR #1537, opened 2020-06-03
Fix compiler warnings

jan-cerny:
This PR aims to fix some of the warnings that appeared with GCC 10. For more details, please see the commit message of each commit.


jan-cerny:
@evgenyz Great idea! I'll check it.

jan-cerny:
I have removed the packed attribute from remaining packed structures. I also run our ABI compliance checker, comparing this with OpenSCAP 1.3.3 and it doesn't report any API/ABI break.

evgenyz:
LGTM! Thanks!

PR #1543, opened 2020-06-15
Fixing pthreads config missing when X-compiling

carlosmmatos:
This is in reference to #1073 in order to support cross-compiling OpenSCAP. See the build output attached for the issue. 

I also tested this change to ensure appveyor build was not affected as well.

[cross-compile-build-log.txt](https://github.com/OpenSCAP/openscap/files/4782392/cross-compile-build-log.txt)


evgenyz:
Thank you for the patch!

PR #1545, opened 2020-06-15
Fixing issue with getopt when using mingw to cross compile

carlosmmatos:
This PR depends on #1543 - Fixes issue with cross compilation error using mingw32-cmake when looking for getopt's library.


carlosmmatos:
Added the changes - I knew something looked off yesterday when I did it lol.. good catch.

evgenyz:
OK, LGTM. Thanks!

PR #1549, opened 2020-06-16
GH Actions

evgenyz:
Set up GH Actions CI.

Fixed Travis MacOS build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,16 +6,52 @@ name: CI
 # events but only for the master branch
 on:
   push:
-    branches: [ master ]
+    branches: [ master, maint-1.3 ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, maint-1.3 ]
+
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   # This workflow contains a single job called "build"
-  build:
+  ubuntu:
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, ubuntu-20.04, ubuntu-latest]
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+
+    # Runs a single command using the runners shell
+    - name: Install packages
+      run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -6,14 +6,14 @@ name: CI
...
+    branches: [ master, maint-1.3 ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, maint-1.3 ]
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   # This workflow contains a single job called "build"
-  build:
+  ubuntu:
```
jan-cerny, line 16: Would it be possible to specify specific versions of Ubuntu?

evgenyz:
> It doesn't build the yamlfilecontent probe. Can you look at it?

Dammit, again :(

.github/workflows/main.yml
evgenyz: It is possible, but maybe it would be great to follow the edge in our CI?

jan-cerny:
> > It doesn't build the yamlfilecontent probe. Can you look at it?
> 
> Dammit, again :(

I can see that it uses --no-recurse-submodules.

.github/workflows/main.yml
ggbecker: what about using matrix to run multiple jobs: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

you can run the action on multiple versions at the same time

evgenyz: I'm not sure how hardly we could abuse GH builders, but okay. What do we want? ubuntu-latest + ubuntu 18.04 + ubuntu 20.04; macos-latest?

evgenyz: Options: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#github-hosted-runners

evgenyz: Hold on a second. I forgot Windows :). I'll add it

jan-cerny: Windows are already covered by AppVeyor and old Ubuntu is already covered by Travis.

evgenyz:
So, yaml probe is in place but it will fail until we fix dpkg probe in 20.04.

evgenyz:
Okay, dpkginfo is now fixed. Let's have it.

evgenyz:
Aha, now password probe test is failing on 20.04. Bummer!

evgenyz:
Finally!

jan-cerny:
Great, let's try it in action, we can always change it later.

Issue #1540, opened 2020-06-10
Error viewing report in epiphany web browser


bkmgit:
#### Description of Problem:

Trouble viewing report in the broswer

#### OpenSCAP Version:

1.3.3
#### Operating System & Version:

Fedora 31

#### Steps to Reproduce:

1.  sudo dnf install openscap openscap-utils epiphany
2.  sudo oscap xccdf eval  --report report.html /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml
3.  epiphany report.html

#### Actual Results:

This page contains the following errors:
error on line 19 at column 40: xmlParseEntityRef: no name
Below is a rendering of the page up to the first error.

![Report rendering error in Epiphany](https://user-images.githubusercontent.com/1679477/84242304-5659c300-ab09-11ea-802d-7f406867d014.png)

#### Expected Results:

View of report. Report opens ok in Chromium web browser

#### Additional Information / Debugging Steps:

Think this is an HTML issue. Is it worth trying to submit a fix?

evgenyz:
Hey, @bkmgit! Thank you for reporting the problem. It is of course worth submitting a fix. Even the description of the exact problem in HTML document could help.

jan-cerny:
This error message seems to be caused by using not escaped `&` within `<script>` element. It looks like it has to be escaped in XHTML or we have to use "CDATA". https://stackoverflow.com/questions/1513871/xhtml-wont-validate-and-in-a-javascript-function. If I fix this, then other errors pop up, for example unpaired tags, which are not allowed in XHTML. I think we should run the validator on the generated report.

bkmgit:
Ok. Ran the [W3C validator](https://validator.w3.org) on the generated report file. The output is below. Might it be worth adding some tests using w3c-markup-validator in Travis-CI? w3c-markup-validator is available in a number of packaged software repositories.

    Error: Attribute xmlns:arf not allowed here.

    From line 1, column 16; to line 1, column 135

    TYPE html><html xmlns="http://www.w3.org/1999/xhtml" xmlns:arf="http://scap.nist.gov/schema/asset-reporting-format/1.1" lang="en"><head>

    Warning: Attribute with the local name xmlns:arf is not serializable as XML 1.0.

    From line 1, column 16; to line 1, column 135

    TYPE html><html xmlns="http://www.w3.org/1999/xhtml" xmlns:arf="http://scap.nist.gov/schema/asset-reporting-format/1.1" lang="en"><head>

[...]

PR #1550, opened 2020-06-19
Make the HTML output valid

jan-cerny:
The HTML reports and guides were invalid. The main problem was the confusing namespace. The HTML output is HTML 5 according to doctype and according to its contents but the elements used the XHTML namespace. But, the HTML reports and guides aren't parseable as XML.

Also, there were some unpaired elements and elements that aren't allowed in the given context.

I managed to reduce the errors reported by W3C validator significantly. But, there remains error 
`CSS: Deprecated media feature max-device-width.`.

For more details, please see the commit message of each commit.

Fixes: #1540

### DIFF
--- a/xsl/xccdf-branding.xsl
+++ b/xsl/xccdf-branding.xsl
@@ -29,9 +29,9 @@ logo to the HTML report / guide header.
 -->
 
 <xsl:stylesheet version="1.1"
-	xmlns="http://www.w3.org/1999/xhtml"
 	xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
-	xmlns:cdf="http://checklists.nist.gov/xccdf/1.2">
+	xmlns:cdf="http://checklists.nist.gov/xccdf/1.2"
+	exclude-result-prefixes="xsl cdf">
 
 <xsl:param name="oscap-version"/>
 
--- a/xsl/xccdf-branding.xsl
+++ b/xsl/xccdf-branding.xsl
@@ -48,7 +48,7 @@ logo to the HTML report / guide header.
 </xsl:template>
 
 <xsl:template name="xccdf-report-header">
-    <nav class="navbar navbar-default" role="navigation">
+    <nav class="navbar navbar-default">
         <div class="navbar-header" style="float: none">
             <a class="navbar-brand" href="#">
                 <xsl:call-template name="xccdf-branding-logo"/>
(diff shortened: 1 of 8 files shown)


evgenyz:
In newer `bootstrap` v4.5.0 there is no `max-device-width` feature.

evgenyz:
BTW, which validator did you use?

jan-cerny:
I have used online validator on https://validator.w3.org/.

There are catches: The validator output actually depends on the specific HTML report or guide. Selections of rules can trigger some of the errors. It also seems that this online validator can't process big files. I used tailored Fedora standard profile that deselected some random rules to make the report smaller. I also tried empty profile.

jan-cerny:
@evgenyz Do you think we should update bootstrap to the latest version?

evgenyz:
> @evgenyz Do you think we should update bootstrap to the latest version?

Yeah, because I think it should fix the problem. But it might break something (current is 4-years(sic!) old).

evgenyz:
> I have used online validator on https://validator.w3.org/.
> 
> There are catches: The validator output actually depends on the specific HTML report or guide. Selections of rules can trigger some of the errors. It also seems that this online validator can't process big files. I used tailored Fedora standard profile that deselected some random rules to make the report smaller. I also tried empty profile.

There is `w3c-markup-validator` package in Fedora. It worth checking if it could handle large files. And we could use it the test suite in some way.

jan-cerny:
The problem with updating Bootstrap is that recently we have done some changes in bootstrap to fix printed HTML report issues. The patch https://github.com/OpenSCAP/openscap/commit/d419d8c2de3cfc97ee9ba388dd0f32aa332940b3 changes the `bootsrap.min.css` and we would have to do the changes again.

jan-cerny:
I think that using a deprecated feature has a very small impact. Since updating the bundled bootstrap isn't very easy, I suggest fixing it in a separate PR. I don't want to block the more important fixes such as the namespace fixes by that.

evgenyz:
That's totally makes sense to fix Bootstrap in a separate PR. But, will #1540 be fixed with this CSS-related error still in the report?

jan-cerny:
@evgenyz Yes, it will, the problem in Epiphany was basically caused by inconsistent namespaces (XHTML namespace in HTML 5 content). Actually, 2d91fed is enough to make the report viewable in Epiphany.

evgenyz:
Great! LGTM then, thank you!

PR #1552, opened 2020-06-25
probes/yamlfilecontent: Bump yaml-filter, extend the schema and probe to be able to work with a set of values in maps

evgenyz:
This introduces YAML Path *selection* `$.blah['key1','key2']` syntax and
enriches yamlfilecontent test and schema with `EntityStateRecordType` and
`EntityItemRecordType` result elements for collecting and checking
complex objects in YAML/JSON documents.

This change also depends on removal of the 'only lower case' restriction
from the `name` attribute of the field element (`EntityXxxFieldType`).


evgenyz:
Example of a check for the "Upgradable" problem:

```xml
    <ind-def:yamlfilecontent_object version="1" id="oval:0:obj:5">
      <ind-def:path>/tmp</ind-def:path>
      <ind-def:filename>openshift-logging.yaml</ind-def:filename>
      <ind-def:yamlpath>.status.conditions[:]['status','type']</ind-def:yamlpath>
    </ind-def:yamlfilecontent_object>

    <ind-def:yamlfilecontent_state version="1" id="oval:0:ste:5">
      <ind-def:result datatype="record">
        <field name="status" datatype="string">True</field>
        <field name="type" datatype="string">Upgradeable</field>
      </ind-def:result>
    </ind-def:yamlfilecontent_state>
```

More details in `tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml`.

CC @JAORMX

evgenyz:
Problems so far: this thing
```xml
<field name="value_of" datatype="string" var_ref="oval:0:var:2" var_check="at least one" entity_check="at least one"/>
```
is not working, no idea why. (CC @jan-cerny)

evgenyz:
As a bonus side of this change – we can now address any map, given that there is no values with complex objects (selection syntax could help if that's not the case).

evgenyz:
This is a possible fix for https://github.com/ComplianceAsCode/content/pull/5822.

yuumasato:
> Problems so far: this thing
> 
> ```
> <field name="value_of" datatype="string" var_ref="oval:0:var:2" var_check="at least one" entity_check="at least one"/>
> ```
> 
> is not working, no idea why. (CC @jan-cerny)

It could be because `record` data type does not allow variables.
https://github.com/OVALProject/Language/blob/master/docs/oval-common-schema.md#---complexdatatypeenumeration---

schemas/oval/5.11.3/independent-definitions-schema.xsd
```diff
@@ -2115,7 +2115,7 @@
...
+                                                            <xsd:documentation>Specifies an YAML Path expression to evaluate against the YAML file specified by the filename entity. This YAML Path expression must evaluate to a list or a map (part of a map) of zero or more scalar values which will be accessible in OVAL via instances of the result entity. Any results from evaluating the YAML Path expression other than a list of scalar values (or a map with scalar values) at the bottom level (e.g. list of lists, list of maps, map with maps or lists) is considered an error. The intention is to drawn scalar values from instance(s) of a single (or multiple) flat element(s). However, an OVAL interpreter is not required to verify this, so the author should define the YAML Path expression carefully. Note that "equals" is the only valid operator for the yamlpath entity.</xsd:documentation>
```
matejak, line 2118: Description remarks:

- I believe that drawn -> draw
- What is a flat element?
- What is an OVAL interpreter not required to verify?

evgenyz:
> > Problems so far: this thing
> > ```
> > <field name="value_of" datatype="string" var_ref="oval:0:var:2" var_check="at least one" entity_check="at least one"/>
> > ```
> > 
> > 
> > is not working, no idea why. (CC @jan-cerny)
> 
> It could be because `record` data type does not allow variables.
> https://github.com/OVALProject/Language/blob/master/docs/oval-common-schema.md#---complexdatatypeenumeration---

Yeah, but it a field (child of it, not record itself). I'm not sure here, but I understood that should be possible for field.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
```diff
@@ -76,11 +76,16 @@
   <states>
 
     <ind-def:yamlfilecontent_state version="1" id="oval:0:ste:1">
-      <ind-def:value_of datatype="string">secureforward-offcluster</ind-def:value_of>
+      <ind-def:result datatype="record">
```
matejak, line 79: This is a UX regression, checks that a particular value is present in a particular place will be quite common, and they are somehow difficult, IIUC the `value_of` in `name` is basically a keyword.
Is it possible to somehow preserve the old approach? I would even consider a new object/test/state for the proposed extension.

schemas/oval/5.11.3/independent-definitions-schema.xsd
```diff
@@ -2158,13 +2158,24 @@
...
                                           </xsd:annotation>
                                     </xsd:element>
-                                    <xsd:element name="value_of" type="oval-def:EntityStateAnySimpleType" minOccurs="0" maxOccurs="1">
+                                    <xsd:element name="result" type="oval-def:EntityStateRecordType" minOccurs="0">
```
yuumasato, line 2164: The OVAL schema specifies that name of the `field` should be lowercase and unique within a record.
https://github.com/OVALProject/Language/blob/master/docs/oval-definitions-schema.md#-entityobjectfieldtype-

Could we have yaml files with upper case letters in the keys?
I guess the `name` attribute in the `field` could have it  lowercase'd.
But what if the mapping already has the lowercased key? (Is it a too "corney" case?)

yuumasato:
> Yeah, but it's a field (child of it, not record itself). I'm not sure here, but I understood that should be possible for field.

Well noted, it makes sense that records cannot be compared to variables, and there is nothing related to fields comparison with variables.

schemas/oval/5.11.3/independent-definitions-schema.xsd
evgenyz: I removed that requirement in this PR. It'd also be proposed to the OVAL community to remove this restriction (the alternative is a new CaseSensitiveField type or something like that).

schemas/oval/5.11.3/oval-definitions-schema.xsd
```diff
@@ -1553,13 +1553,8 @@
             <xsd:extension base="xsd:anySimpleType">
                 <xsd:attribute name="name" use="required">
                     <xsd:annotation>
-                        <xsd:documentation>A string restricted to disallow upper case characters.</xsd:documentation>
```
jan-cerny, line 1556: I'm afraid that changing the core part of schema will be difficult.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
```diff
@@ -41,6 +41,17 @@
...
     </definition>
 
+    <definition class="compliance" version="1" id="oval:0:def:4"> <!-- comment="true" -->
+      <metadata>
+        <title></title>
+        <description></description>
+      </metadata>
+      <criteria operator="AND">
+        <criterion comment="multiple_mapping_values" test_ref="oval:0:tst:5"/>
+        <criterion comment="map_target" test_ref="oval:0:tst:6"/>
```
jan-cerny, line 51: I'd prefer to have a separate definition for each of the tests because that makes the debugging easier.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
jan-cerny: I think the `yamlfilecontent_item` could contain both `value_of` element and `result`. The `value_of` element would be used for scalars and `result` for maps. If that's going to happen the `result` should probably be renamed.

schemas/oval/5.11.3/independent-system-characteristics-schema.xsd
```diff
@@ -606,12 +606,19 @@
...
                                    </xsd:annotation>
                               </xsd:element>
-                              <xsd:element name="value_of" type="oval-sc:EntityItemAnySimpleType" minOccurs="0" maxOccurs="unbounded">
+                              <xsd:element name="result" type="oval-sc:EntityItemRecordType" minOccurs="0" maxOccurs="unbounded">
                                    <xsd:annotation>
-                                        <xsd:documentation>The value_of element checks the value(s) of the text node(s) or attribute(s) found. How this is used is entirely controlled by operator attributes.</xsd:documentation>
+                                        <xsd:documentation>The result entity holds the target of the specified YAML Path. If YAML Path target is a scalar value or a list of scalar values, the field child element(s) would be named 'value_of'. The check is entirely controlled by operator attributes of the field element.</xsd:documentation>
```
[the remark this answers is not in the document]
jan-cerny: Interesting idea. Could you give an example of "actual name of the leaf key"?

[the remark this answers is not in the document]
evgenyz: I thought about that as well, but things like `$.abc[:][:][0:3][1]` ending up in `<filed name="abc">` could be seen as awkward. Or maybe they would not.

schemas/oval/5.11.3/oval-definitions-schema.xsd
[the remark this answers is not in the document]
evgenyz: Alternative is adding `CaseSensitiveField`. Also the change will relax requirements, it is definitely backward-compatible. I hope it could be possible. Or we will add `YamlField` type or something.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
evgenyz: This is possible in theory. Also the `result` elements could be renamed to `values`. Not sure if this polymorphic approach would be in the spirit of OVAL language, though.

evgenyz:
> Would it be possible to split the C code into more functions that would make it more obvious what is going on there? The code is starting to be complex, and it is said that function names document better than comments that tend to become obsolete very quickly.

I will brush it up during the final implementation of all the details we are currently discussing (like having `value_of` entry together with `result` and so on). This version was made with the least amount of code changes in mind to show what is actually going on.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
evgenyz: Got it, OK

schemas/oval/5.11.3/independent-system-characteristics-schema.xsd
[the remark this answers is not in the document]
yuumasato: > $.abc[:][:][0:3][1]

You should know what you are checking for.
I'd argue that it is very hard to know what this yamlpath wants to check, and it is very much order dependent.
The field name would give some guarantee that the yamlpath found the item you wanted. If the field name doesn't match the leaf key name, I think it should result in error.

From [Entity state field](https://github.com/OVALProject/Language/blob/master/docs/oval-definitions-schema.md#-entitystatefieldtype-):
> When analyzing system characteristics an error should be reported for the result of a field that is present in the OVAL State, but not found in the system characteristics Item.

schemas/oval/5.11.3/independent-definitions-schema.xsd
jan-cerny: Yes, "draw" is correct.

Flat element is probably a YAML key that contains only scalar. I think this needs to be clarified.

The OVAL interpreter isn't required to verify that the YAML path expresssion in the given object matches a scalar, a list of scalars or a map of scalars.

evgenyz: Yeah, after reading it couple more times I realized that it needs some rephrasing.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
jan-cerny: Good point. I skimmed through the specification and I can't find existing example of it. The win57_item, activedirectory57_item, ldap57_item or sql_57 return a single element.

Regarding the naming,  In sql57_item and wmi57_item it's called "result". In  activedirectory57_item and ldap57_item it's "value". I like more "value" because YAML path expressions are describing values in some files. But actually it doesn't matter.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
evgenyz: Done.

schemas/oval/5.11.3/oval-definitions-schema.xsd
evgenyz: I'm afraid that it is the only sane way. Anyhow, until OVAL decides something, I think it would be the best option to stick with the simplest implementation.

evgenyz:
> > > Problems so far: this thing
> > > ```
> > > <field name="value_of" datatype="string" var_ref="oval:0:var:2" var_check="at least one" entity_check="at least one"/>
> > > ```
> > > 
> > > 
> > > is not working, no idea why. (CC @jan-cerny)
> > 
> > 
> > It could be because `record` data type does not allow variables.
> > https://github.com/OVALProject/Language/blob/master/docs/oval-common-schema.md#---complexdatatypeenumeration---
> 
> Yeah, but it's a field (child of it, not record itself). I'm not sure here, but I understood that should be possible for field.

So, it just wasn't implemented. I fixed it.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
evgenyz: So, as no other probe does this polymorphic stuff, I think it would be better to stick with only record-type result. We already have enough problems with adoption of the probe.

But, `value` or `result`? Speak up!

schemas/oval/5.11.3/independent-system-characteristics-schema.xsd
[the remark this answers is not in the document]
evgenyz: It seems that `<field name="">` is a legit thing, so I implemented it as a compromise.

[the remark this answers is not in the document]
evgenyz: In case it will yield some problems later, we could fall back to "value_of" or "actual name of the leaf key".

schemas/oval/5.11.3/independent-definitions-schema.xsd
evgenyz: Fixed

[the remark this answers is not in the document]
evgenyz: Fixed

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
matejak: I side with `value` as well.

src/OVAL/results/oval_cmp_impl.h
```diff
@@ -29,16 +29,6 @@
...
-/**
- * Compare state entity (or variable/value) to sysent object collected from system.
- * This function does not support @datatype="record".
- * @param state_value Value defined within state/entity/value or variable/value
- * @param sysent Value collected from system
- * @operation Comparison type operation
- * @returns OVAL Result of comparison
- */
-oval_result_t oval_ent_cmp_str(char *state_data, oval_datatype_t state_data_type, struct oval_sysent *sysent, oval_operation_t operation);
-
```
yuumasato, line 41: Shouldn't we keep this for API compatibility?

evgenyz: Mmm, it looks like a private API.

evgenyz: @jan-cerny Am I right?

evgenyz: `abi-compliance-checker` is OK with this change.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml
evgenyz: Done.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
```diff
@@ -55,14 +75,24 @@
...
+    <ind-def:yamlfilecontent_test version="1" id="oval:0:tst:5" check="all" comment="true">
+      <ind-def:object object_ref="oval:0:obj:5"/>
+      <ind-def:state state_ref="oval:0:ste:5"/>
+    </ind-def:yamlfilecontent_test>
+
+    <ind-def:yamlfilecontent_test version="1" id="oval:0:tst:6" check="all" comment="true">
+      <ind-def:object object_ref="oval:0:obj:6"/>
+      <ind-def:state state_ref="oval:0:ste:6"/>
+    </ind-def:yamlfilecontent_test>
+
```
matusmarhefka, line 95: I would consider adding also a test which doesn't match any map in the yaml file with `comment="false"`.

evgenyz: What is the purpose? What should be different from `get_non_existing_scalar` case?

matusmarhefka: Oh sorry, I didn't expressed it correctly, I meant test which doesn't match anything inside a map, for example `.status.conditions[:]['nonexistent']`. I believe it should exercise a different path in the code compared to the `get_non_existing_scalar` case.

src/OVAL/results/oval_cmp_impl.h
jan-cerny: @evgenyz @yuumasato Yes, this is a private function. Public API functions have `OSCAP_API` decorator and they are usually located in `public` directories.

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_types.sh
```diff
@@ -24,49 +24,49 @@ function test_probes_yamlfilecontent_types {
...
-    assert_exists 1 $sd'/ind-sys:yamlfilecontent_item/ind-sys:value_of[@datatype!="int" and text()="81"]'
+    assert_exists 1 $sd'/ind-sys:yamlfilecontent_item/ind-sys:value/field[@name="" and @datatype!="int" and text()="81"]'
 
     # bool_error_cast, int_error_cast, float_error_cast
     co='/oval_results/results/system/oval_system_characteristics/collected_objects'
     assert_exists 3 $co'/object[@flag="error"]'
     assert_exists 3 $co'/object[@flag="error"]/message'
 
-    rm -f $result
+    #rm -f $result
```
jan-cerny, line 69: ?

evgenyz: Oops :)

tests/probes/yamlfilecontent/test_probes_yamlfilecontent_key.xml
evgenyz: Done.

evgenyz:
All current issues were addressed, waiting for reaction in https://github.com/OVAL-Community/OVAL/issues/91.

JAORMX (approved):

schemas/oval/5.11.3/independent-system-characteristics-schema.xsd
```diff
@@ -606,12 +606,19 @@
...
                               </xsd:element>
-                              <xsd:element name="value_of" type="oval-sc:EntityItemAnySimpleType" minOccurs="0" maxOccurs="unbounded">
+                              <xsd:element name="value" type="oval-sc:EntityItemRecordType" minOccurs="0" maxOccurs="unbounded">
                                    <xsd:annotation>
-                                        <xsd:documentation>The value_of element checks the value(s) of the text node(s) or attribute(s) found. How this is used is entirely controlled by operator attributes.</xsd:documentation>
+                                        <xsd:documentation>The value entity holds the target(s) of the specified YAML Path. A signle scalar value or a list of scalar values (where there is no key to associate) would have the name attribute of the field element set to '#'. Due to the limitation of the record type field names could not contain uppercase letters, they will be converted to the lowercase and escaped using the '^' symbol (the '^' symbol would be escaped as well). For example 'myCamelCase^Key' would be collected as 'my^camel^case^^^key'.</xsd:documentation>
```
jan-cerny, line 614: signle => single

src/OVAL/probes/independent/yamlfilecontent_probe.c
```diff
@@ -265,21 +335,37 @@ static int process_yaml_file(const char *prefix, const char *path, const char *f
...
+					SEXP_t *value_sexp = oscap_iterator_next(item_value_it);
+					SEXP_t *field = probe_ent_creat1("field", NULL, value_sexp);
+					probe_item_attr_add(field, "name", key);
+					SEXP_list_add(result_ent, field);
+				}
+				oscap_iterator_free(item_value_it);
+				SEXP_free_r(&se_tmp_mem);
+			}
+			oscap_htable_iterator_free(record_it);
+			SEXP_list_add(item, result_ent);
```
jan-cerny, line 361: Please check if result_ent needs to be freed.

jan-cerny:
looks great

src/OVAL/probes/independent/yamlfilecontent_probe.c
evgenyz: Yep, thanks

schemas/oval/5.11.3/independent-system-characteristics-schema.xsd
evgenyz: Eagle-eye! Thanks

jan-cerny:
Let's merge it now, any other improvements can be opened separately.

PR #1561, opened 2020-07-13
Add CURLOPT_TRANSFER_ENCODING, enable CURLOPT_VERBOSE with CURLOPT_DEBUGFUNCTION

evgenyz:
Further improvement of #1557.

Adds a request for compressed Transfer Encoding in the outgoing
HTTP request. If the server supports this and so desires, it can
respond with the HTTP response sent using a compressed
Transfer-Encoding that will be automatically uncompressed by
libcurl on reception.

The CURLOPT_DEBUGFUNCTION callback is used for printing headers and
connection information on VERBOSE level (dD).

### DIFF
--- a/src/common/oscap_acquire.c
+++ b/src/common/oscap_acquire.c
@@ -49,6 +49,7 @@
 #include "common/_error.h"
 #include "oscap_string.h"
 #include "oscap_helpers.h"
+#include "debug_priv.h"
 
 #ifndef OSCAP_TEMP_DIR
 #define OSCAP_TEMP_DIR "/tmp"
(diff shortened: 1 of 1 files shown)


src/common/oscap_acquire.c
```diff
@@ -288,6 +289,36 @@ oscap_acquire_url_to_filename(const char *url)
...
+	case CURLINFO_DATA_OUT:
+	case CURLINFO_SSL_DATA_OUT:
+	case CURLINFO_DATA_IN:
+	case CURLINFO_SSL_DATA_IN:
+	default:
+		return 0;
+		break;
+	}
+
+	char *text = strndup(data, size);
```
evgenyz, line 315: I don't like that we would copy the data even if we are not in VERBOSE mode. Any suggestions?

evgenyz:
Example:
```
...
D: oscap: == cURL info:   Trying 93.184.216.34:443...
 [oscap(165266):oscap(7efc15e3b900):oscap_acquire.c:316:_curl_trace]
D: oscap: == cURL info: Connected to www.example.com (93.184.216.34) port 443 (#0)
 [oscap(165266):oscap(7efc15e3b900):oscap_acquire.c:316:_curl_trace]
...
D: oscap: => cURL header (out): GET /security/data/oval/oval.xml.bz2 HTTP/2
Host: www.example.com
accept: */*
connection: TE
te: gzip
accept-encoding: deflate, gzip, br
...
```

src/common/oscap_acquire.c
evgenyz: Scratch that. It seems that despite the warning headers and info are always NULL-terminated. We can leave it as it is now.

matusmarhefka:
LGTM, thanks @evgenyz

PR #1563, opened 2020-07-15
Fix test for password probe

evgenyz:
The `_apt:x:977:972:APT account for owning persistent & cache data:/var/lib/apt:/sbin/nologin`
entry in the /etc/passwd file generates an error:

```
145: Entity: line 863: parser error : xmlParseEntityRef: no name
145:       <gcos>APT account for owning persistent & cache data</gcos>
```


matusmarhefka:
LGTM, thanks @evgenyz

PR #1564, opened 2020-07-20
Add test for cURL "Accept-Encoding" header

evgenyz:


tests/curl/test_curl_encoding.sh
```diff
@@ -0,0 +1,23 @@
...
+set -e -o pipefail
+
+. $builddir/tests/test_common.sh
+
+function curl_accept_encoding {
+	local DF="${srcdir}/ds.xml"
+	local RF="results.xml"
+	local LOG="verbose.log"
+
+	$OSCAP xccdf --verbose=DEVEL eval --fetch-remote-resources --results $RF $DF 2>$LOG || echo "OK"
```
matusmarhefka, line 12: Just a nitpick, we are not checking or working with `$RF` in any way, so maybe we can remove `--results` option.

matusmarhefka: Also, shouldn't it rather be `... && echo "OK"`?

evgenyz: We are not using it because the result is mostly an error (since the remote content is just the github.com main page). But if we would use some real content it might make sense to check the result. I'm not sure here.

In that case won't neutralize the negative result from the scanner itself and test will fail.

matusmarhefka: Oh I see, the command actually fails but we are only interested if the scanner requested encoding through `Accept-Encoding`.

PR #1566, opened 2020-07-31
SCAP 1.3

evgenyz:
This PR is an evolution of #1323.

Details pending.


utils/oscap-ds.c
```diff
@@ -345,6 +349,22 @@ int app_ds_sds_compose(const struct oscap_action *action) {
 		goto cleanup;
 	}
 
+	oscap_document_version_t ver = OSCAP_DOCUMENT_VERSION_UNKNOWN;
+	if (action->f_scap_version) {
+		if (!strcmp(action->f_scap_version, "1.1"))
```
jan-cerny, line 354: I think it would be better to re-factor this block out to a new function, eg. `oscap_document_version_from_cstr`.

PR #1571, opened 2020-08-06
Plug a memory leak

jan-cerny:
Do not create the global name cache.  The ncache instance created in
`__init_once` is always replaced by a new instance of the cache in
`probe_common_main` (probe_main.c:217). The purpose of the removed
code is unclear to me, it seems to be a leftover from the multi-process
architecture used in branch maint-1.2.

Dicsovered when running valgrind on command:
oscap xccdf eval --results-arf /tmp/arf.xml --profile ospp --rule xccdf_org.ssgproject.content_rule_selinux_state /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml

Addressing:
==117734== 5,459 (640 direct, 4,819 indirect) bytes in 8 blocks are definitely lost in loss record 269 of 270
==117734==    at 0x483A809: malloc (vg_replace_malloc.c:307)
==117734==    by 0x4931C18: probe_ncache_new (ncache.c:81)
==117734==    by 0x49328AB: probe_common_main (probe_main.c:213)
==117734==    by 0x4CDA431: start_thread (in /usr/lib64/libpthread-2.31.so)
==117734==    by 0x52A8912: clone (in /usr/lib64/libc-2.31.so)

### DIFF
--- a/src/OVAL/oval_probe_session.c
+++ b/src/OVAL/oval_probe_session.c
@@ -49,74 +49,6 @@
 #include "probe-table.h"
 #include "oval_types.h"
 
-#if defined(OSCAP_THREAD_SAFE)
-#include <pthread.h>
-static pthread_once_t __oval_probe_session_init_once = PTHREAD_ONCE_INIT;
-#else
-static volatile int __oval_probe_session_init_once = 0;
-#endif /* OSCAP_THREAD_SAFE */
-
-/**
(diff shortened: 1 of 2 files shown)


jan-cerny:
Based on the test results this causes a segfault, probably a race condition. We propably need a different fix.

PR #1573, opened 2020-08-18
Change local and remote file system detection

jan-cerny:
The previous way was unreliable because some remote file systems were
detected as local. For example, the SMB mounts that used paths with
backslashes were not recognized as remote file systems.  The `getmntent`
function returns a file system type, so we can use it for detection of
remote file systems.

The new solution is inspired by systemd, function `fstype_is_network`:
https://github.com/systemd/systemd/blob/21fd6bc263f49b57867d90d2e1f9f255e5509134/src/basic/mountpoint-util.c#L290

It also handles FUSE file systems. Also, we are cancelling the TODO idea
of providing the data at the build time and removing the dead code (if
0).

GPFS is also treated as remote system because we have multiple reports that GPFS shouldn't be scanned.

Resolves: RHBZ#1869195, RHBZ#1870087, RHBZ#1840575, RHBZ#1840578

### DIFF
--- a/src/OVAL/probes/fsdev.c
+++ b/src/OVAL/probes/fsdev.c
@@ -94,6 +94,11 @@ static int is_local_fs(struct mntent *ment)
 		return 0;
 	}
 
+	/*
+	 * The following code is inspired by systemd, function fstype_is_network:
+	 * https://github.com/systemd/systemd/blob/21fd6bc263f49b57867d90d2e1f9f255e5509134/src/basic/mountpoint-util.c#L290
+	 */
+
 	const char *fstype = ment->mnt_type;
 	if (oscap_str_startswith(fstype, "fuse.")) {
 		fstype += strlen("fuse.");
--- a/src/OVAL/probes/fsdev.c
+++ b/src/OVAL/probes/fsdev.c
@@ -117,6 +117,7 @@ static int is_local_fs(struct mntent *ment)
 		"gfs",
 		"gfs2",
 		"glusterfs",
+		"gpfs",
 		"pvfs2", /* OrangeFS */
 		"ocfs2",
 		"lustre",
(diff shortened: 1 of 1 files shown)


src/OVAL/probes/fsdev.c
```diff
@@ -97,37 +94,36 @@ static int is_local_fs(struct mntent *ment)
 		return 0;
 	}
 
-	if (ment->mnt_fsname == NULL) {
-		return 0;
+	const char *fstype = ment->mnt_type;
+	if (oscap_str_startswith(fstype, "fuse.")) {
+		fstype += strlen("fuse.");
+	}
```
matejak, line 100: This is great. I would suggest to add a comment that points to the source of the inspiration, so it could be easily revised later.
Alternatively, we could consider calling the systemd function if it is possible, but that may be too risky, so maybe not.

src/OVAL/probes/fsdev.c
```diff
@@ -97,37 +94,36 @@ static int is_local_fs(struct mntent *ment)
...
+		"afs",
+		"ceph",
+		"cifs",
+		"smb3",
+		"smbfs",
+		"sshfs",
+		"ncpfs",
+		"ncp",
+		"nfs",
+		"nfs4",
```
matejak, line 111: We have an issue with `gpfs` filesystems, so I would add that one here. This link indicates that `gpfs` is the right type string: https://www.ibm.com/developerworks/systems/library/es-gpfs/index.html

src/OVAL/probes/fsdev.c
jan-cerny: @matejak I have added a comment.

src/OVAL/probes/fsdev.c
jan-cerny: @matejak I have added gpfs to the list.

matejak:
This is a nice, streamlined approach for the local/network logic. Thank you for the PR!

PR #1575, opened 2020-08-20
Plug a memory leak

evgenyz:
Alternative for #1571.

Do not re-create global name cache, but clear the contents.

Dicsovered when running valgrind on command:
oscap xccdf eval --results-arf /tmp/arf.xml --profile ospp --rule xccdf_org.ssgproject.content_rule_selinux_state /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml

Addressing:
==117734== 5,459 (640 direct, 4,819 indirect) bytes in 8 blocks are definitely lost in loss record 269 of 270
==117734==    at 0x483A809: malloc (vg_replace_malloc.c:307)
==117734==    by 0x4931C18: probe_ncache_new (ncache.c:81)
==117734==    by 0x49328AB: probe_common_main (probe_main.c:213)
==117734==    by 0x4CDA431: start_thread (in /usr/lib64/libpthread-2.31.so)
==117734==    by 0x52A8912: clone (in /usr/lib64/libc-2.31.so)

### DIFF
--- a/src/OVAL/probes/probe/ncache.c
+++ b/src/OVAL/probes/probe/ncache.c
@@ -111,6 +111,23 @@ void probe_ncache_free (probe_ncache_t *cache)
         return;
 }
 
+void probe_ncache_clear (probe_ncache_t *cache)
+{
+        size_t i;
+
+        if (cache == NULL)
+                return;
+
+        if (pthread_rwlock_wrlock(&(cache)->lock))
+                return;
+        for (i = 0; i < cache->real; ++i)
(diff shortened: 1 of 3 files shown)


evgenyz:
@jan-cerny Getting rid of the global cache is not trivial, but I have a different solution for the leak.

evgenyz:
@openscap-ci test this, please

jan-cerny:
@evgenyz That's great! It indeed fixes the leak and doesn't create segfaults elsewhere.

PR #1576, opened 2020-08-20
Enable compressed feeds, and ability to debug them for 1.2

matejak:
This is port of #1557, #1561 and #1564 to the maint-1.2 branch.

### DIFF
--- a/tests/curl/test_curl_encoding.sh
+++ b/tests/curl/test_curl_encoding.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -e -o pipefail
+
+. $builddir/tests/test_common.sh
+
+function curl_accept_encoding {
+	local DF="${srcdir}/ds.xml"
+	local RF="results.xml"
+	local LOG="verbose.log"
+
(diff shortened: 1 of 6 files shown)


matejak:
Some issues with CI are expected:

- The curl compression code uses features of curl that are newer than what RHEL6 supplies, in which case the test fails. I guess that I could disable the test for RHEL6, but I don't think that it is worth the effort.
- The librpm has been updated, and oscap 1.2 doesn't keep up with it. So if it hits e.g. Fedora>=32 in CI, it will fail.

tests/curl/test_curl_encoding.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -e -o pipefail
+
+. $builddir/tests/test_common.sh
```
jan-cerny, line 5: Try to use relative path instead of `$builddir/tests`

```suggestion
. ../test_common.sh
```

jan-cerny (approved):
Hi, I like your patch very much.

FTR The maint-1.2 branch currently doesn't compile on F32, but that isn't caused by this PR, it's caused by incompatibility of the rpm probes code with the latest librpm.

PR #1577, opened 2020-08-26
Free already serialized items

jan-cerny:
When serializing OVAL collected items, the serialized items weren’t freed  from memory. They existed both as a C structure and XML node in memory at the same time.

This PR tries to mitigate this problem by freeing items right after they are converted to XML nodes.

On a common scenario when evaluating a single OVAL definition which matches 10000 of file objects and OVAL results are generated, this change reduces the memory consumption at the highest peak from 180 MB to 159 MB (by 11.66 %).

However, there are problems.

I'm afraid that the change of behavior in `oval_syschar_model_to_dom` is breaking the API. The function used to only read the data from the map in OVAL syschar model (the `syschar_model->sysitem_map`). Now, it removes some items from this map. That leaves the map and the model unusable for further calls.

Also, `oval_syschar_model_to_dom` is called from many places so the change can affect many things. Here is the call graph:

- oval_syschar_model_to_dom
  - oval_syschar_model_export
    - app_collect_oval
  - oval_result_system_to_dom
    - oval_results_to_dom
      - oval_results_model_export_source
        - oval_session_export
          - app_collect_oval
        - oval_results_model_export
          - app_analyse_oval
        - _xccdf_session_export_oval_result_file
          - _build_oval_result_sources
            - xccdf_session_export_oval
              - app_evaluate_xccdf
              - app_xccdf_export_oval_variables
              - app_xccdf_remediate

Some of these symbols are a part of public API :scream: .

A solution could be not to modify an existing function but to introduce new functions. What are your opinions?

### DIFF
--- a/src/OVAL/adt/oval_string_map.c
+++ b/src/OVAL/adt/oval_string_map.c
@@ -213,6 +213,16 @@ void oval_string_map_put(struct oval_string_map *map, const char *key, void *val
         }
 }
 
+void oval_string_map_del(struct oval_string_map *map, const char *key)
+{
+	if (map == NULL || key == NULL) {
+		return;
+	}
+	if (rbt_str_del((rbt_t *)map, key, NULL) != 0) {
+		dD("rbt_str_del: non-zero return code");
+	}
+}
+
 void oval_string_map_put_string(struct oval_string_map *map, const char *key, const char *val)
 {
 	if (map == NULL || key == NULL) {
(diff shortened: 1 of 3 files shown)


evgenyz:
The better way to implement it is to introduce the `oval_syschar_model_to_dom_free` function and use it where appropriate. I tried to make such a change but it is sophisticated (I only covered one scenario, for test sake). And it is not worth it even with more data-rich probes like textfilecontent54. Let's put it into the table until we have a use-case where it would be a deal-breaker.

PR #1578, opened 2020-08-27
Remove superfluous free

jan-cerny:
When cpe_dict_detect_version_priv fails and returns NULL it
frees the reader and then the reader is freed again in the caller
function oscap_source_get_schema_version (oscap_source.c:405).

### DIFF
--- a/src/CPE/cpedict.c
+++ b/src/CPE/cpedict.c
@@ -180,7 +180,6 @@ char *cpe_dict_detect_version_priv(xmlTextReader *reader)
 	const char* elm_name = (const char *) xmlTextReaderConstLocalName(reader);
 	if (!elm_name || strcmp("cpe-list", elm_name)) {
 		oscap_seterr(OSCAP_EFAMILY_OSCAP, "Expected root element name 'cpe-list', found '%s'.", elm_name);
-		xmlFreeTextReader(reader);
 		return NULL;
 	}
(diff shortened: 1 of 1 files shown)


src/CPE/cpedict.c
```diff
@@ -180,7 +180,6 @@ char *cpe_dict_detect_version_priv(xmlTextReader *reader)
 	const char* elm_name = (const char *) xmlTextReaderConstLocalName(reader);
 	if (!elm_name || strcmp("cpe-list", elm_name)) {
 		oscap_seterr(OSCAP_EFAMILY_OSCAP, "Expected root element name 'cpe-list', found '%s'.", elm_name);
-		xmlFreeTextReader(reader);
```
evgenyz, line 183: What will happen here is case `elm_name == NULL`? And, wouldn't it leak in case `elm_name != 'cpe-list'`?

jan-cerny: @evgenyz 

When elm_name is NULL the function returns NULL.  If elm_name != 'cpe-list' it also returns NULL. But regardless the return value, the caller function hits this break: https://github.com/OpenSCAP/openscap/blob/maint-1.3/src/source/oscap_source.c#L386 and then it goes to https://github.com/OpenSCAP/openscap/blob/maint-1.3/src/source/oscap_source.c#L405 where the `xmlFreeTextReader(reader);` is called. So I think the reader is always freed and it shouldn't leak.

Hopefully I'm not missing anything.

PR #1581, opened 2020-08-31
code has changed,but the doc has not changed

pengqb:
code has changed in appveyor.yml
install:
- cmd: vcpkg install curl libxml2 libxslt bzip2 pcre pthreads zlib getopt-win32
cache: c:\tools\vcpkg\installed\
before_build:


evgenyz:
Nice catch. Thank you!

PR #1582, opened 2020-08-31
Avoid duplicating parts of SDS in memory

jan-cerny:
Replaces function ds_doc_from_foreign_node which called
xmlDOMWrapCloneNode by a similar new function which calls
xmlDOMWrapAdoptNode instead.

This should prevent copying parts of DOM trees when registering
individual source datastream components.

### DIFF
--- a/src/DS/sds.c
+++ b/src/DS/sds.c
@@ -240,9 +240,40 @@ static int ds_sds_register_sce(struct ds_sds_session *session, xmlNodePtr compon
 	return ret;
 }
 
-static int ds_sds_register_xmlDoc(struct ds_sds_session *session, xmlDoc* doc, xmlNodePtr component_inner_root, const char *relative_filepath)
+static xmlDoc *ds_sds_doc_from_foreign_node(xmlNode *node, xmlDoc *parent)
 {
-	xmlDoc *new_doc = ds_doc_from_foreign_node(component_inner_root, doc);
+	xmlDOMWrapCtxtPtr wrap_ctxt = xmlDOMWrapNewCtxt();
+	xmlDocPtr new_doc = xmlNewDoc(BAD_CAST "1.0");
+	if (xmlDOMWrapAdoptNode(wrap_ctxt, parent, node, new_doc, NULL, 0) != 0)
+	{
+		oscap_seterr(OSCAP_EFAMILY_XML, "Error when adopting node '%s' while "
+				"dumping component from DataStream", node->name);
+		xmlFreeDoc(new_doc);
+		xmlDOMWrapFreeCtxt(wrap_ctxt);
+		return NULL;
+	}
+	xmlDocSetRootElement(new_doc, node);
+	if (xmlDOMWrapReconcileNamespaces(wrap_ctxt, node, 0) != 0)
+	{
+		oscap_seterr(OSCAP_EFAMILY_XML,
+				"Internal libxml error when reconciling namespaces for node "
+				"'%s' while dumping component.", node->name);
+		xmlFreeDoc(new_doc);
+		xmlDOMWrapFreeCtxt(wrap_ctxt);
+		return NULL;
+	}
+	xmlDOMWrapFreeCtxt(wrap_ctxt);
+	return new_doc;
+}
+
+static int ds_sds_register_xmlDoc(struct ds_sds_session *session, xmlDoc* doc, xmlNodePtr component_inner_root, const char *relative_filepath, bool clone)
+{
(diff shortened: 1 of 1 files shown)


jan-cerny:
This patch is clearly wrong because it removes the nodes from the source datastream document tree but the source datastream needs to remain intact because later when generating ARF we need include the whole source datastream document into ARF `<report_request>` element.

The code that I modified serves to break the source datastream into respective components and creates an in-memory XML document for each component and then populates the XCCDF session which then works transparently as if the components were separate files.

I envision 2 possible options that we can explore:
1. When generating ARF reconstruct again the original source datastream document from all the components documents.
2. Rework XCCDF session logic to enable using pointers to nodes in original source datastream document instead of having to split the source datastream into components.

I guess both of them won't be quick fixes.

jan-cerny:
OpenSCAP is often used to evaluate large datastreams. For example, the RHSA CVE datastream for RHEL 7 is about 16 MB. The biggest datastream that we currently produce is ssg-rhel7-ds.xml which is currently 32 MB. When using such a large datastream, we start to see a big difference between memory consumption when ARF is generated and when ARF isn't generated.

For example, these values were returned by `/usr/bin/time -f %M` when scanning ssg-rhel7-ds.xml of 32 MB:

* evaluating OSPP:
  * without `--results-arf`: 395 MB
  * with `--results-arf`: 682 MB
* evaluating PCI-DSS: 
  * without `--results-arf`: 404 MB
  * with `--results-arf`: 700 MB

So in this example the memory consumption grows by 73 % when `--results-arf` is used.

Next step is to identify how much memory can be saved by avoiding duplication of SDS components in memory. We surely can't make the memory consumption of the `--results-arf` the same as without it.

jan-cerny:
> OpenSCAP is often used to evaluate large datastreams. For example, the RHSA CVE datastream for RHEL 7 is about 16 MB. The biggest datastream that we currently produce is ssg-rhel7-ds.xml which is currently 32 MB. When using such a large datastream, we start to see a big difference between memory consumption when ARF is generated and when ARF isn't generated.
> 
> For example, these values were returned by `/usr/bin/time -f %M` when scanning ssg-rhel7-ds.xml of 32 MB:
> 
>     * evaluating OSPP:
>       
>       * without `--results-arf`: 395 MB
>       * with `--results-arf`: 682 MB
> 
>     * evaluating PCI-DSS:
>       
>       * without `--results-arf`: 404 MB
>       * with `--results-arf`: 700 MB
> 
> 
> So in this example the memory consumption grows by 73 % when `--results-arf` is used.
> 
> Next step is to identify how much memory can be saved by avoiding duplication of SDS components in memory. We surely can't make the memory consumption of the `--results-arf` the same as without it.

@jan-cerny I think your comment is totally irrelevant to this PR. This PR wants to solve the duplication that happens during loading SDS in memory. That happens every time a datastream is evaluated, regardless if user provides `--results-arf` option or not. Instead, data showing how this PR changes  memory consumption should be provided.

jan-cerny:
Fair enough. This PR reduces:
* when scanning a single rule from a datastream: from 215 MB to 161 MB
* when scanning an empty profile from a 32 MB ssg-rhel7-ds.xml: from 385 MB to 299 MB
* when scanning a OSPP profile from a 32 MB ssg-rhel7-ds.xml: from 392 MB to 305 MB.

Saving 100 MB is probably the upper limit that we can achieve in a common scenario.

jan-cerny:
To clarify, the numbers in the previous comment were obtained using the patches that are proposed in this PR. But, as I already mentioned, these patches are breaking things. I expect that a correct implementation would give us a similar gain. Or more likely, the correct implementation would give a slightly lower gain, because the problem with current patches is that ARF is missing some parts of SDS that also take some amount of memory.

From the 2 proposed ways I would choose option 2. It seems that in option 1 it will be hard to reconstruct the exactly same datastream as the input datastream when generating ARF, but ARF should contain a copy of the input datastream inside.

The test coverage is good, so the risk on pursuing either of them is low. However, both outlined solutions aren't easy to implement. Even if we invest the time, we wouldn't get much, maximally 100 MB. 

We think that the biggest memory problem was the memory leak in rpmverifyfileprobe, fixed by https://github.com/OpenSCAP/openscap/commit/860a81c3024b588f523bb9dc9b36fb4e2763aac3. Rpmverifyfile probe is used often in practice, so it can be a common ground of memory problems.

We suggest postponing this item until we get some bug report that this is a severe problem. As of now, we still think that he total consumption would be under minimal requirements for RHEL or Fedora.

PR #1586, opened 2020-09-02
Split test API/XCCDF/applicability

jan-cerny:
This patch splits the large test "all.sh" into smaller chunks.
The split makes it easier to debug, identify problems or
run individual tests when needed.


evgenyz:
Thanks! That would be useful for sure.

PR #1594, opened 2020-09-07
Fix memory leak

### LINKED ISSUE #1588: Memory leak in file probe
#### Description of Problem:
I have discovered this memory leak when running valgrind on evaluation of XCCDF rule `no_files_unowned_by_user`. Running valgrind on this rule takes 6 hours, but fortunately, the leak is also reproducible with a simple OVAL definition.

#### OpenSCAP Version:
current upstream maint-1.3 HEAD as of c1fc1f364d8aa851f47e986966f3b5354482d802

F32

#### Steps to Reproduce:
extract this to get the oval: [file.oval.xml.zip](https://github.com/OpenSCAP/openscap/files/5163071/file.oval.xml.zip)

1. cd build
2. ./run valgrind --leak-check=full utils/oscap oval eval --results /tmp/results.xml ~/work/oval_examples/file.oval.xml

#### Actual Results:
memory leaks

```
==30389== 32 bytes in 1 blocks are definitely lost in loss record 101 of 235                                                                                                                  
==30389==    at 0x483CE14: memalign (vg_replace_malloc.c:906)                                                                                                                                 
==30389==    by 0x483CF14: posix_memalign (vg_replace_malloc.c:1070)                           
==30389==    by 0x49452E8: oscap_aligned_malloc (util.h:348)                                   
==30389==    by 0x4945361: SEXP_val_new (sexp-value.c:36)                                                                                                                                     
==30389==    by 0x4946245: SEXP_rawval_copy (sexp-value.c:557)                                                                                                                                
==30389==    by 0x4942F2D: SEXP_softref (sexp-manip.c:1687)                                    
==30389==    by 0x494190A: SEXP_listref_first (sexp-manip.c:1032)                                                                                                                             
==30389==    by 0x48E49BC: probe_item_attr_add (probe-api.c:141)                               
==30389==    by 0x48E6ACC: probe_ent_attr_add (probe-api.c:1032)                                                                                                                              
==30389==    by 0x48E4C65: probe_itement_setstatus (probe-api.c:218)                           
==30389==    by 0x494CCCF: file_cb (file_probe.c:395)                                                                                                                                         
==30389==    by 0x494D11A: file_probe_main (file_probe.c:510)  
```

#### Expected Results:
memory doesn't leak

I suspect that the call of SEXP_free here https://github.com/OpenSCAP/openscap/blob/maint-1.3/src/OVAL/probes/probe-api.c#L174 should free this memory but it doesn't because condition here https://github.com/OpenSCAP/openscap/blob/maint-1.3/src/OVAL/probes/SEAP/sexp-manip_r.c#L424 is evaluated as false

To clarify, the 32 bytes leak when the simplified scenario is executed. During evaluation of rule `no_file_unowned_by_user` this code was executed repeatedly and the leak was 11,200,096 bytes.

evgenyz:
We should free the replaced list element, returned from the `..._replace` function, not the soft reference to it.

Fixes #1588.

### DIFF
--- a/src/OVAL/probes/probe-api.c
+++ b/src/OVAL/probes/probe-api.c
@@ -138,6 +138,11 @@ SEXP_t *probe_item_attr_add(SEXP_t * item, const char *name, SEXP_t * val)
 {
 	SEXP_t *n_ref, *ns;
 
+	if (val == NULL)
+		ns = SEXP_string_new(name, strlen(name));
+	else
+		ns = SEXP_string_newf(":%s", name);
+
 	n_ref = SEXP_listref_first(item);
 
 	if (SEXP_listp(n_ref)) {
@@ -145,13 +150,7 @@ SEXP_t *probe_item_attr_add(SEXP_t * item, const char *name, SEXP_t * val)
 		 * There are already some attributes.
 		 * Just add the new to the list.
 		 */
-		if (val == NULL)
-			ns = SEXP_string_new(name, strlen(name));
-		else
-			ns = SEXP_string_newf(":%s", name);
-
 		SEXP_list_add(n_ref, ns);
-		SEXP_free(ns);
 
 		if (val != NULL)
 			SEXP_list_add(n_ref, val);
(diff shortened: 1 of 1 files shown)


evgenyz:
Uh, okay. It seems it is not that simple.

jan-cerny:
Yes, I was also destroyed by this.

evgenyz:
So, I fixed memory leak. But the problem with weak ('soft') references is still there somewhere. It would probably require matches and some gasoline to fix it...

jan-cerny:
Great!

PR #1596, opened 2020-09-08
Add g_hash_table_insert to suppressions file

jan-cerny:


evgenyz:
I wonder why GLib is leaking this memory. Is it a native bug or maybe we are using some g_ function (or g_-based) in a wrong way? I suppose it is somewhere in the sytemd-related probe?

jan-cerny:
I'm puzzled by this, because I can't find it anymore in the logs, see for example https://jenkins.complianceascode.io/job/openscap-maint-1.3-valgrind-test/30/artifact/build/tests/valgrind_test.log. So I'm closing it because I can't justify it anymore.

PR #1604, opened 2020-09-21
Test oscap under memory limit

jan-cerny:
This test runs evaluation of a profile under memory constrainsts. The
profile consists of 10 memory intensive rules. The memory limit is set
to 700 MB using cgroups. This test isn't a part of upstream test suite,
because it takes a longer time to complete.  Instead, it will be run by
a new nightly Jenkins job.


matusmarhefka:
@jan-cerny But for this test to work we will need a RHEL system with package group "Server with GUI" or similar as the excessive memory usage is triggered by the large number of files on the filesystem. Will the new Jenkins job handle this?

Also don't forget that for this test to work we will need additional packages `libcgroup` and `libcgroup-tools`.

jan-cerny:
@matusmarhefka Good point. I don't know how to handle this because disk space on Jenkins machines is limited. As of now, the RHEL 7 Wedos machine contains 842 RPM packages and RHEL 8 Wedos machine contains 777 RPM packages. Maybe lowering the limit in the test would help?

jan-cerny:
I have tried to run the proposed test manually in one of the Jenkins workers, specifically on the RHEL 8 worker that runs on Wedos. The test passed.

Then I built OpenSCAP from 1.3.3 tag. That's because 1.3.3 didn't contain any of the memory improvements. I tried to run the `oscap xccdf eval` command which is used in the proposed test and measure the memory consumption using `/usr/bin/time -f %M`. The command has been killed by the system because it consumed roughly 1.7 GB memory and this machine has 2 GB memory.

I also tried to run the test with the 1.3.3 build, however, the test got stuck.

We think that the test can at least guard if the memory consumption doesn't exceed 700 MB which seems a reasonable limit for 10 rules. We can see that it would catch huge memory leaks that we have seen.

matusmarhefka:
Thanks @jan-cerny for the summary, LGTM.

PR #1605, opened 2020-09-22
Split large tests to separate test scenarios

jan-cerny:
Some of the tests use a wrapper shell script, usually called `all.sh`, which calls the test scripts. With CTest, it's possible to call the test scripts directly. The wrapper isn't needed and it actually complicates debugging. Also, it's easier to find the point of failure if we run multiple smaller tests instead of one big testcase.


matusmarhefka:
@jan-cerny Nice improvements, logs are now much more convenient, mainly in cases when some tests would be failing. I am merging.

PR #1608, opened 2020-09-25
Remove removing Python 2

jan-cerny:
It seems that we don't need to remove Python2 anymore.

Fixes the GitHub actions Error message on Mac OS X.
Error: No installed keg or cask with the name "python@2"
Error: Process completed with exit code 1.


matejak:
It seems to serve the purpose, merging.

PR #1609, opened 2020-09-28
Fix export of Profile platform to DOM

yuumasato:
Fix how Profile platforms are converted to DOM.

Similar to [src/XCCDF/rule.c#L967](https://github.com/yuumasato/openscap/blob/46b78146db6ba1fa57926068c4400d876423126b/src/XCCDF/rule.c#L967)


JAORMX (approved):

jan-cerny:
Nice!

PR #1610, opened 2020-09-30
Prevent segfault

jan-cerny:
Prevent segfault if there are circular dependencies between OVAL
definitions.  This segfault happened when an OVAL definition references
itself in the extend_definition element. It can also happen when the
referenced OVAL definition references back to the parent OVAL definition
and therefore  there is a circular dependency between OVAL definitions.
See the test OVALs for examples of such definitions. The circle wasn't
detected and it caused infinite recursion which caused a segfault.

The fix uses a list which stores the IDs of already evaluated
definitions and it checks if a definition isn't evaluated for the second
time.

Fixes: rhbz#1812476


evgenyz:
@jan-cerny It fails to pass its own test on Ubuntu, can you please check it out?

evgenyz:
Okay, we have to add `libdbus-1-dev` to dependencies in our Ubuntu CI. But it still fails.

evgenyz:
Also, can we maybe test it with a less exotic probe/test?

jan-cerny:
OK, I'll look into this. I think any probe can be used, it should be reproducible regardless of the used probe because the problem was on the OVAL level. The systemd probe was used in the original reproducer in the BZ so it was created from that.

jan-cerny:
I tried to change the test to make it more general and to use the OVAL file test insted of OVAL systemdunit test in the test files. Let's take a look on the new CI run now.

src/OVAL/results/oval_resultSystem.c
```diff
@@ -205,14 +214,20 @@ struct oval_result_definition *oval_result_system_get_new_definition
 		char *id = oval_definition_get_id(oval_definition);
 		rslt_definition = oval_result_system_get_definition(sys, id);
 		if (rslt_definition == NULL) {
-			rslt_definition = make_result_definition_from_oval_definition(sys, oval_definition,
+			rslt_definition = make_result_definition_from_oval_definition(sys, oval_definition, visited_definitions,
 				variable_instance ? variable_instance : 1);
+			if (rslt_definition == NULL) {
+				return NULL;
+			}
```
evgenyz, line 221: I would say that these braces are redundant.

src/OVAL/results/oval_resultSystem.c
```diff
@@ -205,14 +214,20 @@ struct oval_result_definition *oval_result_system_get_new_definition
...
 			oval_result_system_add_definition(sys, rslt_definition);
 		}
 		else if (oval_result_definition_get_variable_instance_hint(rslt_definition) != oval_result_definition_get_instance(rslt_definition)) {
 			int hint = oval_result_definition_get_variable_instance_hint(rslt_definition);
 			dI("Creating another result-definition for id=%s based on variable_instance: %d", id, hint);
-			rslt_definition = make_result_definition_from_oval_definition(sys, oval_definition, hint);
+			rslt_definition = make_result_definition_from_oval_definition(sys, oval_definition, visited_definitions, hint);
+			if (rslt_definition == NULL) {
+				return NULL;
+			}
```
evgenyz, line 230: I would say that these braces are redundant.

src/OVAL/results/oval_resultSystem.c
```diff
@@ -376,17 +391,26 @@ struct oval_result_definition *oval_result_system_prepare_definition(struct oval
...
 	}
 
+	struct oscap_list *visited_definitions = oscap_list_new();
         rslt_definition = oval_result_system_get_definition(sys, id);
         if (rslt_definition == NULL) {
-		rslt_definition = make_result_definition_from_oval_definition(sys, oval_definition, 1);
+		rslt_definition = make_result_definition_from_oval_definition(sys, oval_definition, visited_definitions, 1);
+		if (rslt_definition == NULL) {
+			goto cleanup;
+		}
```
evgenyz, line 400: I would say that these braces are redundant.

src/OVAL/results/oval_resultSystem.c
```diff
@@ -376,17 +391,26 @@ struct oval_result_definition *oval_result_system_prepare_definition(struct oval
...
                 oval_result_system_add_definition(sys, rslt_definition);
         }
 	else if (oval_result_definition_get_variable_instance_hint(rslt_definition) != oval_result_definition_get_instance(rslt_definition)) {
 		int hint = oval_result_definition_get_variable_instance_hint(rslt_definition);
 		dI("Creating another result-definition for id=%s based on variable_instance: %d", id, hint);
-		rslt_definition = make_result_definition_from_oval_definition(sys, oval_definition, hint);
+		rslt_definition = make_result_definition_from_oval_definition(sys, oval_definition, visited_definitions, hint);
+		if (rslt_definition == NULL) {
+			goto cleanup;
+		}
```
evgenyz, line 409: I would say that these braces are redundant.

evgenyz:
Otherwise LGTM.

src/OVAL/results/oval_resultSystem.c
jan-cerny: Yes, you're right, the Linux kernel coding style discourages that. I usually do this to prevent forgetting them if the single statement would be replaced by multiple statements in future. I will remove them quickly.

jan-cerny:
I have removed the superfluous braces.

jan-cerny:
@openscap-ci test this please

Issue #1560, opened 2020-07-13
Update Bootstrap


evgenyz:
As a follow-up of #1550, the remaining issue with HTML report:

`CSS: Deprecated media feature max-device-width.`

jan-cerny:
This isn't just about the update, but then you have to make sure that everything still works. You have to check that the layout isn't broken, the generated documents are accessible for visually impaired, that it prints on paper and into PDF. Notice that it affects both HTML guides and HTML reports. We have done some modifications to the official bootstrap to achieve this. They aren't possible to apply to the latest versions as they are, you have to do it again.

PR #1612, opened 2020-10-02
Update xccdf-resources-build.sh

Honny1:
This PR changes tools for build and minifies CSS and JS. Old tools didn't work or are not supported on last versions of Fedora.


jan-cerny:
@Honny1 Welcome!

jan-cerny:
@openscap-ci add to whitelist

evgenyz:
Hey! That is a great start. But there are more problems with JS and CSS resources generation. The most important one is that they are not generated/complied/minified during the build, but reside in the repo in an almost binary form. Would you like to get on this as well?

evgenyz:
Also there is (#1560) a nice-to-have upgrade of the Bootstrap.

jan-cerny:
>  Hey! That is a great start. But there are more problems with JS and CSS resources generation. The most important one is that they are not generated/complied/minified during the build, but reside in the repo in an almost binary form. Would you like to get on this as well?

Great idea, but please in separate PRs.

xsl/xccdf-resources-build.sh
```diff
@@ -1,6 +1,8 @@
-# A horrible little script that builds and minifies CSS and JS we bundle with HTML report and guide
+#!/bin/sh
+# This script builds and minifies CSS and JS that we bundle to HTML report and guide.
+set -e 
 
-for cmd in csstidy slimit; do
+for cmd in yuicompressor; do
```
jan-cerny, line 5: You have only 1 item now, so the `for` cycle isn't necessary.

jan-cerny:
I have tried to generate guide and report on Fedora OSPP profile and the generated documents look fine.

evgenyz:
The one last thing maybe is to mention it in the docs as a build dep.

jan-cerny:
In my opinion it isn't a build dependency because the rebuild isn't a part of build process. The csstidy and slimit also aren't mentioned. I wouldn't require to add it at this point.

PR #1615, opened 2020-10-05
Suppress most of the memory errors

jan-cerny:
When we make the Jenkins job passing it will be easier to detect
freshly introduced memory problems. Until all memory problems
are fixed we risk that new problems will be ignored.


matejak:
I can confirm that this suppression list makes the test go green, which makes it a good start with memory management improvements, and some sort of memory quality gating.
Good work!

PR #1623, opened 2020-11-09
Remove SEXP_lstack

jan-cerny:
The SEXP_lstack module is not used anywhere except in struct
SEXP_ostate which is used but only to be set to NULL.
However, we can't remove struct SEXP_ostate completely because
its type definition is a part of public API.


evgenyz:
Okay, it seems to be safe to do so.

PR #1626, opened 2020-11-09
Added the missing header for chroot() on macOS

### LINKED ISSUE #1624: PR testsuite fails for macOS
Whenever a pull request is submitted a testsuite is executed.  However, due to an issue in https://github.com/OpenSCAP/openscap/blame/maint-1.3/src/OVAL/probes/probe/worker.c#L1013 (an implicit call to `chroot()`) the macOS testsuite fails.

The `worker.c` needs to include `#include <unistd.h>` for macOS to pass the testsuite.

galaxy4public:
Fixes #1624

### DIFF
--- a/src/OVAL/probes/probe/worker.c
+++ b/src/OVAL/probes/probe/worker.c
@@ -39,7 +39,7 @@
 #if defined(OS_OSX)
 /* XXX: please see https://github.com/OpenSCAP/openscap/pull/1626 for the
  *      reason of explicitly defining an external here instead of simply
- *      including the unistd.h headr file.
+ *      including the unistd.h header file.
  */
 extern int chroot(const char *);
 #endif
--- a/src/OVAL/probes/probe/worker.c
+++ b/src/OVAL/probes/probe/worker.c
@@ -31,10 +31,13 @@
 #include <string.h>
 #include <pthread.h>
 #include <errno.h>
-#include <unistd.h>
 
 #if defined(OS_FREEBSD)
 #include <pthread_np.h>
+#endif
+
+#if defined(OS_OSX)
+external int chroot(const char *);
 #endif
 
 #include "probe-api.h"
--- a/src/OVAL/probes/probe/worker.c
+++ b/src/OVAL/probes/probe/worker.c
@@ -37,7 +37,7 @@
 #endif
 
 #if defined(OS_OSX)
-external int chroot(const char *);
+extern int chroot(const char *);
 #endif
 
 #include "probe-api.h"
--- a/src/OVAL/probes/probe/worker.c
+++ b/src/OVAL/probes/probe/worker.c
@@ -37,6 +37,10 @@
 #endif
 
 #if defined(OS_OSX)
+/* XXX: please see https://github.com/OpenSCAP/openscap/pull/1626 for the
+ *      reason of explicitly defining an external here instead of simply
+ *      including the unistd.h headr file.
+ */
 extern int chroot(const char *);
 #endif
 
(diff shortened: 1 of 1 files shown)


src/OVAL/probes/probe/worker.c
```diff
@@ -36,6 +36,10 @@
 #include <pthread_np.h>
 #endif
 
+#if defined(OS_OSX)
```
ggbecker, line 39: I wonder if this really works. And I don't think it should be a problem to include this file in all platforms.

ggbecker: It probably should. Since: https://github.com/OpenSCAP/openscap/blob/maint-1.3/compat/oscap_platforms.h#L52 defines

anyway... it's still failing. You can try if without the `#if` has the same effect

evgenyz:
It is defined in the same `unistd.h` in Linux. I wonder what is the real problem here, why it is working in Linux right now (and for years was working for MacOS right until couple weeks ago). Maybe there is something else that is not properly included alongside with `unistd.h` (some definitions etc)?

galaxy4public:
I am updating my Xcode, so I could test it locally (quicker).  Once it is done, I will update the PR.

galaxy4public:
Looking at https://opensource.apple.com/source/Libc/Libc-320/include/unistd.h.auto.html it seems that `chroot()` is only defined if `_POSIX_SOURCE` is undefined.

P.S. Still waiting for 12GB Xcode to download :)

galaxy4public:
OK, I am at loss.  Perhaps, we need to enable better output from CI tests, so it is possible to understand how a particular object is being compiled?  This is from my local test:
```
galaxy@MacBook-Pro ~ % llvm-gcc -c test-chroot1.c
test-chroot1.c:2:2: error: implicit declaration of function 'chroot' is invalid
      in C99 [-Werror,-Wimplicit-function-declaration]
        chroot("/");
        ^
1 error generated.
galaxy@MacBook-Pro ~ % cat test-chroot2.c        
#include <unistd.h>

int main() {
	chroot("/");
	return 0;
}
galaxy@MacBook-Pro ~ % llvm-gcc -c test-chroot2.c
galaxy@MacBook-Pro ~ %
```
So, I don't know what CI does not like in my patch.  To answer the question why it was working for months before, most likely your Xcode got updated and got stricter default flags.

galaxy4public:
It's 3:30am here, so I am going to have some sleep and will think about this since I am out of ideas right now.

galaxy4public:
Also, please note that unfenced `#include <unistd.h>` breaks the Windows CI test, so I guess we need to wrap it with `#ifdef`'s for macOS.

galaxy4public:
It seems the `extern int chroot(char *)` is a success :)

galaxy4public:
@ggbecker , I understand that it is quite hacky, but with such a limited visibility into what CI is doing + the scope explicitly limited to `OS_OSX`, I think it may be acceptable to "heal" the PR checks.  Should I put a comment just before `extern` pointing back to this PR?

ggbecker:
> @ggbecker , I understand that it is quite hacky, but with such a limited visibility into what CI is doing + the scope explicitly limited to `OS_OSX`, I think it may be acceptable to "heal" the PR checks. Should I put a comment just before `extern` pointing back to this PR?

Yeah, the way it is now should be fine. Thanks

ggbecker (approved):

evgenyz:
Okay, I'm going to merge it to restore CI functionality, but I'm going to keep the bug as we ought to dig to the root of the problem.

PR #1627, opened 2020-11-10
Fix memory allocation

jan-cerny:
We can't assume that size of a structure is a sum of sizes of its
members because padding and alignment can be involved. In fact,
we need to allocate more bytes for the structure than the
sum of sizes of its members.

The wrong assumption caused invalid writes and invalid reads
which can be discovered by valgrind. Moreover, when run with
MALLOC_CHECK_ environment variable set to non-zero value, the
program aborted.

The memory issue happened only when NDEBUG is defined, eg. when cmake
-DCMAKE_BUILD_TYPE=RelWithDebInfo or Release, it doesn't happen if cmake
-DCMAKE_BUILD_TYPE=Debug which we usually use in Jenkins CI. This is
most likely because in debug mode the struct SEXP contains 2 additional
members which are the magic canaries and therefore is bigger.

This commit wants to fix the problem by 2 step allocation in which
first the size of the struct SEXP_val_lblk is used and then the
array of SEXPs is allocated separately.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1891770

### DIFF
--- a/src/OVAL/probes/SEAP/_sexp-value.h
+++ b/src/OVAL/probes/SEAP/_sexp-value.h
@@ -94,7 +94,7 @@ struct SEXP_val_lblk {
         uintptr_t nxsz;
         uint16_t  real;
         uint16_t  refs;
-        SEXP_t    memb[];
+	SEXP_t *memb;
 };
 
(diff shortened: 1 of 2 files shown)


evgenyz:
LGTM. Thanks

PR #1653, opened 2020-11-24
Extend test_fsdev_is_local_fs to verify GPFS

matusmarhefka:
Test for commit 1541487 - add GPFS into fake_mtab (which is
utilized by the `test_fsdev_is_local_fs`) to make sure that it
is not detected as a local filesystem.

For more details also see
https://www.ibm.com/developerworks/systems/library/es-gpfs/index.html


evgenyz:
Nice! Thanks!

PR #1654, opened 2020-11-26
Remove outdated files

jan-cerny:
Removes:
- old SCAP content for RHEL 6 because RHEL 6 is EOL and we had
  better content in SSG
- SCAP content for Fedora 14 because Fedora 14 is historic
- Fedora 14 init files - dtto
- sectool content which can be replaced by using SSG


evgenyz:
LGTM! Thanks!

Issue #1629, opened 2020-11-12
Incorrect form of @href attribute of <xccdf:benchmark> element


jan-cerny:
#### Description of Problem:
SCAP 1.3 specification section 4.5 says:

> 
> To be considered valid SCAP result content, the <xccdf:TestResult> element SHALL meet the following conditions:
> 
 ...
> 
> When the <xccdf:TestResult> is the root XCCDF element, then it will include an
> <xccdf:benchmark> element
> 
 ...
> 
> The <xccdf:benchmark> element SHALL have an @href attribute specified. The
> @href attribute SHALL hold the URI referencing the XCCDF component (either local
> to the data stream collection or remote) that was processed. The URI SHALL be in the
> form specified for the @href attribute in Table 8.
> 

#### Steps to Reproduce:

1. scan with arf results
2.  java -jar /opt/scapval/scapval-1.3.5.jar -scapversion 1.3  -valreportfile scapval.html  -resultfile results_arf.xml
3. 

#### Actual Results:
The href attribute looks like
``` <benchmark href="r400.1.1-datastream.xml" id="xccdf_6gA.9tIT5j86j.XFD97_benchmark_xccdf-for-family-test-validation-content"/>```

This causes SCAPVAL error RES-253-2.

#### Expected Results:
the href attribute conforms with the specification

PR #1655, opened 2020-12-02
Fix TestResult/benchmark/@href attribute

jan-cerny:
Make the href attribute an URI according to SCAP specification Section
4.5. This should fix SCAPVAL error RES-253-2 which occured when
validating ARFs by SCAPVAL. See #1629 for the problem description.

Also, it fixes a problem that xccdf_benchmark_item_clone didn't create
the clusters_dict hash table.

The existing tests are extended to test this property, the
function test_sds_external_xccdf has been split into 2 functions.

Fixes: #1629


evgenyz:
LGTM.

PR #1658, opened 2020-12-03
Hide even more problems

jan-cerny:
Extends suppression file to make the Jenkins job
`openscap-maint-1.3-valgrind-test` green. It's a follow up on
https://github.com/OpenSCAP/openscap/pull/1615

### DIFF
--- a/tests/suppressions.supp
+++ b/tests/suppressions.supp
@@ -64,6 +64,14 @@
    fun:xmlfilecontent_probe_main
    ...
 }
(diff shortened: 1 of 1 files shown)


evgenyz:
This test should be used for gating, but it is very time-consuming. We still need a contingency plan about working in the reverse direction of eliminating problems. I wasn't able to come up with anything sensible, so I'll just merge suppression for now.

Issue #1630, opened 2020-11-12
Atrributes role and severity are missing in xccdf:rule-result element


jan-cerny:
#### Description of Problem:
SCAP 1.3 specification, section 4.5, point 10, says:

> The <xccdf:rule-result> elements report the result of the application of each selected
rule [XCCDF:6.6.2]. The @role, @severity, and @weight attributes of the
<xccdf:rule-result> element SHALL be provided to indicate their values used during
assessment.

#### OpenSCAP Version:
1.3.4

#### Steps to Reproduce:

1. scan with arf results
2.  java -jar /opt/scapval/scapval-1.3.5.jar -scapversion 1.3  -valreportfile scapval.html  -resultfile results_arf.xml

#### Actual Results:
`<rule-result idref="xccdf_gov.nist.validation_rule_r400_1_3" time="2020-11-12T09:37:49+01:00" weight="1.000000">`

The role and severity attributes are missing.

It causes SCAPVAL error RES-392

#### Expected Results:

The `@role`, `@severity`, and `@weight` attributes of the `<xccdf:rule-result>` element SHALL be provided.

PR #1659, opened 2020-12-04
Add severity and role attributes

jan-cerny:
Adds severity and role attributes to rule-result elements in XCCDF
results. These attributes are optional in XCCDF 1.2 specification but if
the XCCDF is used in ARF the attributes are mandatory.  See SCAP 1.3
specification Section 4.5 item 10. We didn't create these attributes if
the input XCCDF didn't contain them. We can fix it by assuming the
default values. According to XCCDF 1.2 specification, the default value
of role element is "full" and the default value of severity element is
"unknown". The presence of role, weight and severity attributes is
checked by SCAPVAL. We had weight attribute but missing role and
severity caused error RES-392.

Fixes: #1630

### DIFF
--- a/src/XCCDF/rule.c
+++ b/src/XCCDF/rule.c
@@ -273,8 +273,6 @@ void xccdf_group_free(struct xccdf_item *group)
 struct xccdf_item *xccdf_rule_new_internal(struct xccdf_item *parent)
 {
 	struct xccdf_item *rule = xccdf_item_new(XCCDF_RULE, parent);
-	rule->sub.rule.role = 0;
-	rule->sub.rule.severity = 0;
 	rule->sub.rule.role = XCCDF_ROLE_FULL;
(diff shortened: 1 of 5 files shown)


src/XCCDF/rule.c
```diff
@@ -275,6 +275,8 @@ struct xccdf_item *xccdf_rule_new_internal(struct xccdf_item *parent)
 	struct xccdf_item *rule = xccdf_item_new(XCCDF_RULE, parent);
 	rule->sub.rule.role = 0;
 	rule->sub.rule.severity = 0;
```
evgenyz, line 277: Shouldn't these 2 lines be removed?

jan-cerny: Yes

evgenyz:
LGTM. Thanks!

PR #1660, opened 2020-12-07
Use "xcode12.2" MacOS X image for Travis CI

evgenyz:


evgenyz:
@jan-cerny So, this fix seems to work for Travis, but we have strange failures in other CI elements.

jan-cerny:
@evgenyz You're amazing!

This PR is surely not the reason of these failures. I'm looking into them!

PR #1663, opened 2020-12-11
Improve valgrind tests output

jan-cerny:
The valgrind output was redirected to a separate file. This change redirects valgrind output to the test output when valgrind fails. It will make the logs easier to read.


matejak:
It works pretty good.
But why should one introduce a wrapper that contains valgrind-specific bits (i.e. checking for the error code), and there is a separate Valgrind wrapper already.
Was there a benefit of introducing the wrapper code that is shared with every test execution?

jan-cerny:
We often use output redirects in our tests to test oscap stdout and stderr. For example:

```
$OSCAP xccdf eval .... > $stdout 2> $stderr
grep "something" $stdout
[ -s $stderr ]
```

If we `cat` the valgrind output log file in the `tests/valgrind_test.sh` the log would contamine the stdout and therefore break tests like this.

I think that's the reason why we need to print the output one level above.

jan-cerny:
I have visited this problem and I noticed one problem:

If the valgrind finds a memory issue the tests/valgrind_test.sh should return a non-zero error code 66. The test then should fail right away because we always set `-e -o pipefail` in our test cases. That means that the stdout and stderr won't be examined and it shouldn't matter that they would be polluted.

I think I should examine this situation more.

jan-cerny:
> That means that the stdout and stderr won't be examined and it shouldn't matter that the stdout is contamined.

That's correct, but still, the valgrind output would be eaten by the redirects and wouldn't be displayed in the test output. As we want to display the valgrind output in the test output we have to read the log after the test failure. Sorry for the noise.

tests/valgrind_test.sh
```diff
@@ -7,29 +7,14 @@
 #      Jan Černý <jcerny@redhat.com>
 
 # set up variables
-valgrind_output=$(mktemp)
 logfile=$builddir/tests/valgrind_test.log
 error_code=66
```
matejak, line 11: The PR moves those definitions to the wrapper, so I would remove those lines as well.

jan-cerny:
Hi, thanks to the feedback @matejak provided I have discovered that this often doesn't work.

First problem is that many test still use the old "test_init", "test_run" and "test_exit" functions which don't account for our special valgrind return code 66. This problem probably can be solved quickly by updating these functions.

Second problem is more difficult to fix. Some tests don't set -e option which means they don't fail immediately when any command fails but they fail only on explicit command. A memory issue can cause the test to fail, but often the valgrind output wouldn't be displayed. We would have to review every test and change some of them.

At this moment it's better to keep the current state and have everything in a separate file rather than to see only part of the output.

Issue #1661, opened 2020-12-08
Centos8.3 Not detected


ichasco:
#### Description of Problem:

Hi, with the update to Centos8.3, the scan doesn't detect the System. With Centos8.2 works

#### OpenSCAP Version:

```
openscap-1.3.4-1.el8.x86_64
openscap-utils-1.3.4-1.el8.x86_64
openscap-scanner-1.3.4-1.el8.x86_64
scap-security-guide-0.1.53-1.el8.noarch
scap-workbench-1.2.0-5.el8.x86_64
openscap-containers-1.3.4-1.el8.noarch
```

#### Operating System & Version:

```
CentOS Linux release 8.3.2011
```

#### Steps to Reproduce:

1. Install Openscap
2. Run: 
```
oscap xccdf eval \
          --profile xccdf_org.ssgproject.content_profile_standard \
          --results-arf /var/tmp/oscap-arf.xml \
          --report /var/tmp/oscap-report.html \
          --fetch-remote-resources \
          /usr/share/xml/scap/ssg/content/ssg-centos8-ds.xml
``` 

#### Actual Results:

get notapplicable and ignore all rules

#### Expected Results:

apply the CPE

Thanks!! :)

evgenyz:
Most likely it belongs to CaC, but I'll leave it here until we will be sure.

jan-cerny:
Actually, it's a problem in both OpenSCAP and SSG. This particular use-case has to be fixed in SSG but the same problem is also in OpenSCAP embed CPE dict. 

I have found that the applicability checks depends on centos-release package. On CentOS 8.3 centos-release has been replaced by centos-linux-release. The change seems to be related to the CentOS Stream initiative. The CentOS Stream features centos-stream-release instead.

evgenyz:
Another argument for switching to `/etc/os-release` and `filecontent58`.

jan-cerny:
@evgenyz Why? We always prefer to use rpm if possible

evgenyz:
For starters, `/etc/os-release` is more or less standard and `centos-release` is definitely not. Then, what would you use for Arch Linux? Also, do you remember RHCOS problems with librpm?

ichasco:
Is there any workaround to fix it for the moment?

Thanks :)

ggbecker:
> Is there any workaround to fix it for the moment?
> 
> Thanks :)

[...]

PR #1664, opened 2020-12-11
Fix CentOS 8 CPE

jan-cerny:
With the CentOS stream effort, the centos-relase RPM package is no
longer available. While CentOS 8.2.2004 features centos-release RPM
package, the CentOS 8.3.2011 features centos-linux-release RPM and
CentOS Stream features centos-stream-release RPM package. We have
decided to parse /etc/os-release file instead which is often present and
doesn't require librpm to be present.

Fixes: #1661


evgenyz:
LGTM. Thanks for the fix.

PR #1665, opened 2020-12-11
Allow oscap-podman to run rootless

robbmanes:
Running oscap-podman as rootless only requires that we run
"podman unshare" at key points to change the namespace.  Check if we're
rootless, and if so, prepend with unshare.

This appears to work with both rootless and root modes without issues:
```
$ wget -O - https://www.redhat.com/security/data/oval/v2/RHEL8/rhel-8.oval.xml.bz2 | bzip2 --decompress > rhel-8.oval.xml

$ podman images
REPOSITORY                                  TAG     IMAGE ID      CREATED       SIZE
registry.redhat.io/ubi8                     latest  33df2983b061  5 weeks ago   208 MB

$ /usr/bin/oscap-podman-rootless 33df2983b061 oval eval --report vulnerability.html rhel-8.oval.xml
Definition oval:com.redhat.rhsa:def:20205237: false
Definition oval:com.redhat.rhsa:def:20205236: false
Definition oval:com.redhat.rhsa:def:20205146: false
Definition oval:com.redhat.rhsa:def:20205100: false
Definition oval:com.redhat.rhsa:def:20205085: false
Definition oval:com.redhat.rhsa:def:20204952: false
- - - - 8< - - - -
Definition oval:com.redhat.rhba:def:20194268: false
Definition oval:com.redhat.rhba:def:20193674: false
Definition oval:com.redhat.rhba:def:20193384: false
Evaluation done.

$ head vulnerability.html; echo "- - - - 8< - - - -"; tail vulnerability.html
<?xml version="1.0"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">

<!--
				Color mapping
				Result enumeration
		Class       |t|f|u|e|na|ne|
		compliance  |G|R|B|Y|- |- |
		inventory   |-|-|B|Y|- |- |
		misc        |-|-|B|Y|- |- |
- - - - 8< - - - -
	  <tr class="resultgoodB">
		<td class="Text" align="center">oval:com.redhat.rhba:def:20193384</td>
		<td class="Text" align="center">false</td>
		<td class="Text" align="center">patch</td>
		<td class="Text" align="center">[<a class="Hover" target="_blank" href="https://access.redhat.com/errata/RHBA-2019:3384">RHBA-2019:3384</a>], [<a class="Hover" target="_blank" href="https://access.redhat.com/security/cve/CVE-2019-8320">CVE-2019-8320</a>], [<a class="Hover" target="_blank" href="https://access.redhat.com/security/cve/CVE-2019-8321">CVE-2019-8321</a>], [<a class="Hover" target="_blank" href="https://access.redhat.com/security/cve/CVE-2019-8322">CVE-2019-8322</a>], [<a class="Hover" target="_blank" href="https://access.redhat.com/security/cve/CVE-2019-8323">CVE-2019-8323</a>], [<a class="Hover" target="_blank" href="https://access.redhat.com/security/cve/CVE-2019-8325">CVE-2019-8325</a>]</td>
		<td class="Text">RHBA-2019:3384: ruby:2.5 bug fix and enhancement update (Moderate)</td>
	  </tr>
	</table>
  </body>
</html>
```

### DIFF
--- a/utils/oscap-podman
+++ b/utils/oscap-podman
@@ -34,10 +34,10 @@ function usage()
     echo "oscap-podman -- Tool for SCAP evaluation of Podman images and containers."
     echo
     echo "Compliance scan of Podman image:"
-    echo "$ sudo oscap-podman [--oscap=<OSCAP_BINARY>] IMAGE_NAME OSCAP_ARGUMENT [OSCAP_ARGUMENT...]"
+    echo "$ oscap-podman [--oscap=<OSCAP_BINARY>] IMAGE_NAME OSCAP_ARGUMENT [OSCAP_ARGUMENT...]"
     echo
     echo "Compliance scan of Podman container:"
-    echo "$ sudo oscap-podman [--oscap=<OSCAP_BINARY>] CONTAINER_NAME OSCAP_ARGUMENT [OSCAP_ARGUMENT...]"
+    echo "$ oscap-podman [--oscap=<OSCAP_BINARY>] CONTAINER_NAME OSCAP_ARGUMENT [OSCAP_ARGUMENT...]"
     echo
     echo "See \`man oscap\` to learn more about semantics of OSCAP_ARGUMENT options."
 }
@@ -58,8 +58,9 @@ else
     invalid "Invalid arguments provided."
 fi
 
+# rootless requires unshare prior to running certain commands
 if [ "$(id -u)" -ne 0 ]; then
-    die "This script cannot run in rootless mode."
+	ROOTLESS_CMD="podman unshare --"
 fi
 if grep -q "\-\-remediate" <<< "$@"; then
     die "This script does not support '--remediate' option."
@@ -92,29 +93,26 @@ fi
 # podman init creates required files such as: /run/.containerenv - we don't care about output and exit code
 podman init $ID &> /dev/null || true
 
-DIR=$(podman mount $ID) || die "Failed to mount."
+DIR=$($ROOTLESS_CMD podman mount $ID) || die "Failed to mount."
 
 if [ ! -f "$DIR/run/.containerenv" ]; then
     # ubi8-init image does not create .containerenv when running podman init, but we need to make sure that the file is there
-    touch "$DIR/run/.containerenv"
+    $ROOTLESS_CMD touch "$DIR/run/.containerenv"
 fi
 
 export OSCAP_CONTAINER_VARS
 OSCAP_CONTAINER_VARS=`podman inspect $ID --format '{{join .Config.Env "\n"}}'`
 
 export OSCAP_PROBE_ROOT
-OSCAP_PROBE_ROOT="$(cd "$DIR" && pwd)" || die "Unable to change current directory to OSCAP_PROBE_ROOT (DIR)."
+OSCAP_PROBE_ROOT="$($ROOTLESS_CMD cd "$DIR" && $ROOTLESS_CMD pwd)" || die "Unable to change current directory to OSCAP_PROBE_ROOT (DIR)."
 export OSCAP_EVALUATION_TARGET="$TARGET"
 shift 1
 
-$OSCAP_BINARY "$@"
+$ROOTLESS_CMD $OSCAP_BINARY "$@"
 EXIT_CODE=$?
 
+$ROOTLESS_CMD podman umount $ID > /dev/null || die "Failed to unmount."
 if [ $CLEANUP -eq 1 ]; then
-    # podman-rm should handle also unmounting of the container filesystem.
-    podman rm -f $ID > /dev/null || die "Failed to clean up."
-else
-    # If a running container is target of the scan just unmount its filesystem.
-    podman umount $ID > /dev/null || die "Failed to unmount."
+    podman rm $ID > /dev/null || die "Failed to clean up."
 fi
 exit $EXIT_CODE


tunacado:
Linking to [Red Hat Bugzilla 1906849](https://bugzilla.redhat.com/show_bug.cgi?id=1906849)

evgenyz:
@openscap-ci test this please

utils/oscap-podman
```diff
@@ -92,29 +93,26 @@ fi
 # podman init creates required files such as: /run/.containerenv - we don't care about output and exit code
 podman init $ID &> /dev/null || true
 
-DIR=$(podman mount $ID) || die "Failed to mount."
+DIR=$($ROOTLESS_CMD podman mount $ID) || die "Failed to mount."
 
 if [ ! -f "$DIR/run/.containerenv" ]; then
     # ubi8-init image does not create .containerenv when running podman init, but we need to make sure that the file is there
-    touch "$DIR/run/.containerenv"
+    $ROOTLESS_CMD touch "$DIR/run/.containerenv"
```
jan-cerny, line 100: I think the command used in the if condition (`[ ! -f "$DIR/run/.containerenv" ]`) also needs to be run with `podman unshare`.

jan-cerny (commented):
This is amazing patch!

Unfortunately, it doesn't work for me, because when I try to scan any profile as a normal user I get every rule evaluated as not applicable. But under `sudo` it gives expected results.

For example:

```
utils/oscap-podman registry.access.redhat.com/rhel7 xccdf eval --profile ospp /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml 
sudo  utils/oscap-podman registry.access.redhat.com/rhel7 xccdf eval --profile ospp /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml 
```
The sudo version produces expected results.

```
utils/oscap-podman registry.access.redhat.com/ubi8 xccdf eval --profile ospp /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml 
sudo  utils/oscap-podman registry.access.redhat.com/ubi8 xccdf eval --profile ospp /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml
```

Again, the sudo version produces expected results.


I'm using podman-2.2.1-1.fc32.x86_64, scap-security-guide-0.1.53-1.fc32.noarch and openscap-1.3.4-2.fc32.x86_64.

What do I miss? Have you experienced similar problems?

evgenyz:
Podman's `unshare` is very specific technique that might be not the only thing we need and maybe not in this exact way.

https://www.redhat.com/sysadmin/rootless-podman-makes-sense

jan-cerny:
> What do I miss? Have you experienced similar problems?

Apparently the problem is about determining the applicability. The applicability is determined using RPM check. If the platforms checks are removed from the datastream, it produces some results, but some other RPM results are broken. I guess it has problems with chroot(). As a non-root user it can't chroot but some code parts in oscap require a chroot. I don't know what to do with that, it might be that we will need further changes in oscap.

evgenyz:
Have you tried it with the elevated capabilities binary (used in tests)?

jan-cerny:
> Have you tried it with the elevated capabilities binary (used in tests)?

Yes, but I still get wrong results

robbmanes:
Just letting you know I didn't forget about this, was on break and back today.  Thank you everybody for testing this and working on it.

evgenyz:
So, `chroot` again. Or, our not exactly correct usage of it because of the `librpm` bugs with root directory override (and it is the only reason for chroot ATM, we were (or could be) able to replace everything else).

evgenyz:
> Have you tried it with the elevated capabilities binary (used in tests)?

I think that `podman --unshare` interferes with `chroot` capabilities.

jan-cerny:
> I think that `podman --unshare` interferes with `chroot` capabilities.

Yes, that's likely.

However, I can't see any failure in the verbose logs. I would expect to see `chroot failed` in the logs. But I can see 
* Switching probe to PROBE_OFFLINE_CHROOT mode.
* I will run rpminfo_probe_main
* Package .... not found.

jan-cerny:
The BZ https://bugzilla.redhat.com/show_bug.cgi?id=1566985 has been closed as WONTFIX and my understanding is that without that we can't get rid of chroot which blocks using the podman unshare command which means we can't merge this PR. @evgenyz Is my understanding correct? Or is there some other information?

evgenyz:
> The BZ https://bugzilla.redhat.com/show_bug.cgi?id=1566985 has been closed as WONTFIX and my understanding is that without that we can't get rid of chroot which blocks using the podman unshare command which means we can't merge this PR. @evgenyz Is my understanding correct? Or is there some other information?

I'd say that RPM's inability to properly operate without chroot is the most definitive pain in the ass in regard of being rootless. OTOH in theory we can try to replace sudo with elevated capabilities of the binary itself. No sure if it would be a great idea in terms of security though.

matejak:
What should be done with this? Would it make sense to update https://bugzilla.redhat.com/show_bug.cgi?id=1566985 with the new reasoning why we would like to have it fixed?

evgenyz:
All the reasoning falls into the category of pushing for the ability to work with librpm without being root. We can do everything we want (well, almost) if we drop this requirement. I believe the desire to be root-less was already expressed in the bug, we won't bring anything new. 

On the other hand I'm not really convinced that we really want to be rootless, because root gives us access to all the data in the system, not only RPMs. We will loose a lot of important information sources without this privilege. Evaluation would be incomplete and inconsistent. 

I have some ideas about how to tackle this RPM-related rules problems by actually not using the RPM probe (or using it in a limited way). But I'd need some time to dig around.

PR #1669, opened 2021-01-06
Fix multiple segfaults and broken test in --stig-viewer feature

jan-cerny:
This PR aims to fix 2 segfaults in the code creating the XML output for DISA STIG Viewer.

The first segmentation fault is a new problem. It has been caused by a mistake in 9b40767967e533bdb340ca4c91f2fd1192694820. This segmentation fault occurs any time an user uses `--stig-viewer`. For example: `oscap xccdf eval --profile ospp --stig-viewer /tmp/xxx /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml`. 

Unfortunately, the problem hasn't been discovered by CI, because the test https://github.com/OpenSCAP/openscap/blob/maint-1.3/tests/API/XCCDF/unittests/test_single_rule_stigw.sh is wrong. https://github.com/OpenSCAP/openscap/blob/1ccdd3bb31b2f1d4da5c7e9b6018597454e879cd/tests/API/XCCDF/unittests/test_single_rule_stigw.sh#L22
This condition ignores any fail in the `oscap` command. A similar mistake is repeated also when the comparison script is executed, therefore the test also didn't ensure that the generated file is correct.

The second segmentation fault has been reported in https://bugzilla.redhat.com/show_bug.cgi?id=1911999 and https://bugzilla.redhat.com/show_bug.cgi?id=1912000. Shortly, the problem is that our code expects that every `xccdf:reference` element has a `href` attribute. But, according to XCCDF 1.2 specification section 6.2.6, the `href` attribute is optional. We don't expect no `href` and we crash.  Some SCAP content uses `xccdf:reference` without `href` attribute, for example DISA STIG for RHEL 7 V3R1. Moreover, his STIG uses Dublin Core elements within `xccdf:reference`. We will extend the aforementioned test case to cover these elements.

### DIFF
--- a/src/XCCDF/result.c
+++ b/src/XCCDF/result.c
@@ -1159,8 +1159,9 @@ void xccdf_result_to_dom(struct xccdf_result *result, xmlNode *result_node, xmlD
 			struct oscap_reference_iterator *references = xccdf_item_get_references(item);
 			while (oscap_reference_iterator_has_more(references)) {
 				struct oscap_reference *ref = oscap_reference_iterator_next(references);
-				if (strcmp(oscap_reference_get_href(ref), DISA_STIG_VIEWER_HREF[0]) == 0 ||
-				    strcmp(oscap_reference_get_href(ref), DISA_STIG_VIEWER_HREF[1]) == 0) {
+				const char *href = oscap_reference_get_href(ref);
+				if (href && (strcmp(href, DISA_STIG_VIEWER_HREF[0]) == 0 ||
+							strcmp(href, DISA_STIG_VIEWER_HREF[1]) == 0)) {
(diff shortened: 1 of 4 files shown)


evgenyz:
@jan-cerny Can you please check the test results. They look very strange.

jan-cerny:
Yes, I can, indeed it's strange

jan-cerny:
he problem with the current solution is that PYTHON variable isn't defined,
so it defaults to "python". On some systems, for example RHEL 8,
"python" doesn't exist, becuse they have only python3. This error was
previously ignored due to masked return code.

evgenyz:
LGTM, thank you!

Issue #1632, opened 2020-11-12
[SCAP.R.800] obtrusive data from probe when evaluating test content


jan-cerny:
#### Description of Problem:
During evaluation of r800-rhel-datastream.xml from [SCAP 1.3 validation test suite](https://csrc.nist.gov/CSRC/media/Projects/scap-validation-program/documents/SCAP1.3ValidationTestContent_1-3.0.0.0.zip) a warning about obtrusive data is produced.

I think that evaluation of a valid content should not produce this type of suspicious messages.

#### OpenSCAP Version:
1.3.4

#### Steps to Reproduce:

1. oscap xccdf eval --profile xccdf_gov.nist_profile_RHEL_MAC-1_Classified --results-arf results_arf.xml             requirements/R800/r800-rhel-datastream.xml

#### Actual Results:
stderr contains

```
W: oscap:     Obtrusive data from probe!
W: oscap:     Obtrusive data from probe!
W: oscap:     Obtrusive data from probe!
W: oscap:     Obtrusive data from probe!
```

#### Expected Results:
no obtrusive data warnings

#### Additional Information / Debugging Steps:

It seems to be the same thing as https://bugzilla.redhat.com/show_bug.cgi?id=1722003

jan-cerny:
I can see the same warning message also when evaluating test suite SCAP source data streams in requirements 2910 and 2920.

jan-cerny:
and also SCAP.R.3005

PR #1678, opened 2021-01-27
Fix how oval_probe_ext_eval checks absence of the response from the probe

evgenyz:
The code was updated in accordance with how empty messages (with sexp == NULL) are encoded in:
https://github.com/OpenSCAP/openscap/blob/d05a0e8f382f758820c7623deabdd0ea2dd28b03/src/OVAL/probes/SEAP/seap-packet.c#L256

Fixes annoying warning `Obtrusive data from probe!`.

Fixes: #1632.


jan-cerny:
It fixes both #1632 and rhbz#1722003 for me. You are amazing.

PR #1679, opened 2021-01-28
Add proper resource management to the handler of empty messages from a probe

evgenyz:
Follow-up for #1678.


evgenyz:
@jan-cerny Sorry for being late

jan-cerny:
aha

PR #1680, opened 2021-02-02
Update cpe test to not depend on internal cpe dictionary

matusmarhefka:
There are 2 tests:
* positive to match the CPE of the tested system
* negative to not match the CPE of the tested system (notapplicable result is expected)


tests/API/XCCDF/default_cpe/cpe-dict.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<cpe-list xmlns="http://cpe.mitre.org/dictionary/2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://cpe.mitre.org/dictionary/2.0 http://cpe.mitre.org/files/cpe-dictionary_2.3.xsd">
+      <cpe-item name="<<CPE>>">
+            <title xml:lang="en-us">Applicable example platform</title>
+            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="cpe-oval.xml">oval:x:def:1</check>
+      </cpe-item>
```
matusmarhefka, line 8: Hmm, do you think there should be one more cpe item with the `cpe:/o:example:not_applicable:5`?

evgenyz: No, don't think so. It won't be applicable if it does not exists (as a CPE entity anywhere), so why bother?

matusmarhefka: Well, my idea was to test the scenario, where you have a content including a CPE dictionary for a certain product (`cpe:/o:example:not_applicable:5` in this case), but this product is not matched as the scanned system has a different CPE.

evgenyz:
Mmm, the test is failing 🤷

matusmarhefka:
Hmm, seems that there is no `/etc/system-release-cpe` on ubuntu 18.04.

evgenyz:
Hmm, how about just 255-ing this test is there is no `/etc/system-release-cpe`?

tests/API/XCCDF/default_cpe/test_default_cpe.sh
```diff
@@ -6,45 +6,31 @@
...
 
 CPE_OVAL="cpe-oval.xml"
-FEDORA_XCCDF="fedoraX-xccdf.xml"
-FEDORA_CPE="cpe:/o:fedoraproject:fedora:"
-RHEL_XCCDF="rhelX-xccdf.xml"
-RHEL_CPE="cpe:/o:redhat:enterprise_linux:"
+CPE_DICT="cpe-dict.xml"
+XCCDF="xccdf.xml"
 SYSTEM_CPE_FILE="/etc/system-release-cpe"
+SYSTEM_CPE=$(cat $SYSTEM_CPE_FILE)
```
jan-cerny, line 12: Do I understand it well that the "cpe-oval.xml" in this directory always passes? If yes, why to bother with system CPE? Instead, you can hardcode a random dummy CPE into this test. wdyt?

matusmarhefka: Great idea, I will do it like that, thank you.

evgenyz:
Great, thank you all.

PR #1681, opened 2021-02-02
Fix "ERROR: The EULA was not accepted. Installation aborted." for 25-msodbcsql17_17.7.1.1-1_amd64.deb

evgenyz:
Should fix problem with CI in #1680. Sigh.


matusmarhefka:
Thanks for fixing this @evgenyz

PR #1682, opened 2021-02-03
Syncronize internal CPE dictionary with the official NIST NVD CPE Dictionary v2.3

evgenyz:
Also, deprecate it. In the next API-breaking version we will remove it (that's it: #1683).


evgenyz:
@matusmarhefka Quid pro quo?

evgenyz:
@openscap-ci test this please

jan-cerny:
Do you know what the dash means and when it will match and when it won't match?

I think when we are in it we can add F34 because it will be branched of rawhide soon.

evgenyz:
> Do you know what the dash means and when it will match and when it won't match?

As far as I was able to get from the specs dash in the update section means just itself, dash.

The `cpe:/o:oracle:linux:8:-` should not match `cpe:/o:oracle:linux:8`, I guess that is the reason you're asking this question.

Now, you might also ask, what would be the best solution here:

- Remove `cpe:/o:oracle:linux:8`
- Change it to `cpe:/o:oracle:linux:8:-`
- Document `cpe:/o:oracle:linux:8` as an exception?

My preference is the second option.

matusmarhefka:
> > Do you know what the dash means and when it will match and when it won't match?
> 
> As far as I was able to get from the specs dash in the update section means just itself, dash.
> 
> The `cpe:/o:oracle:linux:8:-` should not match `cpe:/o:oracle:linux:8`, I guess that is the reason you're asking this question.
> 
> Now, you might also ask, what would be the best solution here:
> 
>     * Remove `cpe:/o:oracle:linux:8`
> 
>     * Change it to `cpe:/o:oracle:linux:8:-`
> 
>     * Document `cpe:/o:oracle:linux:8` as an exception?
> 
> 
> My preference is the second option.

I am a little bit confused about this. Does it mean that `cpe:/o:oracle:linux:8` is deprecated and instead of removing it you propose to add suffix `:-` so it is never matched?

evgenyz:
> I am a little bit confused about this. Does it mean that `cpe:/o:oracle:linux:8` is deprecated and instead of removing it you propose to add suffix `:-` so it is never matched?

The correct form is `cpe:/o:oracle:linux:8:-` (per v2.3 official NIST NBD CPE list). We have an incorrect form `cpe:/o:oracle:linux:8`. They are **not** equal. We have 3 options to deal with this situation.

matusmarhefka:
> > I am a little bit confused about this. Does it mean that `cpe:/o:oracle:linux:8` is deprecated and instead of removing it you propose to add suffix `:-` so it is never matched?
> 
> The correct form is `cpe:/o:oracle:linux:8:-` (per v2.3 official NIST NBD CPE list). We have an incorrect form `cpe:/o:oracle:linux:8`. They are **not** equal. We have 3 options to deal with this situation.

Aha I see. Okay, I think that it is fine like it is now.

cpe/openscap-cpe-dict.xml
```diff
@@ -137,33 +69,9 @@
...
-      </cpe-item>
-      <cpe-item name="cpe:/o:suse:sles:10">
-            <title xml:lang="en-us">SUSE Linux Enterprise Server 10</title>
-            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.sles:def:10</check>
-      </cpe-item>
-      <cpe-item name="cpe:/o:suse:sled:10">
-            <title xml:lang="en-us">SUSE Linux Enterprise Desktop 10</title>
-            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.sled:def:10</check>
-      </cpe-item>
-      <cpe-item name="cpe:/o:suse:linux_enterprise_server:11">
```
matusmarhefka, line 152: Shouldn't we also change to `cpe:/o:suse:linux_enterprise_server:11:-`?

cpe/openscap-cpe-dict.xml
```diff
@@ -218,43 +118,27 @@
...
-      </cpe-item>
-      <cpe-item name="cpe:/o:windriver:wrlinux:8">
-            <title xml:lang="en-us">Wind River Linux 8</title>
-            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.wrlinux:def:8</check>
-      </cpe-item>
-      <cpe-item name="cpe:/o:windriver:wrlinux:1019">
-            <title xml:lang="en-us">Wind River Linux 1019</title>
-            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.wrlinux:def:1019</check>
-      </cpe-item>
-      <cpe-item name="cpe:/o:microsoft:windows_7">
```
matusmarhefka, line 233: I can see that all of the removed CPEs for Windows are still in the official dictionary, but have the `:-` suffix.

evgenyz:
Ooops moment. Continued in a new PR.

cpe/openscap-cpe-dict.xml
evgenyz: I decided to remove all platforms instead of having spotty support.

cpe/openscap-cpe-dict.xml
evgenyz: I decided to remove all platforms instead of having spotty support.

Issue #140, opened 2015-08-14
[SCAP.R.900] Add option to verify signature of signed XML files


iankko:
The `oscap` tool currently supports `--fetch-remote-resources` option to download remote XCCDF / OVAL content from 3rd-party locations depending on the URL in the check.

Since some XML content might be signed (to ensure data integrity), when fetching remote XCCDF / OVAL content, there should be `--verify` option to verify the signature of the signed XML content (IOW verify the signed content is truly the authentic one).

Therefore this request for enhancement is filed in order the new --verify option to be added into the `oscap` command (at appropriate places where `--fetch-remote-resources` is allowed and where signed XML content can be expected to be provided).

The idea is that `--verify` option would automatically detect the key, the content has been signed with. It would try to download the public key for that key, and perform the signature verification.

Optionally the `--key` option could be specified too (if necessary) which would mean to hold the location of the public key, which should be used for signed XML content verification.

In ideal case the `--fetch--remote-resources` option would be able to detect if the XML content is signed or not. In the case it's not signed, it would proceed as currently. In the case the content is signed, and `--verify` option was provided, it would try to download the corresponding public key for that content, and verify the integrity of the content. If it wouldn't be able to locate the public key, it would issue `Unable to locate public key for content verification. Use --key option to specify the location.` message or something similar.

Thank you for the consideration.

Regards, Jan.

matejak:
This is a feature request that is essential for SCAP 1.3 support.

PR #1684, opened 2021-02-03
XML Signature Validation

jan-cerny:
Adds ability to validate signatures in digitally signed SCAP source data streams. Validation of other XML formats than SCAP source data streams is out of scope of this patch.

Needs dependency on xmlsec library (xmlsec1-openssl-devel).

The signature validation will be performed automatically when loading the source data stream file. Users can skip it using the `--skip-signature-validation` option.

This doesn't support keys and certificates form external files, because it isn't a requirement in certification requirements. However, the implementation leaves some room for adding these options in future, because they can make sense for users.

The old tests are replaced by new more focused tests. The documentation is updated.  As OpenSCAP implements the validation of digital signatures in SCAP source data stream, using xmlsec1 is no longer described in user manual.

Fixes: #140


jan-cerny:
@openscap-ci test this please

evgenyz (approved):
Awesome PR. That is a good start, let's merge and use it as the base for the research of corner cases. It would also be great to have it earlier for other platforms to catch build bugs (@tom-seewald I would like to kindly ask you to take a look).

docs/manual/manual.adoc
```diff
@@ -901,62 +901,17 @@ NOTE: Please note that for the rest of `oscap` functionality, unless you specify
...
-respective elements. This is due to the fact that there are way too many options
-when it comes to keystores and crypto choices. Instead we recommend users to use
-{xmlsec}[xmlsec1] to verify their SCAP content.  Safely evaluating signed
-content (with signature verification) involves the following steps:
+When evaluating a digitally signed SCAP source data stream OpenSCAP validates
+the digital signature of the data stream. The signature validation is performed
+automatically while loading the file. Data streams with invalid signature are
+rejected and OpenSCAP doesn't evaluate the content. OpenSCAP uses
+https://www.aleksey.com/xmlsec/[XML Security Library] with OpenSSL backend to
+validate the digital signature.
```
matejak, line 911:
```suggestion
validate the digital signature.

The signature validation only checks that the datastream hasn't been altered since its latest signature. OpenSCAP doesn't address trustworthiness of certificates or public keys that are part of the `KeyInfo` signature element and that are used to verify the signature. You should verify those keys yourself to prevent evaluation of datastreams that have been modified and signed by bad actors.
```

jan-cerny:
Hi @evgenyz and @matejak , thank you for the review. I have updated the PR. What is your opinion on the test? Tell me some suggestions for testing.

evgenyz:
I think that we are good with existing test for now. I might add another test with signature in ARF file (if it won't be a part of ARF-specific test unit, I'm still considering). But otherwise I think we should move forward with case-by-case approach.

utils/oscap.8
```diff
@@ -184,6 +184,11 @@ Selects a component ref from any datastream that references a component with XCC
 Do not validate input/output files.
 .RE
 .TP
+\fB\-\-skip-signature-validation\fR
```
jan-cerny, line 187: @jan-cerny Please add this to `dist/bash_completion.d/oscap`.

evgenyz:
LGTM. Thank you!

PR #1686, opened 2021-02-08
Add NIST tests for SCAP1.3

evgenyz:


jan-cerny:
Can you explain what is actually checked?

For example, R.400. We're supposed to run 4 different datastreams, the first is valid, second is invalid due to schema, third is a malformed XML and fourth is invalid according to schematron. We're supposed to check that the first is evaluated and the others are rejected by oscap. Trying to automate this using catalog would be nice, but currently the R400-1.3/catalog.xml describes one one of these 4 datastreams.

And there are also other requirements where we need to checks specific text on stdout/stderr or find contents of specific elements in the ARF or other result files.

PR #1687, opened 2021-02-08
Syncronize internal CPE dictionary with the official NIST NVD CPE Dictionary v2.3

evgenyz:
Continuation of #1682.


matusmarhefka:
LGTM. I am basically okay with any of the first two options proposed in https://github.com/OpenSCAP/openscap/pull/1682#issuecomment-773257049

PR #1689, opened 2021-02-16
Add --skip-validation option synonym for --skip-valid

evgenyz:
First step on the way to deprecate and then remove misleading option `--skip-valid` (why on earth should we skip something that is valid?).


jan-cerny:
@evgenyz What about deprecating it right now? I mean keep `--skip-valid` working, but document it as deprecated and prefer `--skip-validation` everywhere.

evgenyz:
> @evgenyz What about deprecating it right now? I mean keep `--skip-valid` working, but document it as deprecated and prefer `--skip-validation` everywhere.

I'm not against deprecating it in the documentation (I kind of already moved the accent away from it in this PR), but emitting a warning could break some tests. Also I changed all `--skip-valid` invocations in the project to `--skip-validation`.

jan-cerny:
It is a known issue: https://github.com/actions/virtual-environments/issues/2703 and they say they will roll out new image soon.

evgenyz:
> It is a known issue: [actions/virtual-environments#2703](https://github.com/actions/virtual-environments/issues/2703) and they say they will roll out new image soon.

QA? No, never heard about it.

Issue #1650, opened 2020-11-23
[SCAP.R.4000] create a test for AI


jan-cerny:
#### Description of Problem:
SCAP.R.4000 requires us to produce Asset Identification fields in ARF. We already do that, but I can't find any test that checks it. It would be nice to create a test for it.

#### OpenSCAP Version:
1.3.4

#### Operating System & Version:
f32

#### Steps to Reproduce:

1. ctest
2. 
3. 

#### Actual Results:
presence and completenes of AI fields is not tested

#### Expected Results:
presence and completenes of AI fields is tested

#### Additional Information / Debugging Steps:
no

jan-cerny:
this is not a blocker, but I want to use the label

PR #1691, opened 2021-02-16
tests: Add a test for the contents of the arf:asset entity

evgenyz:
Check that we are filling required Asset Identification Fields for the ARF result. Fixes: #1650.


evgenyz:
I have one more change that might be useful:

In line https://github.com/OpenSCAP/openscap/blob/3fb0134bbc06f3f37e04bc8b524695d55e056137/src/DS/rds.c#L342 change
`urn:xccdf:fact:ethernet:MAC` to `urn:xccdf:fact:asset:identifier:mac`. But it is a bit controversial in terms of backward compatibility. What do you thing @jan-cerny?

src/XCCDF/result.c
```diff
@@ -261,25 +278,67 @@ void xccdf_result_fill_sysinfo(struct xccdf_result *result)
...
+					xccdf_target_fact_set_name(fact, "urn:xccdf:fact:asset:identifier:fqdn");
+					xccdf_target_fact_set_string(fact, p->ai_canonname);
+					/* store FQDN under XCCDF1.2 (6.6.3) predefined name */
+					_xccdf_result_add_target_fact_uniq(result, fact);
+					if (!fqdn)
+						fqdn = strdup(p->ai_canonname);
+				}
+				freeaddrinfo(info);
+			} else {
+				dI("Unable to get FQDN(s) via getaddrinfo: %s\n", gai_strerror(gai_res));
```
jan-cerny, line 315: The `dI` macro will print end of line automatically.

```suggestion
				dI("Unable to get FQDN(s) via getaddrinfo: %s", gai_strerror(gai_res));
```

src/XCCDF/result.c
```diff
@@ -313,23 +376,28 @@ void xccdf_result_fill_sysinfo(struct xccdf_result *result)
...
+					if (getnameinfo(ifa->ifa_addr, sizeof(struct sockaddr_in),
+							hostip, sizeof(hostip), NULL, 0, NI_NUMERICHOST))
+						goto out2;
+				} else if (family == AF_INET6) {
+					struct sockaddr_in6 *sin6;
+
+					sin6 = (struct sockaddr_in6 *) ifa->ifa_addr;
+					if (!inet_ntop(family, (const void *) &sin6->sin6_addr,
+							hostip, sizeof(hostip)))
+						goto out2;
```
jan-cerny, line 391: This kind of reminds me the old rule  "if you need more than 3 levels of indentation, you are screwed, and you should fix your program" :smile: In other words, do you think this can be extracted to a separate function?

tests/API/XCCDF/unittests/test_xccdf_results_arf_asset.sh
```diff
@@ -0,0 +1,46 @@
...
+    assert_exists 1 $asset'/ai:fqdn[text()="'$fqdn'"]'
+
+    local macs=`ifconfig -a | grep ether | uniq | awk -F ' ' '{print toupper($2);}'`
+    for mac in $macs; do
+        echo $asset'/ai:connections/ai:connection/ai:mac-address[text()="'${mac}'"]'
+        assert_exists 1 $asset'/ai:connections/ai:connection/ai:mac-address[text()="'$mac'"]'
+    done
+
+    local ip4s=`ifconfig -a | grep 'inet ' | uniq | awk -F ' ' '{print $2;}'`
+    for ip in $ip4s; do
```
jan-cerny, line 38: and the IPv6 addresses?

jan-cerny (commented):
This is amazing improvement. Thanks for going above or beyond.


> change urn:xccdf:fact:ethernet:MAC to urn:xccdf:fact:asset:identifier:mac. But it is a bit controversial in terms of backward 

According to XCCDF 1.2 specification page 41 the correct URL is `urn:xccdf:fact:asset:identifier:mac`. Why do you think is controversial? Maybe I'm missing something. I think this code is used when you generate ARF which you generate during the scan. Maybe it could break `oscap ds rds-create` if the XCCDF results file was created by an older version of OpenSCAP, but I don't know who would do that.

tests/API/XCCDF/unittests/test_xccdf_results_arf_asset.sh
evgenyz: I was just too lazy to write a conversion function from ifconfig's IPv6 representation into what we store in `ai:...` fields (ipcalc is not much of a help). And, since the source of all addresses is the same it felt a bit redundant.

src/XCCDF/result.c
evgenyz: This whole function cries for refactoring. But we have bigger fish to fry at this moment, I was trying to make the least invasive change.

src/XCCDF/result.c
evgenyz: Yeah, missed that one

evgenyz:
> Maybe I'm missing something. I think this code is used when you generate ARF which you generate during the scan. Maybe it could break `oscap ds rds-create` if the XCCDF results file was created by an older version of OpenSCAP, but I don't know who would do that.

Exactly the reason I was thinking about. It was working somehow before and we could break this implicit API, not a good thing to do for a minor version.

jan-cerny:
> Exactly the reason I was thinking about. It was working somehow before and we could break this implicit API, not a good thing to do for a minor version.

I think it's artificial fear. I can't imagine a use-case in which this would matter.  Also, in the worst case, the AI section would be missing a field, but the information would still exist in the XCCDF part.

tests/API/XCCDF/unittests/test_xccdf_results_arf_asset.sh
evgenyz: Okay, it's there

src/XCCDF/result.c
evgenyz: Fixed

evgenyz:
> > Exactly the reason I was thinking about. It was working somehow before and we could break this implicit API, not a good thing to do for a minor version.
> 
> I think it's artificial fear. I can't imagine a use-case in which this would matter. Also, in the worst case, the AI section would be missing a field, but the information would still exist in the XCCDF part.

Okay, changed it.

tests/API/XCCDF/unittests/test_xccdf_results_arf_asset.sh
jan-cerny: ok

jan-cerny:
lgtm

Issue #1644, opened 2020-11-18
[SCAP.R.100] Update user manual with regards to SCAP 1.3


jan-cerny:
#### Description of Problem:
SCAP.R.100: The product’s documentation (printed or electronic) MUST assert that it uses SCAP
and its component specifications and explain relevant details to the users of the product.

I think we follow this requirement but we mostly describe and reference SCAP 1.2 so I would like to have a review and refresh regarding SCAP 1.3. For example change links to specifications to 1.3, update examples etc.

Also, add examples for RHEL 8 and check whether everything makes sense on RHEL 8.

#### OpenSCAP Version:
1.3.4

#### Operating System & Version:
all

#### Steps to Reproduce:

1. open user manual
2. 
3. 

#### Actual Results:
some outdated information

#### Expected Results:
SCAP 1.3 is mentioned and everything is up-to-date

#### Additional Information / Debugging Steps:
no

PR #1693, opened 2021-02-16
Refresh OpenSCAP user manual

jan-cerny:
This commit wants to improve OpenSCAP user manual, make it
up-to-date and more understandable for the readers.
- update for SCAP 1.3
- fix broken links
- improve document structure
- remove duplicate information
- remove outdated information
- document oscap-podman
- promote using data streams over separate XCCDF and OVAL files
- fix adoc syntax
- fix style
- fix some grammar
- use more real or currently used examples
- fixed misleading information about oscap xccdf generate fix
- added generating Ansible Playbooks

Reviewers, please, I know it's tempting to check just for typos and grammar. But try to change your mind and focus on the content instead. Check on the documented facts. It's really important to avoid misleading information and make sure that there aren't wrong or not working commands or mistakes in command line options. Also, try to pretend that you are an user.

Fixes: #1644

### DIFF
--- a/docs/manual/manual.adoc
+++ b/docs/manual/manual.adoc
@@ -4,6 +4,7 @@
 :sce_web:       https://www.open-scap.org/features/other-standards/sce/
 :openscap_web: https://open-scap.org/
 :oscap_git:     https://github.com/OpenSCAP/openscap
+:devel_manual:  https://github.com/OpenSCAP/openscap/blob/maint-1.3/docs/developer/developer.adoc
 :ssg_git:       https://github.com/OpenSCAP/scap-security-guide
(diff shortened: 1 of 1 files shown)


matejak:
As the documentation has been basically refactored, I would propose to explicitly mark the "fixed" result as an alias for XCCDF "pass" result. All other results in the table are in line with the XCCDF standard.

matejak:
The PR refactors the documentation, updates it with SCAP 1.3 content, and provides it with updates.

The manual itself is still not readable in a sequential manner, but it is a detailed resource if one is able to searching for interesting subjects.

PR #1697, opened 2021-02-24
Fix ARF asset information test and the way we collect and store these entites

evgenyz:


evgenyz:
We can ignore `Arm64` test result, it is a no-relevant glitch.

src/DS/rds.c
```diff
@@ -271,39 +271,13 @@ static xmlNodePtr ds_rds_add_ai_from_xccdf_results(xmlDocPtr doc, xmlNodePtr ass
...
-			char *delimiter = strchr(content, '.');
-			if (delimiter)
-				*delimiter = '\0';
-
-			xmlNewTextChild(computing_device, ai_ns, BAD_CAST "hostname", BAD_CAST content);
-
-			free(content);
-		}
-		else if (strcmp((const char*)(test_result_child->name), "target-address") == 0)
+		if (strcmp((const char*)(test_result_child->name), "target-address") == 0)
```
jan-cerny, line 280:
```suggestion
		if (strcmp((const char *)(test_result_child->name), "target-address") == 0)
```

jan-cerny (approved):

Issue #1638, opened 2020-11-16
[SCAP.R.2850] XML namespace isn't ignored


jan-cerny:
#### Description of Problem:

When evaluating r2850-rhel-datastream.xml from [SCAP 1.3 validation test suite](https://csrc.nist.gov/CSRC/media/Projects/scap-validation-program/documents/SCAP1.3ValidationTestContent_1-3.0.0.0.zip) I get unexpected results. All rules are evaluated as notapplicable by OpenSCAP. The CPE checks from this datastream are evaluated as not applicable. They consist of xmlfilecontent tests. The items described by xmlfilecontent_objects are not found. It's because the xmlfilecontent_objects contain simple XPath expressions without namespace, but the scanned XML file has a namespace. The test suite expects that the XPath expressions will be evaluated ignoring the namespace.

#### OpenSCAP Version:
1.3.4

#### Operating System & Version:
f32

#### Steps to Reproduce:

1.  cd requirements/R2850
2. cd configuration_scripts/
3. sudo python3 r2850_rhel_config_1.py
 (this script will copy the XML files gov.nist.secauto-scapval-1.3.2.swidtag from this directory to /scap_validation_content/R2850 that will be scanned by oscap in the next step, so you can review them here as well)
4. cd ..
5. oscap xccdf eval --results-arf results_arf.xml --profile xccdf_gov.nist.validation_profile_r2850-rhel r2850-rhel-datastream.xml

#### Actual Results:
xccdf_gov.nist.validation.r2850-rhel_rule_100 notapplicable
xccdf_gov.nist.validation.r2850-rhel_rule_200 notapplicable
xccdf_gov.nist.validation.r2850-rhel_rule_300 notapplicable

#### Expected Results:
xccdf_gov.nist.validation.r2850-rhel_rule_100 pass
xccdf_gov.nist.validation.r2850-rhel_rule_200 fail
xccdf_gov.nist.validation.r2850-rhel_rule_300 notapplicable

#### Additional Information / Debugging Steps:

There are elements like `<xpath>/SoftwareIdentity/@name</xpath>` in the datastream. They don't contain namespace. 

The file `requirements/R2850/configuration_scripts/gov.nist.secauto-scapval-1.3.2.swidtag` has an xmlns in the root element.
```
<SoftwareIdentity 
  xmlns="http://standards.iso.org/iso/19770/-2/2015/schema.xsd" 
  xml:lang="en-US" 
  name="SCAP Validation Tool" 
[...]

PR #1700, opened 2021-03-01
Ignore namespace in XPath expressions

jan-cerny:
The OVAL xmlfilecontent_test doesn't provide any way to register
namespaces or namespace prefixes. The only way to match namespaced XML
elements in an XPath expression would be to use wildcard and
local-name() function. However, the SCAP 1.3 validation test suite
expects in R2850 that the namespaces of the XML document are ignored.
And the existing SCAP content counts with this behavior.

Read this mailing list thread:
https://lists.cisecurity.org/empathy/thread/ZEJKSY7QMLVXNZ7QRQ555N7W2T2SDNTO

We haven't found any way to enable the behavior of ignoring namespaces
in libxml2. Therefore our solution is to remove namespaces from the
given XML file before examination. This is implemented using an one-off
XSLT transformation.

Fixes: #1638


evgenyz:
LGTM, thanks!

PR #1703, opened 2021-03-08
Add SCAP 1.3 source data stream schematron

jan-cerny:
This PR attempts to convert SCAP 1.3 source data stream XML schematron to use XPath 1.0 expressions. The original schematron contains XPath 2.0 but libxml can't process them. So this PR tries to workaround the lack of support for it.
We probably won't be able to convert all of the assertions to XPath 1.0, but let's try at least something.

The original schematron is located in `schemas/sds/1.3/source-data-stream-1.3.sch.original`.

The rules are added step-by step into `schemas/sds/1.3/source-data-stream-1.3.sch`.

~At this moment, please generate the XSLT file yourself! I won't regenerate it and commit the generated code after each change in the schematron file. When the PR will be ready for review we will of course generate the XSLT and commit it.~

### DIFF
--- a/schemas/sds/1.3/source-data-stream-1.3.sch
+++ b/schemas/sds/1.3/source-data-stream-1.3.sch
@@ -114,6 +114,9 @@
       <sch:assert id="scap-general-xccdf-status-rule-value-date-1" test="text()='draft' or text()= 'accepted'">SRC-5-1</sch:assert>
       <sch:assert id="scap-general-xccdf-status-rule-value-date-2" test="@date!=''">SRC-5-2</sch:assert>
     </sch:rule>
+    <sch:rule id="scap-cce-check-rule-1" context="xccdf:ident">
+      <sch:assert flag="WARNING" test="not(@system='http://cce.mitre.org' or @system='CCE') or text() != ''" id="scap-cce-check-assert-1">A-16-1|xccdf:Rule <sch:value-of select="ancestor::xccdf:Rule/@id"/></sch:assert>
+    </sch:rule>
   </sch:pattern>
 
 </sch:schema>
--- a/schemas/sds/1.3/source-data-stream-1.3.sch
+++ b/schemas/sds/1.3/source-data-stream-1.3.sch
@@ -100,6 +100,7 @@
     </sch:rule>
     <sch:rule id="scap-general-oval-generator" context="oval-def:generator/oval-com:schema_version">
       <sch:assert id="scap-general-oval-version" test="@platform or text()='5.3' or text()='5.4' or text()='5.5' or text()='5.6' or text()='5.7' or text()='5.8' or text()='5.9' or text()='5.10' or text()='5.10.1' or text()='5.11' or text()='5.11.1' or text()='5.11.2'">SRC-216-1</sch:assert>
+      <sch:assert id="scap-general-oval-platform-version" test="not(@platform) or text()='5.11.1:1.0' or text()='5.11.1:1.1' or text()='5.11.1:1.2' or text()='5.11.2:1.0' or text()='5.11.2:1.1' or text()='5.11.2:1.2'">SRC-216-2</sch:assert>
     </sch:rule>
   </sch:pattern>
 
--- a/schemas/sds/1.3/source-data-stream-1.3.sch
+++ b/schemas/sds/1.3/source-data-stream-1.3.sch
@@ -83,6 +83,7 @@
     </sch:rule>
     <sch:rule id="scap-general-xccdf-check" context="xccdf:Rule/xccdf:check">
       <sch:assert id="scap-general-xccdf-check-sys-req" test="@system = 'http://oval.mitre.org/XMLSchema/oval-definitions-5' or @system = 'http://scap.nist.gov/schema/ocil/2'">SRC-118-1|xccdf:Rule <sch:value-of select="ancestor::xccdf:Rule[1]/@id"/></sch:assert>
+      <sch:assert id="scap-general-xccdf-check-patches-ref-oval-only" test="not(current()/parent::xccdf:Rule[substring(@id, string-length(@id) - string-length('security_patches_up_to_date') + 1) = 'security_patches_up_to_date']) or current()/@system = 'http://oval.mitre.org/XMLSchema/oval-definitions-5'">SRC-169-2|xccdf:Rule <sch:value-of select="parent::xccdf:Rule/@id"/></sch:assert>
     </sch:rule>
   </sch:pattern>
 
--- a/schemas/sds/1.3/source-data-stream-1.3.sch
+++ b/schemas/sds/1.3/source-data-stream-1.3.sch
@@ -110,6 +110,10 @@
       <sch:assert id="scap-general-signature-sig-third-ref-manifest" test="dsig:Object/dsig:Manifest[concat('#',@Id) = current()/dsig:SignedInfo/dsig:Reference[3]/@URI]">SRC-288-1|dsig:Signature <sch:value-of select="@Id"/></sch:assert>
       <sch:assert id="scap-general-signature-sig-key-info" test="dsig:KeyInfo">SRC-290-1|dsig:Signature <sch:value-of select="@Id"/></sch:assert>
     </sch:rule>
+    <sch:rule id="scap-general-xccdf-status-rule-value-date" context="xccdf:status">
+      <sch:assert id="scap-general-xccdf-status-rule-value-date-1" test="text()='draft' or text()= 'accepted'">SRC-5-1</sch:assert>
+      <sch:assert id="scap-general-xccdf-status-rule-value-date-2" test="@date!=''">SRC-5-2</sch:assert>
+    </sch:rule>
   </sch:pattern>
 
 </sch:schema>
--- a/schemas/sds/1.3/source-data-stream-1.3.sch
+++ b/schemas/sds/1.3/source-data-stream-1.3.sch
@@ -108,6 +108,7 @@
       <sch:assert id="scap-general-signature-sig-first-ref-ds" test="ancestor::ds:data-stream-collection[1]/ds:data-stream[concat('#',@id) = current()/dsig:SignedInfo/dsig:Reference[1]/@URI]">SRC-286-1|dsig:Signature <sch:value-of select="@Id"/></sch:assert>
       <sch:assert id="scap-general-signature-sig-ref-sig-prop" test="dsig:Object/dsig:SignatureProperties[concat('#',@Id) = current()/dsig:SignedInfo/dsig:Reference[2]/@URI]">SRC-287-1|dsig:Signature <sch:value-of select="@Id"/></sch:assert>
       <sch:assert id="scap-general-signature-sig-third-ref-manifest" test="dsig:Object/dsig:Manifest[concat('#',@Id) = current()/dsig:SignedInfo/dsig:Reference[3]/@URI]">SRC-288-1|dsig:Signature <sch:value-of select="@Id"/></sch:assert>
+      <sch:assert id="scap-general-signature-sig-key-info" test="dsig:KeyInfo">SRC-290-1|dsig:Signature <sch:value-of select="@Id"/></sch:assert>
     </sch:rule>
   </sch:pattern>
 
--- a/schemas/sds/1.3/source-data-stream-1.3.sch
+++ b/schemas/sds/1.3/source-data-stream-1.3.sch
@@ -33,6 +33,7 @@
       <sch:assert id="scap-general-xccdf-benchmark-description" test="xccdf:description">SRC-10-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
       <sch:assert id="scap-general-xccdf-benchmark-lang-required" test="@xml:lang">SRC-2-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
       <sch:assert id="scap-general-xccdf-no-check-content-check" test="not(.//xccdf:check-content)">SRC-25-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
+      <sch:assert id="scap-general-xccdf-benchmark-val-forbid" test="not(.//xccdf:Value//xccdf:source) and not(.//xccdf:Value//xccdf:complex-value) and not(.//xccdf:Value//xccdf:complex-default) and not(.//xccdf:Value//xccdf:choices//xccdf:complex-choice)">SRC-276-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
     </sch:rule>
     <sch:rule id="scap-general-xccdf-profile" context="xccdf:Profile">
       <sch:assert id="scap-general-xccdf-profile-description" test="xccdf:description">SRC-10-1|xccdf:Profile <sch:value-of select="@id"/></sch:assert>
--- a/schemas/sds/1.3/source-data-stream-1.3.sch
+++ b/schemas/sds/1.3/source-data-stream-1.3.sch
@@ -30,101 +30,101 @@
 
   <sch:pattern id="scap-general">
     <sch:rule id="scap-general-xccdf-benchmark" context="xccdf:Benchmark">
-      <sch:assert id="scap-general-xccdf-benchmark-description" test="xccdf:description">SRC-10-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
-      <sch:assert id="scap-general-xccdf-benchmark-lang-required" test="@xml:lang">SRC-2-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
-      <sch:assert id="scap-general-xccdf-no-check-content-check" test="not(.//xccdf:check-content)">SRC-25-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
-      <sch:assert id="scap-general-xccdf-benchmark-val-forbid" test="not(.//xccdf:Value//xccdf:source) and not(.//xccdf:Value//xccdf:complex-value) and not(.//xccdf:Value//xccdf:complex-default) and not(.//xccdf:Value//xccdf:choices//xccdf:complex-choice)">SRC-276-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
-      <sch:assert id="scap-general-xccdf-benchmark-version-time-attribute-req" test="xccdf:version/@time">SRC-3-2|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
-      <sch:assert id="scap-general-xccdf-benchmark-no-xinclude" test="not(.//xinclude:include)">SRC-339-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
-      <sch:assert id="scap-general-xccdf-benchmark-version-update-req" test=".//xccdf:version[string(@update)]">SRC-341-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
-      <sch:assert id="scap-general-xccdf-benchmark-no-set-complex-value" test="not(.//xccdf:set-complex-value)">SRC-343-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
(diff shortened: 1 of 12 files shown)


schemas/sds/1.3/source-data-stream-1.3.sch
```diff
@@ -0,0 +1,131 @@
...
+  <sch:ns prefix="xccdf" uri="http://checklists.nist.gov/xccdf/1.2"/>
+  <sch:ns prefix="xcf" uri="nist:scap:xslt:function"/>
+  <sch:ns prefix="xinclude" uri="http://www.w3.org/2001/XInclude"/>
+  <sch:ns prefix="xlink" uri="http://www.w3.org/1999/xlink"/>
+  <sch:ns prefix="xml" uri="http://www.w3.org/XML/1998/namespace"/>
+  <sch:ns prefix="xsd" uri="http://www.w3.org/2001/XMLSchema"/>
+
+  <sch:pattern id="scap-general">
+    <sch:rule id="scap-general-xccdf-benchmark" context="xccdf:Benchmark">
+      <sch:assert id="scap-general-xccdf-benchmark-description" test="xccdf:description">SRC-10-1|xccdf:Benchmark <sch:value-of select="@id"/></sch:assert>
```
jan-cerny, line 33: @ggbecker I think that if an user will see `SRC-10-1` in `oscap` output they won't know what is that. I think that we should try to add some description. Or at least add everywhere `Schematron validation error` or something. What do you think?

ggbecker: Does NIST provide a document with the requirements? I think we can at least point to this document and throw the generic "Schematron validation error"

schemas/sds/1.3/source-data-stream-1.3.sch
```diff
@@ -0,0 +1,131 @@
...
+      <sch:assert id="scap-general-signature-sig-key-info" test="dsig:KeyInfo">SRC-290-1|dsig:Signature <sch:value-of select="@Id"/></sch:assert>
+    </sch:rule>
+    <sch:rule id="scap-general-xccdf-status-rule-value-date" context="xccdf:status">
+      <sch:assert id="scap-general-xccdf-status-rule-value-date-1" test="text()='draft' or text()= 'accepted'">SRC-5-1</sch:assert>
+      <sch:assert id="scap-general-xccdf-status-rule-value-date-2" test="@date!=''">SRC-5-2</sch:assert>
+    </sch:rule>
+    <sch:rule id="scap-cce-check-rule-1" context="xccdf:ident">
+      <sch:assert flag="WARNING" test="not(@system='http://cce.mitre.org' or @system='CCE') or text() != ''" id="scap-cce-check-assert-1">A-16-1|xccdf:Rule <sch:value-of select="ancestor::xccdf:Rule/@id"/></sch:assert>
+      <sch:assert flag="ERROR" test="not(@system='http://cce.mitre.org' or @system='CCE') or startswith(text(), 'CCE-')" id="scap-cce-check-assert-2">A-17-1|<sch:value-of select="."/></sch:assert>
```
ggbecker, line 119:
```suggestion
      <sch:assert flag="ERROR" test="not(@system='http://cce.mitre.org' or @system='CCE') or starts-with(text(), 'CCE-')" id="scap-cce-check-assert-2">A-17-1|<sch:value-of select="."/></sch:assert>
```
`startswith` function doesn't exist. It's called `starts-with`

```
xmlXPathCompOpEval: function startswith not found
XPath error : Unregistered function
OpenSCAP Error: Could not apply XSLT
...
```

schemas/sds/1.3/source-data-stream-1.3.sch
```diff
@@ -0,0 +1,131 @@
...
+    </sch:rule>
+    <sch:rule id="scap-cce-check-rule-2" context="oval-def:reference">
+      <sch:assert flag="WARNING" test="not(@source='http://cce.mitre.org' or @source='CCE') or @ref_id!=''" id="scap-cce-check-assert-5">A-16-1|oval-def:definition <sch:value-of select="ancestor::oval-def:definition/@id"/></sch:assert>
+      <sch:assert flag="ERROR" test="not(@source='http://cce.mitre.org' or @source='CCE') or startswith(@ref_id, 'CCE-')" id="scap-cce-check-assert-6">A-17-1|<sch:value-of select="@ref_id"/></sch:assert>
```
ggbecker, line 123:
```suggestion
      <sch:assert flag="ERROR" test="not(@source='http://cce.mitre.org' or @source='CCE') or starts-with(@ref_id, 'CCE-')" id="scap-cce-check-assert-6">A-17-1|<sch:value-of select="@ref_id"/></sch:assert>
```
`startswith` function doesn't exist. It's called `starts-with`

```
xmlXPathCompOpEval: function startswith not found
XPath error : Unregistered function
OpenSCAP Error: Could not apply XSLT
...
```

schemas/sds/1.3/source-data-stream-1.3.sch
evgenyz: We should emit the same message original 2.0 Schematron does. Is it possible to do?

jan-cerny: Yes, they provide, it's in the same zip from NIST as the schematron file, it's `requirements/scapval-scap-1.3-requirements.xml`

As you can see in the original schematron there are these short messages - they come from there. I have mostly just copied them.

jan-cerny: we should at least enable distinguishing between warnings and errors

ggbecker:
I've done a round of [unit tests](https://github.com/usnistgov/scapval/tree/master/scap-schematron-rules/src/test/resources/unit-tests/scap-1.3) from scapval using the schematron proposed here and all the implemented asserts are passing the tests.

[scapval_unit_tests_using_converted_schematron.txt](https://github.com/OpenSCAP/openscap/files/6129408/scapval_unit_tests_using_converted_schematron.txt)

evgenyz:
> I've done a round of [unit tests](https://github.com/usnistgov/scapval/tree/master/scap-schematron-rules/src/test/resources/unit-tests/scap-1.3) from scapval using the schematron proposed here and all the implemented asserts are passing the tests.
> 
> [scapval_unit_tests_using_converted_schematron.txt](https://github.com/OpenSCAP/openscap/files/6129408/scapval_unit_tests_using_converted_schematron.txt)

What about DSes from NIST's SCAP1.3ValidationTestContent_1-3.0.0.0?

ggbecker:
> > I've done a round of [unit tests](https://github.com/usnistgov/scapval/tree/master/scap-schematron-rules/src/test/resources/unit-tests/scap-1.3) from scapval using the schematron proposed here and all the implemented asserts are passing the tests.
> > [scapval_unit_tests_using_converted_schematron.txt](https://github.com/OpenSCAP/openscap/files/6129408/scapval_unit_tests_using_converted_schematron.txt)
> 
> What about DSes from NIST's SCAP1.3ValidationTestContent_1-3.0.0.0? Especially R400 and R1300.

No, no yet. I guess this can be the next step.

At least we already know that the implemented asserts work as expected.

jan-cerny:
@evgenyz @ggbecker  I have checked the data stream that should be invalid according to the schematron for source data stream. The oscap produces multiple findings. When I compared that with SCAPVal report I noticed that all of them are shown as warnings in the report. However, we don't have these that should be errors, in this case the SRC-236-2 (scap-use-case-conf-verification-benchmark-one-rule-ref-oval-ocil) and SRC-346-1 (scap-general-scap-content-xccdf-check-content-ref-name-not-req).

This leads me to the problem that when we use the schematron directly the users don't know the message category (warning, error). Do you think we should add it?

ggbecker:
> @evgenyz @ggbecker I have checked the data stream that should be invalid according to the schematron for source data stream. The oscap produces multiple findings. When I compared that with SCAPVal report I noticed that all of them are shown as warnings in the report. However, we don't have these that should be errors, in this case the SRC-236-2 (scap-use-case-conf-verification-benchmark-one-rule-ref-oval-ocil) and SRC-346-1 (scap-general-scap-content-xccdf-check-content-ref-name-not-req).
> 
> This leads me to the problem that when we use the schematron directly the users don't know the message category (warning, error). Do you think we should add it?

SCAPVal seems to use: [scapval-scap-1.3-requirements.xml](https://github.com/usnistgov/scapval/blob/master/scap-schematron-rules/src/main/resources/requirements/scapval-scap-1.3-requirements.xml) to decide whether an assertion is required or not. I don't know if we can use that resource in OpenSCAP.

Additionally, I think should be able to differentiate between warning and failures and so the final result (e.g exit code) can be properly produced. Where else would this schematron verification be used? Is it only during `oscap xccdf validate --schematron` ?

evgenyz:
> > @evgenyz @ggbecker I have checked the data stream that should be invalid according to the schematron for source data stream. The oscap produces multiple findings. When I compared that with SCAPVal report I noticed that all of them are shown as warnings in the report. However, we don't have these that should be errors, in this case the SRC-236-2 (scap-use-case-conf-verification-benchmark-one-rule-ref-oval-ocil) and SRC-346-1 (scap-general-scap-content-xccdf-check-content-ref-name-not-req).
> > This leads me to the problem that when we use the schematron directly the users don't know the message category (warning, error). Do you think we should add it?


> SCAPVal seems to use: [scapval-scap-1.3-requirements.xml](https://github.com/usnistgov/scapval/blob/master/scap-schematron-rules/src/main/resources/requirements/scapval-scap-1.3-requirements.xml) to decide whether an assertion is required or not. I don't know if we can use that resource in OpenSCAP.

It is a part of `scap-schematron-rules-1.3.5.zip`, provided by NIST. As far as I know, all work done by the government is a public domain, unless stated otherwise. I guess we can use it.
 
> Additionally, I think should be able to differentiate between warning and failures and so the final result (e.g exit code) can be properly produced. Where else would this schematron verification be used? Is it only during `oscap xccdf validate --schematron` ?

Yes. And it will stay optional.

jan-cerny:
@ggbecker I have added check for SRC-236-2

src/source/schematron.c
```diff
@@ -100,8 +107,342 @@ struct oscap_schema_table_entry OSCAP_SCHEMATRON_TABLE[] = {
...
+	xmlNodePtr component = _xcf_get_component(component_ref, context);
+	if (component == NULL) {
+		dD("component not found\n");
+		return false;
+	}
+
+	char *name = (char *) xmlGetProp(check_content_ref_node, BAD_CAST "name");
+
+	/* exists(xcf:get-component(xcf:get-component-ref($m/cat:catalog, $o/@href))//oval-def:definition[@id eq $o/@name and matches(@class, '^(compliance|patch)$')]) or exists(xcf:get-component(xcf:get-component-ref($m/cat:catalog, $o/@href))//ocil:questionnaire[@id eq $o/@name])) */
+	bool res = (_req_src_236_2_sub5(component, context, name, rule_id) || _req_src_236_2_sub6(component, context, name, rule_id));
```
ggbecker, line 292: I believe you should test which kind the retrieved component is.

For example, if you try to run this xpath query `".//ocil:questionnaire[@id='%s']"` using an OVAL component it yields:

`XPath error : Undefined namespace prefix`

The datastream: `R1300/r1300.2.2-datastream.xml.result.xml` line 108 contains the href:

`<xccdf:check-content-ref href="495dA_374l7P7640Z8oZuU27CQp049194N6-0585Ob1621ui41540dv2Dt"`, if you remove the last char `t` it will correctly find the component (OVAL one) and the fail does not occur in the line schematron.c:283 anymore. I don't know if this typo in the datastream is intended, but it contains the OVAL definition name that is missing from the OVAL component, which is the core assert of SRC-346-1 (to check if a OVAL definition exists in the OVAL component)

jan-cerny: I think the problem was that I haven't registered the `ocil:` namespace prefix. I have fixed  that below.

> I believe you should test which kind the retrieved component is.

I think that if `_req_src_236_2_sub5` or `_req_src_236_2_sub6` don't find the element they will return false. And their calls are a part of a disjunction.

ggbecker: That seems to fix the problem. But do you think that the potential typo in the `R1300/r1300.2.2-datastream.xml.result.xml` is intentional or not? with that character in place, the datastream is unable to be processed and find the correct OVAL component. Nevertheless, the implementation seems to be working in detecting if the oval definition is present or not.

jan-cerny: I think that's intentional there to check if the tested product is able to detect this.

ggbecker: Okay, it sort of makes sense.

ggbecker:
The schematron test is failing: tests/DS/schematron/schematron.sh

It seems that the combination of results is not being correctly produced, for example the output for `simple_ds.xml` schematron validation:

`./oscap_wrapper  --verbose DEVEL xccdf validate --schematron /home/ggasparb/workspace/github/openscap/tests`
```/DS/schematron/simple_ds.xml
I: oscap: Identified document type: data-stream-collection [oscap(1416542):oscap(7f1971228d80):doc_type.c:96:oscap_determine_document_type_reader]
D: oscap: Validating SCAP Source Datastream (1.3) document from /home/ggasparb/workspace/github/openscap/tests/DS/schematron/simple_ds.xml. [oscap(1416542):oscap(7f1971228d80):oscap_source.c:350:oscap_source_validate]
I: oscap: Identified document type: oval_definitions [oscap(1416542):oscap(7f1971228d80):doc_type.c:96:oscap_determine_document_type_reader]
Starting schematron validation of 'test_single_rule.oval.xml':
Schematron validation of 'test_single_rule.oval.xml': PASS

I: oscap: Identified document type: Benchmark [oscap(1416542):oscap(7f1971228d80):doc_type.c:96:oscap_determine_document_type_reader]
Starting schematron validation of 'scap_org.open-scap_cref_test_single_rule.xccdf.xml':
<?xml version="1.0"?>
Warning: The 'Benchmark' element has no platform specified, which implies the benchmark applies to all platforms. Applicable platforms should be indicate if possible. See the XCCDF 1.2.1 specification, Section 6.2.5.

Schematron validation of 'scap_org.open-scap_cref_test_single_rule.xccdf.xml': FAIL

I: oscap: Identified document type: oval_definitions [oscap(1416542):oscap(7f1971228d80):doc_type.c:96:oscap_determine_document_type_reader]
Starting schematron validation of 'scap_org.open-scap_cref_test_single_rule.oval.xml':
Schematron validation of 'scap_org.open-scap_cref_test_single_rule.oval.xml': PASS

Starting schematron validation of '/home/ggasparb/workspace/github/openscap/tests/DS/schematron/simple_ds.xml':
D: oscap: Checking ds:component-ref id='scap_org.open-scap_cref_test_single_rule.xccdf.xml' [oscap(1416542):oscap(7f1971228d80):schematron.c:527:_req_src_346_1_sub1]
D: oscap: Found OVAL definition id='oval:x:def:1' [oscap(1416542):oscap(7f1971228d80):schematron.c:433:_req_src_346_1_sub5]
<?xml version="1.0"?>
Warning: SRC-207-1|oval-def:definition oval:x:def:1
Warning: SRC-3-2|xccdf:Benchmark xccdf_com.example.www_benchmark_dummy
Warning: SRC-341-1|xccdf:Benchmark xccdf_com.example.www_benchmark_dummy
Warning: SRC-4-1|xccdf:Benchmark xccdf_com.example.www_benchmark_dummy
Warning: SRC-251-1|xccdf:Rule xccdf_com.example.www_rule_test-pass
Warning: A-26-1|xccdf:Rule xccdf_com.example.www_rule_test-pass


Schematron validation of '/home/ggasparb/workspace/github/openscap/tests/DS/schematron/simple_ds.xml': PASS
```

I would expect the validation to fail since it shows `Schematron validation of 'scap_org.open-scap_cref_test_single_rule.xccdf.xml': FAIL`

jan-cerny:
I have squashed some of the commits and I have rebased on maint-1.3 branch.

schemas/sds/1.3/source-data-stream-1.3.sch.original
```diff
@@ -0,0 +1,329 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jan-cerny, line 1: This is the original file from NIST. Do we want to keep this file or should it be removed? This file isn't used at the moment by OpenSCAP. But it's good for a reference.

src/source/oscap_source.c
```diff
@@ -358,8 +358,16 @@ int oscap_source_validate(struct oscap_source *source, xml_reporter reporter, vo
 
 int oscap_source_validate_schematron(struct oscap_source *source, const char *outfile)
 {
-	return oscap_source_validate_schematron_priv(source, oscap_source_get_scap_type(source),
-			oscap_source_get_schema_version(source), outfile);
+	FILE *outfile_fd = stdout;
+	if (outfile == NULL)
```
jan-cerny, line 362: This function is called only at 2 places and always with outfile = NULL. I don't see a benefit in this parameter. Do you think we can mark this function as deprecated in the API and introduce a new function? Or would it be a "scope creep"?

evgenyz: I see it as a poor man's way to return the validation result back to the user's code without capturing the output. Coming up with a better return mechanism would definitely be a scope creep. 

On the other hand marking it as deprecated with a further note (maybe in a form of an issue) to provide a better way of doing the same should be OK, IMHO.

jan-cerny: But you shouldn't add something as OSCAP_DEPRECATED without providing a new solution at the same time because it would cause compiler warnings at the 2 places where the old function is called.

evgenyz: I would vote for a different PR then.

jan-cerny: OK

jan-cerny: https://github.com/OpenSCAP/openscap/issues/1720

schemas/sds/1.3/source-data-stream-1.3.sch.original
evgenyz: There is no harm in having it in the repo.

src/source/schematron.c
```diff
@@ -100,27 +107,577 @@ struct oscap_schema_table_entry OSCAP_SCHEMATRON_TABLE[] = {
...
+	 * assert scap-use-case-conf-verification-rule-ref-oval.
+	 *
+	 * <sch:assert id="scap-use-case-conf-verification-rule-ref-oval" test="if( @use-case eq 'CONFIGURATION' ) then every $m in ds:checklists/ds:component-ref satisfies every $n in xcf:get-component($m)//xccdf:Rule satisfies (if(exists($n/ xccdf:check[@system eq 'http://oval.mitre.org/XMLSchema/oval-definitions-5' or @system eq 'http://scap.nist.gov/schema/ocil/2']/xccdf:check-content-ref[exists(@name) and not(xcf:is-external-ref($m/cat:catalog, @href) cast as xsd:boolean)])) then some $o in $n/xccdf:check[@system eq 'http://oval.mitre.org/XMLSchema/oval-definitions-5' or @system eq 'http://scap.nist.gov/schema/ocil/2']/xccdf:check-content-ref[exists(@name) and not(xcf:is-external-ref($m/cat:catalog, @href) cast as xsd:boolean)] satisfies (exists(xcf:get-component(xcf:get-component-ref($m/cat:catalog, $o/@href))//oval-def:definition[@id eq $o/@name and matches(@class,'^(compliance|patch)$')]) or exists(xcf:get-component(xcf:  get-component-ref($m/cat:catalog, $o/@href))//ocil:questionnaire[@id eq $o/@name])) else true()) else true()">SRC-236-2|scap:data-stream <sch:value-of select="@id"/></sch:assert>
```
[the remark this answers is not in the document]
jan-cerny: It is for easier understanding of the code below/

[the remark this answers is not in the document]
evgenyz: Okay, I won't insists

src/source/oscap_source.c
```diff
@@ -358,8 +358,16 @@ int oscap_source_validate(struct oscap_source *source, xml_reporter reporter, vo
...
+		outfile_fd = stdout;
+	else
+		outfile_fd = fopen(outfile, "w");
```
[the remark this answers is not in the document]
jan-cerny: I think it should fail instead of falling back because it would do something different that it's asked to do and documented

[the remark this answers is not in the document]
evgenyz: OK

evgenyz:
Okay, I think we are good to go. Thank you all for the great job you did here!

PR #1704, opened 2021-03-08
Add a missing space

jan-cerny:
Thanks @ggbecker !


evgenyz:
Looks like it would take a bit more to fix this :)

jan-cerny:
apparently depends on the timezone

evgenyz:
Nice! Thanks for the fix.

PR #1706, opened 2021-03-15
Fix some of the coverity warnings

jan-cerny:
To get some information please read commit messages of every commit.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1938830


compat/dev_to_tty.c
```diff
@@ -303,6 +303,7 @@ unsigned dev_to_tty(char *restrict ret, unsigned chop, dev_t dev_t_dev, int pid,
   if((flags&ABBREV_PTS) && !strncmp(tmp,"pts/", 4) && tmp[4]) tmp += 4;
   /* gotta check before we chop or we may chop someone else's memory */
   if(chop + (unsigned long)(tmp-buf) <= sizeof buf)
```
evgenyz, line 305: What if `chop` would be equal `TTY_NAME_SIZE` and `tmp` == `buf`? I think that the problem here is in the `<=`, it should be `<`.

evgenyz:
LGTM. Thanks.

PR #1708, opened 2021-03-16
Update man page and help texts

jan-cerny:
Minor formatting and factual fixes in our man pages and help texts and
outputs. Often fixes incorrect usage of "data stream".


evgenyz:
LGTM.

PR #1709, opened 2021-03-17
Fix behavior of StateType operator

yuumasato:
- Fix behavior of the [StateType](https://github.com/OVALProject/Language/blob/master/docs/oval-definitions-schema.md#-statetype-) `operator`.
  - The result of a  `check_existence` was being added as an entity result.
    This was causing wrong evaluation of Items in tests with a State using `operator="OR"` when none of the State Entity matched. Would also have issues with States and operators when a difference in 1 state conformance would be relevant (for example with operators `ONE` and `XOR`)

In `file_permissions` template, when allowing stricter permissions, [a state `operator="OR"` is used](https://github.com/ComplianceAsCode/content/blob/11ee4fc5cf92acca36295fa495a5c17154e611b9/shared/templates/file_permissions/oval.template#L25).
With the template, a file with adequate file permissions is expected to fail all State entity evaluations and not match the State.
Without this patch, a file with adequate file permissions fails to satisfy the State entity evaluations, but still matches the state :exploding_head:.

Related to https://github.com/ComplianceAsCode/content/pull/6707 and https://github.com/ComplianceAsCode/content/pull/6709


evgenyz:
Huh. Looks like we starting to hit the very bottom of the OVAL language. None of the official tests caught this problem. Thanks for the fix.

yuumasato:
Thank you for the review, that was fast! :D

yuumasato:
A side note, it seems [the `filter action="include"`](https://github.com/OpenSCAP/openscap/blob/0a5492e004bbaa11c58c6d3a9d7b7e2307288191/src/OVAL/probes/probe-api.c#L266~L300) is not affected by this.

The `check_existence` is not relevant when filtering objects, only in the scope of the OVAL test.

PR #1710, opened 2021-03-17
Fix broken HTML report generation

jan-cerny:
Broken by fbb762b358861e52e850404cefa1de03c492a8f9.
xsltCleanupGlobals() is already executed in oscap_cleanup()
in src/common/oscapxml.c, line 122

Addressing:
[jcerny@localhost openscap{maint-1.3}]$ c_oscap xccdf eval  --report report.html /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml
xmlXPathCompOpEval: function node-set not found
XPath error : Unregistered function
runtime error: file /home/jcerny/work/git/openscap/xsl/xccdf-report.xsl line 83 element variable
Failed to evaluate the expression of variable 'last_test_time'.
OpenSCAP Error: Could not apply XSLT /home/jcerny/work/git/openscap/xsl/xccdf-report.xsl to XML file: NONEXISTENT [/home/jcerny/work/git/openscap/src/source/xslt.c:177]


evgenyz:
Nice. That is exactly the bug I stumbled upon while I was working on schematron update.

Issue #1701, opened 2021-03-02
Document environment variables honored by the OpenSCAP


evgenyz:
In our documentation we have some mentions scattered around, but there is no single place to read about all the variables that cold affect the behavior.
---

- [ ] SOURCE_DATE_EPOCH (#1699)

PR #1712, opened 2021-03-18
Document used environment variables

jan-cerny:
Fixes: #1701

### DIFF
--- a/docs/manual/manual.adoc
+++ b/docs/manual/manual.adoc
@@ -1513,6 +1513,18 @@ the presence of a matching SWID tag:
 $ oscap oval eval --results results.xml r2860-rhel-oval.xml
 ----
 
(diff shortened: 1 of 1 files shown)


evgenyz:
Great addition to the documentation. But there is another plain to this task: we use some libraries that could be affected by the environment as well, for example `cURL`: https://curl.se/libcurl/c/libcurl-env.html. It would be nice to at least mention these libraries with corresponding links to their documentation.

evgenyz:
I went through the deps and think that the cURL is probably the only interesting library in terms of altering the behavior via environment (proxy, headers etc). But maybe I'm over-complicating it. What do you think, @jan-cerny?

evgenyz:
LGTM. Thanks!

PR #1715, opened 2021-03-22
Ignore "Bad address" errors from the sysctl probe

evgenyz:
This problem manifests in the Travis CI environment: https://travis-ci.org/github/OpenSCAP/openscap/jobs/763932418.

`E: oscap:     An error ocured when reading from "/proc/sys/kernel/spl/hostid" (fp=0xaaaaea9d6cf0): l=0, 14, Bad address`.


jan-cerny:
I haven't investigated the code, but I have a question. Have you considered handling the issue in the probe code itself? Is that possible? Would it cause problems?

evgenyz:
> I haven't investigated the code, but I have a question. Have you considered handling the issue in the probe code itself? Is that possible? Would it cause problems?

I think that probe behaves correctly. If there is a "file" with content that would not be able to fit into the buffer we don't want to deal with such a value.

PR #1717, opened 2021-03-26
utils: Fix handling of '--arg=val'-styled common options

evgenyz:
Fixes RHBZ#1939425: https://bugzilla.redhat.com/show_bug.cgi?id=1939425


matusmarhefka:
@evgenyz Nice, this does the job. Thanks for fixing it and providing a test.

Issue #1713, opened 2021-03-18
Add a safeguard in form of a command line switch to avoid processing unsigned data streams


evgenyz:
### Problem

Scanner (according to the SCAP1.3 requirements) would process all unsigned data steams. For a user of the scanner, who would like to rely on signing procedure to protect from malicious content, there is no way to force OpenSCAP to evaluate only signed DSes.

### Proposal

For all modules that check the signature of the input data stream add `--enforce-signature` (or similar) switch, which would force the scanner to reject all unsigned documents. This option will conflict with `--skip-signature-validation` and should be added alongside (modules: `eval`, `guide`, `fix`).

PR #1719, opened 2021-04-06
utils: Add --enforce-signature option for eval, guide, and fix modules

evgenyz:
With this option set all unsigned data streams would be considered invalid. It is the antipode of `--skip-signature-validation` option.

Fixes: #1713.


utils/oscap.8
```diff
@@ -191,6 +191,11 @@ Do not validate input/output files.
 Do not validate digital signatures in digitally signed SCAP source data streams.
 .RE
 .TP
+\fB\-\-enforce-signature\fR
+.RS
+Process only digitally signed SCAP source data streams.
```
jan-cerny, line 196: Maybe add for clarity that data streams with no signature will be rejected.

src/source/signature.c
```diff
@@ -159,6 +159,11 @@ static int _oscap_signature_validate_doc(xmlDocPtr doc, oscap_document_type_t sc
 	xmlNodePtr root = xmlDocGetRootElement(doc);
 	node = xmlSecFindNode(root, xmlSecNodeSignature, xmlSecDSigNs);
 	if (node == NULL) {
+		if (enforce_signature) {
+			oscap_seterr(OSCAP_EFAMILY_XML, "Signature not found");
+			dW("Signature node not found");
```
jan-cerny, line 164: It's kind of duplicate.

utils/oscap.8
```diff
@@ -191,6 +191,11 @@ Do not validate input/output files.
 Do not validate digital signatures in digitally signed SCAP source data streams.
 .RE
 .TP
+\fB\-\-enforce-signature\fR
```
jan-cerny, line 194: Please add it also to `dist/bash_completion.d/oscap`.

jan-cerny (changes requested):
It is great

utils/oscap.8
evgenyz: Damn, always forgetting it. Thanks.

src/source/signature.c
evgenyz: Well, I agree.

utils/oscap.8
evgenyz: Done

PR #1726, opened 2021-04-12
OVAL/probes: Make 'proc' and 'sysfs' filesystems non-local

evgenyz:
More recently, we've come to use the -l switch in the df command to identify the "local" filesystems. We feel this is probably a better reflection of the intention behind the local attribute value.

The df utility considers 'proc' and 'sysfs' as "dummy" and does not produce results for them.

### DIFF
--- a/docs/manual/manual.adoc
+++ b/docs/manual/manual.adoc
@@ -1516,6 +1516,32 @@ the presence of a matching SWID tag:
 $ oscap oval eval --results results.xml r2860-rhel-oval.xml
 ----
 
+
+=== Notes on specifics of OVAL implementation
+
+==== Excluding non-local filesystems using the recurse_file_system="local" attribute of a FileBehaviors entity
+
+The scanner loosely follows the OVAL's idea behind this attribute to behave like
+the coreutils utility *df* (`df -l`). This is the list of filesystems, that are
+not considered local by the scanner:
+
+* proc, sysfs
+* afs
+* ceph
+* cifs
+* smb3, smbfs
+* sshfs
+* ncpfs, ncp
+* nfs, nfs4
+* gfs, gfs2
+* glusterfs
+* gpfs
+* pvfs2
+* ocfs2
+* lustre
+* davfs
+
+
 == List of accepted environment variables
 
 * `OSCAP_CHECK_ENGINE_PLUGIN_DIR` - Defines path to a directory that contains plug-in libraries implementing additonal check engines, eg. SCE.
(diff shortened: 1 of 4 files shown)


docs/manual/manual.adoc
```diff
@@ -1516,6 +1516,32 @@ the presence of a matching SWID tag:
 $ oscap oval eval --results results.xml r2860-rhel-oval.xml
 ----
 
+
+=== Notes on specifics of OVAL implementation
+
+==== unix-def:FileBehaviors with recurse_file_system="local"
+
+The scanner loosely follows the OVAL's idea behind this attibute to behave like
```
jan-cerny, line 1524:
```suggestion
The scanner loosely follows the OVAL's idea behind this attribute to behave like
```

docs/manual/manual.adoc
```diff
@@ -1516,6 +1516,32 @@ the presence of a matching SWID tag:
...
+the coreutils utility *df* (`df -l`). This is the list of filesystems, that are
+considered non-local by the scanner:
```
jan-cerny, line 1526:
```suggestion
not considered local by the scanner:
```

docs/manual/manual.adoc
```diff
@@ -1516,6 +1516,32 @@ the presence of a matching SWID tag:
 $ oscap oval eval --results results.xml r2860-rhel-oval.xml
 ----
 
+
+=== Notes on specifics of OVAL implementation
+
+==== unix-def:FileBehaviors with recurse_file_system="local"
```
jan-cerny, line 1522: It's not just unix-def namespace, it's also in texfilecontent54 which belongs to independent and also others.

Maybe it would be better to start the paragraph with a sentence mentioning all OVAL tests that are affected. Them you can change the title to something like 

```suggestion
==== Excluding non-local filesystems using the recurse_file_system="local" attribute
```

jan-cerny (commented):
Overall it looks great, I have only minor comments.

docs/manual/manual.adoc
evgenyz: It is still not clear for me how the difference in `ind-def:FileBehaviors`, `win-def:FileBehaviors` and `unix-def:FileBehaviors` should be handled. But you're right, on linux we behave in the same way both for ind and unix.

docs/manual/manual.adoc
evgenyz: Oops. Thanks!

PR #1728, opened 2021-04-12
Add error message

jan-cerny:
This adds the same error message as in oscap_source_validate.
If the document ins ivalid accrding to schematron and schematron
validation has been requested by the user, an error should be
reported.

### DIFF
--- a/src/source/oscap_source.c
+++ b/src/source/oscap_source.c
@@ -366,10 +366,20 @@ int oscap_source_validate_schematron(struct oscap_source *source, const char *ou
 			return -1;
 		}
 	}
-	int ret = oscap_source_validate_schematron_priv(source, oscap_source_get_scap_type(source),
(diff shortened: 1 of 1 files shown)


src/source/oscap_source.c
```diff
@@ -366,10 +366,20 @@ int oscap_source_validate_schematron(struct oscap_source *source, const char *ou
 			return -1;
 		}
 	}
-	int ret = oscap_source_validate_schematron_priv(source, oscap_source_get_scap_type(source),
-			oscap_source_get_schema_version(source), outfile_fd);
+	oscap_document_type_t scap_type = oscap_source_get_scap_type(source);
+	const char *schema_version = oscap_source_get_schema_version(source);
+	if (!schema_version) {
+			schema_version = "unknown schema version";
```
evgenyz, line 372:
```suggestion
		schema_version = "unknown schema version";
```

PR #1730, opened 2021-04-14
utils: Make schematron-based validation enabled by default for validate command of oval and xccdf modules

evgenyz:
Also introduce disabling option `--skip-schematron`. The `--schematron` option is now obsolete; it is removed from the documentation and would be deprecated in the next API-breaking release.

### DIFF
--- a/dist/bash_completion.d/oscap
+++ b/dist/bash_completion.d/oscap
@@ -3,7 +3,7 @@
 # returns number of params
 function _oscap_noarg {
     case "$1" in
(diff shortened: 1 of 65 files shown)


jan-cerny:
The changes look fine.

But when I think of it from the user perspective, he isn't informed about the XSD validation result other than by the exit code.

evgenyz:
> But when I think of it from the user perspective, he isn't informed about the XSD validation result other than by the exit code.

Right, right. Should be better now.

jan-cerny:
I think it's OK.

PR #1732, opened 2021-04-14
Distinguish between warnings and errors in XCCDF schematron

jan-cerny:
In the XCCDF 1.2 schematron there are warnings and errors.
Warnings will be displayed but they won't cause the validation
to fail, but errors will render the document invalid.

### DIFF
--- a/tests/DS/schematron/schematron.sh
+++ b/tests/DS/schematron/schematron.sh
@@ -3,11 +3,22 @@
 . $builddir/tests/test_common.sh
 set -e -o pipefail
 
+# both XCCDF and SDS schematrons find only warnings but not errors
 output="$(mktemp)"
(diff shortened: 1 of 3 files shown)


tests/DS/schematron/schematron.sh
```diff
@@ -3,11 +3,22 @@
 . $builddir/tests/test_common.sh
 set -e -o pipefail
 
+# both XCCDF and SDS schematrons find only warnings but not errors
 output="$(mktemp)"
-$OSCAP xccdf validate --schematron "$srcdir/simple_ds.xml" >"$output" || ret=$?
+$OSCAP xccdf validate --schematron "$srcdir/simple_ds.xml" >"$output"
```
evgenyz, line 8:
```suggestion
$OSCAP xccdf validate "$srcdir/simple_ds.xml" >"$output"
```

evgenyz:
Nice, thanks!

Issue #1729, opened 2021-04-14
[SCAP.R.3005] xccdf:rule-result doesn't contain role attribute


jan-cerny:
#### Description of Problem:

ARF file created by OpenSCAP when evaluating the r3005-datastream-01.xml from SCAP 1.3 validation test suite is invalid according to SCAPVAL.

#### OpenSCAP Version:
current usptream maint-1.3 as of HEAD 2e804671d765475138c41393f552c193882fbe01

#### Operating System & Version:
F 33

#### Steps to Reproduce:

1. oscap xccdf eval --profile xccdf_gov.nist.validation_profile_r3005_01 --results-arf /tmp/arf.xml R3005/r3005-datastream-01.xml
2. sudo java -jar /opt/scapval/scapval-1.3.2.jar -resultfile /tmp/arf.xml -scapversion 1.3
3. firefox validation-report.html

#### Actual Results:
Some of the rule-result elements in the generated ARF don't contain role attribute and therefore they fail scapval tool validation specifically RES-392-1

```The @role, @severity, and @weight attributes of the <xccdf:rule-result> element SHALL be provided to indicate their values used during assessment. ```

#### Expected Results:
the attribute isn't missing and scapval doesn't report problems

PR #1733, opened 2021-04-14
Set the default value of role attribute

jan-cerny:
By default the value of role member of struct xccdf_refine_rule was set
to zero when created which happened usually when parsing a
xccdf:refine-rule element. But the value of 0 doesn't map to the default
value of the role attribute of xccdf:refine-rule element. The default
value of this attribute is "full" according to the XCCDF specification.
As a result when creating XCDF results or ARF results the role attribute
of xccdf:rule-result hasn't been created. That's a problem according to
scapval.
Fixes: #1729


evgenyz:
LGTM, thanks!

PR #1734, opened 2021-04-15
Fix CMake warnings

evgenyz:
Problems addressed: 

```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```
```
CMake Warning (dev) at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:426 (message):
  The package name passed to `find_package_handle_standard_args` (DBUS) does
  not match the name of the calling package (DBus).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  cmake/FindDBus.cmake:59 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:104 (find_package)
```
```
CMake Error at utils/CMakeLists.txt:24 (target_link_libraries):
  The plain signature for target_link_libraries has already been used with
  the target "oscap".  All uses of target_link_libraries with a target must
  be either all-keyword or all-plain.
  The uses of the plain signature are here:
   * utils/CMakeLists.txt:11 (target_link_libraries)
```


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 2.8...3.19)
```
jan-cerny, line 1: Why do you set that 3.19 is the maximum? What will happen in future? Have you tried the 3.20? What are the problems of the current latest release?

jan-cerny (changes requested):
PR has no description.

CMakeLists.txt
evgenyz: It does not work like that. It basically says that we require the CMake to be *at least* somewhere between 2.8...3.19. There is no cap on the version, it is still *minimum required*.

evgenyz: Yeah, and 3.19 is what I have ATM and can confirm that it is working.

JAORMX (approved):

evgenyz:
Oh, hey Ozz, what wind brought you here? :)

JAORMX:
@evgeni got interested in reviewing some OpenSCAP PRs; that's all.

evgenyz:
@evgeni Hello, again :)

evgeni:
Hi! Long time no see :)

jan-cerny:
@evgenyz There is a similar message in https://jenkins.open-scap.org/job/openscap-pull-requests-multi/752/label=fedora/console also for Perl:

```
-- Found PCRE: /usr/lib64/libpcre.so  
CMake Warning (dev) at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (Perl) does
  not match the name of the calling package (PerlLibs).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
```

evgenyz:
Hmm, wonder why I don't have it

PR #1740, opened 2021-04-19
Remove codecov.io

jan-cerny:
Due to lack of proper unit tests in our project it would not
produce useful results.


evgenyz:
Okay

PR #1742, opened 2021-04-22
tests: Fix minor problems

evgenyz:


tests/DS/test_ds_misc.sh
```diff
@@ -253,8 +253,8 @@ function test_ds_continue_without_remote_resources() {
 function test_source_date_epoch() {
 	local xccdf="$srcdir/sds_multiple_oval/multiple-oval-xccdf.xml"
 	local result="$(mktemp)"
-	local timestamp="2020-03-05T12:09:37"
-	export SOURCE_DATE_EPOCH="1583410177"
+	local timestamp="2015-03-05T12:09:37"
+	export SOURCE_DATE_EPOCH="1425557377"
```
jan-cerny, line 257: I really don't understand the fix. I think that you have only changed the date but the rest works the same. Can you explain it?

evgenyz: The problem is in here:
```c
		char *source_date_epoch = getenv("SOURCE_DATE_EPOCH");
		if (source_date_epoch == NULL ||
			(mtime = (time_t)strtoll(source_date_epoch, NULL, 10)) <= 0 ||
			mtime > file_stat.st_mtime)
                                mtime = file_stat.st_mtime;
```
You have a "newer" source on your machine than in the repo.

evgenyz: Well, now I'm not even sure. Since I made the scratch build from my sources it might be that I have something off in terms of timestamps.

jan-cerny:
@openscap-ci test this please

tests/DS/test_ds_misc.sh
evgenyz: Okay, we don't need it. It was my screw-up.

jan-cerny: @evgenyz Do you know any details? I'm curious.

evgenyz: Well, it looks like I was trying to break the test at some point and succeeded. And forgot about it.

Issue #1743, opened 2021-04-22
Add missing XCCDF Profile selection remark to HTML report


cipherboy:
#### Description of Problem:

I was discussing how to add commentary to a tailoring file the other day on `#openscap`. I suggested adding a `remark` under the `select` element of a `profile`. However:

OpenSCAP doesn't display the `<xccdf:remark />` child of a `<xccdf:select />` inside a `<xccdf:Profile />` in the corresponding HTML report.

This is allowed [by the standard](https://csrc.nist.gov/CSRC/media/Publications/nistir/7275/rev-4/final/documents/nistir-7275r4_updated-march-2012_clean.pdf) -- pdf page 43, noted as page 36. 

Take for instance this DS:

```xml
      <xccdf-1.2:Profile id="xccdf_org.ssgproject.content_profile_cis_level1_server">
        <xccdf-1.2:title xml:lang="en-US" override="true">CIS Ubuntu 20.04 Level 1 Server Benchmark</xccdf-1.2:title>
        <xccdf-1.2:description xml:lang="en-US" override="true">This baseline aligns to the Center for Internet Security
Ubuntu 20.04 LTS Benchmark, v1.0.0, released 07-21-2020.</xccdf-1.2:description>
        <xccdf-1.2:select idref="xccdf_org.ssgproject.content_rule_file_groupowner_cron_allow" selected="true">
            <xccdf-1.2:remark>My testing remark on disabling groupowner_chron_allow</xccdf-1.2:remark>
        </xccdf-1.2:select>
        ....
      </xccdf-1.2:Profile>
```

This eventually gets rendered into a results.xml like the following:

```xml
  <Profile id="xccdf_org.ssgproject.content_profile_cis_level1_server">
    <title xmlns:xhtml="http://www.w3.org/1999/xhtml" xml:lang="en-US" override="true">CIS Ubuntu 20.04 Level 1 Server Benchmark</title>
    <description xmlns:xhtml="http://www.w3.org/1999/xhtml" xml:lang="en-US" override="true">This baseline aligns to the Center for Internet Security
Ubuntu 20.04 LTS Benchmark, v1.0.0, released 07-21-2020.</description>
    <select idref="xccdf_org.ssgproject.content_rule_file_groupowner_cron_allow" selected="true">
      <remark xml:lang="en-US">My testing remark on disabling groupowner_chron_allow</remark>
    </select>
    ...
  </Profile>
```

And a HTML report like follows:

[...]

PR #1744, opened 2021-04-26
Added remark under a select of a profile to the HTML XCCDF report

Mab879:
This PR adds remark under a select of a profile as discussed in #1743.

Willing to fix any issues you all might find.

Big thanks to @cipherboy for pointing me to this issue.

### DIFF
--- a/xsl/xccdf-report-impl.xsl
+++ b/xsl/xccdf-report-impl.xsl
@@ -852,9 +852,13 @@ Authors:
                         </xsl:apply-templates>
                 </div></td></tr>
             </xsl:if>
-            <xsl:if test="$profile/cdf:select[@idref=$item/@id]/cdf:remark/text()">
-                <tr><td>Remark</td><td class="remark">
-                    <xsl:value-of select="$profile/cdf:select[@idref=$item/@id]/cdf:remark/text()" />
+            <xsl:if test="count($profile/cdf:select[@idref=$item/@id]/cdf:remark) > 0">
+                <tr><td>Remarks</td><td class="remarks">
+                    <xsl:for-each select="$profile/cdf:select[@idref=$item/@id]/cdf:remark">
+                        <blockquote class="small">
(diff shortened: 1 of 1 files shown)


jan-cerny:
Thank you. It works great.

I have two suggestions:
1) It would be nice to have the same feature also in HTML guide. Sometimes the people review the HTML guides before applying the profile, so they would be interested in the remarks. It would also support https://github.com/ComplianceAsCode/content/issues/6889
2) I think it can be useful to enable it also for tailoring files - ie. for profiles that are children of `<xccdf:Tailoring>`. However it won't be as easy because you will have to pass the tailoring element as another parameter to the template.

xsl/xccdf-report-impl.xsl
```diff
@@ -852,6 +852,11 @@ Authors:
                         </xsl:apply-templates>
                 </div></td></tr>
             </xsl:if>
+            <xsl:if test="$profile/cdf:select[@idref=$item/@id]/cdf:remark/text()">
```
cipherboy, line 855: Just an FYI, I was reading the standard tonight:

> The `<xccdf:select>` element MAY have one or more `<xccdf:remark>`
> elements (strings) containing explanatory material or other prose. Each
> instance of `<xccdf:remark>` MAY have an `@xml:lang` attribute (see
> Section 6.2.10).

IIUC, this doesn't handle having multiple remarks under the selected item. You'll probably need to iterate over each remark and add it to a separate section.

cipherboy (changes requested):
Sadly, according to the XCCDF spec, we'll need to support multiple `<xccdf:remark />` elements and IIUC this doesn't yet. :/

xsl/xccdf-report-impl.xsl
Mab879: This should fixed in my latest commit.

jan-cerny:
@openscap-ci test this please

PR #1749, opened 2021-05-06
Introduce OSBuild Blueprint fix type

evgenyz:


utils/oscap-xccdf.c
```diff
@@ -887,10 +888,12 @@ int app_generate_fix(const struct oscap_action *action)
 			template = "urn:xccdf:fix:script:ignition";
 		} else if (strcmp(action->fix_type, "kubernetes") == 0) {
 			template = "urn:xccdf:fix:script:kubernetes";
+		} else if (strcmp(action->fix_type, "blueprint") == 0) {
+			template = "urn:redhat:osbuild:blueprint";
 		} else {
 			fprintf(stderr,
 					"Unknown fix type '%s'.\n"
-					"Please provide one of: bash, ansible, puppet, anaconda, ignition, kubernetes.\n"
+					"Please provide one of: bash, ansible, puppet, anaconda, ignition, kubernetes, blueprint.\n"
```
jan-cerny, line 896: Please add it also to the man page.

jan-cerny (changes requested):

utils/oscap-xccdf.c
evgenyz: Done

PR #1753, opened 2021-05-11
Replace getlogin by cuserid

jan-cerny:
The getlogin() is used here to fill in the xccdf:identity element which
shall contain information about the system identity or user employed
during application of the benchmark. But, the getlogin() can return NULL
when there is no controlling terminal. This happened when testing oscap
on a test system with no pty.  As an alternative, the system provides
also cuserid() function which gets the effective user ID of the process.
However, these 2 values differ when the program is executed under sudo.
From the user experience point of view, it would be better to have
displayed there the user logged in on the controlling terminal. As a
compromise, we will first attempt to obtain the name using getlogin()
and if that fails we will run cuserid().


matusmarhefka:
@jan-cerny :bomb: thanks for the fix!

PR #1756, opened 2021-05-17
oval/yamlfilecontent: Add 'null' values handling

evgenyz:
For now null values would be represented as string '(null)' as record's field could not be attributed as nil="true" yet.


jan-cerny (approved):
It looks amazing, I have run the evaluation of the test content and I can see the strings in the results as expected.

PR #1757, opened 2021-05-26
Various HTML improvements

jan-cerny:
**This is only an idea how it can be improved.**

**The main purpose of the PR is to get feedback, collect your ideas and point to problems. Please write all your opinions.** I hope that together we can make the HTML outputs less confusing.

My feedback is:
* The big header in HTML guides is "OpenSCAP Security Guide", but in fact the content of the HTML guide doesn't come from the OpenSCAP project. The content usually comes from ComplianceAsCode or often from some third party which has no relation with OpenSCAP. This is confusing. And sometimes people when they read a HTML guide they report problems to OpenSCAP instead to the author of the SCAP content. I know and I'm very well aware that this header was added to promote OpenSCAP, but I don't think it's the best promotion, because it effectively puts our branding on top of somebody else's content. I feel that the text "Generated using OpenSCAP" in the footer is a sufficient promotion. Before you will tell me that anyone can insert his own branding in `/xsl/xccdf-branding.xsl` I would like to point out that even the biggest known user, ComplianceAsCode, doesn't do that and distrubutes the generated HTMLs everywhere with the OpenSCAP header. I propose replacing "OpenSCAP Security Guide" with the actual XCCDF Benchmark title. For example, for `ssg-rhel8-ds.xml`, that would be "Guide to the Secure Configuration of Red Hat Enterprise Linux 8".
* The profile title and description are rendered in a small grey font but in fact these are very important information. I propose making them as big as the other text.
* The HTML report shows the same introduction as the HTML guide. This actually depends on the actual wording in the SCAP content, but based on what is used in scap-security-guide I think it doesn't belong to an evaluation report, I think that evaluation report should mainly focus on results of evaluation. I propose removing that from the report.
* The visualization of XCCDF front matter element is weird. I haven't changed that in the PR yet. What are your opinions on this? Maybe to put it into a box? Or remove it completely?

### DIFF
--- a/xsl/xccdf-branding.xsl
+++ b/xsl/xccdf-branding.xsl
@@ -72,12 +72,17 @@ logo to the HTML report / guide header.
 </xsl:template>
 
 <xsl:template name="xccdf-guide-header">
+    <xsl:param name="benchmark"/>
     <nav class="navbar navbar-default">
         <div class="navbar-header" style="float: none">
             <a class="navbar-brand" href="#">
                 <xsl:call-template name="xccdf-branding-logo"/>
             </a>
-            <div><h1>OpenSCAP Security Guide</h1></div>
+            <div>
+                <h1>
+                    <xsl:value-of select="$benchmark/cdf:title/text()"/>
+                </h1>
+            </div>
         </div>
     </nav>
 </xsl:template>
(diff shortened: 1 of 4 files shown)


evgenyz:
Can you pretty please attach an example report?

evgenyz:
@Honny1 Pay attention to this WIP PR for possible improvements in `oscap-report`.

PR #1764, opened 2021-06-10
GH Actions: Add Fedora container-based job

evgenyz:


jan-cerny:
@evgenyz I can see that there are some tests skipped. I assume we need to skip the offline mode tests, but I guess the yamlfilecontent probe tests or SCE tests can be enabled. Could you take a look into this?

evgenyz:
Okay, I think that I holed this ball, we can ignore results from Jenkins

evgenyz:
Err, hold on a sec, rpm probes are skipped in F34. The hell?

jan-cerny (approved):

jan-cerny:
amazing

PR #1765, opened 2021-06-10
Do not set Rpath

jan-cerny:
See: https://docs.fedoraproject.org/en-US/packaging-guidelines/#_beware_of_rpath

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1967200

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -482,25 +482,7 @@ else()
 endif()
 set(OSCAP_TEMP_DIR "/tmp" CACHE STRING "use different temporary directory to execute sce scripts (default=/tmp)")
 
-# ---------- RPATHS for linking
 
-# see https://cmake.org/Wiki/CMake_RPATH_handling
-
-# use, i.e. don't skip the full RPATH for the build tree
-set(CMAKE_SKIP_BUILD_RPATH FALSE)
-
-# when building, don't use the install RPATH already
-# (but later on when installing)
-set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
-
-set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR})
-
-# add the automatically determined parts of the RPATH
-# which point to directories outside the build tree to the install RPATH
-set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-
-# Turn on RPATH for OSX for policy warning
-set(CMAKE_MACOSX_RPATH ON)
 # ---------- CONFIGURATION
 
 configure_file("config.h.in" "config.h")


jan-cerny:
@openscap-ci  test this please

jan-cerny:
The patch in Debian https://salsa.debian.org/philou/openscap/-/blob/debian/1.3.4-1/debian/patches/011_remove_custom_rpath.patch hints that the current upstream state isn't correct.

CMakeLists.txt
```diff
@@ -482,27 +482,6 @@ else()
 endif()
 set(OSCAP_TEMP_DIR "/tmp" CACHE STRING "use different temporary directory to execute sce scripts (default=/tmp)")
 
-# ---------- RPATHS for linking
-
-# see https://cmake.org/Wiki/CMake_RPATH_handling
-
-# use, i.e. don't skip the full RPATH for the build tree
-set(CMAKE_SKIP_BUILD_RPATH FALSE)
```
jan-cerny, line 490: When reading https://cmake.org/cmake/help/latest/variable/CMAKE_SKIP_BUILD_RPATH.html I think that we should rather set CMAKE_SKIP_BUILD_RPATH to TRUE than removing it completely because the variable doesn't have a default. What do you think?

jan-cerny:
They have pushed some temporary disablement of the RPATH check in https://src.fedoraproject.org/rpms/openscap/c/b7f22e2de4ddea05a5e38d4c7ee165caf8810b9f?branch=rawhide but I believe we should fix it for real and ideally in upstream.

CMakeLists.txt
evgenyz: I agree.

jan-cerny: I have removed the removal and replaced it by replacement of the value.

jan-cerny: But it has broken all the tests.

matusmarhefka: Hmm, it looks like it breaks the `add_oscap_test_executable` from `tests/CMakeLists.txt`.

jan-cerny: it worked locally because unfortunately it linked to the system installed openscap :scream:

jan-cerny:
I was reading the wiki again and I think that we need the default settings as described in https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling#default-rpath-settings in section Default RPATH settings:

```
By default if you don't change any RPATH related settings, CMake will
link the executables and shared libraries with full RPATH to all used
libraries in the build tree. When installing, it will clear the RPATH of
these targets so they are installed with an empty RPATH.
```

This is what we need. I think because we will have RPATH locally so that the tests and oscap executable would link to our local built libopenscap.so. And the default will remove it when installing and installation happens during rpm packaging so it should solve the problem with bugzilla and RPM build. It says  "if you don't change any RPATH related settings" which means that we can remove the block.

matusmarhefka (approved):

PR #1766, opened 2021-06-11
utils/oscap.8: fix indentation of --enforce-signature option in man page

matusmarhefka:
The `--enforce-signature` option was not indented properly in man page under `generate guide` and `generate fix`.


jan-cerny:
This is only a documentation change therefore it shouldn't affect the CI.

jan-cerny (approved):

PR #1768, opened 2021-06-16
Add additional tests for signature validation

matusmarhefka:
Add more tests for signature validation feature to cover also `generate guide` and `generate fix` XCCDF operations.


evgenyz (approved):
Test are passing, LGTM. Thanks!

PR #1769, opened 2021-06-17
Allow providing pre-downloaded components

jan-cerny:
OpenSCAP can download remote SCAP source data stream components from the
internet if the --fetch-remote-resources option is set. When this
command line option isn't set, it is possible to download the remote
component separately, save it as file to the filesystem which name is
equal to the name of the remote component and OpenSCAP would use this
local file instead of just skipping the checks. It currently works by
accident, it wasn't intended to work this way or at least I haven't
heard it before, but somobody found out that it works and documented it
as a KSC: https://access.redhat.com/solutions/5185891. However, it
produces warning messages if the approach used in the KSC is used.

This patch promotes the workaround to a feature. And it changes the logic so that it won't print the warnings.

People will be able to download their remote components and save them to a local directory and pass a path to the directory using the `--local-files <dir>` option and OpenSCAP will pick the downloaded files instead of connecting to the network.

Fixes: rhbz#1970527


jan-cerny:
@jan-cerny Please document this in the documentation. Also, consider making this feature optional, don't automatically load the file if the user doesn't request it.

jan-cerny:
> @jan-cerny Please document this in the documentation. Also, consider making this feature optional, don't automatically load the file if the user doesn't request it.

Thank you very much for your feedback. I have changed the PR. Now, it works only if the user explicitly requests it by adding the `--use-local-file` option which I have added in this PR. I have also extended the tests and I have mentioned it in the user manual.

ggbecker (changes requested):
Why does the local file need to be in the same directory as the data stream?

For example:
```
./oscap_wrapper xccdf eval --use-local-file --profile ospp ../../content/build/ssg-rhel8-ds.xml
OpenSCAP Error: Unable to open file: '../../content/build/security-data-oval-com.redhat.rhsa-RHEL8.xml' [/home/ggasparb/workspace/github/openscap/src/source/oscap_source.c:288]
Could not extract scap_org.open-scap_cref_ssg-rhel8-xccdf-1.2.xml with all dependencies from datastream. [/home/ggasparb/workspace/github/openscap/src/DS/ds_sds_session.c:228]
```
People would need to copy the local file to `/usr/share/xml/scap/ssg/content/` in order to make this work. And that doesn't seem the best approach.

I would expect that the oscap could find the file in the current directory that it is being executed. Or we could specify the filepath as a parameter.

jan-cerny:
It currently takes it as a path relative to the data stream. I will take a look into what we can do with that.

ggbecker:
> I have found that we can easily change it from relative to the datastream to relative to the current working directory. I like the idea of providing the path on the command line. But we need to find out if that should be a path to a file or to a directory and what would happen if there are multiple remote components in the data stream.

Because of potential multiple remote components in the data stream, then it makes sense to be a path to a directory.

jan-cerny:
@ggbecker

I have also another thing to consider: Currently, the file has to have the same name as the `name` attribute of `uri` element within the `catalog` element. That's a hard thing to figure out for a user. The `oscap info` command output can't help them to figure that out because in the `Referenced check files:` section it doesn't signalize which checks are remote . And the warnings shown on stderr don't show the value of the `name` attribute.

Maybe the user should have an option to replace the remote component by arbitrary file. But then, they would have to define a mapping because `oscap` needs to know files to replace the component.

ggbecker:
> @ggbecker
> 
> I have also another thing to consider: Currently, the file has to have the same name as the `name` attribute of `uri` element within the `catalog` element. That's a hard thing to figure out for a user. The `oscap info` command output can't help them to figure that out because in the `Referenced check files:` section it doesn't signalize which checks are remote . And the warnings shown on stderr don't show the value of the `name` attribute.
> 
> Maybe the user should have an option to replace the remote component by arbitrary file. But then, they would have to define a mapping because `oscap` needs to know files to replace the component.

I think oscap could print out which of the files can be used for remote content. For example, if they try to use this option (`--use-local-file`) and none of the corresponding files are found, then oscap throws an error and prints the name of the remote content files that it could accept for the given data stream. Maybe even a nice table indicating which remote web address correspond to which remote component filename. Something like:

```
You are using --use-local-file but none of the remote content files can be found locally:

Remote Address - https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL8.xml
    Expected Local File: security-data-oval-com.redhat.rhsa-RHEL8.xml
.
.
.
etc
```

Another thing is that we should probably consider other util tools to be updated, for example: https://github.com/OpenSCAP/openscap/blob/5a09d769deeeabb3b52f28013856ce049c04b2e3/utils/oscap-ssh#L55

jan-cerny:
I have tried to address the feedback.

utils/oscap-xccdf.c
```diff
@@ -171,6 +172,7 @@ static struct oscap_module XCCDF_EVAL = {
 		"                                   (only applicable for source data streams)\n"
 		"   --enforce-signature           - Process only signed data streams.\n"
 		"   --fetch-remote-resources      - Download remote content referenced by XCCDF.\n"
+		"   --local-files <dir>       - Use locally downloaded copies of remote resources stored in the given directory.\n"
```
ggbecker, line 175: It's missing some spaces/tabs here. Same as below.

utils/oscap.8
```diff
@@ -72,6 +72,11 @@ For XCCDF or SCAP source data stream files, the info module prints out IDs of in
 Allow download of remote components referenced from data stream.
 .RE
 .TP
+\fB\-\-local-files DIRECTORY\fR
+.RS
+Instead of downloading remote data stream components from the network, use data stream components stored locally as files in the given directory. In place of the remote data stream component OpenSCAP will attempt to use a file whose file name is equal to @name attribute of the uri element within the catalog element within the component-ref element in the data stream if such file exists.
```
ggbecker, line 77:
```suggestion
Instead of downloading remote data stream components from the network, use a data stream components stored locally as files in the given directory. In place of the remote data stream component OpenSCAP will attempt to use a file whose file name is equal to @name attribute of the uri element within the catalog element within the component-ref element in the data stream if such file exists.
```

utils/oscap.8
```diff
@@ -201,6 +206,11 @@ Process only digitally signed SCAP source data streams. Data streams without a s
 Allow download of remote OVAL content referenced from XCCDF by check-content-ref/@href.
 .RE
 .TP
+\fB\-\-local-files DIRECTORY\fR
+.RS
+Instead of downloading remote data stream components from the network, use data stream components stored locally as files in the given directory. In place of the remote data stream component OpenSCAP will attempt to use a file whose file name is equal to @name attribute of the uri element within the catalog element within the component-ref element in the data stream if such file exists.
```
ggbecker, line 211:
```suggestion
Instead of downloading remote data stream components from the network, use a data stream components stored locally as files in the given directory. In place of the remote data stream component OpenSCAP will attempt to use a file whose file name is equal to @name attribute of the uri element within the catalog element within the component-ref element in the data stream if such file exists.
```

utils/oscap.8
```diff
@@ -303,6 +318,11 @@ Select a particular profile from XCCDF document.
 Allow download of remote OVAL content referenced from XCCDF by check-content-ref/@href.
 .RE
 .TP
+\fB\-\-local-files DIRECTORY\fR
+.RS
+Instead of downloading remote data stream components from the network, use a data stream component stored locally in a file. In place of the remote data stream component OpenSCAP will attempt to use a file whose file name is equal to @name attribute of the uri element within the catalog element within the component-ref element in the data stream if such file exists.
```
ggbecker, line 323:
```suggestion
Instead of downloading remote data stream components from the network, use data stream component stored locally in a file. In place of the remote data stream component OpenSCAP will attempt to use a file whose file name is equal to @name attribute of the uri element within the catalog element within the component-ref element in the data stream if such file exists.
```

ggbecker, line 323:
```suggestion
Instead of downloading remote data stream components from the network, use data stream components stored locally as files in the given directory. In place of the remote data stream component OpenSCAP will attempt to use a file whose file name is equal to @name attribute of the uri element within the catalog element within the component-ref element in the data stream if such file exists.
```

utils/oscap.8
```diff
@@ -477,6 +497,9 @@ Do not validate input/output files.
 .TP
 \fB\-\-fetch-remote-resources\fR
 Allow download of remote components referenced from data stream.
+.TP
+\fB\-\-local-files DIRECTORY\fR
+Instead of downloading remote data stream components from the network, use a data stream component stored locally in a file. In place of the remote data stream component OpenSCAP will attempt to use a file whose file name is equal to @name attribute of the uri element within the catalog element within the component-ref element in the data stream if such file exists.
```
ggbecker, line 502:
```suggestion
Instead of downloading remote data stream components from the network, use data stream components stored locally as files in the given directory. In place of the remote data stream component OpenSCAP will attempt to use a file whose file name is equal to @name attribute of the uri element within the catalog element within the component-ref element in the data stream if such file exists.
```

utils/oscap.8
```diff
@@ -638,6 +661,9 @@ Do not validate input/output files.
...
```
ggbecker, line 666:
```suggestion
Instead of downloading remote data stream components from the network, use data stream components stored locally as files in the given directory. In place of the remote data stream component OpenSCAP will attempt to use a file whose file name is equal to @name attribute of the uri element within the catalog element within the component-ref element in the data stream if such file exists.
```

ggbecker (changes requested):
Few documentation items need to be addressed first. But overall LGTM.

jan-cerny:
updated again with the man page and whitespace problems

ggbecker (commented):
I tested oscap-ssh and unfortunately you cannot use `./` and `../` as a directory since `ssh` doesn't work with that but using a folder name works perfectly.

ggbecker (approved):
LGTM

PR #1773, opened 2021-06-25
Workaround rpminspect problem

jan-cerny:
rpminspect produces this problem:

```
xml-files:
----------
1) File /usr/share/openscap/xsl/oval-results-report.xsl is a malformed XML file on x86_64
Result: VERIFY
Waiver Authorization: Anyone

Details:
No declaration for element stylesheet

Suggested Remedy: Correct the reported errors in the XML document
```

I assume that it's caused by mixing the DTD and schema - it probably
expects that the DTD will contain a declaration of the root element
as well. The workaround simply expands both entities by substituting
them by their contents.

### DIFF
--- a/xsl/oval-results-report.xsl
+++ b/xsl/oval-results-report.xsl
@@ -1,10 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE xsl:stylesheet [
-<!-- check symbol -->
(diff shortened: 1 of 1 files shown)


matejak:
LGTM, although the PR effectively disconnects usage of those symbols, the damage is not big.

matejak (approved):

PR #1775, opened 2021-06-29
Fix failing test

jan-cerny:
The test fails becuse the OVAL content in
`test_remediation_simple.oval.xml` used in rule
`xccdf_moc.elpmaxe.www_rule_1` in
`test_profile_selection_by_suffix.xccdf.xml` expects that a file named
`test_file` exists in the current working directory.

This test doesn't fail when executed as a part of complete test suite
run. I guess that it's because some other test creates the `test_file`
file and doesn't delete it. Unfortunately, I can't find which test
creates it. There are many test cases that use a file `test_file`
and it is also created often by remediation executed in some tests.

### DIFF
--- a/tests/API/XCCDF/unittests/test_profile_selection_by_suffix.sh
+++ b/tests/API/XCCDF/unittests/test_profile_selection_by_suffix.sh
@@ -13,6 +13,9 @@ echo "Stderr file = $stderr"
 echo "Result file = $result"
 ret=0
 
+touch test_file
+[ -f test_file ]
+
 # Multiple matches should result in failure
 $OSCAP xccdf eval --profile common $benchmark 2> $stderr || ret=$?
 [ $ret -eq 1 ]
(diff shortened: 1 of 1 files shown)


jan-cerny:
@matusmarhefka @evgenyz I think that the solution with using a file with the same name (`test_file`) in many tests is error prone. For example, some test fails so it terminates before cleaning the `test_file` which will influence some other test executed later that expects that `test_file` doesn't exist. Maybe when the tests that use `test_file` would run in parallel they could interfere and cause "random fails". Perhaps we should leave this approach and use a unique temporary file different in each test. What is your opinion?

evgenyz:
It would require templating all test content files. I'm not opposed to the idea, but I'd prefer to have a function or something to unify this if possible.

matusmarhefka:
@jan-cerny I agree, it would be best if we can have some shared function which can be used in every test to replace templated file names in our test content.

matusmarhefka (approved):

PR #1779, opened 2021-07-08
Initialize crypto API only once

jan-cerny:
The function `crapi_init` calls `gcry_check_version` which must be
called before any other function from the Libgcrypt library. That might
be violated when multiple threads executing multiple probes are running.
The mitigation proposed in this PR is to call `crapi_init` only once
when the session is initialized which means before any threads are
spawned.

See also: https://www.gnupg.org/documentation/manuals/gcrypt/Multi_002dThreading.html#Multi_002dThreading

Resolves: RHBZ#1959570

### DIFF
--- a/src/OVAL/oval_probe_session.c
+++ b/src/OVAL/oval_probe_session.c
@@ -93,6 +93,11 @@ static void oval_probe_session_libinit(void)
 	SEXP_free((SEXP_t *)exp);
 
         ncache_libinit();
+	/*
+	 * Initialize crypto API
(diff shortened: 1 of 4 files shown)


jan-cerny:
so this time tests/API/XCCDF/unittests/test_remediate_simple.sh passed therefore which means that it  fails unreliably

jan-cerny:
rebased on maint-1.3

evgenyz (approved):
LGTM, thanks!

PR #1781, opened 2021-07-15
Allow disablement of SHA-1 and MD5

jan-cerny:
This patch adds 2 new CMake options `OPENSCAP_ENABLE_SHA1` and
`OPENSCAP_ENABLE_MD5`.  By setting these CMake options to `OFF` we can
disable SHA-1 and MD5 hashing algorithm in OpenSCAP. If they are set to
`OFF`, the code using SHA-1 and MD5 won't be built.  Linux distributions
that wish to disable SHA-1 and MD5 can use these CMake options in their
spec files from now on.

Conditional compilation using C preprocessor macros is used to achieve
it.  The approach with conditional compilation has been preferred over
the hard removal of the code because future versions of OpenSCAP can get
into currently stable Linux distributions where a removal would have to
be reverted by a downstream patch to avoid feature regression.

The consequence of disabling MD5 and SHA-1 is that filehash probe will
not be built but filehash_test is a deprecated OVAL test.  Second
consequence is that people won't be able to use SHA-1 and MD5 in their
filehash58_test. That means people will have to rewrite their SCAP
content to use SHA-256 or SHA-512 in their filehash58_objects.

Resolves: rhbz#1936619


tests/probes/filehash58/test_probes_filehash58.sh
```diff
@@ -96,7 +96,9 @@ function test_probes_filehash58_chroot_fail {
 
 test_init
 
-test_run "test_probes_filehash58" test_probes_filehash58
+if [[ "$OPENSCAP_ENABLE_MD5" == "ON" && "$OPENSCAP_ENABLE_SHA1" == "ON" ]] ; then
+    test_run "test_probes_filehash58" test_probes_filehash58
+fi
```
jan-cerny, line 101: @jan-cerny I assume you should implement a new test for the probe that would test the other algos.

tests/API/crypt/test_crapi_mdigest.c
```diff
@@ -24,6 +24,8 @@
 #include <config.h>
 #endif
 
+#if (defined OPENSCAP_ENABLE_MD5 && defined OPENSCAP_ENABLE_SHA1)
+
```
jan-cerny, line 28: What is your opinion on the test `test_crapi_mdigest`?

CMakeLists.txt
```diff
@@ -251,11 +251,14 @@ if (WIN32)
 endif()
 cmake_dependent_option(ENABLE_PROBES_WINDOWS "build OVAL probes for the Windows OVAL tests" ${IS_WIN32} "ENABLE_PROBES" OFF)
 
+option(OPENSCAP_ENABLE_SHA1 "Enable using the SHA-1 algorithm" ON)
+option(OPENSCAP_ENABLE_MD5 "Enable using the MD5 algorithm" ON)
+
```
Jakuje, line 256: my only concern would be the defaults here. While SHA1 still might be considered usable, I would argue that the MD5 should not be enabled by default.

jan-cerny: The impact of such change is unclear for me. Currently in upstream both SHA1 and MD5 are enabled. The question is their usage in existing SCAP content. The https://github.com/ComplianceAsCode/content/ project uses file hashing only in 2 rules (etcd_unique_ca and openssl_use_strong_entropy) and in both of them they use SHA-256. But I have no idea about third-party providers of content. They might be using these old algorithms. So I rather don't change current default behavior if I don't know the impact to the users. Can anybody shed a light on this?

evgenyz: I'd say that this is not a change to consider for patch version. Even maybe not for minor (because we would kick out deprecated but still legit probe, from SCAP standard PoV).

evgenyz: Anyhow, this is the vanilla build flags, all distros are free to break compatibility as they wish

tests/probes/filehash58/test_probes_filehash58.sh
jan-cerny: > @jan-cerny I assume you should implement a new test for the probe that would test the other algos.

See the new commit

tests/API/crypt/test_crapi_digest.c
```diff
@@ -58,88 +58,69 @@ static int mem2hex (uint8_t *mem, size_t mlen, char *str, size_t slen)
...
-                abort ();
-        }
+	uint8_t *dst = malloc(dstlen);
 
-        if (lseek (fd, 0, SEEK_SET) == (off_t)-1) {
-                perror ("lseek");
-                return (2);
-        }
+		size_t comp_sum_len = (dstlen * 2) + 1;
+		char *comp_sum = malloc(comp_sum_len);
```
jan-cerny, line 109: @jan-cerny these 2 lines should be indented only by a 1 tab

yuumasato (commented):
The code changes look sane, and I was able to use the scanner without SHA-1 and MD-5.

But I noticed that when I scan using a content with `filehash58_test` using SHA-1 the result is just `fail`. This can be confusing, there is no warning that the algorithm is not supported. A result of `notchecked` seems more accurate to me.

Also, would it be possible to include something in the oval and/or xccdf results stating the support for the requested algorithm was not built? Or something else that would give a hint about why the `filehash58_test` is not working

The fact that the scanner still accepts/expects SHA-1 can contribute to the confusion.
I added a typo in the test and I see this:
```shell
$ ./oscap_wrapper oval eval --results ./oval-results.xml ../tests/probes/filehash58/check_filehash_simple.xml 
File '../tests/probes/filehash58/check_filehash_simple.xml' line 36: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#independent}hash_type': [facet 'enumeration'] The value 'SHA1' is not an element of the set {'MD5', 'SHA-1', 'SHA-224', 'SHA-256', 'SHA-384', 'SHA-512', ''}.
OpenSCAP Error: Invalid OVAL Definition (5.11) content in ../tests/probes/filehash58/check_filehash_simple.xml. [/home/wsato/git/openscap/src/source/oscap_source.c:353]
```

Another point is how do I know if the probe was compiled with support for SHA-1 or MD-5?
Could we add some new section in `oscap -V`? Either a new section or more detail in the probe line, for example:
`independent   filehash58                   probe_filehash58 (no SHA-1, no MD-5)`

CMakeLists.txt
evgenyz: Here is the reminder: #1784

evgenyz:
>  `independent filehash58 probe_filehash58 (no SHA-1, no MD-5)`

I would rather go with `independent filehash58 probe_filehash58 (SHA-224, SHA-256, SHA-384, SHA-512)` if possible.

evgenyz:
> But I noticed that when I scan using a content with `filehash58_test` using SHA-1 the result is just `fail`. This can be confusing, there is no warning that the algorithm is not supported. A result of `notchecked` seems more accurate to me.

+1 here.

jan-cerny:
@yuumasato @evgenyz I have updated it according to the review

yuumasato (approved):
Works great for me.
Code changes look ok, but I defer to @evgenyz for extra review.

`oscap -V` shows the supported algorithms:
`independent   filehash58                   probe_filehash58 (MD5, SHA-224, SHA-256, SHA-384, SHA-512)`

Scanning with content with not supported algorithm results in `error` and a warning.
```
Title
	OpenSSL uses strong entropy source
Rule
	xccdf_org.ssgproject.content_rule_openssl_use_strong_entropy
Ident
	CCE-82721-2
W: oscap:     This version of OpenSCAP doesn't support the 'SHA-1' hash algorithm.
Result
	error
```
report_filehash58_unsupported_algorithm

evgenyz:
LGTM, thank you @jan-cerny!

PR #1782, opened 2021-07-20
Add offline mode for password probe

jan-cerny:
This will allow us to use OVAL password_test in offline mode.

Fixes: #1783


evgenyz:
LGTM, thanks!

evgenyz (approved):

PR #1785, opened 2021-07-27
Fix covscan issues

jan-cerny:
Addressing some covscan issues in RHEL9. See the commit messages for details.

### DIFF
--- a/tests/API/crypt/test_crapi_digest.c
+++ b/tests/API/crypt/test_crapi_digest.c
@@ -40,6 +40,10 @@ static int mem2hex (uint8_t *mem, size_t mlen, char *str, size_t slen)
 {
         const char ch[] = "0123456789abcdef";
         register size_t i;
+
+	if (str == NULL) {
+		return -1;
+	}
         
         if (slen < (mlen * 2) + 1) {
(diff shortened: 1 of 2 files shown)


tests/API/crypt/test_crapi_digest.c
```diff
@@ -40,6 +40,10 @@ static int mem2hex (uint8_t *mem, size_t mlen, char *str, size_t slen)
 {
         const char ch[] = "0123456789abcdef";
         register size_t i;
+
+	if (str == NULL) {
```
evgenyz, line 44: Can we please have consistent indentation here?

jan-cerny: I can't find which line is indented wrongly. I can see that all 3 indented lines are indented by tabs.

evgenyz: Yeah, but the code around them is indented with spaces. You can see it even in the GH diff. It is really painful to skim through such a code. As far as I remember we agreed on keeping the indentation style of the original code if changes are small.

evgenyz (approved):
Nice! Thank you!

PR #1788, opened 2021-07-28
Add a regression test for rhbz#1959570

jan-cerny:
The bug was a segmentation fault in filehash58 probe which happened
in openscap-1.3.3-6.el8_3.

The bug was fixed by https://github.com/OpenSCAP/openscap/pull/1779
and this patch adds a very small test.


jan-cerny:
the fail is a random segfault in API/XCCDF/variable_instance/all.sh, it's unrelated to this PR

ggbecker (approved):
LGTM

PR #1799, opened 2021-08-17
Fix 1798

DoctorPepper12:


evgenyz:
Hi, @DoctorPepper12. We are still considering your approach for probes "pluginization", but here is the list of red flags we currently see:
- It's windows only (though it's more of an orange flag, we might conclude that is it not very interesting feature for *nix, jury is still out on this one);
- A lot of functions suddenly became public API, like `oscap_string_to_enum`, we would prefer not to make such a commitment;
- This feature has no test coverage whatsoever.

Also, I would kindly ask you to properly re-base the PR against `maint-1.3` and squash appropriate parts of your initial development history, there is no benefit in it for the project. 

Please try to adhere to the https://github.com/OpenSCAP/openscap/blob/maint-1.3/docs/contribute/contribute.adoc, especially in terms of code style (we try to preserve existing code style if we are making small changes for the sake of code readability).

evgenyz:
I don't quite get the purpose of `plugin/OVAL/probes/windows/file_probe`, it's just a dummy. Maybe it is an experimental piece of code slipped into the PR? It should be removed then.

jan-cerny:
Closing because of inactivity.

PR #1800, opened 2021-08-19
Compile "environmentvariable_probe" on Windows

OSCAPC:
By default, the independent "environmentvariable_probe" is not compiled on
Windows Systems. By deleting the NOTWIN32 variable in the CMakeLists.txt,
we can now compile and use the independent "environmentvariable_probe" on Windows.

Configuration used :
- Dev OS : Windows 10
- Tested OS : Windows 10
- OVAL file test : ind-def_environmentvariable_test.xml

Exemples of environmentvariable_item  obtained :

```
          <ind-sys:environmentvariable_item id="11542837" status="exists">
            <ind-sys:name>TMP</ind-sys:name>
            <ind-sys:value>C:\Users\User\AppData\Local\Temp</ind-sys:value>
          </ind-sys:environmentvariable_item>
          <ind-sys:environmentvariable_item id="11542836" status="exists">
            <ind-sys:name>TEMP</ind-sys:name>
            <ind-sys:value>C:\Users\User\AppData\Local\Temp</ind-sys:value>
          </ind-sys:environmentvariable_item>
          <ind-sys:environmentvariable_item id="11542835" status="exists">
            <ind-sys:name>SystemRoot</ind-sys:name>
            <ind-sys:value>C:\Windows</ind-sys:value>
          </ind-sys:environmentvariable_item>
```


evgenyz:
Huh? @jan-cerny any idea why it was disabled on Windows?

jan-cerny:
@evgenyz It was disabled on Windows because we had the following approach:  We first disabled everything and we expected that when people will evaluate or modify the probe to compile on Windows they would submit a PR with a patch enabling the probe.

CMakeLists.txt
```diff
@@ -255,7 +255,7 @@ option(OPENSCAP_ENABLE_SHA1 "Enable using the SHA-1 algorithm" ON)
 option(OPENSCAP_ENABLE_MD5 "Enable using the MD5 algorithm" ON)
 
 # INDEPENDENT PROBES
-cmake_dependent_option(OPENSCAP_PROBE_INDEPENDENT_ENVIRONMENTVARIABLE "Independent environmentvariable probe" ON "ENABLE_PROBES_INDEPENDENT; NOT WIN32" OFF)
```
jan-cerny, line 258: Amazing!

Please move it away from the list of `Independent probes incompatible with WIN32`.

OSCAPC: @jan-cerny  Thank you for reviewing the code. Fixed in commit  : https://github.com/OSCAPC/openscap/tree/01143823c9870a8b5b11dd427339ca0ce1c9a04a .

evgenyz:
Well, it's all nice and cool, but `environmentvariable58_test` is
`Deprecated As Of Version: 5.8
Reason: Replaced by the environmentvariable58_test.
Comment: This object has been deprecated and may be removed in a future version of the language.`

Can we do something about 5.8 version of this probe as well?

jan-cerny:
> Well, it's all nice and cool, but `environmentvariable58_test` is
> `Deprecated As Of Version: 5.8 Reason: Replaced by the environmentvariable58_test. Comment: This object has been deprecated and may be removed in a future version of the language.`
> 
> Can we do something about 5.8 version of this probe as well?

We can, that will be amazing, but I consider it to be out of scope of this PR.

The difference between environmentvariable_test and environmentvariable58_test is that the former collects the current process's environment whereas the latter allows you to collect environment of a given process specified by a PID. As a consequence, the implementation of environmentvariable_test is different than environmentvariable58_test. The former reads the platform-independent global variable `environ` (https://docs.microsoft.com/en-us/cpp/c-runtime-library/environ-wenviron?view=msvc-160 and https://man7.org/linux/man-pages/man7/environ.7.html). On the other nahd the latter reads from `/proc` which is specific for Unix. So you would have to write a windows code that would be analogous to that to make the environemtnvariable58 probe ready for Windows.

evgenyz:
Fair enough. Okay.

evgenyz (approved):

PR #1801, opened 2021-08-23
Add Service Probe for Windows

OSCAPC:
Creation of a new probe for OVAL service_test for Windows systems.
The probe use Windows API to reach services information.

Configuration used :
- Dev OS : Windows 10 Pro 21H1
- Tested OS : Windows 10 Pro 21H1

I can provide an OVAL test file for this probe. What is the best manner to do it ? gist, pull request, ...


evgenyz:
Would you be interested in creating a CTest-based test suite (for probes) for Windows? You can get some inspiration from existing *NIX test sute, but I doubt it would be meaningful to try and make it portable.

OSCAPC:
Hi, I never created CTest-based test suite. Have you got some example for Windows ? If you got some, I can try to create one for the probe, else, I'll need more explanation on how CTest-based test suite works.

evgenyz:
CTest is part of CMake and it is really simple (https://gitlab.kitware.com/cmake/community/-/wikis/doc/ctest/Testing-With-CTest). 

We have the implementation for *NIX OSes (tests subdir), the initialization is here: https://github.com/OpenSCAP/openscap/blob/2f27ffca9e042ed4399bcf3c618b48004993ab48/CMakeLists.txt#L332.  It is turned off for Windows because the whole test suite is based on linux-specific tools and commands. And I would not ask anyone to make it portable or recreate using Windows' tools. It is OK that platform-independent code is checked only on Linux (better to check everywhere, but the effort if too big).

But! It is important to be able to test Windows-only things, like Windows probes. I'm pretty much OK with merging two your PRs as they are, because they won't affect Linux version in any way. But you would be the one who will suffer when somebody would accidentaly ruin these probes without knowing it.

OSCAPC:
I'm sorry to answer this late, I can't work on the project anymore. You can use my work as you wish. Thank you for your time.

marcusburghardt:
@evgenyz and @jan-cerny  , assuming the tests were not created and the author can't work on it anymore, would you agree to close this PR and the other, also related?

evgenyz:

PR #1803, opened 2021-08-31
Lower memory limits and improve their checking

jan-cerny:
This patch attempts to mitigate problems caused by a large amount of
collected objects such as rhbz#1932833.

Specifically, these changes are made:
- Lower the threshold so that the amount of used memory is checked when
  only 1000 items are collected for the given OVAL object. That's
  because 32768 items (the original value) is already a large amount which
  occupies a lot of memory during further processing.
- Lower the memory usage ratio limit for the probe to 10 %. We have
  found experimentally that giving the probe 15 % or more will cause the
  oscap process to be killed when processing the collected data and
  generating results.
- In the calling function probe_item_collect, distinguish between return
  codes which means different behavior when there is insufficient memory
  than when the memory consumption can't be checked.
- Improve the warning message to show greater details about memory
  consumption to the user.
- Remove the check for the absolute amount of remaining free memory. As
  we can see on the example of rhbz#1932833, on systems with large
  amount of memory the remaining memory of 512 MB isn't enough memory for
  openscap to process the collected data. At the same time, if we lowered
  the usage ratio, we don't need this anymore.
- Remove useless message "spt:" from the verbose log because it's
  produced many times and pollutes the log extremely.

### DIFF
--- a/src/OVAL/probes/probe/worker.c
+++ b/src/OVAL/probes/probe/worker.c
@@ -37,6 +37,10 @@
 
 #include "worker.h"
 
+/* default max. memory usage ratio - used/total */
+/* can be overridden by environment variable OSCAP_PROBE_MEMORY_USAGE_RATIO */
+#define OSCAP_PROBE_MEMORY_USAGE_RATIO_DEFAULT 0.1
+
 extern bool  OSCAP_GSYM(varref_handling);
 extern void *OSCAP_GSYM(probe_arg);
 
@@ -923,6 +927,14 @@ SEXP_t *probe_worker(probe_t *probe, SEAP_msg_t *msg_in, int *ret)
 		SEXP_t *varrefs, *mask;
 
 		pctx.offline_mode = probe->selected_offline_mode;
+
+		pctx.max_mem_ratio = OSCAP_PROBE_MEMORY_USAGE_RATIO_DEFAULT;
+		char *max_ratio_str = getenv("OSCAP_PROBE_MEMORY_USAGE_RATIO");
+		if (max_ratio_str != NULL) {
+			double max_ratio = strtod(max_ratio_str, NULL);
+			if (max_ratio != 0)
+				pctx.max_mem_ratio = max_ratio;
+		}
 
 		/* simple object */
                 pctx.icache  = probe->icache;
--- a/src/OVAL/probes/probe/worker.c
+++ b/src/OVAL/probes/probe/worker.c
@@ -932,7 +932,7 @@ SEXP_t *probe_worker(probe_t *probe, SEAP_msg_t *msg_in, int *ret)
 		char *max_ratio_str = getenv("OSCAP_PROBE_MEMORY_USAGE_RATIO");
 		if (max_ratio_str != NULL) {
 			double max_ratio = strtod(max_ratio_str, NULL);
-			if (max_ratio != 0)
+			if (max_ratio > 0)
 				pctx.max_mem_ratio = max_ratio;
 		}
 
--- a/docs/manual/manual.adoc
+++ b/docs/manual/manual.adoc
@@ -1925,6 +1925,7 @@ behaviour.
 
 * *OSCAP_FULL_VALIDATION=1* - validate all exported documents (slower)
 * *SEXP_VALIDATE_DISABLE=1* - do not validate SEXP expressions (faster)
+* *OSCAP_PROBE_MEMORY_USAGE_RATIO* - maximum memory usage ratio (used/total) for OpenSCAP probes, default: 0.1
 
 
 
--- a/src/OVAL/probes/probe/icache.c
+++ b/src/OVAL/probes/probe/icache.c
@@ -457,9 +457,8 @@ int probe_icache_nop(probe_icache_t *cache)
         return (0);
 }
 
-#define PROBE_RESULT_MEMCHECK_CTRESHOLD  32768  /* item count */
-#define PROBE_RESULT_MEMCHECK_MINFREEMEM 512    /* MiB */
-#define PROBE_RESULT_MEMCHECK_MAXRATIO   0.8   /* max. memory usage ratio - used/total */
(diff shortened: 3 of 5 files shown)


jan-cerny:
When I think of it again, this proposal is "very far from being ideal".

The limit 10 % of memory usage ratio is applicable only in certain situations. I was focused on the situation when the file probe collects a large amount of files, which is the situation happening in the respective BZ (rhbz#1932833). On a RHEL 7.9 system with 8 GB of RAM this seems to be ideal for the file probe. But, on a system with only 2 GB of RAM it cuts off right after it reaches 1000 collected items, with a ratio 28 %. That means that on systems with a small amount of memory it stops sooner that needed and on systems with a very large amount of memory it stops later than needed.

I'm afraid of unwanted consequences such as producing incomplete results in certain special configuraions.

The memory limit mechanism is effective for all probes. But, different OVAL items have different size. For example, the `textfilecontent54 _item` contains matched text from text files which can be arbitrary size. Also, different probes allocate different amout of memory for their operation.

Moreover, the amount of memory needed for processing the collected data doesn't depend on the system memory size but on the amount of the data to be processed.

So we have multiple options:

- go with the proposal in this PR
- make it simple and stop after collecting certain (fixed) amount of items, the threshold would be hardcoded
- stop after collecting certain amount of items where the threshold would be computed from the available memory size
- measure the size of collected data (in bytes) and stop after this reaches some threshold

evgenyz:
@jan-cerny Let's concentrate on the goal we have here. So, if we are scanning the system where we can't collect (and properly represent) all the objects for a certain rule, what would be the best thing we can do? 

Processing and storing partial result for the rule won't be much of a help to determine what we *were not* able to collect and it definitely won't be helpful for the whole evaluation.

The natural course of thinking brings us to the conclusion that if we can't properly process current rule's result we can just drop all the data and mark the rule as "not processed" and move on, doing that with all further rules that "won't fit". I'm not sure about the ratio between internal and xml representation, but if it is somewhere around 1:2 then we could stick with 30% of now-available-memory for every rule (30internal +60xml + 10extra_margin). In case the rule would fail to fit, we'll just drop everything and try to collect data for other rules.

evgenyz:
@jan-cerny After reading your comments in the PR and the code again I think that we won't be able to do much here anyway. The limits are sane enough. 

The only thing we can try and adjust is the free RAM threshold. You said that when there is more than 15% of RAM allowed to use by the probe OSCAP crashed on processing. But what happens in 
> But, on a system with only 2 GB of RAM it cuts off right after it reaches 1000 collected items, with a ratio 28 %.
scenario?

evgenyz:
Anyhow, we can introduce an env. variable for `PROBE_RESULT_MEMCHECK_MAXRATIO` for cases where our default value would be sub-optimal. That could give a user some room for maneuvering.

jan-cerny:
> @jan-cerny Let's concentrate on the goal we have here. So, if we are scanning the system where we can't collect (and properly represent) all the objects for a certain rule, what would be the best thing we can do?

We definitely want to avoid the situation of being killed. We want to be able to produce report and result files. For the rules where we can't collect all the objects we want to signalize that in the results file.

> 
> Processing and storing partial result for the rule won't be much of a help to determine what we _were not_ able to collect and it definitely won't be helpful for the whole evaluation.

Yes, there are rules for which we can't determine the results if we don't have all the data. But, there are some rules where we can compute the results. For example, if we evaluate a rule that requires that at least one object must exist, once we find a single object we can say that we passed and collecting other matching objects won't  affect the rule result.

> 
> The natural course of thinking brings us to the conclusion that if we can't properly process current rule's result we can just drop all the data and mark the rule as "not processed" and move on, doing that with all further rules that "won't fit". I'm not sure about the ratio between internal and xml representation, but if it is somewhere around 1:2 then we could stick with 30% of now-available-memory for every rule (30internal +60xml + 10extra_margin). In case the rule would fail to fit, we'll just drop everything and try to collect data for other rules.

This sounds reasonable, but I don't have data about the ratio between internal and xml representation.

jan-cerny:
> @jan-cerny After reading your comments in the PR and the code again I think that we won't be able to do much here anyway. The limits are sane enough.
> 
> The only thing we can try and adjust is the free RAM threshold. You said that when there is more than 15% of RAM allowed to use by the probe OSCAP crashed on processing. But what happens in
> 
> > But, on a system with only 2 GB of RAM it cuts off right after it reaches 1000 collected items, with a ratio 28 %.
> 
> scenario?

we get:

```
W: probe_file: Memory usage ratio limit reached! limit=0.100000, current=0.312033, used=573 MB, free=131 MB, total=1837 MB, count of items=1001
Result  fail
```

and then it normally continues and finishes successfully.

jan-cerny:
This is now improved by introducing environment variable OSCAP_PROBE_MEMORY_USAGE_RATIO which allows users to customize the ratio threshold.

evgenyz:
LGTM. Are you done with it?

src/OVAL/probes/probe/worker.c
```diff
@@ -924,6 +928,14 @@ SEXP_t *probe_worker(probe_t *probe, SEAP_msg_t *msg_in, int *ret)
 
 		pctx.offline_mode = probe->selected_offline_mode;
 
+		pctx.max_mem_ratio = OSCAP_PROBE_MEMORY_USAGE_RATIO_DEFAULT;
+		char *max_ratio_str = getenv("OSCAP_PROBE_MEMORY_USAGE_RATIO");
+		if (max_ratio_str != NULL) {
+			double max_ratio = strtod(max_ratio_str, NULL);
+			if (max_ratio != 0)
```
evgenyz, line 935: Maybe
```suggestion
			if (max_ratio > 0)
```

jan-cerny: great because a negative value wouldn't make sense

evgenyz:
Nice!

evgenyz (approved):

PR #1805, opened 2021-09-10
Fail download on HTTP errors

jan-cerny:
When the HTTP server returns status code greater than or equal 400,
the download will fail.

Resolves: rhbz#2002733


evgenyz (approved):
LGTM, thank you!

PR #1806, opened 2021-09-10
Add an alternative source of hostname

jan-cerny:
If /etc/hostname can't be read, we will try to open /proc/sys/kernel/hostname instead.

Resolves: rhbz#1977668


evgenyz (approved):
Sure, why not? LGTM, thanks!

PR #1808, opened 2021-09-17
Change some errors to warnings

jan-cerny:
A fail during the conversion of integer or string might be expected in
some cases.  There can exist a valid OVAL check which tries to compare
integer with a string and yet ends up evaluated as true.  If an error is
produced in such situation it leads to user confusion.  We can make it
better by promoting this to a warning.

The commit also adds a small test which is based OVAL OVAL check for
rule accounts_passwords_pam_faillock_unlock_time from Fedora SSG.

Resolves: rhbz#1945580


evgenyz (approved):
LGTM

Issue #1819, opened 2021-10-27
xpath missing in during install on Debian 11 bullseye


egberts:
#### Description of Problem:

attempted source install on Debian 11 (bullseye) using `git clone` from here.

#### OpenSCAP Version:

default branch is refs/heads/maint-1.3

#### Operating System & Version:

Debian 11 (bullseye)

#### Steps to Reproduce:

1.  Follow install instructions for source-based cmake.
2. run ctest -V
2. 
3. 

#### Actual Results:

`I require xpath tool but it is not installed. Aborting.`

All of some 29 test failed.

#### Expected Results:

pass at least 1 test.

#### Additional Information / Debugging Steps:

Debian 11 workaround is to run:

    apt install libxml-xpath-perl

Tested, passed all tested, reesolved.

evgenyz:
Err, `libxml-xpath-perl` or `perl-XML-XPath` in case of Fedora is the build-time or devel dependency. It is used in tests, which are obviously not part of the binary package. If you want to develop openscap (and run tests), install build-time dependencies.

PR #1820, opened 2021-10-27
add missing package for Debian 11

egberts:
Issue #1819

### DIFF
--- a/docs/developer/developer.adoc
+++ b/docs/developer/developer.adoc
@@ -89,6 +89,12 @@ libxml-xpath-perl libperl-dev libbz2-dev librpm-dev g++ libapt-pkg-dev libyaml-d
 libxmlsec1-dev libxmlsec1-openssl
 ----
 
+On Debian 11 (bullseye), in addition to Debian 10 above, additional build dependencies are:
+
+----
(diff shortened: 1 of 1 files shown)


evgenyz:
I don't get it. First of all, there is no Debian 10 "above". And even if it was there, does it (and also all Ubuntus and other Debians in the group) cover the dependency in a different way?

evgenyz:
And finally, any particular reason to have #1821 on top of it?

jan-cerny:
> I don't get it. First of all, there is no Debian 10 "above". And even if it was there, does it (and also all Ubuntus and other Debians in the group) cover the dependency in a different way?

Moreover, the libxml-xpath-perl is already listed in the list above.

egberts:
well, when i attempted to do a pull request, Github automatically created another issue.

egberts:
shall i redo the pull request by including all of Debian 10 dependent package names to the Debian 11 list ... instead?

evgenyz:
> shall i redo the pull request by including all of Debian 10 dependent package names to the Debian 11 list ... instead?

How about adding Debian 11 to the list of other Debians instead of creating a new section? I'm still not sure what you are trying to achieve.

PR #1826, opened 2021-11-09
Print warning for local files

jan-cerny:
This will explicitely display users that they're using local
files instead of the remote resource.

See https://bugzilla.redhat.com/show_bug.cgi?id=1970529#c6


matusmarhefka:
LGTM.

matusmarhefka (approved):

PR #1832, opened 2021-11-25
Select rules on the command line

jan-cerny:
We introduce some changes to let users modify the list of rules to be evaluated more easily by command line options. Often users want to skip some rules or evaluate just a couple of rules. Currently, they need to create a tailoring file or build a new profile to do that. With this PR, they will be able to do it on the fly by adding command line options.

Specifically, this will
* Enable multiple `--rule` options to allow users to cherry-pick more rules at once (rhbz#2020581)
* Add new `--skip-rule` option to allow users to skip some rules (rhbz#2020580)

For more details, please read commit message of every commit.


evgenyz:
I have an important question. How this feature supposed to interact with `requires` and `conflicts`?

For example, if in `/tests/API/XCCDF/unittests/test_xccdf_requires_conflicts.sh` you change the command line to `$OSCAP xccdf eval --rule xccdf_moc.elpmaxe.www_rule_2 --results "$result" "$source"`, what should be the result? And what about `$OSCAP xccdf eval --rule xccdf_moc.elpmaxe.www_rule_2 --rule xccdf_moc.elpmaxe.www_rule_3 --results "$result" "$source"`?

jan-cerny:
Great question. The `--rule` feature is another layer on the top of resolved selections, profile selections, requires, and conflicts. It doesn't influence the requires and conflicts resolution. The scanner resolves the profiles, the requires and conflict as usual, according to specification, and then simply skips the rules that are not listed in `--rule`. That means:
* if the rule is not a part of the selected profile it won't be evaluated because it wouldn't be evaluated even without the `--rule` option
* if the rule conflicts with some other rule it won't be evaluated because it wouldn't be evaluated even without the `--rule` option
* if rule A requires rule B and user provides only `--rule A` it will evaluate only rule A and skip rule B

Now your answers are a consequence of this behavior:
- `oscap xccdf eval --rule xccdf_moc.elpmaxe.www_rule_2` - even if rule 2 is selected in the default profile, oscap won't evaluate anything because rule 2 conflicts with rule 3 which is also selected in the default profile.
- `oscap xccdf eval --profile xccdf_org.open-scap_profile_override --rule xccdf_moc.elpmaxe.www_rule_2` - same as previous, the profile doesn't change the selections of the rule 2 or 3, so the result is the same
- `oscap xccdf eval --rule xccdf_moc.elpmaxe.www_rule_2 --rule xccdf_moc.elpmaxe.www_rule_3` evaluates only rule 3 , rule 2 isn't evaluated because rule 2 conflicts with rule 3, rule 4 isn't evaluated despite it's required by rule 3 because it wasn't listed on the command line, once you provide a `--rule` option, oscap evaluates only the rules explicitly listed
- `oscap xccdf eval --rule xccdf_moc.elpmaxe.www_rule_2 --rule xccdf_moc.elpmaxe.www_rule_3 --profile xccdf_org.open-scap_profile_override`  evaluates only rule 3 because rule 2 conflicts with rule 3, the profile doesn't change the selections of the rule 2 or 3, so the result is the same
- `oscap xccdf eval --rule xccdf_moc.elpmaxe.www_rule_3` evaluates only rule 3, despite the fact that rule 3 requires rule 4, but the rule 4 got skipped because it isn't listed, once you provide a `--rule` option, oscap evaluates only the rules explictely listed
- `oscap xccdf eval --rule xccdf_moc.elpmaxe.www_rule_3 --rule xccdf_moc.elpmaxe.www_rule_4 ` evaluates both rule 3 and 4, it doesn't evaluate rule 1, despite rule 4 requires rule 1, but rule 1 wasn't listed on command line
- `oscap xccdf eval --rule xccdf_moc.elpmaxe.www_rule_3 --rule xccdf_moc.elpmaxe.www_rule_4 --profile xccdf_org.open-scap_profile_override` evaluates only rule 3, it doesn't evaluate rule 4, because rule 4 requires rule 1, but rule 1 is unselected by the `override` profile therefore rule 4 can't be evalauted

Now the question is whether the behavior is expected or isn't stupid.

For example, in the aforementioned situation `oscap xccdf eval --rule xccdf_moc.elpmaxe.www_rule_3`, an user will be confused why rule 4 wasn't evaluated if rule 3 has the `<requires idref="rule 4">` element. But if we change the behavior and oscap will evaluate both rules 3 and 4 in this situation, a different user will be confused, because he wanted to evaluate just rule 3 and will wonder why rule 4 is evaluated.

For example, in the aforemenation situation `oscap xccdf eval --rule xccdf_moc.elpmaxe.www_rule_2` an user will be confused why the rule 2 wasn't evaluated if he explicetly provided the `--rule` option. But if we change the behavior to evaluate the rule 2 in this situation, a different user will be confused because the rule is in conflict with other rule and yet it's evaluated.


My impression is that the goal of the feature was to provide user a simple way to select a single rule or 2-3 rules so maybe we might want to ignore the requires and conflicts here?

evgenyz:
I have no idea how this feature should work, it's better to ask stakeholders (who they are BTW?). But I could suggest either extending `test_xccdf_requires_conflicts.sh` with `--rule` cases or adding some `requires` and `conflicts` to `test_skip_rule` and co. Whatever path we choose we must ensure that it won't be broken later.

jan-cerny:
There is a discussion with the BZ reporter in comment 4 https://bugzilla.redhat.com/show_bug.cgi?id=2020581#c4

Actually I like his view for development and testing:

> For my development and testing scenarios, evaluating all rules specified with --rule (irrespectively of selections and conflicts) and including required rules would make the most sense.

tests/API/XCCDF/unittests/test_xccdf_requires_conflicts.sh
```diff
@@ -23,7 +23,200 @@ assert_exists 1 '//rule-result[@idref="xccdf_moc.elpmaxe.www_rule_11"]/result[te
...
+assert_exists 1 '//rule-result[@idref="xccdf_moc.elpmaxe.www_rule_4"]/result[text()="notselected"]'
+assert_exists 1 '//rule-result[@idref="xccdf_moc.elpmaxe.www_rule_6"]/result[text()="notselected"]'
+assert_exists 1 '//rule-result[@idref="xccdf_moc.elpmaxe.www_rule_8"]/result[text()="notselected"]'
+assert_exists 1 '//rule-result[@idref="xccdf_moc.elpmaxe.www_rule_10"]/result[text()="notselected"]'
+assert_exists 1 '//rule-result[@idref="xccdf_moc.elpmaxe.www_rule_12"]/result[text()="notselected"]'
+assert_exists 1 '//rule-result[@idref="xccdf_moc.elpmaxe.www_rule_14"]/result[text()="notselected"]'
+assert_exists 1 '//rule-result[@idref="xccdf_moc.elpmaxe.www_rule_17"]/result[text()="notselected"]'
+
+
+# test the "conflicts" lemeent eith the --rule option
```
evgenyz, line 39: WUT? `lemeent eith`?

evgenyz:
```suggestion
# test the "conflicts" element with the --rule option
```

evgenyz (commented):
Otherwise, LGTM.

jan-cerny:
@evgenyz The only thing missing is the automated addition of required rules. Is that correct?

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -716,11 +721,12 @@ static int _xccdf_policy_report_rule_result(struct xccdf_policy *policy,
 	} else
 		xccdf_check_free(check);
 
-	/* If policy selects only one rule, skip reporting for the other
-	 * unselected rules - only the selected rule will be reported. */
-	if (policy->rule != NULL) {
+	/* If at least one --rule option has been provided by the user on the
+	 * command line skip reporting for the other rules - only the selected
+	 * rule(s) will be reported. */
+	if (oscap_htable_itemcount(policy->rules) > 0) {
```
evgenyz, line 727: Why not `_user_specified_rule_mode(policy)`?

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -1018,23 +1046,33 @@ _xccdf_policy_rule_evaluate(struct xccdf_policy * policy, const struct xccdf_rul
...
+	/* If the rule is requested to be skipped by the user using --skip-rule on
+	 * the command line we will skip the evaluation of this rule. */
+	if (oscap_htable_get(policy->skip_rules, rule_id) != NULL) {
+		return _xccdf_policy_report_rule_result(policy, result, rule, NULL, XCCDF_RESULT_NOT_SELECTED, NULL);
+	}
+
+	/* If user wants to evaluate only specific rules and the rule currently
+	 * being evaluated is not among these rules, do not evaluate it and mark it
+	 * as notselected. */
+	if (_user_specified_rule_mode(policy) > 0) {
```
evgenyz, line 1058: Or use `oscap_htable_itemcount(policy->rules) > 0` here and get rid of that function.

utils/oscap.8
```diff
@@ -108,7 +108,12 @@ Select a particular profile from XCCDF document. If "(all)" is given a virtual p
 .TP
 \fB\-\-rule RULE\fR
 .RS
-Select a particular rule from XCCDF document. Only this rule will be evaluated. Rule will use values according to the selected profile. If no profile is selected, default values are used.
+Select a particular rule from XCCDF document. Only this rule will be evaluated. Rule will use values according to the selected profile. If no profile is selected, default values are used. This option can be used multiple times to specify multiple rules at once.
```
evgenyz, line 111: Maybe worth mentioning that dependencies won't be evaluated?

utils/oscap.8
```diff
@@ -108,7 +108,7 @@ Select a particular profile from XCCDF document. If "(all)" is given a virtual p
 .TP
 \fB\-\-rule RULE\fR
 .RS
-Select a particular rule from XCCDF document. Only this rule will be evaluated. Rule will use values according to the selected profile. If no profile is selected, default values are used. This option can be used multiple times to specify multiple rules at once.
+Select a particular rule from XCCDF document. Only this rule will be evaluated. Any other rules required by this rule rule won't be evaluated. Rule will use values according to the selected profile. If no profile is selected, default values are used. This option can be used multiple times to specify multiple rules at once.
```
evgenyz, line 111:
```suggestion
Select a particular rule from XCCDF document. Only this rule will be evaluated. Any other rules required by this rule won't be evaluated. Rule will use values according to the selected profile. If no profile is selected, default values are used. This option can be used multiple times to specify multiple rules at once.
```

evgenyz (approved):

evgenyz:
@jan-cerny Can you please rebase it?

jan-cerny:
@evgenyz rebased, thanks

Issue #1683, opened 2021-02-03
Remove internal CPE dictionary


evgenyz:
According to the SCAP standard scanner does not have to provide inbuilt CPE dictionary. We see more harm than good from having it. It will be removed (emptied) in the next API-breaking version.

We will however retain the functionality, so that a package maintainer or the end user could be able to customize the workflow.

jan-cerny:
This looks like a clever idea: we won't have to maintain it, and people will still be able to provide their own dictionaries. It shouldn't affect common use cases because the SCAP source data streams contain their own built-in dictionaries.

PR #1835, opened 2021-12-14
Add Circle Linux 8 CPE

James6xie:
Signed-off-by: James Xie <james@cclinux.org>


evgenyz:
Hey! I'm sorry but we can not accept this change. We are getting rid of the CPE dictionary in OpenSCAP: https://github.com/OpenSCAP/openscap/issues/1683. We believe that the target detection mechanism belongs to the SCAP content, where it could be precisely crafted and customized independently of the scanner.

PR #1838, opened 2022-01-06
The first-boot remediation service

evgenyz:
This PR introduces two things:
- `--progress-full` output mode for the scanner;
- `oscap-remediate-offline-service` wrapper for systemd's Offline System Update mode (`/system-update`).

The `oscap-remediate-offline` script could be used to trigger the remediation, see usage. 

But the real interface of the service is in the configuration file (which is the target of `/system-update` symlink). Basename of the file must be `oscap-remediate-offline.conf.sh`, this is the way the service recognizes its configuration. It is a *Bash* script that would be `source`'d by the service to load various settings in the form of environment variables:

```sh
# Mandatory
OSCAP_REMEDIATE_DS=/some/data_stream.xml
OSCAP_REMEDIATE_PROFILE_ID=some_profile
# Optional
OSCAP_REMEDIATE_DATASTREAM_ID=some_ds_id
OSCAP_REMEDIATE_XCCDF_ID=some_xccdf_id
OSCAP_REMEDIATE_BENCHMARK_ID=some_bench_id
OSCAP_REMEDIATE_TAILORING=/some/tailoring.xml
OSCAP_REMEDIATE_TAILORING_ID=tailoring_id
OSCAP_REMEDIATE_ARF_RESULT=/some/arf_res.xml
OSCAP_REMEDIATE_HTML_REPORT=/some/report.html
OSCAP_REMEDIATE_VERBOSE_LOG=/var/some_verbose.log
# Optional even if OSCAP_REMEDIATE_VERBOSE_LOG is provided, defaults to INFO
OSCAP_REMEDIATE_VERBOSE_LEVEL=INFO
```
The `oscap-remediate-offline` would only require the user to provide data stream and profile id. But it will also set
```sh
OSCAP_REMEDIATE_VERBOSE_LOG=/var/tmp/oscap-verbose.log
OSCAP_REMEDIATE_ARF_RESULT=/var/tmp/oscap-arf_result.xml
OSCAP_REMEDIATE_HTML_REPORT=/var/tmp/oscap-report.html
```
for convenience. All other variables in the configuration would be empty (unset).

It looks like this during the boot:
Screenshot from 2022-01-17 00-06-13


utils/oscap-remediate-offline-service
```diff
@@ -0,0 +1,78 @@
...
+    systemd-notify --ready --status="Skipped"
+    exit 0
+fi
+
+rm -f /system-update
+
+log "Found config: ${config}"
+
+source "$config"
+
```
jan-cerny, line 27: What if the config file is broken and doesn't contain OSCAP_REMEDIATE_DS or OSCAP_REMEDIATE_PROFILE_ID ?

evgenyz: I've put all these checks in `oscap-remediate-offline` in the beginning. But then recently I realized that it is inherently not the only way to prime the service (think Image Builder) and we have to re-check it in the service itself. Hence `TODO: Check` entries there.

evgenyz: See: https://github.com/OpenSCAP/openscap/pull/1838/files#diff-ca5a0596040ff7754e2ce486cfab14fbd60d1b804d5a7dcbf9649529031fac49R52

utils/oscap-remediate-offline-service
```diff
@@ -0,0 +1,78 @@
...
+	if [[ "${line}" =~ "---remediation" ]]; then
+		message="OpenSCAP is remediating the system to be compliant using the"$'\n'"${profile_title}"$'\n\n'
+		log "Remediating..."
+		continue
+	fi
+	# TODO: Format is "rule_id|Rule Title|result"
+	plymouth display-message --text="${message}${line#*	}"
+	log "${line#*	}"
+done < <(${command})
+
```
jan-cerny, line 73: In my `journalctl` output I can see multiple messages like this:

```
Jan 06 14:52:18 fedora oscap-remediate-offline-service[13975]: plymouth: ply-boot-client.c:465: ply_boot_client_queue_request: Assertion `request_argument == NULL || strlen (request_argument) <= UCHAR_MAX' failed.
Jan 06 14:52:18 fedora oscap-remediate-offline-service[554]: /usr/libexec/oscap-remediate-offline-service: line 72: 13975 Aborted                 (core dumped) plymouth display-message --text="${message}${line#*        }"
```

jan-cerny (commented):
I used the packit's build on a F 35 VM and it all works perfectly. During the reboot I have seen some messages and after that I have found the report and results in /var/lib as expected,

utils/oscap-remediate-offline
```diff
@@ -0,0 +1,69 @@
...
+    usage
+    exit 1
+}
+
+usage() {
+    echo "oscap-remediate-offline -- Setup script for the SCAP evaluation and remediation using systemd'd update service."
+    echo
+    echo "Usage:"
+    echo
+    echo "$ oscap-remediate-offline PROFILE INPUT_CONTENT"
```
matejak, line 29: If you take a look at usage of OpenSCAP in OAA e.g. near https://github.com/OpenSCAP/oscap-anaconda-addon/blob/rhel8-branch/org_fedora_oscap/common.py#L259, then more arguments will be likely needed:
- tailoring file,
- datastream ID, XCCDF ID,
- path for results, report

Although you provide a meaningful default in your script, it makes sense that the addon should be able control the location where reports are generated.

If you don't like the idea of supporting multiple arguments, some of them optional, then [Argbash](https://argbash.readthedocs.io/en/latest/) may help - wink wink.

evgenyz: Actually the interface between OpenSCAP and Image Builder or OAA is the configuration file along with the symlink `/system-update` pointing to it. The `oscap-remediate-offline` was never meant to be used by them because it would definitely be of no help to IB, and I suspect for Anaconda it would be easier to create a symlink in a proper place rather thank juggle with the chroot environment and try executing the script.

It is just a convenient way of testing the thing. Now I'm not even sure it is a good idea to advertise it (via man page or the user manual).

With that being said, extending the configuration with tailoring, DS and XCCDF ids and a path for the result is definitely a good idea. The question is, what should be optional:
- ds file name (definitely required)
- profile_id
- ds_id
- xccdf_id
- path for logs (definitely optional)
- path for results (should it be separated from the logs)?

utils/oscap-remediate-offline-service
evgenyz: Yeah, so we'd better cut down the text on the boot screen. The whole message with the profile name and nice wording and current rule title hitting the message length limit :(

utils/oscap-remediate-offline-service
```diff
@@ -0,0 +1,78 @@
...
+log_dir="/var/lib/ssg/$(date +%Y.%m.%d-%H:%M:%S)"
+
+mkdir -p "${log_dir}" && cd "${log_dir}"
+
+#TODO: Check?
+
+command="oscap xccdf eval --profile ${OSCAP_REMEDIATE_PROFILE_ID}\
+    --results xccdf-results.xml\
+    --results-arf arf-results.xml\
```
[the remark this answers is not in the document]
evgenyz: Hmm. Okay, makes sense.

utils/oscap-remediate-offline.8
```diff
@@ -0,0 +1,22 @@
+.TH oscap-remediate-offline "8" "Dec 2020" "Red Hat, Inc." "System Administration Utilities"
```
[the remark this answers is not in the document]
evgenyz: Now I'm not sure I actually like the idea to advertise this script. See https://github.com/OpenSCAP/openscap/pull/1838#discussion_r779790042

utils/oscap-remediate-offline-service
evgenyz: Should be fixed now

utils/oscap-remediate-offline.8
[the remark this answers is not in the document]
evgenyz: Packaged. Although my hesitation remains.

utils/oscap-remediate-offline-service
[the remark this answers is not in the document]
evgenyz: I've removed the `--results` argument. But now I see that OAA uses it instead of `--results-arf` and Workbench has them both for some reason. @matejak @jan-cerny Do we need it or not? Should it be optional?

utils/oscap-remediate-offline-service
evgenyz: Fixed

utils/oscap-remediate-offline-service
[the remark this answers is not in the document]
jan-cerny: I would say that we don't need it because it contains the same data that are present in the ARF file.

Regarding the OAA and Workbench I think they use --results just for a historic reason because in very old versions oscap didn't support ARF.

utils/oscap-remediate-offline.8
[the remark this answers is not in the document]
jan-cerny: Aha, it's mainly for development / testing purposes. I think that the man page isn't mandatory. I think we should be consistent - so far we ship all man pages we have. So either remove the man page completely or package it. You can also advertise the script in a way that you will add to the man page that it's a "convenient utility for testing and debugging the service"

utils/oscap-remediate-offline
matejak: I think that having a ready-to-use way of scheduling the remediation for the next boot makes sense. If I understand it right, a tool would have to make those substitutions in the config file, copy stuff in place, and create a symlink - looks like a good job for a script.

Regarding which args should be optional - probably all except datastream and profile ID.
And what do you mean by logs? The scanner produces a report and results/ARF, I don't see it capturing logs. Anyway, I think that the option to specify independently where to save those or that we don't want them (probably by providing an empty filename if the default is to produce them somewhere) again makes sense.

utils/oscap-remediate-offline-service
```diff
@@ -0,0 +1,101 @@
...
+header="OpenSCAP is checking the system for compliance using"$'\n'"${profile_title}"$'\n\n'"Evaluating..."
+display "$header"
+
+command="oscap xccdf eval --profile ${OSCAP_REMEDIATE_PROFILE_ID}\
+	--results-arf arf-results.xml\
+	--report report.html\
+	--progress-full\
+	--remediate ${OSCAP_REMEDIATE_DS}"
+
+while IFS= read -r line; do
```
matejak, line 78: Is there a need to manipulate `IFS`? `read` supports the `-d` option, would it make sense to use it to avoid interaction with IFS?

evgenyz: Meh, yes and no. IFS has nothing to do with -d in this case, but is not needed here anymore. Nice catch.

utils/oscap-remediate-offline
evgenyz: While it may seem convenient to have this script, its introduction bears some implications.

First of all it defies declarative over imperative approach we are trying to employ. It also creates a dependency on a bash script (and also the bash itself) and the necessity of executing something at the configuration (priming) step. And I would rather put some effort to discourage usage of this script beyond testing than to make it more versatile.

Moreover, it is an API. And we would be expanding API surface without a good reason (if we no longer consider it as an interface for OAA or IB).

As for the value for the end-user, I can't see any benefits in choosing systemd service over a regular scan and remediation. If nothing else it would move the process into less controllable and observable environment.

matejak: I can see the implication of expanding the API, the script would come up with an API of a higher level of abstraction. We have `oscap-podman`, `oscap-chroot` and so on that do exactly this - they hide implementation details of various execution modes of the scanner. I don't see any concern with expanding our "API surface", as the script is a facade that don't get complicated. Is there something that you have in mimd what could go wrong?
I would also argue that the addon shouldn't have such low-level knowledge of how to schedule the firstboot scan. But I also disagree with the imperative vs declarative claim - although Bash is not a declarative language, the script's interface is the closest to declarative that I can imagine - provided arguments define the result, and you don't care what needs to be done to get that result - isn't that what declarative is supposed to mean?

utils/oscap-xccdf.c
```diff
@@ -443,6 +447,39 @@ static int callback_scr_result_progress(struct xccdf_rule_result *rule_result, v
 	return 0;
 }
 
+static int callback_scr_rule_progress_full(struct xccdf_rule *rule, void *arg)
+{
+	const char * rule_id = xccdf_rule_get_id(rule);
```
jan-cerny, line 452:
```suggestion
	const char *rule_id = xccdf_rule_get_id(rule);
```

utils/oscap-xccdf.c
```diff
@@ -443,6 +447,39 @@ static int callback_scr_result_progress(struct xccdf_rule_result *rule_result, v
...
+
+	if (!xccdf_policy_is_item_selected((struct xccdf_policy *)arg, rule_id))
+		return 0;
+
+	const char *title = xccdf_policy_get_readable_item_title((struct xccdf_policy *)arg, (struct xccdf_item *)rule, NULL);
+	printf("%s|%s|", rule_id, title);
+	free((char *)title);
```
jan-cerny, line 459: This shouldn't be `const` if you modify it by `free()`. The function `xccdf_policy_get_readable_item_title` returns just `char *`

utils/oscap-xccdf.c
```diff
@@ -1172,6 +1225,7 @@ bool getopt_xccdf(int argc, char **argv, struct oscap_action *action)
 		{"enforce-signature", no_argument, &action->enforce_signature, 1},
 		{"fetch-remote-resources", no_argument, &action->remote_resources, 1},
 		{"progress", no_argument, &action->progress, 1},
+		{"progress-full", no_argument, &action->progress, 2},
```
jan-cerny, line 1228: Consider replacing magic constants 1 and 2 by an enum.

jan-cerny (commented):
It works amazing, I have seen the new location and the report containing fixed results. It seems that the only remaining thing is to solve whether to have the initialization script or not.

utils/oscap-remediate-offline
jan-cerny: While I found the script very convenient for performing my review, I slightly lean towards removing the script. 

To make it working properly we would have to improve the `oscap info` tool to be able to verify that the provided parameters are correct and will lead to a meaningful remediation. Improving `oscap info` wouldn't be simple. 

Keep in mind that in imagebuilder we can't execute any bash script, so we will have to declare the config file options as a public API anyway. So the script doesn't help us with the abstraction that we would want, it doesn't allow us to fully hide implementation details from the outer world.

matusmarhefka: I don't have a strong opinion about the script, the advantage of the script I see is that whenever we change something in API we won't need to update OAA component. And regarding the argument about validating parameters of the script, I think we could do just some basic sanity check and let it fail during boot in case something is invalid, user will see it failed from the logs - that would be the same as if one does some mistake when writing parameters directly into a config file.

Also as mentioned the API needs to be extended to accept more parameters, I think what @evgenyz listed looks good:
```
Required:
ds file name
profile_id

Optional:
ds_id
xccdf_id
tailoring
path for XML results
path for HTML report
path for logs
```

utils/oscap-remediate-offline-service
evgenyz: Fixed.

utils/oscap-xccdf.c
evgenyz: Fixed.

utils/oscap-xccdf.c
evgenyz: Done.

utils/oscap-xccdf.c
evgenyz: Fixed.

evgenyz:
> It works amazing, I have seen the new location and the report containing fixed results. It seems that the only remaining thing is to solve whether to have the initialization script or not.

There is a high chance that this PR would be dragged around in form of patches. I'd like to squash it a bit before merging. Don't just merge it before that happen even if you would think that it is good enough.

utils/oscap-remediate
```diff
@@ -0,0 +1,146 @@
...
+if [[ "$(basename ${config})" != "${OSCAP_REMEDIATE_CONF_BASENAME}" ]]; then
+	log "The ${SYSTEMD_SYSTEM_UPDATE_LINK} symlink does not point to oscap offline remediation configuration, ignoring"
+	exit 0
+fi
+
+rm -f "${SYSTEMD_SYSTEM_UPDATE_LINK}"
+log "Removed ${SYSTEMD_SYSTEM_UPDATE_LINK}"
+log "Found config: ${config}"
+
+source "$config"
```
jan-cerny, line 44: @evgenyz  We will need to have the format of the config file documented somewhere - ie. document which variables the service accepts and what is their meaning so that the people/software using this service know how to set this service up. Ideally in this service's documentation.

evgenyz:
Okay we are nearing the apex. Remaining things: squashing and documentation.

utils/oscap-remediate
evgenyz: Done

jan-cerny (approved):
I have successfully set up a F35 VM using OSPP profile.

PR #1839, opened 2022-01-19
Change default value for OSCAP_PROBE_MEMORY_USAGE_RATIO to 0.33

evgenyz:
After some testing we decided that this ratio is more usable in real-world cases.


jan-cerny (approved):
FTR We have discovered that the test for original rhbz#1932833 (which we fixed by introducing this limit) requires this to be set to 0.25 or lower. Accepting this change means that we might start crashing in situations similar like that again. So this change is a balance with some other use cases where the value of 0.1 causes a lot of unknown results and warning messages. Users might need to export the environment variable to override.

jan-cerny:
@evgenyz please next time create branches in your fork

PR #1844, opened 2022-02-01
Prevent fails of test_ds_misc.sh

jan-cerny:
Other files from which the datastream is composed might also
affect the timestamp attributes in result document depending
on their mtime.


evgenyz (approved):
Mmm, okay

jan-cerny:
yes, it didn't happen before despite I tried multiple times :O

PR #1845, opened 2022-02-02
Prevent file permissions errors

jan-cerny:
The sysctl setting `fs.protected_regular` doesn't allow `O_CREAT` open
on regular files that we don't own in world writable sticky directories
(think `/tmp`). This causes permission denied error when writing HTML
report to a temporary files created by the `mktemp` command executed as
a normal user and then executing `sudo oscap`.

See https://bugzilla.redhat.com/show_bug.cgi?id=2048571

This fixes only creation of the HTML report using the `--report` option,
We will have to create a smimilar patch also for other output options
such as `--results` or `--results-arf`.

### DIFF
--- a/src/source/xslt.c
+++ b/src/source/xslt.c
@@ -29,6 +29,7 @@
 #include <libxslt/xsltutils.h>
 #include <libexslt/exslt.h>
 #include <string.h>
+#include <sys/stat.h>
 #ifdef OS_WINDOWS
 #include <io.h>
 #else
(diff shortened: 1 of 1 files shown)


jan-cerny:
The proposal checks if the file exists and if yes it removes it and open a new one under same name and then sets its owner and group owner to the same values as original file. Now the  questions are if it makes sens or is dangerous or breaks anything else. Do you have any recommendation.

jan-cerny:
Supersed by #1846

PR #1846, opened 2022-02-03
Prevent file permissions errors

jan-cerny:
The sysctl setting `fs.protected_regular` doesn't allow `O_CREAT` open
on regular files that we don't own in world writable sticky directories
(think `/tmp`). This causes permission denied error when writing HTML
report to a temporary files created by the `mktemp` command executed as
a normal user and then executing `sudo oscap`.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2048571

If OpenSCAP fails to open the file because of permissions, it will retry
to open the file without O_CREAT flag.

Supersedes #1845 .

### DIFF
--- a/src/common/elements.c
+++ b/src/common/elements.c
@@ -227,29 +227,9 @@ int oscap_xml_save_filename(const char *filename, xmlDocPtr doc)
 		xmlCode = xmlSaveFormatFileEnc(filename, doc, "UTF-8", 1);
 	}
 	else {
-#ifdef OS_WINDOWS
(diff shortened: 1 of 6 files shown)


evgenyz:
Yeah, great, thank you, that's definitely better from race conditions POV. But I guess we can do better. We could wrap all this into an utility function, can we?

jan-cerny:
I have installed it on a Fedora 35 virtual machine and I have executed a sudo remote scan from SCAP Workbench running on the host. It fixes the problem, both scan and remediation ended successfully and I could view the report. So I switch this to review.

PR #1849, opened 2022-03-08
Skip non-text files in textfilecontent54 probe

jan-cerny:
The probe will guess whether a file is a text file and if it isn't a
text file it will skip the file. It's based on assumption that non-text
files don't contain any meaningful text to be matched by a
textfilecontent54_object. This is to filter our binary files. Up until
now binary files are processed but they cause an error produced by
pcre_exec which can't process non-UTF8 string but binary files usually
contain an invalid UTF8 string. Therefore people couldn't use it anyway.

This will prevent producing errors when you specify file path using a
regex in your OVAL content and the regex matches also non-text file paths.
For example the rule xccdf_org.ssgproject.content_rule_accounts_umask_interactive_users

Resolves: [rhbz#2033246](https://bugzilla.redhat.com/show_bug.cgi?id=2033246)

The main disadvantage is that using `magic` is only guessing, it can guess wrong so it can filter out some files that should be processed.


marcusburghardt (approved):
LGTM. I have tested the PR with the rule `accounts_umask_interactive_users` and it worked as expected. Binary files are now ignored and the error is no longer present. Thanks @jan-cerny

jan-cerny:
@marcusburghardt I remember that @evgenyz did some investigation and he had some objections which he was about to post.

evgenyz:
This behavior of the probe is against the OVAL specification and it will create more problems in the future.

The actual problem is that the probe is unable to operate on non-unicode files. This error should be clearly communicated to the OVAL result, instead of masking it with an artificial error "File is not a text file" (which is by the way just a guesswork).

Just imagine that we are checking config files in `/etc/sshd_config.d/*`. One of them is a malignant binary file that causes buffer overflow in configuration parser and forces it to provide passwordless root access. This fix will allow the scanner to just ignore it and happily report that the system is compliant.

False-negative is way worse than false-positive.

evgenyz:
Binary file in the place you're expecting to contain only text files is a problem all by itself. If you expect some binary files there — refine your regexp.

marcusburghardt:
> Binary file in a place you're expecting to contain only text files is a problem all by itself. If you expect some binary files there — refine your regexp.

Unfortunately, cases like in the mentioned rule are not possible to be solved with `regexp` on the content side. : (

evgenyz:
> The OVAL specification says that individual blocks of text from regular files should be processed:
> 
>     * https://oval-community-guidelines.readthedocs.io/en/latest/oval-schema-documentation/independent-definitions-schema.html?highlight=textfilecontent54_test#textfilecontent54-test-1

> The magic library is not guessing about text files. ...
> ... some files which can't be safely defined by its signature and are "guessed" by the library.

Please don't make self-contradictory statements, they are of no help to anyone. Guesswork is guesswork.

> The patch is ignoring everything which is not a known "text" file. So, it could skip a bad formatted or even corrupted text file and this looks reasonable, 

Where the OVAL spec defines such a behavior?

> but will not guess about the confirmed "text" files. Once a text file is identified, it will be processed. Since the content of a text file is not necessarily tied to its mime signature, if the known text file has any content which can't be processed by the probe, then it is expected to raise an error, even in the mentioned ssh example. In short, the patch is not ignoring the error itself, but is ignoring non-text files, which looks pretty reasonable according to the OVAL definition and realistic scenarios.

None of this is defined by the OVAL spec. It's your speculation.

> So, I don't think this patch will generate false-positives or even create future problems for the `textfilecontent54 probe`.

I just gave you an example of a *false-negative* error. False-positive was the error this PR is trying to fix. There would be *less* false-positive errors, but far more false-negatives. And these are is way way worse.

evgenyz:
> > Binary file in a place you're expecting to contain only text files is a problem all by itself. If you expect some binary files there — refine your regexp.
> 
> Unfortunately, cases like in the mentioned rule are not possible to be solved with `regexp` on the content side. : (

Unfortunately it is not a reason for swiping the problem under the carpet.

evgenyz:
What about `xmlfilecontent` and `yamfilecontent` probes? Are we going to be consistent and silently ignore files that have parse errors?

marcusburghardt:
> What about `xmlfilecontent` and `yamfilecontent` probes? Are we going to be consistent and silently ignore files that have parse errors?

They are also "text" files according to their mime signatures, so they will still be processed and parse errors will still be shown. The point is not to ignore errors, but ignore non-text files, one step behind.

evgenyz:
> > What about `xmlfilecontent` and `yamfilecontent` probes? Are we going to be consistent and silently ignore files that have parse errors?
> 
> They are also "text" files according to their mime signatures, so they will still be processed and parse errors will still be shown. The point is not to ignore errors, but ignore non-text files, one step behind.

There is no mention of filtering *text files* based on *mime signatures* in the OVAL spec for these probes. Stop speculating.

evgenyz:
> Unfortunately, cases like in the mentioned rule are not possible to be solved with `regexp` on the content side. : (

How about instead of going through all files in the home directory you'll start targeting actual configuration files?
```
~/.bash_profile
~/.bash_login
~/.profile
~/.bashrc
~/.bash_logout
~/.inputrc
```

marcusburghardt:
> > The OVAL specification says that individual blocks of text from regular files should be processed:
> > ```
> > * https://oval-community-guidelines.readthedocs.io/en/latest/oval-schema-documentation/independent-definitions-schema.html?highlight=textfilecontent54_test#textfilecontent54-test-1
> > ```
> 
> > The magic library is not guessing about text files. ...
> > ... some files which can't be safely defined by its signature and are "guessed" by the library.
> 
> Please don't make self-contradictory statements, they are of no help to anyone. Guesswork is guesswork.

The library can guess about what is not known and even return "unknown", as some examples mentioned in its documentation. But in cases where the mime signature is known, it is not a guess. It is simply how the library works.

> 
> > The patch is ignoring everything which is not a known "text" file. So, it could skip a bad formatted or even corrupted text file and this looks reasonable,
> 
> Where the OVAL spec defines such a behavior?

I shared the relevant OVAL spec above.
Here I am talking about the implementation behaviour, which is not covered in this level of detail by the OVAL definition.
Sounds logical to skip non-text files and we are actually disucssing this implementation just because the OVAL is not clear on that. : )

> 
> > but will not guess about the confirmed "text" files. Once a text file is identified, it will be processed. Since the content of a text file is not necessarily tied to its mime signature, if the known text file has any content which can't be processed by the probe, then it is expected to raise an error, even in the mentioned ssh example. In short, the patch is not ignoring the error itself, but is ignoring non-text files, which looks pretty reasonable according to the OVAL definition and realistic scenarios.
> 
> None of this is defined by the OVAL spec. It's your speculation.

I don't think it is expetect this to be defined in OVAL. Anyways, I am not talking about OVAL.
This is not speculation. Fortunatelly I had opportunitty to work with forensic of files and know pretty well how their mimes are formed. So, I still don't see the point to process non-text files looking for blocks of text. So, why not skipping them? 

> 
> > So, I don't think this patch will generate false-positives or even create future problems for the `textfilecontent54 probe`.
> 
> I just gave you an example of a _false-negative_ error. False-positive was the error this PR is trying to fix. There would be _less_ false-positive errors, but far more false-negatives. And these are way way worse.

Still, if a parse error happen in a text file, it should be caught. It was not the intention of this patch but I agree that better error handling is still relevant to be included, but not necessarily in this patch.

evgenyz:
> > Please don't make self-contradictory statements, they are of no help to anyone. Guesswork is guesswork.
> 
> The library can guess about what is not known and even return "unknown", as some examples mentioned in its documentation. But in cases where the mime signature is known, it is not a guess. It is simply how the library works.

You can't be 80% pregnant, you either are or you are not.

> > > The patch is ignoring everything which is not a known "text" file. So, it could skip a bad formatted or even corrupted text file and this looks reasonable,
> > 
> > Where the OVAL spec defines such a behavior?
> 
> I shared the relevant OVAL spec above. Here I am talking about the implementation behaviour, which is not covered in this level of detail by the OVAL definition. Sounds logical to skip non-text files and we are actually disucssing this implementation just because the OVAL is not clear on that. : )

No, it is not logical.

> > > but will not guess about the confirmed "text" files. Once a text file is identified, it will be processed. Since the content of a text file is not necessarily tied to its mime signature, if the known text file has any content which can't be processed by the probe, then it is expected to raise an error, even in the mentioned ssh example. In short, the patch is not ignoring the error itself, but is ignoring non-text files, which looks pretty reasonable according to the OVAL definition and realistic scenarios.
> > 
> > None of this is defined by the OVAL spec. It's your speculation.
> 
> I don't think it is expetect this to be defined in OVAL. Anyways, I am not talking about OVAL. This is not speculation. Fortunatelly I had opportunitty to work with forensic of files and know pretty well how their mimes are formed. So, I still don't see the point to process non-text files looking for blocks of text. So, why not skipping them?

Because we are not supposed to. And content authors will expect that we won't. I think that is enough of a reason.

> > > So, I don't think this patch will generate false-positives or even create future problems for the `textfilecontent54 probe`.
> > 
> > I just gave you an example of a _false-negative_ error. False-positive was the error this PR is trying to fix. There would be _less_ false-positive errors, but far more false-negatives. And these are way way worse.
> 
> Still, if a parse error happen in a text file, it should be caught. It was not the intention of this patch but I agree that better error handling is still relevant to be included, but not necessarily in this patch.

If you silently exclude all the broken files beforehand there won't be any. This patch hides the problem by silently ignoring it. This is unacceptable.

evgenyz:
So, this file will set `umask` to `0077` but won't be detected by this solution.
[.bashrc.zip](https://github.com/OpenSCAP/openscap/files/8460000/default.bashrc.zip)

evgenyz:
Just make sure that you don't have umask string in any other file.

jan-cerny:
@marcusburghardt @evgenyz Thanks for summarizing all the pros and cons.

Things now got more complicated by a new BZ we received today: https://bugzilla.redhat.com/show_bug.cgi?id=2074037 where the reporter has a problem that the probe can't process non-utf8 files.

Remember that there was also https://bugzilla.redhat.com/show_bug.cgi?id=1285757 where we had a false negative because of processing a file with non-utf8 enconding.

evgenyz:
Any type of guesswork is not acceptable because we would never know if our guesswork matches the guesswork (if any) of the end product (the utility that would read the file).

evgenyz:
But bash will interpret it without any problems. Again, the matter of interpretation and guesswork.

yuumasato:
> My opinion is that we should not ignore broken files. We could change the result of the scan to something else (NOTCHECKED?), but it should be something that won't result in a compliant state (because we were not able to check the system).

I think a result of `error` or `unknown` is more appropriate than 'notchecked`, see XCCDF definitions below:
```
error [E] The checking engine could not complete the evaluation, therefore the status of the target’s
compliance with the <xccdf:Rule> is not certain. This could happen, for example, if a testing
tool was run with insufficient privileges and could not gather all of the necessary information.

unknown [U] The testing tool encountered some problem and the result is unknown. For example, a result of
‘unknown’ might be given if the testing tool was unable to interpret the output of the checking engine (the output has no meaning to the testing tool).

notchecked [K] The <xccdf:Rule> was not evaluated by the checking engine. This status is designed for
<xccdf:Rule> elements that have no <xccdf:check> elements or that correspond to an
unsupported checking system. It may also correspond to a status returned
```

PR #1850, opened 2022-03-17
Packit: add CentOS Stream 8 and 9 build targets

matusmarhefka:
Summary of updates in this PR:
* remove depracated `synced_files` key
* enable CentOS Stream build and tests targets
* enable testing on Testing Farm


jan-cerny:
Great!

Where is the test that we see in the testing farm defined?

matusmarhefka:
> Great!
> 
> Where is the test that we see in the testing farm defined?

There is only one test in the `tests` job which verifies that package can be installed on the selected composes: https://gitlab.com/testing-farm/tests/-/tree/main/packit

jan-cerny:
@matusmarhefka So now what remains to be done here?

matusmarhefka:
> @matusmarhefka So now what remains to be done here?

I would wait for the fix of the CentOS Stream 9 compose which needs to be done by Testing Farm team. It is in progress and when the issue is fixed I will rerun the job to verify it.

matusmarhefka:
@jan-cerny Issue with CentOS Stream 9 is fixed and the PR is ready for review/merge.

jan-cerny (approved):
amazing and great contribution

PR #1861, opened 2022-05-09
Reset errno before call to strtoll

ajaynair:
While debugging https://github.com/OpenSCAP/openscap/issues/1858, I see that strtoll was being called before resetting errno. In some cases, it can result in false positives.


evgenyz (approved):
Thanks!

PR #1864, opened 2022-05-24
Use correct includes

jan-cerny:
rpmvercmp is defined in rpm/rpmver.h
risdigit is defined in rpm/rpmstring.h

This fixes a FTBFS in Rawhide: https://bugzilla.redhat.com/show_bug.cgi?id=2080210
See a scratch build https://koji.fedoraproject.org/koji/taskinfo?taskID=87439260 that uses this patch.


cmake/FindRPM.cmake
```diff
@@ -30,6 +30,9 @@ set(RPM_VERSION ${RPM_PKGCONF_VERSION})
 if(RPM_VERSION)
 	string(COMPARE GREATER "4.6" ${RPM_VERSION} RPM46_FOUND)
 	string(COMPARE GREATER "4.7" ${RPM_VERSION} RPM47_FOUND)
+	if(NOT (RPM_VERSION VERSION_LESS "4.18"))
```
matusmarhefka, line 33: What about rather using `VERSION_GREATER_EQUAL`?

evgenyz (approved):

havardAasen:
`VERSION_GREATER_EQUAL` was introduced in CMake version 3.7, and seems to break with the minimum required version 2.8.
https://github.com/OpenSCAP/openscap/blob/e4091e124666269a00356bfb38d6d5fb516c1d32/CMakeLists.txt#L1

At the same time, CMake 2.8 was released in 2009 and is now 13 years old, is it necessary to "support" such an old version?

PR #1869, opened 2022-07-02
Move firstboot (offline remediation) service and wrapper to a sub-package

evgenyz:
This functionality is not considered as part of the primary user experience. But we still want users and developers to be able to experiment with this feature.

### DIFF
--- a/openscap.spec
+++ b/openscap.spec
@@ -82,6 +82,14 @@ The %{name}-scanner package contains oscap command-line tool. The oscap
 is configuration and vulnerability scanner, capable of performing
 compliance checking using SCAP content.
 
+%package        remediate-offline
+Summary:        OpenSCAP Offline Remediation Service
+Requires:       %{name}%{?_isa} = %{epoch}:%{version}-%{release}
+Requires:       %{name}-scanner%{?_isa} = %{epoch}:%{version}-%{release}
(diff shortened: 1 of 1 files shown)


openscap.spec
```diff
@@ -82,6 +82,14 @@ The %{name}-scanner package contains oscap command-line tool. The oscap
 is configuration and vulnerability scanner, capable of performing
 compliance checking using SCAP content.
 
+%package        remediate-offline
```
jan-cerny, line 85: We don't ship the oscap-remediate-offline script, right? Shouldn't the package name be just `remediate` or `remediate-service`?

evgenyz: Actually I don't like any combination with `remediate`. Because it might give an impression that the sub-package is needed for the scanner to be able to remediate the system. The feature of systemd that we utilize is called `system-update`. How about `openscap-system-update` sub-package?

evgenyz: I'll put this question on the consilia.

jan-cerny: My point was to avoid confusion with the "oscap-remediate-offline" executable. But you have a good point that we can have a different confusion as well.

evgenyz:
Nah, we don't need it: #1988.

PR #1872, opened 2022-07-12
Generate graphs, when creating Doxygen documentation

havardAasen:
When building documentation with Doxygen, and `dot` is found, generate diagrams and graphs.

I changed the image format to svg, for smaller file size and better scaling.

`dot` is normally found in `graphviz` package and is entirely optional to install.


CMakeLists.txt
```diff
@@ -469,6 +469,7 @@ message(STATUS " ")
 message(STATUS "Documentation:")
 message(STATUS "enabled: ${ENABLE_DOCS}")
 message(STATUS "doxygen: ${DOXYGEN_EXECUTABLE}")
+message(STATUS "graphviz: ${DOXYGEN_DOT_EXECUTABLE}")
```
evgenyz, line 472:
```suggestion
message(STATUS "dot (graphviz): ${DOXYGEN_DOT_EXECUTABLE}")
```

evgenyz:
@jan-cerny What do you thing about diagrams in our API docs?

CMakeLists.txt
havardAasen: Updated with the applied suggestion

jan-cerny:
@evgenyz They look like they won't help you much, you would need some more knowledgable diagram, that would have to be created manually I think.

jan-cerny (approved):
Thank you!

Issue #1867, opened 2022-06-23
Scanning of certain rules produces memory error messages


matejak:
#### Description of Problem:

Scanning of `accounts_user_dot_no_world_writable_programs`, a rule that got its content for the first time in Q2/2022, reliably triggers an error when scanned on a RHEL7 system.

#### OpenSCAP Version:

1.3.6

#### Steps to Reproduce:

Obtain a recent-enough datastream that implements a check for `accounts_user_dot_no_world_writable_programs`.

1. `oscap xccdf eval --profile '(all)' --rule xccdf_org.ssgproject.content_rule_accounts_user_dot_no_world_writable_programs /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml`
2. Watch the logs - you should see something like
   ```
    ...
    Title   User Initialization Files Must Not Run World-Writable Programs
    Rule    xccdf_org.ssgproject.content_rule_accounts_user_dot_no_world_writable_programs
    Ident   CCE-80523-4
    E: probe_file: Failed to check available memory
    E: probe_file: Failed to check available memory
    E: probe_file: Failed to check available memory
    E: probe_file: Failed to check available memory
    E: probe_file: Failed to check available memory
    E: probe_file: Failed to check available memory
    E: probe_file: Failed to check available memory
   ```

#### Expected Results:

No messages like these.

The scanner doesn't crash, but validity of produced results is questionable.

jan-cerny:
If you remove the `prodtype` from the rule `accounts_user_dot_no_world_writable_programs` and build Fedora content, then a very similar issue would be reproducible also on Fedora 35 using openscap-1.3.6-3.fc35.x86_64. I say very similar because in addition of large amount of messages ` E: probe_file: Failed to check available memory` it also produces a large amount of messages `E: oscap:     Function pcre_exec() failed to match a regular expression with return code -10 on string`.

This "more verbose" issue is also reproducible with the 1.3.6 upstream release complied from git on Fedora 35. However, the message `E: probe_file: Failed to check available memory` isn't reproducible with current upstream maint-1.3 branch (as of HEAD 72dd8ba0fc45cc9e5989fdb345c5717d7ddb5bae). With current upstream maint-1.3 branch (as of HEAD 72dd8ba0fc45cc9e5989fdb345c5717d7ddb5bae), only the excessive `E: oscap:     Function pcre_exec() failed to match a regular expression with return code -10 on string` remain.

[...]

PR #1874, opened 2022-07-22
Add unit test for read_common_sizet function

jan-cerny:
The unit test will cover the missing set errno
to 0 which was the root cause of:
https://github.com/OpenSCAP/openscap/issues/1867

Therefore, this test can be used during verification of:
https://bugzilla.redhat.com/show_bug.cgi?id=2109485

### DIFF
--- a/tests/API/probes/test_memusage.c
+++ b/tests/API/probes/test_memusage.c
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2022 Red Hat Inc., Durham, North Carolina.
+ * All Rights Reserved.
+ *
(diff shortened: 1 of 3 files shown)


tests/API/probes/test_memusage.c
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2022 Red Hat Inc., Durham, North Carolina.
```
evgenyz, line 2: We should really stop copy-pasting that. https://spdx.dev/ids/

evgenyz (approved):

PR #1881, opened 2022-08-21
oscap-ssh: simplify to allow limited sudo rule

maage:
With this patch set I can use very limited sudo rule when running `oscap xccdf eval`.

Also address some shellcheck warnings.

Without this to try to limit how sudo needs to be configured, sudoers rule needs to have `/usr/bin/sh` permission, exec permission (`sh -c`), and run `chown` without known path. This is too broad.

`chown` is not needed as `umask 022` + `rm -rf dir` is enough to remove files as directory is user owned.

My example rule:
- Note: in Fedora binary is named oscap.
- Only allow `xccdf eval` under sudo, example user is named `oscap-eval-user`.
- Limit what arguments `eval` are accepted, tune for your solution
- Sudo currently must run `oscap` as `unconfined_t`. This is default, but rule just ensures it works even if SELinux user (`staff_u`) are used.
- Relax `requiretty` only for this user, so there is no issues, but not expose other users. 
- With NOEXEC, ensures oscap is not able to run external commands, I guess this needs to be removed if SCE is used.

```
Cmnd_Alias CMND_OSCAP = \
	/usr/bin/oscap xccdf eval *, \
	! /usr/bin/oscap * --fetch-remote-resources *, \
	! /usr/bin/oscap * --remediate *, \
	! /usr/bin/oscap * --skip-signature-validation *, \
	! /usr/bin/oscap * --skip-valid *, \
	! /usr/bin/oscap * --skip-validation *
Defaults:oscap-eval-user !requiretty
oscap-eval-user ALL = ( root ) ROLE=unconfined_r NOEXEC: NOPASSWD: CMND_OSCAP
```

### DIFF
--- a/utils/oscap-ssh
+++ b/utils/oscap-ssh
@@ -303,7 +303,7 @@ echo "Starting the evaluation..."
 # changing directory because of --oval-results support. oval results files are
 # dumped into PWD, and we can't be sure by the file names - we need controlled
 # environment
-ssh_execute_with_command_and_options "cd $REMOTE_TEMP_DIR; umask 022; ${OSCAP_SUDO[@]} scap $(command_array_to_string "${oscap_args[@]}")" "$SSH_TTY_ALLOCATION_OPTION"
+ssh_execute_with_command_and_options "cd $REMOTE_TEMP_DIR || exit 1; umask 022; ${OSCAP_SUDO[@]} scap $(command_array_to_string "${oscap_args[@]}")" "$SSH_TTY_ALLOCATION_OPTION"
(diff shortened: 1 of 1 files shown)


maage:
Two test enva failed at `Start 199: probes/fwupdsecattr/test_probes_fwupdsecattr_mock.sh`. I don't see connection with this patchset and that tests.

evgenyz:
Hey, @maage. Can you please rebase the PR?

Mab879:
No response in over 2 years, closing.

PR #1885, opened 2022-09-12
ASAN, UBSAN, compiler warning fixes

maage:
Assortment of changes I saw after I enabled ASAN and UBSAN.

Target here was for me to enable sanitizers and see that ctest passes and then see that my tailored DS passes too. I was not able to make ASAN to work with some plugins so I needed to modify testing etc to skip those.


jan-cerny (approved):

maage:
Thanks

PR #1886, opened 2022-09-12
Test case skipping if setup is not full vm

maage:
Skip / tune tests
- if no internet access
- if no full systemd environment
- if chroot
- check SELinux using selinuxenabled

### DIFF
--- a/tests/DS/test_ds_misc.sh
+++ b/tests/DS/test_ds_misc.sh
@@ -251,6 +251,9 @@ function test_ds_continue_without_remote_resources() {
 }
 
 function test_ds_error_remote_resources() {
+	# --fetch-remote-resources uses internet
+	require_internet || return 255
+
 	local DS="${srcdir}/$1"
 	local PROFILE="$2"
 	local result=$(mktemp)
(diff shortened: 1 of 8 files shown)


maage:
Seems there is some issue with failed checks: `error: %changelog entries must start with *`.

Maybe issue is that test environment does not handle `%autochangelog`. 

https://download.copr.fedorainfracloud.org/results/packit/OpenSCAP-openscap-1886/centos-stream-8-x86_64/04863323-openscap/openscap.spec
```
%changelog
%autochangelog
```
Maybe `%autochangelog` is not supported in those distros? Feature is quite new, it seems.

https://fedoraproject.org/wiki/Changes/rpmautospec

I think in test context it provides very little if any value. Maybe drop it in spec and try again.

jan-cerny:
Yes, I have noticed the same thing, originally I thought it's a temporary problem, because it used to work before, but now I can see this in all recent PRs.

jan-cerny (approved):
Thanks!

maage:
Thanks

Issue #1890, opened 2022-10-31
xmlfilecontent probe produces invalid OVAL results


jan-cerny:
#### Description of Problem:
During productization of ComplianceAsCode on 2022-10-31 (as of ComplianceAsCode upstream head 4b5551f) we have found that OpenSCAP produces invalid OVAL results.

#### OpenSCAP Version:
openscap-1.3.6-4.el9.x86_64.rpm

#### Operating System & Version:
RHEL 9.1

#### Steps to Reproduce:

1. oscap xccdf eval         --profile xccdf_org.ssgproject.content_profile_ism_o --report  xccdf_org.ssgproject.content_profile_ism_o.html         --results-arf xccdf_org.ssgproject.content_profile_ism_o-xccdf-arf-results.xml         --results xccdf_org.ssgproject.content_profile_ism_o-xccdf-results.xml --oval-results         /tmp/ssg-rhel9-ds.xml
2. oscap oval validate --results --schematron /var/lib/libvirt/images/ism_o/ssg-rhel9-oval.xml.result.xml

#### Actual Results:

OVAL results are invalid

```
File '/var/lib/libvirt/images/ism_o/ssg-rhel9-oval.xml.result.xml' line 65535: Element '{http://oval.mitre.org/XMLSchema/oval-system-characteristics-5#independent}value_of', attribute 'datatype': [facet 'enumeration'] The value '' is not an element of the set {'binary', 'boolean', 'evr_string', 'fileset_revision', 'float', 'ios_version', 'int', 'ipv4_address', 'ipv6_address', 'string', 'version'}.
OpenSCAP Error: Invalid OVAL Results (5.11) content in /var/lib/libvirt/images/ism_o/ssg-rhel9-oval.xml.result.xml. [/builddir/build/BUILD/openscap-1.3.6/src/source/oscap_source.c:353]
```

Notice the `value_of` element's attribute `datatype`:
```
79191           <ind-sys:xmlfilecontent_item id="105091439" status="does not exist">
79192             <ind-sys:filepath>/usr/lib/firewalld/zones/trusted.xml</ind-sys:filepath>
79193             <ind-sys:path>/usr/lib/firewalld/zones</ind-sys:path>
79194             <ind-sys:filename>trusted.xml</ind-sys:filename>
79195             <ind-sys:xpath>/zone/service[@name='ssh']</ind-sys:xpath>
79196             <ind-sys:value_of datatype="" status="does not exist"/>
[...]

PR #1891, opened 2022-11-02
Don't emit items if XPath doesn't match

jan-cerny:
This commit fixes the behavior of the xmlfilecontent probe in situation when the XPath query in xmlfilecontent_object doesn't match any node in the given XML file and the query returns an empty node set. Currently, in this situation, we emit an item in which we add an empty value_of element. However, this value_of element has its datatype attribute set to an empty string, which is invalid according to the OVAL schema. When we try to make the OVAL results valid, we face the problem that it isn't clear what should be the value of the datatype attribute for empty elements. But as we can realize the XPath doesn't match anything means that the requested object doesn't exist on the system, so a better behavior would be to not produce a xmlfilecontent54_item.  That is consistent with eg. situation when a regular expression matched nothing in textfilecontent54_object.  This commit therefore stops the item generation in this situation.

This commit also extends the existing test to cover the situation of XPath queries for nonexistent element and nonexistent attribute.

Fixes: #1890, rhbz#2138884, rhbz#2139060

### DIFF
--- a/tests/probes/xmlfilecontent/test_xmlfilecontent_probe.sh
+++ b/tests/probes/xmlfilecontent/test_xmlfilecontent_probe.sh
@@ -6,9 +6,15 @@ set -e -o pipefail
 cp $srcdir/example.xml /tmp/
 result=$(mktemp)
 $OSCAP oval eval --results $result $srcdir/test_xmlfilecontent_probe.xml
+# Even if OSCAP_FULL_VALIDATION is set, an invalid OVAL result doesn't cause
+# the "oscap oval eval" to return a non-zero value, so let's run validation
+# as a separate command
+$OSCAP oval validate "$result"
 assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:1" and @result="true"]'
 assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:2" and @result="true"]'
 assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:3" and @result="true"]'
 assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:4" and @result="true"]'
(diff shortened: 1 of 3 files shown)


tests/probes/xmlfilecontent/test_xmlfilecontent_probe.sh
```diff
@@ -6,9 +6,15 @@ set -e -o pipefail
...
+# as a separate command
+$OSCAP oval validate "$result"
 assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:1" and @result="true"]'
 assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:2" and @result="true"]'
 assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:3" and @result="true"]'
 assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:4" and @result="true"]'
 assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:5" and @result="true"]'
+assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:6" and @result="true"]'
+assert_exists 1 '/oval_results/results/system/definitions/definition[@definition_id="oval:x:def:7" and @result="true"]'
 rm -f $result
```
evgenyz, line 20: Could you also please add `newline` while you're at it?

jan-cerny:
Originally, there were infrastructure problems in testing farm. But now, I think that we have a real issue in our test suite. I have reported the problem in https://github.com/OpenSCAP/openscap/issues/1897.

jan-cerny:
I have rebased this PR on the top of the latest upstream maint-1.3 and now we can watch the CI results.

jan-cerny:
I have rebased this PR on the top of the latest upstream maint-1.3.

The issue with OS X seems to be in the `brew` step and shouldn't be related to changes in this PR.

evgenyz (approved):

PR #1894, opened 2022-11-11
fix: oval_cmp_evr_string: fixed undefined reference to risdigit when rpm is not available.

vogoltsov:
If `HAVE_RPMCMP` is not defined, build fails with ``ld: src/libopenscap.so.25.2.0: undefined reference to `risdigit'``.

### DIFF
--- a/src/OVAL/results/oval_cmp_evr_string.c
+++ b/src/OVAL/results/oval_cmp_evr_string.c
@@ -50,11 +50,12 @@
 #include <alloca.h>
 #endif
 static int rpmvercmp(const char *a, const char *b);
+#endif
+
 static int risdigit(int c) {
 	// locale independent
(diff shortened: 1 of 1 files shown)


evgenyz:
No, it ain't gonna fly this way. It has to be guarded properly in case librpm is present.

vogoltsov:
You're correct about `risdigit` being present in `librpm`.

The problem I'm encountering is building `openscap` within yocto. I will look into why the `HAVE_RPMVERCMP` macro is set, but no `risdigit()` is available.

vogoltsov:
Okay, totally my bad. Seems the issue was with absent include of `rpm/rpmstring.h` and has been fixed in 8dec1bb5e9546e75ae6e7b7cf94cf00197ce3e5d. Thank you!

Closing this PR.

PR #1895, opened 2022-11-15
Allow uname probe in offline mode

akarim22:
Signed-off-by: akarim22 <akarim@microsec.ai>

### DIFF
--- a/src/OVAL/probes/probe-table.c
+++ b/src/OVAL/probes/probe-table.c
@@ -331,7 +331,7 @@ static const probe_table_entry_t probe_table[] = {
 	{OVAL_UNIX_SYSCTL, NULL, sysctl_probe_main, NULL, NULL},
 #endif
 #ifdef OPENSCAP_PROBE_UNIX_UNAME
-	{OVAL_UNIX_UNAME, NULL, uname_probe_main, NULL, NULL},
+	{OVAL_UNIX_UNAME, NULL, uname_probe_main, NULL, uname_probe_offline_mode_supported},
 #endif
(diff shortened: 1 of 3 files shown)


evgenyz:
So, what data this probe (based on `sysinfo` call) would collect in the offline mode?

akarim22:
> 

it's sysname, nodename etc
`(gdb) p buf
$1 = {sysname = "Linux", '\000' <repeats 59 times>, nodename = "akarim-IdeaPad-3-15ITL05", '\000' <repeats 40 times>, release = "5.15.0-52-generic", '\000' <repeats 47 times>, 
  version = "#58-Ubuntu SMP Thu Oct 13 08:03:55 UTC 2022", '\000' <repeats 21 times>, machine = "x86_64", '\000' <repeats 58 times>, domainname = "(none)", '\000' <repeats 58 times>}
`
This is what i got from debugging it. I am not very much familiar with internals of it. but i kept getting these warnings
`Requested offline mode is not supported by uname probe` 
so i looked into other probes e.g. file and tried to mimic same for uname probe. 

any guidance is highly appreciated. TIA

evgenyz:
Well, the problem here is that the information would be collected from the *host* system and would not be entirely relevant to the system that being evaluated (defined with `OSCAP_PROBE_ROOT`). Look at the probes that support offline mode to understand what it really means.

evgenyz:
Can you please provide a bit more context to the problem? 

What kind of content you're trying to evaluate? Why uname probe is used there? Where does it come from? What is the target?

evgenyz:

Issue #1897, opened 2022-11-25
Test tests/probes/process58/command_line_extra.sh hangs infinitely


jan-cerny:
#### Description of Problem:
During review of the PR https://github.com/OpenSCAP/openscap/pull/1891 we have experienced issues with testing-farm jobs. There were initially some infrastructure problems, but after the infrastructure was fixed and the jobs were rerun, we have discovered that the test fail because of timeout. However, the timeout is very generous - it's 3 hours. So the test didn't finish in time. 

I have executed the test suite locally and I reproduced the issue, the tests were taking huge amount of time. They were stuck at 
tests/probes/process58/command_line_extra.sh for more than 30 minutes. This test never completed.

#### OpenSCAP Version:
current upstream maint-1.3 branch as of 2022-11-25 as of HEAD d10c40e43e1c627912374b8fbdfa1a84967fcc92

#### Operating System & Version:
F36, F 37

#### Steps to Reproduce:

1. cd build
2. cmake .. && make
3. ctest --verbose -R probes/process58/command_line_extra.sh

#### Actual Results:
the test is stuck and never finishes

see the full output with `set -x` at this link: https://pastebin.com/m6Y79Kcr

#### Expected Results:
the test passes and is completed in a finite time

#### Additional Information / Debugging Steps:
It's stuck in the `clean_processes` function which is a cleanup function executed in `trap` call when the test script process exits.

When the `trap` is commented out, it's stuck in the `exit` call.

evgenyz:
It is reproducible only on F37?

jan-cerny:
I have tried only F 37 but from the testing farm results in PR #1891 we can see that now it happens also on F 36 and rawhide systems.

jan-cerny:
I have successfully reproduced this on F 36 virtual machine. I have updated the issue description, I also added there a link to the output of `set -x` and I have also add the second additional information to the "Additional Information / Debugging Steps" section.

jan-cerny:
It reminds me this famous commit :smiley: https://github.com/openscap/openscap/commit/07bd511cc86c66c857b7379933e7c9d2515feded

jan-cerny:
> I wonder if we could instead of juggling zombies use the OpenSCAP process itself for testing command line.

I think the point of the test is to verify whether oscap correctly handles the special case of zombie process.

[...]

PR #1899, opened 2022-11-29
Fix stuck test command_line_extra.sh

jan-cerny:
Use correct command line, on some systems, the zombied process isn't "bash" but it can be "sleep" so we need to check for both these names.

Fixes: #1897


jan-cerny:
so now the broken jobs started to be fine but the OS X stopped to build

jan-cerny:
The issue with OS X will be addressed at https://github.com/OpenSCAP/openscap/pull/1900

evgenyz (approved):

PR #1900, opened 2022-11-30
Install pcre on OS X CI job

jan-cerny:
With the latest version of OS X image in GitHub CI, the pcre library isn't present by default but needs to be installed explicitely. The pcre library is mandatory for the build.


jan-cerny:
The issue with stuck test command_line_extra.sh isn't caused by this PR and is already tracked by issue #1897 .

evgenyz (approved):

evgenyz:
Gating / Build, Test on MacOS X Latest is not failing.

PR #1901, opened 2022-12-19
oval_cmp_evr_string: Make epoch comparison less strict for dpkg

dodys:
dpkg by default sets epoch to 0 if no epoch is present.


evgenyz:
Hi! It would be nice to have a test that covers the difference.

dodys:
> Hi! It would be nice to have a test that covers the difference.

I added some tests, let me know what you think about it.

I also think the unittest test_evr_string_missing_epoch.oval.xml should be re-made to actually use evr_string with rpminfo_object instead of dpkginfo_object.

evgenyz:
Sorry for the delay, I'm back and going to review it soon.

evgenyz (approved):
I've played with this change a bit and can confirm that it as good as it comes. Just one small request: can you please re-base the PR so that we could have all gating tests green here.

dodys:
> I've played with this change a bit and can confirm that it as good as it comes. Just one small request: can you please re-base the PR so that we could have all gating tests green here.

done!

PR #1903, opened 2022-12-19
tests: Rename blacklist to denylist

evgenyz:
There are a lot of references to blacklist in sample DS files, coming mostly from modprobe. Since we only need these sample files to test parsing capabilities the content itself doesn't have to be real-life bound.

So, despite the fact that modprobe does not have denylist as an alias for the blacklist keyword we can change it in our sample DS files.


matejak (approved):
I agree, moreover the test content is bloated, so it probably doesn't matter at all whether those rules and groups are part of it.

PR #1904, opened 2022-12-19
Rename blacklist in tests

evgenyz:
Rename misleading `whitelisted` variable in `tests/codestyle/all.sh`.

Rename `SYSCTL_BLACKLIST*` variables in `tests/probes/sysctl/test_sysctl_probe_all.sh`


matejak (approved):
LGTM, thanks for improving other naming as well.

PR #1905, opened 2022-12-20
tests: Rename master/slave keywords in sample configuration entries for bind nameserver

evgenyz:
The bind nameserver has `primary` and `secondary` synonymic keywords.


matejak:
As long as the change is consistent in both sides (input and subjects) of the test data, it even doesn't matter what is the real behavior of the package in question, because it is just the scanner that we are testing. As it turns out, the test data were too extensive, as that section was not used at all, so only the input (scanning content) needed to be altered.
Tests pass, so good job!

matejak (approved):

PR #1912, opened 2023-01-03
Update Lua to a newer version

jan-cerny:
The DEB package lua50 isn't available in Ubuntu 22.04.

This fixes the broken CI gating on Ubuntu machines.


evgenyz:
Great, thanks!

evgenyz (approved):

PR #1914, opened 2023-01-04
Remove Windows from release Guide

jan-cerny:
According to https://github.com/OpenSCAP/openscap/blob/maint-1.3/docs/windows.md we don't produce release builds.


evgenyz:
Nice catch.

evgenyz (approved):

PR #1916, opened 2023-01-11
Fix error when processing OVAL filters

jan-cerny:
This commit fixes this problem: We have an OVAL `<object>`, this object contains a `<filter>` that is specified by a `<state>`. This `<state>` in one of its child elements references a `<local_variable>`. This `<local_variable>` references a second `<object>`. If the second `<object>` describes an entity that doesn't exist on the system, an error message "Failed to convert OVAL state to SEXP" occurs.

If an `<object>` referenced from a `<state>` used in `<filter>` doesn't exist, the filter shouldn't apply, which is already happening in the current code.

This fix doesn't change the results of the OVAL evaluation, it only prevents triggering the aforementioned error message.

The fix makes the processing of variable references in `oval_state_to_sexp` consistent with the processing of variable references in `oval_object_to_sexp`.

Resolves: rhbz#2126882
Resolves: rhbz#2126883


jan-cerny:
I can't reproduce the CI fail in test_ds_misc.sh on Fedora latest locally. So I have temporary add set -x to the failing test to get more information.

jan-cerny:
After removing what I added, the problem stopped.

jan-cerny:
I have tested this PR using the steps by the reporter in the BZ description on a RHEL 8.7 virtual machine using the build from the packit CI in this PR and the error message doesn't occur.

evgenyz (approved):

PR #1921, opened 2023-02-01
Fix -Wint-conversion error building with clang

tstellar:
openscapPYTHON_wrap.c:3734:14: error: incompatible integer to pointer conversion returning 'int' from a function with result type 'char *' [-Wint-conversion]
      return 1;
             ^

It seems like return 1 is a copy-and-paste error and the function should return NULL here.  It's possible this function (sub_callback_wrapper) is unused and can be deleted but I wasn't sure.


jan-cerny (approved):
amazing, thanks!

PR #1922, opened 2023-02-01
Fix compile error with future versions of gcc

tstellar:
Future versions of GCC (estimated version is 14) will have the behavior similar to setting the flag `-Werror=implicit-function-declaration`.
Related bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91092

/builddir/build/BUILD/openscap-1.3.6/src/OVAL/probes/unix/linux/rpminfo_probe.c: In function 'rpminfo_probe_fini': /builddir/build/BUILD/openscap-1.3.6/src/OVAL/probes/unix/linux/rpminfo_probe.c:307:9: error: implicit declaration of function 'rpmFreeCrypto'
  307 |         rpmFreeCrypto();
      |         ^~~~~~~~~~~~~

### DIFF
--- a/src/OVAL/probes/unix/linux/rpm-helper.h
+++ b/src/OVAL/probes/unix/linux/rpm-helper.h
@@ -30,6 +30,7 @@
 #include <rpm/rpmts.h>
 #include <rpm/rpmmacro.h>
 #include <rpm/rpmlog.h>
+#include <rpm/rpmcrypto.h>
 #include <rpm/header.h>
 
 #include <pthread.h>


evgenyz:
@tstellar Can you please elaborate on what this future version of gcc is (version, disto version where it would be introduced etc). Also it would be very helpful to know what change in gcc is causing this error.

evgenyz:
> I hope we can make the change in GCC 14, to be released in about a year.

Great, thanks for the heads-up. We'll plan the fix release accordingly.

Can you please also give a link to the description of the change itself? (Dev PR, RFC, etc)

evgenyz:
Superseded by #1967

jan-cerny:
I hope I understand it well that this PR has been superseded by #1967 which is merged so it can be closed.

PR #1929, opened 2023-02-10
Closes #1928 gh-action to run shellcheck static analyzer

### LINKED ISSUE #1928: please add shellcheck github action
When I scanned the Ubuntu package openscap-1.3.7+dfsg-1 I found many shellcheck notes, warnings and errors. The `error`s I examined appeared serious and cause OpenSCAP to not work properly. `note`s do not seem significant, but some `warning`s are. They need their own triage and possibly assigned issues or CVEs. Here is the output of my local shellcheck scan: https://gist.github.com/eslerm/1430755c97570e46462d03c050e48c24

I would like to propose adding a shellcheck GitHub action to track these and new results.

eslerm:
This is a minimal gh-action to run shellcheck.

Please see the ShellCheck step for example output: https://github.com/eslerm/openscap/actions/runs/4147791418/jobs/7175079791

shellcheck exits with an error if issues are found.

If the PR is accepted I'll add false positives for shellcheck to ignore.

I set the severity level to `warning`, since `note`s are mostly noise.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,25 @@
+name: "ShellCheck"
+
+# please read https://github.com/ludeeus/action-shellcheck 
+# to configure output
+# this tool is more useful when false positives are supressed
+
+on:
+  push:
+    branches: [ 'maint-1.3' ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ 'maint-1.3' ]
+  schedule:
+    - cron: '32 17 * * 0'
+
+jobs:
+  shellcheck-analyze:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jan-cerny:
Hi @eslerm I agree that shellcheck is very useful. But I'm afraid that now when the code is full of problems and the job is red people will ignore it. Ideally we should make it green or configure in a way that it would show the new issues in the PRs.

eslerm:
Hi @jan-cerny that sounds reasonable to me.

This GitHub Action might be more useful if it always passed, but saved reports somewhere. I'm not sure if or how you would want that implemented though.

Mostly I wanted to bring these reports to your attention, and I had hoped these PRs would be more helpful than a deluge of reports :D

PR #1930, opened 2023-02-14
Remove reference to PROC_CHECK

0intro:
PROC_CHECK is not defined anymore, since the removal of automake.


jan-cerny (approved):
Nice catch! Thank you

jan-cerny:
The CI fail in testing-farm:fedora-38-x86_64 is caused by infrastructure issue related to f38 branching and isn't caused by the contents of this PR.

PR #1933, opened 2023-02-17
Fix leak of session->skip_rules

0intro:
The list skip_rules is allocated in xccdf_session_new_from_source, but was not freed in xccdf_session_free.


jan-cerny:
I'm trying to investigate the CI problems but they seem unrelated to the contents of this PR.

jan-cerny:
The CI fail in testing-farm:fedora-38-x86_64 is caused by infrastructure issue related to f38 branching and isn't caused by the contents of this PR. The OS X fail is also not related to this PR and will be fixed in https://github.com/OpenSCAP/openscap/pull/1936.

PR #1934, opened 2023-02-20
Fix leak of dpkginfo_reply_t fields

0intro:
The arch, epoch, releasem version and evr strings are allocated in dpkginfo_get_by_name, but were not freed in dpkginfo_free_reply.


jan-cerny (approved):

jan-cerny:
The CI fail in testing-farm:fedora-38-x86_64 is caused by infrastructure issue related to f38 branching and isn't caused by the contents of this PR. The OS X fail is also not related to this PR and will be fixed in https://github.com/OpenSCAP/openscap/pull/1936.

PR #1935, opened 2023-02-20
Fix codeql workflow

paulcacheux:
Fix the install deps step, sourcing package from https://github.com/OpenSCAP/openscap/blob/maint-1.3/docs/developer/developer.adoc#building-openscap-on-linux.

Notes:
- `python-dev` -> `python2-dev`
- fixing this will create a lot of new alerts, this is out of scope for this PR

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -32,10 +32,9 @@ jobs:
     - name: Install Deps
       run: |
         sudo apt-get update
-        sudo apt-get upgrade
         sudo apt-get install -y cmake libdbus-1-dev libdbus-glib-1-dev libcurl4-openssl-dev \
           libgcrypt20-dev libselinux1-dev libxslt1-dev libgconf2-dev libacl1-dev libblkid-dev \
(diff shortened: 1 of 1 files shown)


evgenyz:
@paulcacheux  Why `python2-dev`? It could (and IMHO should) be built with Python 3.

paulcacheux:
@evgenyz I simply followed the instructions provided in https://github.com/OpenSCAP/openscap/blob/maint-1.3/docs/developer/developer.adoc#1-building-openscap-on-linux subsection "On Ubuntu 16.04, Debian 8 or Debian 9" (the only one talking about ubuntu), and changed `python-dev` to `python2-dev` because on ubuntu 16.04 python is python2.
I think it could be built with python3 as well, but this should be explicit in the documentation first

jan-cerny:
@evgenyz I'm confused, is this PR still needed or do the recent changes in the workflow file fix the same thing?

paulcacheux:
I think the recent changes are enough, closing this

PR #1936, opened 2023-02-20
Fix OS X build

jan-cerny:
OS X build can't find pkg-config. For example:
https://github.com/OpenSCAP/openscap/actions/runs/4202194906/jobs/7332722182

Addressing:
- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,6 +93,7 @@ jobs:
         brew update
         brew install doxygen
         brew install opendbx
+        brew install pkg-config
         brew install popt
         brew install swig
         brew install libxmlsec1


ggbecker (approved):
Interesting how this line was already present in scap-workbench's openscap build: https://github.com/OpenSCAP/scap-workbench/pull/302/files#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R54

ggbecker:
I can see that the centos9 package was built properly: https://copr.fedorainfracloud.org/coprs/packit/OpenSCAP-openscap-1936/build/5548392/

The gh action doesn't seem to get updated with the results.

CodeQL checks are being fixed on a different PR: https://github.com/OpenSCAP/openscap/pull/1936

@jan-cerny Do you think that the Fedora 38 tests are something expected or should be addressed first? https://artifacts.dev.testing-farm.io/dba9443a-0143-497c-96a4-f91d28981da5/

something is wrong with cmake:

```
:: [ 13:46:09 ] :: [  BEGIN   ] :: Running 'cmake --build  --target test -- ARGS='--output-on-failure''
Usage: cmake --build <dir>             [options] [-- [native-options]]
       cmake --build --preset <preset> [options] [-- [native-options]]
Options:
  <dir>          = Project binary directory to be built.
  --preset <preset>, --preset=<preset>
                 = Specify a build preset.
  --list-presets[=<type>]
                 = List available build presets.
  --parallel [<jobs>], -j [<jobs>]
                 = Build in parallel using the given number of jobs. 
                   If <jobs> is omitted the native build tool's 
                   default number is used.
                   The CMAKE_BUILD_PARALLEL_LEVEL environment variable
                   specifies a default parallel level when this option
                   is not given.
  -t <tgt>..., --target <tgt>...
                 = Build <tgt> instead of default targets.
  --config <cfg> = For multi-configuration tools, choose <cfg>.
  --clean-first  = Build target 'clean' first, then build.
                   (To clean only, use --target 'clean'.)
  --resolve-package-references={on|only|off}
                 = Restore/resolve package references during build.
  -v, --verbose  = Enable verbose output - if supported - including
                   the build commands to be executed. 
  --             = Pass remaining options to the native tool.
:: [ 13:46:09 ] :: [   FAIL   ] :: Command 'cmake --build  --target test -- ARGS='--output-on-failure'' (Expected 0, got 1)
```

jan-cerny:
@ggbecker I have seen the similar problem with the Fedora 38 tests in multiple other recent PRs. I think that it isn't caused by the contents of this PR because it only changes the OS X CI play settings. I have chatted about the issue with @matusmarhefka and we think that it's an infrastructure problem and that it's related to recent branching of Fedora 38 from Rawhide. The root cause is that Testing farm can't access the Fedora 38 build from Packit, but the build in Packit for Fedora 38 has finished OK. I propose to ignore this test for now and hopefully it will get fixed itself.

PR #1937, opened 2023-02-20
Fix leak of component_colls in _oval_component_evaluate_CONCAT

0intro:
component_colls are allocated in _oval_component_evaluate_CONCAT, but were not freed when not_finished is null.


evgenyz (approved):
In case an OVAL component has sub-components without values in them. Unlikely, but definitely possible. Makes sense, thank you.

PR #1941, opened 2023-02-23
Fix build on FreeBSD.

cschuber:
Now that FreeBSD has imported LLVM15, new errors are flagged.

Ideally the code that implicily converts integers should be fixed but a first step is to use a compiler option that ignores the errors like prior versions of LLVM and GCC do.

Additionaly, as in Linux, read_common_sizet() is also undefined in FreeBSD.

### DIFF
--- a/src/common/memusage.c
+++ b/src/common/memusage.c
@@ -56,7 +56,7 @@
 #include "memusage.h"
 #include "bfind.h"
 
-#if defined(OS_LINUX)
+#if defined(OS_LINUX) || defined(__FreeBSD__)
 static int read_common_sizet(void *szp, char *strval)
 {
 	char *end;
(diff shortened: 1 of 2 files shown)


src/common/memusage.c
```diff
@@ -56,7 +56,7 @@
 #include "memusage.h"
 #include "bfind.h"
 
-#if defined(OS_LINUX)
+#if defined(OS_LINUX) || defined(__FreeBSD__)
```
evgenyz, line 59: Can you please add Solaris as well to this fix?

```suggestion
#if defined(OS_LINUX) || defined(__FreeBSD__) || defined(OS_SOLARIS)
```

evgenyz: See: https://github.com/OpenSCAP/openscap/issues/1945

orcl-jlana: I can confirm that the fix works for the Solaris operating system

PR #1953, opened 2023-03-07
Include correct header for rpmFreeCrypto()

havardAasen:
In RPM version < 4.18, `rpmFreeCrypto()` was located in _rpmgpg.h_. This function and various other generic crypto API has now been moved to _rpmcrypto.h_.

Since the header doesn't exists in earlier version, it was wrapped in an ifdef, to retain compatibility with older version of RPM.

This can of course also be moved into the header _rpm-helper.h_ if you prefer.

### DIFF
--- a/src/OVAL/probes/unix/linux/rpminfo_probe.c
+++ b/src/OVAL/probes/unix/linux/rpminfo_probe.c
@@ -60,6 +60,9 @@
 #include <regex.h>
 
 /* RPM headers */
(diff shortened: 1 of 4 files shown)


evgenyz:
Superseded by https://github.com/OpenSCAP/openscap/pull/1967

evgenyz:
The `rpmgpg.h` is still the way to include `rpmFreeCrypto` as far as I see it.

PR #1954, opened 2023-03-07
GH Actions: Fix deprecation warnings and build on Ubuntu

havardAasen:
GH Actions: Bump actions/checkout to version 3

Version 2 is based on Node.js 12 and is deprecated. Version 3 is based
on Node.js 16. There are some minor changes but none that affects
OpenSCAP. The action for 'CodeQL' already uses version 3.

---
GH Actions: Stop upgrading packages on Ubuntu

Don't upgrade packages before we install new ones. This patch also
replaces an environment variable for a command line option for the
'apt-get' command.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,13 +32,10 @@ jobs:
 
     # Runs a single command using the runners shell
     - name: Install Deps
-      env:
-        ACCEPT_EULA: Y
       run: |
         sudo apt-get update
-        sudo env ACCEPT_EULA=Y apt-get upgrade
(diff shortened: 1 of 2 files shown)


havardAasen:
I initially only wanted to fix the deprecation warning generated by GH Actions, but one thing led to another. I now see that part of the last two commits is similar to what has been proposed with #1935.

evgenyz:
Can you please also get rid of Ubuntu 18 on top of these changes (it'll disappear from GH in one month). And turn off JavaScript in CodeQL. Otherwise LGTM.

evgenyz (approved):
LGTM. Thank you!

Issue #1942, opened 2023-02-28
oscap-podman and oscap-docker evaluate all rules as notapplicable


jan-cerny:
#### Description of Problem:
When scanning an UBI 8 image using oscap-podman or oscap-docker, all rules are evaluated as notapplicable.

This problem has been discovered during a review of https://github.com/OpenSCAP/openscap/pull/1931 but isn't caused by that PR.

#### OpenSCAP Version:
openscap-1.3.7-1.fc37.x86_64
podman-4.3.1-1.fc37.x86_64

#### Operating System & Version:
F 37

#### Steps to Reproduce:

with oscap-podman:

1. sudo podman pull registry.access.redhat.com/ubi8/ubi
2. sudo oscap-podman registry.access.redhat.com/ubi8/ubi:latest xccdf eval --profile xccdf_org.ssgproject.content_profile_stig /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml

with oscap-docker:

1. sudo systemctl start podman
2. sudo podman pull registry.access.redhat.com/ubi8/ubi
3. sudo oscap-docker image registry.access.redhat.com/ubi8/ubi:latest xccdf eval --profile xccdf_org.ssgproject.content_profile_stig /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml

#### Actual Results:
all rules are notapplicable

#### Expected Results:
some rules are applicable and evaluated

#### Additional Information / Debugging Steps:
```
[jcerny@thinkpad ~]$ sudo podman images registry.access.redhat.com/ubi8/ubi:latest
REPOSITORY                           TAG         IMAGE ID      CREATED      SIZE
registry.access.redhat.com/ubi8/ubi  latest      270f760d3d04  13 days ago  214 MB
```

candrews:
I figured out the cause of this problem and submitted a PR to fix it: https://github.com/ComplianceAsCode/content/pull/10320

candrews:
At https://github.com/ComplianceAsCode/content/pull/10320#issuecomment-1463847468 it was asked:
> Could the problem be instead related to the RPM DB in Fedora might be incompatible with the RPM DB in UBI 8?

I did some testing on my Fedora 37 system using:
```sh
ID="$(podman create $IMAGE)" && DIR="$(podman mount $ID)" && rpm --root "$DIR" -q bash && podman rm -f "$ID"
```

[...]

PR #1956, opened 2023-03-12
Always use /var/lib/rpm as rpm dbpath

candrews:
Fedora >=36 changed the default dbpath in librpm from /var/lib/rpm to /usr/lib/sysimage/rpm Therefore, when running openscap on a Fedora >=36 system scanning another systems (such as RHEL, SLES, Fedora<36) openscap's librpm will try to read the rpm db from /usr/lib/sysimage/rpm which doesn't exist and therefore won't work. In implementing this change, /var/lib/rpm is still a symlink to /usr/lib/sysimage/rpm so /var/lib/rpm still works. So /var/lib/rpm is a dbpath that will work on all systems. Therefore, set the dbpath to be /var/lib/rpm, allow openscap running on any system to scan any system.

See: https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr
Fixes: https://github.com/OpenSCAP/openscap/issues/1942

`cd build && cmake ../ && make -j4 && ctest` shows all tests as passing on my Fedora 37 system.

### DIFF
--- a/src/OVAL/probes/unix/linux/rpminfo_probe.c
+++ b/src/OVAL/probes/unix/linux/rpminfo_probe.c
@@ -294,6 +294,17 @@ void *rpminfo_probe_init(void)
 		return ((void *)g_rpm);
         }
 
+        /*
+        * Fedora >=36 changed the default dbpath in librpm from /var/lib/rpm to /usr/lib/sysimage/rpm
+        * See: https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr
+        * Therefore, when running openscap on a Fedora >=36 system scanning another systems (such as RHEL, SLES, Fedora<36)
+        * openscap's librpm will try to read the rpm db from /usr/lib/sysimage/rpm which doesn't exist and therefore won't work.
+        * In implementing this change, /var/lib/rpm is still a symlink to /usr/lib/sysimage/rpm
+        * so /var/lib/rpm still works. So /var/lib/rpm is a dbpath that will work on all systems.
+        * Therefore, set the dbpath to be /var/lib/rpm, allow openscap running on any system to scan any system.
+        */
+        rpmPushMacro(NULL, "_dbpath", NULL, "/var/lib/rpm", RMIL_CMDLINE);
+
 	g_rpm->rpmts = rpmtsCreate();
 	pthread_mutex_init (&(g_rpm->mutex), NULL);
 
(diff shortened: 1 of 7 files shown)


candrews:
This change is a different approach to the attempt to fix this issue made in commit b0d013aedab978e7d945ae8016e0df0a9dd0bf68 by @matusmarhefka

jan-cerny:
Thanks for the PR! It looks great, but unfortunately, the tests are failing. I'm trying to investigate the fails and I will let you know once I will have any results.

candrews:
Phenomenal! I've updated this MR with that change; I'm eagerly awaiting the "all tests pass" result :+1:

candrews:
@jan-cerny everything looks good now - shall this be merged? :crossed_fingers:

jan-cerny:
@candrews Thank you! We are almost there! The last failing CI job is the Ubuntu job where unfortunately we use some RPM related content for testing the generic data stream processing and the changes in this PR uncovered this. Ideally, we should make the failing tests distribution-agnostic. For this moment we can simply skip these tests on systems where RPM isn't installed. I was able to skip them  and make it green on Ubuntu by this change:

```diff
diff --git a/tests/DS/test_ds_misc.sh b/tests/DS/test_ds_misc.sh
index 4d3b1ec55..8f2b6cb96 100755
--- a/tests/DS/test_ds_misc.sh
+++ b/tests/DS/test_ds_misc.sh
@@ -77,6 +77,7 @@ sds_add_multiple_twice(){
 
 function test_eval {
     probecheck "rpminfo" || return 255
+    [ -e "/var/lib/rpm" ] || return 255
     local stderr=$(mktemp -t ${name}.out.XXXXXX)
     $OSCAP xccdf eval "${srcdir}/$1" 2> $stderr
     diff /dev/null $stderr; rm $stderr
diff --git a/tests/DS/test_sds_eval.sh b/tests/DS/test_sds_eval.sh
index fc9da3cc5..606bb9150 100755
--- a/tests/DS/test_sds_eval.sh
+++ b/tests/DS/test_sds_eval.sh
@@ -12,6 +12,7 @@ set -e -o pipefail
 
 function test_eval {
     probecheck "rpminfo" || return 255
+    [ -e "/var/lib/rpm" ] || return 255
     local stderr=$(mktemp -t ${name}.out.XXXXXX)
     $OSCAP xccdf eval "${srcdir}/$1" 2> $stderr
     diff /dev/null $stderr; rm $stderr
```

But I haven't tried if it works on CI.

candrews:
> But I haven't tried if it works on CI.

I added the change this PR, we'll soon find out!

candrews:
> we'll soon find out!

All checks have passed :partying_face:

PR #1957, opened 2023-03-17
Free xmlDoc structure at the end of xccdf_session_load

0intro:
When calling the xccdf_session_load_(xccdf|cpe|oval) functions, the session->source structure is recopied to
session->ds.session->component_sources, including the source->xml.doc field, which contains the XML DOM.

This change adds a new oscap_source_free_xmlDoc function to the OSCAP source API and call it at the end of the xccdf_session_load function, so session->source->xml.doc is freed when it's not used anymore.

### DIFF
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -716,7 +716,9 @@ int xccdf_session_load(struct xccdf_session *session)
 			return ret;
 		}
 	}
-	return xccdf_session_load_tailoring(session);
+	ret = xccdf_session_load_tailoring(session);
+	oscap_source_free_xmlDoc(session->source);
+	return ret;
 }
 
 static int _reporter(const char *file, int line, const char *msg, void *arg)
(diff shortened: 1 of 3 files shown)


jan-cerny:
The CI fail is legit. There is a segmentation fault when using tailoring files. It's reproducible locally, the failing test is `test_api_xccdf_tailoring_profile_generate_fix`. There is an invalid read in the function `ds_sds_lookup_datastream_in_collection`. I tried to investigate the problem but I haven't found any reason so far.

Is this patch trying to address a memory leak or is it only an intended optimization?

0intro:
This patch is an intended optimization.

The goal was to free the XML DOM from `source->xml.doc` after it was copied to `session->ds.session->component_sources`, to prevent having to keep two copies
of the XML DOM in memory while running the checks.

For example, when running the Ubuntu 22.10 checks, this change reduces the
memory footprint of the session by more than a third.

I'll take a look to the issue and try to find a proper solution.

0intro:
The issue should be fixed now.

jan-cerny (approved):
The test `API/XCCDF/tailoring/all.sh` now passes.

Thanks for the fix!

PR #1966, opened 2023-04-15
Add 'auristorfs' to list of network fileystems

kenh:
Auristor is a commercial version of the AFS filesystem and has type of 'auristorfs'.  Add it to the list of network fileystems so it is excluded when the scan is restricted to local filesystems.


kenh:
I know the network filesystem test is different on the 'main' branch and I was going to prepare a pull request for that as well.

evgenyz (approved):

evgenyz:
The fedora-rawhide-x86_64 test failure is irrelevant.

PR #1967, opened 2023-04-21
Fix implicitly declared function

evgenyz:
The function 'rpmFreeCrypto' is implicitly declared. This will be forbidden in future versions of GCC (the `-Werror=implicit-function-declaration` behaviour will be enforced).

On top of that, as an anti-regression measure, the flag `-Werror=implicit-function-declaration` is now part of the CMake config.

Follow-up on #1922. Kudos @tstellar and @fweimer-rh.


evgenyz:
Something got broken in systemd sessions in F38 (related to fwupd mock test). We can ignore it for this PR I guess.

jan-cerny (approved):

PR #1968, opened 2023-04-26
Add entry for XCCDF tailoring files in Schematron table

litios:
When validating an XCCDF tailoring file, oscap will complain about not finding the Schematron rules. 

This happens because the `OSCAP_DOCUMENT_XCCDF_TAILORING` is not listed in the table.

### DIFF
--- a/src/source/schematron.c
+++ b/src/source/schematron.c
@@ -107,6 +107,7 @@ struct oscap_schema_table_entry OSCAP_SCHEMATRON_TABLE[] = {
 	{OSCAP_DOCUMENT_OVAL_DIRECTIVES,        "5.11.2",       "oval/5.11.2/oval-directives-schematron.xsl"},
 	{OSCAP_DOCUMENT_OVAL_DIRECTIVES,        "5.11.3",       "oval/5.11.3/oval-directives-schematron.xsl"},
 	{OSCAP_DOCUMENT_XCCDF,                  "1.2",          "xccdf/1.2/xccdf_1.2-schematron.xsl"},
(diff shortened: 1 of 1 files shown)


jan-cerny:
Thanks for opening the PR. It definetely fixes the error message. But, the results of schematron validation seem to be quite off. For example, when running on a valid tailoring file:

```
[jcerny@thinkpad openscap{pr/1968}]$ build/oscap_wrapper xccdf validate tests/API/XCCDF/unittests/test_single_rule-tailoring.xml
<?xml version="1.0"?>
Error: A 'Profile' element with an @extends attribute must extend another 'Profile' element. 
Error: The given @idref attribute 'xccdf_com.example.www_rule_test-fail' must match a the @id or @cluster-id attributes of a 'Rule' or 'Group' element. See the XCCDF 1.2.1 specification, Section 6.5.3.

OpenSCAP Error: Invalid XCCDF Tailoring (1.2) content in tests/API/XCCDF/unittests/test_single_rule-tailoring.xml. [/home/jcerny/work/git/openscap/src/source/oscap_source.c:381]
```

In this example, the error message that a `Profile` element within `Tailoring` must extend another `Profile` element doesn't make sense because the profile that is being extended is defined in the outside file. Also, the check about `idref` doesn't make sense because the rule that is references is defined in the outside file.

Now, we need to investigate whether the schematron in `schemas/xccdf/1.2/xccdf_1.2-schematron.xsl` can actually be used on Tailoring files. On contrary, there are some XPaths in that file that explicitly match `Tailoring` elements.

evgenyz:

PR #1883, opened 2022-09-04
Build 1

maage:
Because my other PR got dinged by some other build issues, here is my take to sligghtly improve spec file and testing.

https://github.com/OpenSCAP/openscap/pull/1881

I tried to change some tests to skip in my `mock` build environment as it runs w/o network, systemd and is chrooted env.

Then I tried to make tests actually run parallel, but it was no go. Too many of them use shared file names like `test_file`. And beca
use they are not parametritized it was too much work. So I modified spec not to use parallel ctest runs.

I simplified bugs fixed under Fedora. As I was unable to test these under CentOS I used if block.

I converted `BuildRequires` to use pkgconfig and tried to minimize list.

As oscap is really complex C/C++ program I decided to compile it with ASAN/UBSAN and found couple of simple issues.

PR #1970, opened 2023-05-10
Build OpenSCAP with clang sanitizers

jan-cerny:
This PR will create a special COPR build where OpenSCAP will be built with the Clang compiler with special options.

The special option is a ThreadSanitizer https://clang.llvm.org/docs/ThreadSanitizer.html  is a tool that detects data races.

This way we will be able to use and download the build and use it instead of normal OpenSCAP package to debug issues like freezing or thread crashes. The `oscap` will work as usual but at the same time it will print warnings when a data race happens.

To get  meaningful warnings you should install also the openscap-debuginfo rpm package, you will get tracebacks with exact file names and line numbers of the problematic code.

I propose creating COPR builds for EPEL 8 and 9.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -12,16 +12,32 @@ srpm_build_deps:
 jobs:
 - job: copr_build
   trigger: pull_request
-  metadata:
-    targets:
-    - fedora-all-x86_64
-    - centos-stream-8-x86_64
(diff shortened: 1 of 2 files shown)


maage:
I modified my spec to have a part of sanitizers I hope are also in scope here, #1883

PR #1976, opened 2023-06-01
Add offline capabilities to the sysctl OVAL probe

evgenyz:
Although the probe collects the data from `/proc/sys/...` filesystem it is useful for the so-called hybrid mode (OCP, etc).

Note: There are a lot of changes to the indentation, but they all are concentrated in the first commit, it makes sense to review just the second.

### DIFF
--- a/src/OVAL/probes/unix/sysctl_probe.c
+++ b/src/OVAL/probes/unix/sysctl_probe.c
@@ -53,6 +53,7 @@
 #include <ctype.h>
 #include <errno.h>
 #include <limits.h>
+#include <dirent.h>
 #include "oval_fts.h"
 #include "common/debug_priv.h"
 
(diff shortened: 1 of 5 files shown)


jan-cerny:
@evgenyz

What is the reason for the indentation change? I also don't like that the result is 8 spaces because the project guidelines is a tab.

What will happen if the guest system doesn't have the proc directory? What is the difference in the behavior between usage in oscap-podman and OpenShift?

evgenyz:
> What is the reason for the indentation change? I also don't like that the result is 8 spaces because the project guidelines is a tab.

https://github.com/OpenSCAP/openscap/blob/maint-1.3/docs/contribute/contribute.adoc#code-style

The guideline is to follow the formatting unless you drastically refactor the module and change the whole file formatting. Unfortunately the file had a messy mix of tabs and spaces formatting. There were less tabs than spaces, but I can change everything to tabs if you prefer.

evgenyz:
> What will happen if the guest system doesn't have the proc directory? What is the difference in the behavior between usage in oscap-podman and OpenShift?

That's a good question. The answer is quite straightforward, but it made me thinking about what we'd really want the behaviour to be.

And now the answer is: it just won't yield any results like the `process58` probe. Also, It will throw a warning if the prefix (`OSCAP_PROBE_ROOT`) can't be opened in addition to the regular `PROBE_EFATAL`.

evgenyz:
@jan-cerny Ping.

src/OVAL/probes/unix/sysctl_probe.c
```diff
@@ -128,12 +129,21 @@ int sysctl_probe_main(probe_ctx *ctx, void *probe_arg)
...
+                if (prefix != NULL) {
+                        DIR *d = opendir(prefix);
+                        if (d != NULL) {
+                                closedir(d);
+                                return PROBE_ESUCCESS;
+                        }
+                        dW("Can't open prefix directory '%s': %s", prefix, strerror(errno));
+                }
+
+                dE("oval_fts_open_prefixed(%s, %s) (prefix: %s) failed", PROC_SYS_DIR, ".\\+", prefix);
```
jan-cerny, line 146: What is this `".\\+"` here?

jan-cerny (approved):

PR #1977, opened 2023-06-05
tests/DS: Fix DS misc test to not include a check for cURL error message

evgenyz:
The string representation of the cURL error message is not under our control and therefore should not be checked in our tests.

The test fails in current rawhide because cURL library changed the message.


matejak:
The test is problematic, because we assume that the server exists, it serves stuff, and returns 404. It could also return another http error, or even host not found error.

matejak (approved):

PR #1980, opened 2023-06-08
OVAL/probes/systemd*: Use ListUnitFiles D-Bus method to fetch all units

evgenyz:
The method `ListUnits` that was used before has a flaw: the unit won't be listed if is not enabled, masked or a member of other modules' conflicts/depends property.

The `ListUnitFiles` method, on the other hand, returns all the modules based on the files present in the respective directories, including disabled units.

Fix for:  https://github.com/ComplianceAsCode/content/issues/10424

The drawback (or a feature?) is that instead of templated unit instances it would return templates (`user-runtime-dir@1000.service` -> `user-runtime-dir.service`).


evgenyz:
Uh, Ubuntu 20.04 is not happy :(

mildas:
In machine hardening test case (harden running machine, reboot, scan), it passes. Service is correctly identified that it's not masked (fail -> remediate -> pass) now.

In kickstart test case (OAA hardening, boot, scan), it still fails. Service state is not found in OAA scan (pass -> fail). But that's expected in Anaconda env, right?

evgenyz:
> In kickstart test case (OAA hardening, boot, scan), it still fails. Service state is not found in OAA scan (pass -> fail). But that's expected in Anaconda env, right?

Yeah, in Anaconda we can't make it work ATM.

tests/common/test_oscap_util.c
```diff
@@ -0,0 +1,66 @@
...
+}
+
+int main (int argc, char *argv[])
+{
+	int retval = 0;
+
+	if ((retval = test_oscap_strrm()) != 0)
+		return retval;
+
+	return retval;
```
jan-cerny, line 65: It seems duplicate to have the 2 returns, there could be just 1 return and no if.

src/common/util.h
```diff
@@ -313,6 +313,24 @@ static inline char *oscap_strdup(const char *str) {
 #endif
 }
 
+/**
+ * Removes all occurrences of substr from str by shifting the contents to
+ * the left. If string is NULL, does nothing.
+ * @param str String we want to process
+ * @param substr Sub-string we want to get rid of
+ */
+static inline void oscap_strrm(char *str, const char *substr) {
```
jan-cerny, line 322: What is the reason for defining the function inline in a header file?

jan-cerny (approved):
I have seen the common/test_oscap_util.sh passed in the Fedora Latest GH CI job

tests/common/test_oscap_util.c
evgenyz: The code could look like this in the future:

```c

	int retval = 0;

	if ((retval = test_oscap_strrm()) != 0)
		return retval;

	if ((retval = test_oscap_strdup()) != 0)
		return retval;

	if ((retval = test_oscap_foobar()) != 0)
		return retval;

        ...

        return retval;
```

evgenyz:
@jan-cerny Ping.

PR #1989, opened 2023-06-21
CMakeLists.txt: make 2 variables configurable

kkang-wr:
Variables PREFERRED_PYTHON_PATH and PYTHON3_PATH are set with ${PYTHON_EXECUTABLE}. For cross compile, ${PYTHON_EXECUTABLE} may point to other path rather thab standard dir such as /usr/bin. Then the generated library file contains such path which should NOT. Update to make variables PREFERRED_PYTHON_PATH and PYTHON3_PATH configurable to avoid such issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,8 +125,8 @@ endif()
 find_package(PythonInterp 3)
 find_package(PythonLibs 3)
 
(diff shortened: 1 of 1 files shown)


kkang-wr:
It seems not work as expected.

PR #1990, opened 2023-06-21
v2: CMakeLists.txt: make 2 variables configurable

kkang-wr:
Variables PREFERRED_PYTHON_PATH and PYTHON3_PATH are set with ${PYTHON_EXECUTABLE}. For cross compile, ${PYTHON_EXECUTABLE} may point to other path rather than standard dir such as /usr/bin. Then the generated library file contains such path which should NOT. Update to make variables PREFERRED_PYTHON_PATH and PYTHON3_PATH configurable to avoid such issue.

v2:
* change variables' types to `PATH` and update docstring

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,8 +125,8 @@ endif()
 find_package(PythonInterp 3)
 find_package(PythonLibs 3)
 
-set(PREFERRED_PYTHON_PATH "${PYTHON_EXECUTABLE}")
-set(PYTHON3_PATH "${PYTHON_EXECUTABLE}")
+set(PREFERRED_PYTHON_PATH "${PYTHON_EXECUTABLE}" CACHE PATH "Path to preferred Python")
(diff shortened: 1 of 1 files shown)


jan-cerny (approved):
I have checked the generated files:

```
[jcerny@fedora build{pr/1990}]$ grep -r PREFERRED_PYTHON_PATH
.cmake/api/v1/reply/cache-v2-55fe1ac3834f256a286f.json:			"name" : "PREFERRED_PYTHON_PATH",
config.h:#define PREFERRED_PYTHON_PATH "/usr/bin/python3"
CMakeCache.txt:PREFERRED_PYTHON_PATH:PATH=/usr/bin/python3
[jcerny@fedora build{pr/1990}]$ grep -r PYTHON3_PATH
.cmake/api/v1/reply/cache-v2-55fe1ac3834f256a286f.json:			"name" : "PYTHON3_PATH",
config.h:#define PYTHON3_PATH "/usr/bin/python3"
CMakeCache.txt:PYTHON3_PATH:PATH=/usr/bin/python3
```

Issue #1817, opened 2021-10-15
Every other unselect argument in autotailor is ignored.


watsonjm:
#### Description of Problem:
I'm using autotailor to remove a few rules for scanning some containers using oscap-podman. It appears that every other (2/4/6) rule is skipped. 

#### OpenSCAP Version:
OpenSCAP command line tool (oscap) 1.3.4,

#### Operating System & Version:
RHEL 8.4 on an Azure VM

#### Steps to Reproduce:
1. `autotailor -u package_usbguard_installed -u network_configure_name_resolution -u no_files_unowned_by_user -u configure_usbguard_auditbackend -u package_rng-tools_installed -u ssh_client_rekey_limit -p stig_tailored -o tailoring.xml /usr/share/xml/scap/ssg/content/ssg-rhel8-ds-1.2.xml xccdf_org.ssgproject.content_profile_stig`
2.  `sudo oscap-podman <IMAGE ID> xccdf eval --fetch-remote-resources --profile xccdf_org.ssgproject.content_profile_stig_tailored --results results.xml --report report.html --tailoring-file tailoring.xml /usr/share/xml/scap/ssg/content/ssg-rhel8-ds-1.2.xml`

#### Actual Results:
network_configure_name_resolution, configure_usbguard_auditbackend, ssh_client_rekey_limit will all be ignored and still checked in the rules. If you swap all of the even & odd rules in your autotailor command, those 3 will be checked and the new odd-numbered rules will be ignored and still checked. 

#### Expected Results:
All rules should be unselected instead of only odd-numbered ones.

#### Additional Information / Debugging Steps:
You can use this command for autotailor to compare your results:
`autotailor -u network_configure_name_resolution -u no_files_unowned_by_user -u configure_usbguard_auditbackend -u package_rng-tools_installed -u ssh_client_rekey_limit -u package_usbguard_installed -p stig_tailored -o tailoring.xml /usr/share/xml/scap/ssg/content/ssg-rhel8-ds-1.2.xml xccdf_org.ssgproject.content_profile_stig`

watsonjm:
It looks like running the autotailor output through xmllint fixes the issue
`xmllint --format tailoring.xml > tailoring_linted.xml`
and then using that in the eval command:
[...]

PR #1991, opened 2023-06-21
Fix issues when parsing profiles

jan-cerny:
When there is no white space between `<xccdf:select>` elements in a profile in a benchmark, or in a custom profile in a tailoring, every odd element was ignored because the XML node was skipped by OpenSCAP when parsing.

This is caused by the situation that `xmlTextReaderRead` in `oscap_to_start_element` in `xccdf_parse_remarks` jumps to the next node but then `xmlTextReaderRead` at the end of the `while loop` in `xccdf_profile_parse` jumps again to the next node. If there is a white space node in between the element nodes, this works fine, but if there is no node in between, an element gets skipped.

The commit also adds some unit tests:
- `test_no_newline_between_select_elements` checks if no elements are skipped when parsing `<xccdf:select>` elements within `<xccdf:Profile>` elements
- `test_single_line_tailoring` checks if a tailoring file with no newlines between XML elements is processed correctly.

This is related to #1817.

Question:
Although this fixes the case covered by the included test scenarios, I assume a similar problem might occur also elsewhere, because we use the combination of `oscap_to_start_element`, `while` and `xmlTextReaderRead` in multiple places. What strategy do you recommend to help tackle the possible other issues?


evgenyz (approved):
LGTM. We can chase other cases of such behavior in a reactive manner.

PR #1994, opened 2023-06-23
Install Perl files in correct path

havardAasen:
The file destination variables `PERL_VENDORLIB` and `PERL_VENDORARCH` has been mixed. `PERL_VENDORARCH` is where architecture specific files should be placed, which is the file `SWIG_MODULE_openscap_pm_REAL_NAME`.

The behavior is also more in line with how the files is installed on Apple and FreeBSD.


havardAasen:
These variables don't respect neither `CMAKE_INSTALL_PREFIX` nor `CMAKE_INSTALL_LIBDIR`, #1999 could be a better path forward.

PR #1996, opened 2023-07-08
OVAL/sysctl: Fix offline mode

evgenyz:
The initial implementation was buggy: after correctly traversing prefixed PREFIX/proc/sys directory tree it would incorrectly read the data from the non-prefixed directory tree.


evgenyz:
Rawhide failure is related to https://github.com/OpenSCAP/openscap/issues/1997

jan-cerny:
@evgenyz I think we need a test to avoid this situation.

evgenyz:
> @evgenyz I think we need a test to avoid this situation.

I've updated the test to cover that scenario.

PR #1999, opened 2023-07-12
make default perl install path respect CMAKE_INSTALL_PREFIX

sarnold:
I have a local tox workflow and this was trying to install to a system path, so make sure the path starts with PREFIX.

Just be clear, any non-root user installs using cmake install prefix are currently *broken*.


evgenyz:
@sarnold Can you please re-base the PR to fix CI problems?

sarnold:
Sorry I didn't catch this sooner :/  Lemme do that now.

Issue #1997, opened 2023-07-08
The pathfix.py is going away from Python


evgenyz:
#### Description of Problem:

The pathfix.py is going away from Python:
https://lwn.net/Articles/910898/

We use it in our spec file. (https://discuss.python.org/t/remove-outdated-tools-scripts-scripts/19571/35).

Rawhide is affected. There is no solution yet.

#### OpenSCAP Version:

1.3.8

jan-cerny:
@evgenyz In Fedora there is a solution: https://fedoraproject.org/wiki/Changes/Python3.12#pathfix.py_tool_will_be_removed

In Rawhide, they have updated our spec:
https://src.fedoraproject.org/rpms/openscap/pull-request/17#request_diff

I think we can do a similar change in upstream

PR #2001, opened 2023-07-14
Avoid grep warning on oscap-podman invocation

dirkmueller:
grep: warning: stray \ before -

is printed by GNU grep 3.11. This can be fixed by terminating the argument parsing before by using ' -- ' which then avoids the quoting alltogether.


jan-cerny (approved):
This is a nice catch! Thank you!

The CI fail on Rawhide is tracked in #1997 and isn't caused by the contents of this PR.

PR #2003, opened 2023-07-19
Fix leak of item in probe_item_collect

0intro:
This change fixes leak of `item` in `probe_item_collect` when `probe_cobj_memcheck` didn't return `0` (error or when memory contraints are reached).


jan-cerny (approved):
LGTM

The CI fail on Rawhide is tracked in https://github.com/OpenSCAP/openscap/issues/1997 and isn't caused by the contents of this PR.

PR #2004, opened 2023-07-19
Fix leak of filename in oval_agent_new_session

0intro:
This change fixes leak of `ag_sess->filename` in `oval_agent_new_session` when `oval_probe_query_sysinfo` returned `!= 0`.


jan-cerny (approved):
Thanks for this patch!

The CI fail on Rawhide is tracked in #1997 and isn't caused by the contents of this pull request.

PR #2005, opened 2023-07-20
Fix pathfix usage in spec

jan-cerny:
Python 3.12 removes the pathfix utility but in Fedora it's just moved to python3-rpm-macros.

Fixes broken build and broken CI on Rawhide.

Described in: https://fedoraproject.org/wiki/Changes/Python3.12#pathfix.py_tool_will_be_removed
Similar to: https://src.fedoraproject.org/rpms/openscap/pull-request/17#request_diff

Fixes: #1997


jan-cerny:
It fixed the build on Rawhide; but the testing farm has a different problem.

evgenyz:
LGTM. Thanks!

evgenyz (approved):

PR #2006, opened 2023-07-20
Consider cgroup when determining memory usage

0intro:
This change modifies the `oscap_sys_memusage` function to take into account the memory constraints of the cgroup. Both cgroup and cgroup2 are supported.


evgenyz:
Hey, @0intro! Can you please rebase the PR?

0intro:
I've just rebased the change.

evgenyz:
Okay, I understand the change, and I'm not particularly against it. But it will affect every invocation of the scanner everywhere as the scanner has a default memory limit for all operations. What we have right now could be not exactly correct, but it might be what is expected by the users however weird it is.

And we don't really have a test even for the old behavior. How about you add a test that checks your particular use-case (the problem you're trying to solve)? We can then reason about preferred strategies on the new behavior.

Mab879:
Over 2 years and no response, closing.

PR #2013, opened 2023-07-29
fix un-expanded variable in xccdf report output

sarnold:
I was making some openembedded patches and I saw this:
Screenshot from 2023-07-28 19-57-39

so I made a patch for that too.


jan-cerny (approved):
LGTM, Thank you!

The CI fail on Rawhide is tracked in https://github.com/OpenSCAP/openscap/issues/1997 and isn't caused by the content of this PR.

PR #2016, opened 2023-08-31
Update README

Mab879:
A few clean up items in the README.

I need to double check the mailing list address, once that is addressed I will remove the draft status.


jan-cerny (approved):
I think it's correct.

jan-cerny:
The CI fail on Ubuntu is unrelated to the contents of this PR because this PR changes only the README file.

Issue #1938, opened 2023-02-21
Session reuse


0intro:
I’d like to evaluate a list of rule and get the results. However, I'd like to reuse the session between rules, instead of calling `xccdf_session_new` and `xccdf_session_free` before and after evaluating each rule.

I had to modify the public API of the OpenSCAP library a little bit, so I could access the list of results:

```diff
--- a/src/XCCDF/public/xccdf_session.h
+++ b/src/XCCDF/public/xccdf_session.h
@@ -624,6 +624,9 @@ OSCAP_API int xccdf_session_generate_guide(struct xccdf_session *session, const
  */
 OSCAP_API int xccdf_session_export_all(struct xccdf_session *session);
 
+struct xccdf_rule_result_iterator;
+OSCAP_API struct xccdf_rule_result_iterator *xccdf_session_get_rule_results(const struct xccdf_session *item);
+
 /// @}
 /// @}
 #endif
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -1860,6 +1860,11 @@ bool xccdf_session_contains_fail_result(const struct xccdf_session *session)
 	return false;
 }
 
+struct xccdf_rule_result_iterator *xccdf_session_get_rule_results(const struct xccdf_session *session)
+{
+	return xccdf_result_get_rule_results(session->xccdf.result);
+}
+
 int xccdf_session_remediate(struct xccdf_session *session)
 {
 	int res = 0;
```

This issue is that `xccdf_session_get_rule_results` returns the list of the results that were evaluated during the previous run, while I’d like `xccdf_session_get_rule_results` to return only the result corresponding to last evaluated rule.

Is there a way to cleanup the rule results from the session, and reset the session to its initial state, without having to free and allocate a new session again, which would imply to parse the XML again?

For example, when I run the following program:

```c
#include <stdio.h>
#include <stdlib.h>

#include <xccdf_session.h>
#include <oscap_error.h>

#define nelem(x) (sizeof(x)/sizeof((x)[0]))

typedef struct Rule Rule;

struct Rule {
	char *xccdf;
	char *profile;
	char *rule;
};

Rule rules[] = {
	{
		"/tmp/ssg-ubuntu2204-ds-1.2.xml",
		"xccdf_org.ssgproject.content_profile_cis_level1_server",
		"xccdf_org.ssgproject.content_rule_file_owner_cron_hourly",
	},
	{
[...]

PR #2020, opened 2023-09-04
Implement xccdf_session_result_reset function in XCCDF session API

0intro:
This change adds a new public `xccdf_session_result_reset` function to the XCCDF session API.

This function resets XCCDF session results, so `xccdf_session_evaluate` could be called again from a clean state, without having to call `xccdf_session_free`, which would imply to parse the XCCDF file again.

This function resets XCCDF policies, session rules, skipped rules, OVAL system characteristics and OVAL results.

### DIFF
--- a/src/OVAL/oval_agent.c
+++ b/src/OVAL/oval_agent.c
@@ -255,6 +255,21 @@ int oval_agent_reset_session(oval_agent_session_t * ag_sess) {
 	return 0;
 }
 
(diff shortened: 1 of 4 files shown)


evgenyz:
For reference: https://github.com/OpenSCAP/openscap/issues/1938

jan-cerny (approved):
The CI fail is caused by incompatibility of the sysctl probe with the kernel in the workers and therefore isn't related to the contents of this PR.

PR #2021, opened 2023-09-04
Implement xccdf_session_get_rule_results function in XCCDF session API

0intro:
This change adds a new public `xccdf_session_get_rule_results` function to the XCCDF session API.

This function returns rule results.

Results could be iterated using the `xccdf_rule_result_iterator_has_more` function from the XCCDF benchmark API.

### DIFF
--- a/src/XCCDF/public/xccdf_session.h
+++ b/src/XCCDF/public/xccdf_session.h
@@ -575,6 +575,18 @@ OSCAP_API unsigned int xccdf_session_get_cpe_oval_agents_count(const struct xccd
 OSCAP_API bool xccdf_session_contains_fail_result(const struct xccdf_session *session);
 
 /**
+ * @struct xccdf_rule_result_iterator
+ */
+struct xccdf_rule_result_iterator;
+
+/**
+ * Get rule results.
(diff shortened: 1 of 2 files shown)


0intro:
A use case of this API function is presented in the program example available in issue #1938.

src/XCCDF/public/xccdf_session.h
```diff
@@ -574,6 +574,17 @@ OSCAP_API unsigned int xccdf_session_get_cpe_oval_agents_count(const struct xccd
...
+/**
+ * @struct xccdf_rule_result_iterator
+ */
+struct xccdf_rule_result_iterator;
+
+/**
+ * Get rule results.
+ * @memberof xccdf_session
+ * @param session XCCDF Session
+OSCAP_API struct xccdf_rule_result_iterator *xccdf_session_get_rule_results(const struct xccdf_session *session);
```
jan-cerny, line 586: missing end of comment

0intro: Fixed. Thanks.

jan-cerny (approved):

PR #2028, opened 2023-09-10
Be able to delete temporary files on a different mount point

havardAasen:
We are able to create files on a different mount point, but not delete them. This is at most inconvenient, except when running tests.

A lot of the tests just checks of anything has been written to `stderr`, if it has, the test fail. Since we are not allowed to delete files, the parent directory is not empty when it is supposed to be removed, and an error message is printed to `stderr`.

If we are able to create files, we should be allowed to delete them as well.


evgenyz (approved):
Totally makes sense, thank you!

PR #2030, opened 2023-09-12
Update build dependencies in Developer guide

jan-cerny:
The package apt-devel isn't present in RHEL systems.


marcusburghardt (approved):
Thanks for the update

PR #2032, opened 2023-09-20
Fix memory leaks reported by Coverity

0intro:
This change fixes memory leaks reported by Coverity.


jan-cerny (approved):
I have reviewed the code in this PR. The CI fail is a problem is sysctl probe which become incompatible with the kernel, the CI fail isn't related to the contents of this PR. Thanks for your patch.

PR #2037, opened 2023-09-22
Fix incorrect openscap-cpe-oval result filename

rmetrich:
When using `oscap xccdf eval --oval-results` command, the oval result file is not expected one (`openscap-cpe-oval.xml.result.xml`), but `%2Fusr%2Fshare%2Fopenscap%2Fcpe%2Fopenscap-cpe-oval.xml.result.xml`.

This patch fixes the issue.


rmetrich:
Coresponding RHEL Jira: https://issues.redhat.com/browse/RHEL-7050

PR #2038, opened 2023-09-28
Add AlmaLinux CPE

metalefty:
Hi, AlmaLinux developer here. Let me add AlmaLinux CPEs.


evgenyz:
Thank you, but no: https://github.com/OpenSCAP/openscap/issues/1683

metalefty:
Thanks, I didn't know that but it makes sense.

PR #2040, opened 2023-10-04
Fix probe_reset

0intro:
The probe_reset function didn't work properly, because its second argument was always NULL.

This changes fixes the probe_reset function by passing the probe as the second argument.

This change also call probe_ncache_clear to clear the ncache instead of calling probe_ncache_free and probe_ncache_new, like it was done in change 057873ac7c816b5e067c49442bd4f55b77121e9c.


evgenyz (approved):
Thanks!

PR #2046, opened 2023-10-24
Rewrite dpkginfo probe without using APT

0intro:
This change rewrites the dpkginfo probe without using the APT library.

The dpkginfo now parses the list of installed package (`/var/lib/dpkg/status`) directly, instead of relying on the APT library.

This prevents loading the full list of packages in memory and various issues related to the use of the APT library.

The dpkginfo probe is now stateless and doesn't require init and fini functions. Also, the `dpkginfo_get_by_name` function can now be called from multiple threads without having to be protected by a lock.

The dependency on the APT library has been removed from OpenSCAP.


evgenyz:
Are there any potential behaviour changes?

0intro:
> Are there any potential behaviour changes?

There shouldn't be. The new `dpkginfo_get_by_name` function should behave exactly like the former one.

One thing that the APT library did was to return only the packages installed for the current architecture. I don't think it really matters in practice, because the multiple architecture support is only used to install 32-bit libraries on 64-bits systems. Also, it didn't even work properly before, because the APT configuration (`/etc/apt`) was loaded from the root file system, instead of the directory specified by `OSCAP_PROBE_ROOT` (then the apt-config stuff was re-rooted on the right root directory to retrieve the package database).

src/OVAL/probes/unix/linux/dpkginfo-helper.c
```diff
@@ -0,0 +1,174 @@
...
+		if (reply->release == NULL)
+			goto err;
+	}
+	reply->version = strdup(version);
+	if (reply->version == NULL)
+		goto err;
+
+	free(evr);
+	return 0;
+err:
```
evgenyz, line 63: There are a lot of reasons for this (and its caller) function to fail but there is no discrimination. I'd suggest to at least add debug-level logs for key junction points that have a significant chance of happening in a real environment. Otherwise it'd be very hard to understand why the probe does a wrong thing in production. I can't quite pinpoint where it should be placed, but I had a bad exp. with logless probes in the past. Just for you to consider.

evgenyz (changes requested):

src/OVAL/probes/unix/linux/dpkginfo-helper.c
```diff
@@ -0,0 +1,174 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <ctype.h>
+#include <limits.h>
+
+#include "debug_priv.h"
+#include "dpkginfo-helper.h"
+
+#define MAX_LEN 4096
```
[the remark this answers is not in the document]
0intro: Done.

src/OVAL/probes/unix/linux/dpkginfo-helper.c
0intro: I've added some debug prints, which, along with the debug prints already existing in `dpkginfo.c`, should help to understand the potential issues.

evgenyz (changes requested):
LGTM! Thanks!

But, I'd like to ask you to also update the docs, CI config and generic spec file (remove libapt dep) within the same PR. (See: https://github.com/search?q=repo%3AOpenSCAP%2Fopenscap%20apt-devel&type=code)

src/OVAL/probes/unix/linux/dpkginfo-helper.c
evgenyz: Makes sense.

0intro:
> But, I'd like to ask you to also update the docs, CI config and generic spec file (remove libapt dep) within the same PR. (See: https://github.com/search?q=repo%3AOpenSCAP%2Fopenscap%20apt-devel&type=code)

I've just updated the docs, CI config and spec file. Let me know if I have missed anything.

evgenyz:
Hold on a bit. Python 3.12 kicked us right in the fork.

evgenyz:
And new kernel :(

evgenyz:
Okay, we are good to go. Can you please rebase and have another go for CI?

evgenyz (approved):
Waiting for green CI.

0intro:
It's green now (except CodeQL).

evgenyz:
Thank you!

PR #2048, opened 2023-11-07
Fix removal of imp module in Python 3.12

evgenyz:
Python 3.12 got `imp` module removed: https://docs.python.org/3/whatsnew/3.12.html#imp

The workaround is based on `importlib` module. Python 2 branch is removed because we don't care about Python 2 anymore.

### DIFF
--- a/swig/openscap_api.py
+++ b/swig/openscap_api.py
@@ -31,31 +31,24 @@ import re
 
 logger = logging.getLogger("openscap")
 
-from sys import version_info
-if version_info >= (2, 6, 0):
-    def _import_helper():
-        from os.path import dirname
-        import imp
-        fp = None
-        try:
-            fp, pathname, description = imp.find_module(
-                '_openscap_py', [dirname(__file__)])
-        except ImportError:
-            import _openscap_py as OSCAP
-            return OSCAP
-        if fp is not None:
-            try:
-                _mod = imp.load_module(
-                    '_openscap_py', fp, pathname, description)
-            finally:
-                fp.close()
-            return _mod
-    OSCAP = _import_helper()
-    del _import_helper
-else:
-    import _openscap_py as OSCAP
 
-del version_info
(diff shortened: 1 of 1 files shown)


swig/openscap_api.py
```diff
@@ -31,31 +31,23 @@
...
+
+def _import_helper():
+    from os.path import dirname
+    import importlib.machinery
+    import importlib.util
+    spec = importlib.machinery.PathFinder.find_spec('_openscap_py', [dirname(__file__)])
+    if not spec:
+        spec = importlib.machinery.PathFinder.find_spec('_openscap_py')
+    if not spec:
+        return None
```
jan-cerny, line 43: Should we raise an exception/print a message that import failed?

evgenyz: We'll get one after trying to access module which is None. Traceback is pretty much self-explanatory.

evgenyz: But I can change that into `raise ImportError("Unable to import _openscap_py module, extra path: %s." % dirname(__file__))`

jan-cerny (approved):
The CI fail on Rawhide

```
238/277 Test #238: probes/sysctl/test_sysctl_probe_all.sh .............................................***Failed    0.29 sec
Result file: test_sysctl_probe_all.res.out.5fl6Id
Our names file: test_sysctl_probe_all.our.out.nnyTvH
Sysctl names file: test_sysctl_probe_all.sysctl.out.bTJnYG
Errors file: test_sysctl_probe_all.err.out.Ygjjz9
Diff (sysctlNames / ourNames): ------
152d151
< kernel.sched_energy_aware
```

isn't caused by the contents of this PR because this PR is only related to SWIG. It's probably caused by update in kernel as we saw in past.

jan-cerny:
I have reported the problem to upstream in https://github.com/OpenSCAP/openscap/issues/2049

Issue #2049, opened 2023-11-07
Sysctl element kernel.sched_energy_aware cause test fail on Rawhide


jan-cerny:
#### Description of Problem:
The test `test_sysctl_probe_all.sh` fails on Rawhide

#### OpenSCAP Version:
current upstream maint-1.3 branch as of 2023-11-07 as of HEAD 30c8c395e547fa44a44080fa209ce4d27d3c3f95

#### Operating System & Version:
Fedora Rawhide as of 2023-11-07

#### Steps to Reproduce:

1.  run our ctest tests

#### Actual Results:
```
238/277 Test #238: probes/sysctl/test_sysctl_probe_all.sh .............................................***Failed    0.29 sec
Result file: test_sysctl_probe_all.res.out.5fl6Id
Our names file: test_sysctl_probe_all.our.out.nnyTvH
Sysctl names file: test_sysctl_probe_all.sysctl.out.bTJnYG
Errors file: test_sysctl_probe_all.err.out.Ygjjz9
Diff (sysctlNames / ourNames): ------
152d151
< kernel.sched_energy_aware
```

#### Expected Results:
passed test

#### Additional Information / Debugging Steps:
Manifests also in Testing farm, eg. https://artifacts.dev.testing-farm.io/472eea03-0b74-4043-87cf-d0628926f355/

Most likely the kernel.sched_energy_aware needs to be added on a special list in the systemctl probe.

evgenyz:
Last time it happened I was considering making the list dynamic somehow. Maybe now it is time to do so, as it happens more and more often.

PR #2050, opened 2023-11-07
Fix tests/probes/sysctl_all

evgenyz:
The test will now automatically adapt to the list of readable variables and won't break every time a new root-readable variable is introduced in kernel.

Fixes: #2049

### DIFF
--- a/tests/probes/sysctl/test_sysctl_probe_all.sh
+++ b/tests/probes/sysctl/test_sysctl_probe_all.sh
@@ -8,101 +8,57 @@ set -e -o pipefail
 # non root users
 PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin
 
-# non root users are not able to access some kernel params, so they get excluded
-SYSCTL_EXCLUDE='
-	dev.parport.parport0.autoprobe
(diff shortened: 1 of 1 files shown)


tests/probes/sysctl/test_sysctl_probe_all.sh
```diff
@@ -8,101 +8,57 @@ set -e -o pipefail
...
+
+	case $(uname) in
+		FreeBSD)
+			sysctl -aN 2> /dev/null > "$sysctlNames"
+			;;
+		Linux)
+			# sysctl has duplicities in output
+			# hide permission errors like: "sysctl: permission denied on key 'fs.protected_hardlinks'"
+			# kernel parameters might use "/" and "." separators interchangeably - normalizing
+			sysctl -a --deprecated 2> /dev/null | tr "/" "." | cut -d "=" -f 1 | tr -d " " | sort -u > "$sysctlNames"
```
evgenyz, line 36: The key is the removal of `N` argument. If we try and query for the values as well we'll get the list consistent with what the probe could get access to.

jan-cerny (approved):
I can see the test `probes/sysctl/test_sysctl_probe_all.sh` passing in the testing farm rawhide test run. Thanks for your change.

PR #2051, opened 2023-11-16
Introduce a limit of collected items

jan-cerny:
This PR introduces possibility to set a limit of items collected by probes.
If a single OVAL object matches more than given limit, we will stop collecting these items and mark the object as incomplete.

Users will set the limit by defining the `OSCAP_PROBE_MAX_COLLECTED_ITEMS` environment variable.
By default, the limit isn't set which means the probes will collect all the items matching the evaluated OVAL object.

Setting this limit can be useful in some situations in which the produced results are too large and cause memory problems or inability to generate the HTML report.

Related to: https://issues.redhat.com/browse/RHEL-11925, https://issues.redhat.com/browse/RHEL-4141

This change isn't a silver solution of these tickets. It's rather one of the measures that could improve our situation in combination with other measures both on the scanner and on the content side.


evgenyz:
This won't fix the original problem (XSLT will still have problems if OVAL author concatenates multiple results), and on top of that it will introduce failing tests where HTML is not rendered at all (OCP4 Compliance Operator).

In my opinion a correct fix would modify XSLT template or, at least, limit amount of **rendered** objects.

evgenyz:
Even if this functionality could be useful for someone in some cases, I'm strongly against putting it as a default behavior: OVAL does not define any kind of limit, which makes this an extremely unexpected feature without a control knob.

src/OVAL/probes/probe/icache.c
```diff
@@ -39,11 +39,14 @@
 #include "probe-api.h"
 #include "common/debug_priv.h"
 #include "common/memusage.h"
+#include "oscap_helpers.h"
 
 #include "probe.h"
 #include "icache.h"
 #include "_sexp-ID.h"
 
+#define PROBE_ITEM_COLLECT_MAX 1000
```
marcusburghardt, line 48: Here was used `1.000` but in the commit description is mentioned `10.000`. Also, do you have information about how much memory one item consumes? I would like to understand how this proposed limit was calculated.

jan-cerny:
> This won't fix the original problem (XSLT will still have problems if OVAL author concatenates multiple results), and on top of that it will introduce failing tests where HTML is not rendered at all (OCP4 Compliance Operator).

My understanding of the original problem in https://issues.redhat.com/browse/RHEL-11925 is that the ARF tree produced by oscap which is the input of the XSLT template is so huge that libxslt has problems processing it.

I think that introducing a limit for the collected items will directly reduce the size of the produced ARF, so it will lower the chance that libxslt won't be able to process the tree.

You're right that it won't exactly fix the problem, It also doesn't aim to fix it by this PR, but the aim is to mitigate it or make it happening less often. We're running multiple efforts that could improve the situation when applied together, eg. #2052, changes of well-known rules in ComplianceAsCode/content.

I agree with you that XSLT will still have problems if OVAL definition uses multiple objects that lead to a large item collection. In that situation, I think having the limit still could help mitigate the issue. For example, if a definition contains 3 objects and each of them corresponds to 100000 items, then we would have 300000 items if no limit is in place but only 3000 items if there is a limit 1000 per object. Another aspect is the evaluated XCCDF profile and the amount of XCCDF rules in the profile and the nature of their individual OVAL checks. If we have a "bad chance" and we have many rules with these OVAL objects, the output will be reduced but it won't be sufficient. A different solution would be a global limit, ie. limit that doesn't apply per object but on the whole evaluation run.

Regarding the introduction of failing tests, I believe it depends on how specific OVAL definition is constructed. Can you elaborate on what you have in mind?

> 
> In my opinion a correct fix would modify XSLT template or, at least, limit amount of **rendered** objects.

The amount of rendered items in HTML report is limited to 100 items per OVAL object. See

https://github.com/OpenSCAP/openscap/blob/c27e1085a3e211ffca970c475e880952f2207c5f/xsl/xccdf-report-oval-details.xsl#L105-L119


> Even if this functionality could be useful for someone in some cases, I'm strongly against putting it as a default behavior: OVAL does not define any kind of limit, which makes this an extremely unexpected feature without a control knob.

I really like the idea to make the limit configurable in run time, eg. by command line option.

evgenyz:
> Regarding the introduction of failing tests, I believe it depends on how specific OVAL definition is constructed. Can you elaborate on what you have in mind?

OCP Compliance Operator (for example) does not render HTML and does not care about XSLT problems. If this feature will be enabled by default OCP evaluation results could become incomplete instead of passing.
 
> > In my opinion a correct fix would modify XSLT template or, at least, limit amount of **rendered** objects.
> 
> The amount of rendered items in HTML report is limited to 100 items per OVAL object. See
> 
> https://github.com/OpenSCAP/openscap/blob/c27e1085a3e211ffca970c475e880952f2207c5f/xsl/xccdf-report-oval-details.xsl#L105-L119

What I mean here is to try and throw out elements from the tree before trying to render them with XSLT instead of not collecting them at all. With this approach the ARF and XCCDF results would stay correct.
 
> > Even if this functionality could be useful for someone in some cases, I'm strongly against putting it as a default behavior: OVAL does not define any kind of limit, which makes this an extremely unexpected feature without a control knob.
> 
> I really like the idea to make the limit configurable in run time, eg. by command line option.

Could it be better to actually use `OSCAP_PROBE_MAX_COLLECTED_OBJECTS` env. variable (or smth like that)? This is the way we usually influence the behaviour of OVAL probes.

evgenyz:
@jan-cerny We do have quite a lot of env. variables that could affect oscap behavior. What do you think about having a special section somewhere in the main module, that would print all of them on some debug level? It might be useful to know all the input just from the log, without asking extra questions. The same, actually, goes for the command line arguments.

jan-cerny:
@evgenyz Yes, great idea! I believe it would be useful especially if the log output is attached to bug reports. Can you create a ticket for that?

jan-cerny:
I think the fail on Rawhide is caused by removal of xmlParseMemory from libxml2, I have found this thread: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/AI7QXMRXYQ3KAZTLIGICEX6YHFTM5RBR/

jan-cerny:
> What I mean here is to try and throw out elements from the tree before trying to render them with XSLT instead of not collecting them at all. With this approach the ARF and XCCDF results would stay correct.

I think removing elements from ARF before passing the ARF to XSLT would be in practice implemented by XSLT or a combination of XPath and DOM operations. I can imagine that this approach could lead a similar problem with memory issues.

Currently, OpenSCAP internally creates an ARF XML tree, applies XSLT on that to create HTML report, saves the HTML to a file and then saves the ARF file.

I think we can easily limit the amount of the items in the ARF during the serialization into OVAL results DOM. I look into that and it could be done by modifying `oval_syschar_model_to_dom` and `oval_syschar_to_dom`. See `/src/OVAL/oval_sysModel.c` around line 360.  Due to the way how it works now it would cause a problem that also ARF file would be incomplete, which you probably don't want. But we can avoid it by creating a dedicated  ARF tree just for the purpose of HTML creation, ie. first a full ARF tree would be created and saved to file and then another tree with reduced items would be created passed to XSLT to create a HTML output.

jan-cerny:
I have added a option to configure this using an environment variable `OSCAP_PROBE_MAX_COLLECTED_ITEMS`.

I also have extracted the proposal by @evgenyz  from here to https://github.com/OpenSCAP/openscap/issues/2058.

jan-cerny:
The CI fail on Rawhide is tracked in https://github.com/OpenSCAP/openscap/issues/2059 and isn't caused by the contents of PR.

src/OVAL/probes/probe/icache.c
```diff
@@ -528,6 +529,31 @@ static int probe_cobj_memcheck(size_t item_cnt, double max_ratio)
 	return (0);
 }
 
+static int _mark_collected_object_as_incomplete(struct probe_ctx *ctx, const char *message)
+{
+
```
evgenyz, line 534: Redundant \n

evgenyz (approved):
The implementation is sound from my PoV. Just a couple of notes regarding style and names.

jan-cerny:
I have removed empty line and renamed "memory" directory to "probe_behavior".

evgenyz (approved):

PR #2052, opened 2023-11-16
Skip certain paths from scanning

jan-cerny:
This PR adds an option to skip selected paths from content evaluation. This will allow users to avoid pollution of the scan results from container guests, various mounts and other filesystems.

Related to: https://issues.redhat.com/browse/RHEL-11925, https://issues.redhat.com/browse/RHEL-4141

Users can configure the paths they want to be skipped by setting the `OSCAP_PROBE_IGNORE_PATHS` environment variable. If multiple paths should be skipped they need to be separated by a colon.

**Review hints:**

- take a look at the test that is part of this PR
- set the OSCAP_PROBE_IGNORE_PATHS and see how it affects the results, compare the results with this variable set and unset, for example:
  1. `$OSCAP xccdf eval --profile '(all)' --rule xccdf_org.ssgproject.content_rule_selinux_state --report /tmp/report1.html /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml`
  2. `OSCAP_PROBE_IGNORE_PATHS=/etc/selinux $OSCAP xccdf eval --profile '(all)' --rule xccdf_org.ssgproject.content_rule_selinux_state --report /tmp/report2.html /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml`
  3. `firefox /tmp/report1.html /tmp/report2.html`


evgenyz:
Not only this fix won't really amend the initial problem (I expect XSLT would still have troubles if someone concatenate results from multiple tests in OVAL), it'd add another one: out of the blue failed tests where HTML is not generated at all (OCP4 Compliance Operator for example).

IMHO, to properly fix HTML rendering problem one need to try and fix XSLT template or, at least, limit objects amount when **rendering**, not at the collection step.

evgenyz:
To my taste `--ignore-path` or `--skip-path` sound better. I'd maybe even say that `OSCAP_PROBE_IGNORE_PATH=/path1/:/path2/path3` is more suitable.

marcusburghardt:
There are already some workarounds for the demand to exclude directories from an oscap scan, as described in this article https://access.redhat.com/solutions/4193501. So there are valid cases when this behavior might be required.

I believe this new option would make it much easier, compared to other workarounds, for users to consciously avoid some directories by whatever reason. Since the idea is to introduce this option without defaults, there is no change of behavior by default while the tool will still offer the option for the user.

jan-cerny:
One concern with this proposal that I have is that we won't be able to "block" the selected paths completely as it's hard and laborous to track all the places across the code where a file path is opened. Also, the results of doing it this way can be different than a solution with wrapping OpenSCAP to a systemd service with paths excluded. This solution prevents probes from collecting items whose path matches the given path. OTOH the systemd service blocks the open calls which return errors. Probes and follow up logic react to these situations differently.

evgenyz:
> One concern with this proposal that I have is that we won't be able to "block" the selected paths completely as it's hard and laborous to track all the places across the code where a file path is opened. Also, the results of doing it this way can be different than a solution with wrapping OpenSCAP to a systemd service with paths excluded. This solution prevents probes from collecting items whose path matches the given path. OTOH the systemd service blocks the open calls which return errors. Probes and follow up logic react to these situations differently.

That's exactly why we haven't tried to actually implement this feature in the past (it has been discussed multiple times). On the other hand, since this feature is not standardized in any way, we could do *something* and then improve based on the feedback. We should just have to make sure it is clear that the feature is experimental and technically violates OVAL.

The systemd approach is more consistent, but less convenient. I can suggest exploring the option of doing what systemd is doing, but on the scanner side (which might not be easy as well).

evgenyz:
> The systemd approach is more consistent, but less convenient. I can suggest exploring the option of doing what systemd is doing, but on the scanner side (which might not be easy as well).

But, to be noted, the issue with users is not connected to the fact that systemd service is not convenient, but rather to the idea that skipping some paths is not acceptable, as far as I understand. This, if true, makes the whole effort a bit questionable.

marcusburghardt:
> > The systemd approach is more consistent, but less convenient. I can suggest exploring the option of doing what systemd is doing, but on the scanner side (which might not be easy as well).
> 
> But, to be noted, the issue with users is not connected to the fact that systemd service is not convenient, but rather to the idea that skipping some paths is not acceptable, as far as I understand. This, if true, makes the whole effort a bit questionable.

We should not expect this option will be used in all cases. The users are free to decide if they want or not to use this new option for their cases. Users who would benefit from this option currently have to use an inconvenient workaround. This option will make their experiences better.

Based on the @jan-cerny concerns, I think makes a lot of sense to communicate this option as experimental. It doesn't need to be perfect now. We can improve on demand.

jan-cerny:
I have slightly updated this draft. Now, the paths can be specified using the environment variable `OSCAP_PROBE_IGNORE_PATHS` and need to be separated by a colon. I have removed the command line option that was originally here.

I also started a test in the test suite.

jan-cerny:
I have add the feature to other relevant probes. I improved the test. Also, I have rebased this PR on the top of the latest upstream maint-1.3 branch.

jan-cerny:
The CI TF fail on RH will be fixed by https://github.com/OpenSCAP/openscap/pull/2067.

docs/manual/manual.adoc
```diff
@@ -1619,6 +1619,7 @@ not considered local by the scanner:
 * `SOURCE_DATE_EPOCH` - Timestamp in seconds since epoch. This timestamp will be used instead of the current time to populate `timestamp` attributes in SCAP source data streams created by `oscap ds sds-compose` sub-module. This is used for reproducible builds of data streams.
 * `OSCAP_PROBE_MEMORY_USAGE_RATIO` - maximum memory usage ratio (used/total) for OpenSCAP probes, default: 0.1
 * `OSCAP_PROBE_MAX_COLLECTED_ITEMS` - maximal count of collected items by OpenSCAP probe for a single OVAL object evaluation
+* `OSCAP_PROBE_IGNORE_PATHS` - Skip given paths during evaluation. If multiple paths should be skipped they need to be separated by a colon.
```
evgenyz, line 1622: Do we have any constraints on the path, like absolute only, etc?

tests/API/OVAL/skip_paths/test.xml
```diff
@@ -0,0 +1,3 @@
+<parent>
+    <child>text</child>
+</parent>
```
evgenyz, line 3: Newline.

tests/API/OVAL/skip_paths/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_oscap_test("test_skip_paths.sh")
```
evgenyz, line 1: I'm not insisting on it, but could it make more sense to have it in `/tests/probe_behavior/`? Just FYI.

evgenyz (changes requested):
There are some quick nitpicks. I'll take a bit more time to test it thoroughly.

Question right from the top of my head, though: does it work in offline mode?

src/OVAL/probes/probe-api.c
```diff
@@ -1794,4 +1794,52 @@ SEXP_t *probe_obj_getmask(SEXP_t *obj)
     SEXP_free(objents);
     return (mask);
 }
+
+static bool path_startswith(const char *path, const char *prefix)
```
evgenyz, line 1798: I do feel like having this function in `common/util.c` instead. We have a group of path-related functions there already.

Also, unlike counterparts from util.c, this one does not account for Windows environment. Is this an intended limitation?

jan-cerny:
Yes, it should work offline. I have add an offline version of the test. Then I have add a missing EOF, updated the documentation about the paths and I have moved the path_startswith function to utils module.

docs/manual/manual.adoc
```diff
@@ -1619,6 +1619,7 @@ not considered local by the scanner:
 * `SOURCE_DATE_EPOCH` - Timestamp in seconds since epoch. This timestamp will be used instead of the current time to populate `timestamp` attributes in SCAP source data streams created by `oscap ds sds-compose` sub-module. This is used for reproducible builds of data streams.
 * `OSCAP_PROBE_MEMORY_USAGE_RATIO` - maximum memory usage ratio (used/total) for OpenSCAP probes, default: 0.1
 * `OSCAP_PROBE_MAX_COLLECTED_ITEMS` - maximal count of collected items by OpenSCAP probe for a single OVAL object evaluation
+* `OSCAP_PROBE_IGNORE_PATHS` - Skip given paths during evaluation. If multiple paths should be skipped they need to be separated by a colon. The paths should be absolute.
```
evgenyz, line 1622: So, I've checked some paths and found out that paths like `/a/b/../b/c` does not work (as well as `/a/./b/c`).

Which brings me to the conclusion that the requirement is actually to provide the *canonical* path, not one of absolute paths.

src/common/util.c
```diff
@@ -442,3 +442,35 @@ int oscap_open_writable(const char *filename)
 	}
 	return fd;
 }
+
+bool oscap_path_startswith(const char *path, const char *prefix)
```
evgenyz, line 446: This function is a pivoting point of the path blocking. And I would really like its behavior to be documented and validated in a form of a unit test (absolute vs canonical path, empty, invalid etc).

A good place for such a test is `tests/common/test_oscap_util.c` module.

evgenyz (changes requested):
So, in addition to the `oscap_path_startswith` function behavior, I have a couple more concerns regarding the whole feature:

1) Windows support. We don't have to provide it, but silently doing a bad thing is not very friendly to our users. We can either implement it correctly or ignore the environment variable with a warning that it is being ignored because the feature is not implemented foe Windows platform.

2) Silent magic. Imagine receiving a log with an ARF file and a question like "why all my files in /etc are ignored". User will obviously forget to mention the environment variable.
I will not insist on adding anything to this PR, but I'll be **very** hesitant to release this feature (and https://github.com/OpenSCAP/openscap/pull/2051) without an implementation of https://github.com/OpenSCAP/openscap/issues/2063.

evgenyz:
@jan-cerny Do you have any ideas about `probe_behavior/collect_limit.sh` failure in `cenos-8`?

evgenyz (approved):
LGTM. Thank you!

PR #2053, opened 2023-11-17
Fix parsing of Status in dpkginfo probe

0intro:
This change fixes a mistake made during a last-minute change
in 557ddeed1e3e234a655ad77a691869554064b293.

The parsing the Status was incorrect and all
packages were considered as deinstalled.

Sorry for that.


evgenyz:
I just realized that we don't really have a test for dpkg probe. Care to work on it?

evgenyz (approved):

0intro:
> I just realized that we don't really have a test for dpkg probe. Care to work on it?

Yes, it would be nice to have. I will work on it when I have time.

PR #2055, opened 2023-11-23
Update html report with oval details

Xeicker:
With this update it is clearer when a test fails due to a few objects from many collected

Attaching some examples to show the suggested change in action, and how can it be useful:
image

image

Note:
If there is a cleaner way and it is preferable let me know. I'm not quite familiarized with xsl

### DIFF
--- a/xsl/xccdf-report-oval-details.xsl
+++ b/xsl/xccdf-report-oval-details.xsl
@@ -99,7 +99,9 @@ Authors:
             <table class="table table-striped table-bordered">
                 <!-- table head (possibly item-type-specific) -->
                 <thead>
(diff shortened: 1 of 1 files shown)


jan-cerny (changes requested):
Hi, thank you very much for opening this PR.

While adding this can be useful in many situations, it's affected by the problem that our details section doesn't show the full structure of the OVAL tree. Specifically, if you add these results, the users don't know if there is a state element in the given test and what this state contains. That can cause confusion. I'm not sure how to address this, though.

I suggest at least to be more clear by renaming the `Result` column to `Result of item-state comparison`. Hopefully, renaming it will slightly lower the confusion.

Also on a related note I would like to invite you to check the https://github.com/OpenSCAP/openscap-report project which aims to generate a more verbose and modern HTML report than what we have here.

PR #2054, opened 2023-11-21
make default perl install path respect CMAKE_INSTALL_PREFIX

sarnold:
Rebased as requested, same name as the original PR (https://github.com/OpenSCAP/openscap/pull/1999) (sorry about any confusion).

PR #2056, opened 2023-11-23
Use arched path for .so perl module

xambroz:
PERL_VENDORLIB is putting the openscap_pm.so to /usr/share/perl... no architecture dependent libraries should go there

Instead in should go to /lib64/... and that is what  PERL_VENDORARCH does.


evgenyz:
I receive mixed signals here: #2054. So what is the correct form?

sarnold:
The point of using CMAKE_INSTALL_PREFIX is to *not* install directly to a system path unless prefix points to /usr and the install process has root permissions.   PERL_VENDORARCH is *not* what you're looking for...

PR #2068, opened 2024-01-02
Build Windows on GitHub Actions

Mab879:
Add GitHub Actions to build Windows version and will upload the artifact so it can be used elsewhere.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
 
   build-macos:
     # The type of runner that the job will run on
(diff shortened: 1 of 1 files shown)


jan-cerny (approved):
The CI fail on Testing farm CentOS stream 8 isn't related to the contents of this PR because this PR only changes the GitHub actions workflow definition YAML.

The code looks sane. I have verified that it produces "something looking as a MSI" in the artifacts but unfortunately I don't have a Windows machine so I can't confirm that the MSI works.

PR #2071, opened 2024-01-08
Improve references in HTML guides and reports

jan-cerny:
In this commit we will leverage the reference URI-to-title mapping within the `xccdf:Benchmark` element to visually improve the references in HTML guide and report.

This improvements needs SCAP content to contain reference URI-to-title mapping within the `xccdf:Benchmark` element to work nicely.  The ComplianceAsCode content upstream project started to provide this mapping in the SCAP content recently, starting from 62513c391dc5a3fafd12741bd02565ca0e1e8db2.

However, the enhancement is compatible with SCAP content that doesn't contain the reference URI-to-title mapping. If this mapping isn't present, URIs are shown instead of titles in the HTML guides and report. But it still provides a singificant UX improvement.

Moreover, the `Group by` dropdown feature will now also use the URI-to-title mapping in the content as an additional source of dropdown labels next to the hardcoded list. That means that addition of new reference type to ComplianceAsCode will no longer require a need to update OpenSCAP code to get nice labels for the new type to the `Group by` dropdown.

### DIFF
--- a/xsl/xccdf-guide-impl.xsl
+++ b/xsl/xccdf-guide-impl.xsl
@@ -254,11 +254,9 @@ Authors:
                         </tr>
 
 
-                        <tr><td>Identifiers and References</td><td class="identifiers">
-                            <xsl:call-template name="item-idents-refs">
-                                <xsl:with-param name="item" select="$item"/>
-                            </xsl:call-template>
-                        </td></tr>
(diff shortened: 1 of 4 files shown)


Mab879 (approved):
LGTM. 

Thanks!

Issue #2063, opened 2023-11-30
Dump all env. variables that affects the behaviour on a DEBUG or INFO log level


evgenyz:
#### Description of Problem:

We have a set of environment variables that would change the way the scanner operates. It might be useful the have them listed in the log to have better understanding of the behaviour.

See: https://github.com/OpenSCAP/openscap/blob/maint-1.3/docs/manual/manual.adoc#11-list-of-accepted-environment-variables

PR #2073, opened 2024-01-16
Print environment variables

jan-cerny:
If OpenSCAP is executed with verbosity level INFO or DEVEL all environment variable values that can affect OpenSCAP behavior will be listed at the beginning of the log.

Fixes: #2063


evgenyz:
It looks like the `188 - probe_behavior/collect_limit.sh (Failed)` test is unstable and fragile for some reason. :(

evgenyz (approved):

PR #2079, opened 2024-01-26
Introduce JSON tailoring import option for `autotailor`

evgenyz:
The `--json-tailoring` option will provide support for importing https://github.com/ComplianceAsCode/schemas/tree/main/tailoring.

- [X] Code
- [X] Docs
- [x] Finalize schema
- [x] Test


tests/utils/custom.json
```diff
@@ -0,0 +1,23 @@
...
+      "role": "unchecked",
+      "severity": "unknown"
+    }
+  },
+  "variables": {
+    "V1": {
+      "value": "New Value"
+    },
+    "V2": {
+      "select": "some"
```
jan-cerny, line 20: What is `select` in the context of variables?

evgenyz: In XCCDF it is a selector ID. For a Value with option `<... selector="some">100</>` it will set the final value to 100 for `{"V2": {"select": "some"}}`.

evgenyz: In XCCDF Tailoring it is rendered as `<value-refine>`.

ggbecker: Considering we are mirroring XCCDF attributes, why not to use `selector` instead?

ggbecker: I prefer `selector` or `variable_selector`

ggbecker: "base_profile_id" `additionalProperties`, should we stick to a single case format?

jan-cerny: "Selector" sounds a better name than "select" because it suggests what should be the value of the attribute.

evgenyz: Okay, after reading around, I propose this:
```json
{
  "id": "tailored_profile",
  "title": "Test",
  "profile_id": "original_profile",
  "rules": {
    "some_rule_id_here": {
      "evaluate": true
    },
    "different_rule_id_here": {
      "role": "unscored",
      "severity": "high"
    }
  },
  "variables": {
    "a_variable_id": {
      "value": "New Value"
    },
    "3rd_variable_id": {
      "option_id": "var3_value_100"
    }
  }
}
```
Reasoning:
- **base_profile_id -> profile_id**: We tailor profiles, and tailoring becomes a new profile. There is no reasonable way to join profiles together (and unlikely there will ever be) or use an additional profile in this context. Let's keep it simple.
- **selected -> evaluate**: It directly conveys the message that the rule will be forcibly evaluated (with tailoring). Better that select (select to do what?).
- **select -> option_id**: Why not _selector_? It is a very XCCDF-specific term, that has more functions and meaning than just a reference to a predefined value (it could be used to orchestrate rules/clusters behavior as well). On the other hand, every XCCDF Value with predefined values has them in _options_ (pretty common name in UI/UX and CS in general), so we are either set the value or refer to a value using option identifier.

evgenyz: Not sure if we should have `"severity": "HIGH"` instead of `"severity": "high"` to stress the fact that it is a ENUMERATION_CONSTANT.

evgenyz: > We kinda trying the opposite: abstracting XCCDF implementation and terminology.

What do you mean, I don't get it.

evgenyz: > "base_profile_id" `additionalProperties`, should we stick to a single case format?

What do you mean, I don't get it.

evgenyz:
@ggbecker If I got you right, this is a variation for multiple tailored profiles in a single JSON tailoring:
```json
{
  "profiles": [
    {
      "id": "tailoring_id",
      "title": "Test",
      "base_profile_id": "original_profile_id",
      "rules": {
        "some_rule_id": {
          "evaluate": true,
          "severity": "high"
        },
        "different_rule_id": {
          "role": "unscored"
        }
      },
      "variables": {
        "a_variable_id": {
          "value": "New Value"
        },
        "another_variable_id": {
          "value": 42
        },
        "3rd_variable_id": {
          "option_id": "some_option_id"
        }
      }
    }
  ]
}
```

evgenyz:
@jan-cerny @ggbecker ping.

tests/utils/custom.json
evgenyz: Thank you for your valuable input, gentlemen. I've update the PR.

docs/manual/manual.adoc
```diff
@@ -1,6 +1,7 @@
 = OpenSCAP User Manual
 :imagesdir: ./images
 :workbench_url: https://www.open-scap.org/tools/scap-workbench/
+:json_tailoring_url: https://github.com/ComplianceAsCode/schemas/tree/main/tailoring
```
jan-cerny, line 4: @evgenyz It would be great if the directory https://github.com/ComplianceAsCode/schemas/tree/main/tailoring contains a README with a short description :smiley:

utils/autotailor
```diff
@@ -198,27 +247,66 @@ class Tailoring:
...
         self.rule_refinements_to_xml(profile)
+        self.value_refinements_to_xml(profile)
 
         root_str = ET.tostring(root)
         pretty_xml = xml.dom.minidom.parseString(root_str).toprettyxml()
         with open(location, "w") if location != "-" else sys.stdout as f:
             f.write(pretty_xml)
 
+    def import_json_tailoring(self, json_tailoring):
+        import json
```
jan-cerny, line 260: Move the import to the top of the file.

utils/autotailor
```diff
@@ -198,27 +247,66 @@ class Tailoring:
...
+        with open(json_tailoring, "r") as jf:
+            all_tailorings = json.load(jf)
+
+        if 'profiles' in all_tailorings and all_tailorings['profiles']:
+            # We currently support tailoring of one profile only
```
jan-cerny, line 265: Can we check and raise if there are multiple profiles?

utils/autotailor
```diff
@@ -198,27 +247,66 @@ class Tailoring:
...
+        self._add_value_overrides(profile)
         self.rule_refinements_to_xml(profile)
+        self.value_refinements_to_xml(profile)
 
         root_str = ET.tostring(root)
         pretty_xml = xml.dom.minidom.parseString(root_str).toprettyxml()
         with open(location, "w") if location != "-" else sys.stdout as f:
             f.write(pretty_xml)
 
+    def import_json_tailoring(self, json_tailoring):
```
jan-cerny, line 259: Split this method to multiple smaller ones to reduce code complexity.

utils/autotailor
```diff
@@ -198,27 +247,66 @@ class Tailoring:
...
         "datastream", metavar="DS_FILENAME",
         help="The tailored data stream filename.")
     parser.add_argument(
-        "profile", metavar="BASE_PROFILE_ID",
+        "profile", metavar="BASE_PROFILE_ID", nargs='?', default="",
         help="Specify ID of the base profile. ID of the profile can be "
         "either its full ID, or the suffix, in which case the "
         "'xccdf_<id-namespace>_profile' prefix will be prepended internally.")
+    parser.add_argument(
+        "--json-tailoring", metavar="JSON_TAILORING_FILENAME", default="",
```
jan-cerny, line 307: What is your opinion on adding a short variant of this option, for example `-j`? I think that would simplify and make the usage of the tool more comfortable.

utils/autotailor
```diff
@@ -198,27 +247,66 @@ class Tailoring:
...
 
         root_str = ET.tostring(root)
         pretty_xml = xml.dom.minidom.parseString(root_str).toprettyxml()
         with open(location, "w") if location != "-" else sys.stdout as f:
             f.write(pretty_xml)
 
+    def import_json_tailoring(self, json_tailoring):
+        import json
+        with open(json_tailoring, "r") as jf:
+            all_tailorings = json.load(jf)
```
jan-cerny, line 262: Question: Can we somehow validate the input file if it's correct? Can the JSON schema be used for this validation?

jan-cerny (commented):
I have used the autotailor to generate an XCCDF tailoring file using a JSON output. I tried to use the `tests/utils/custom.json` and I also experimented with various modifications.

docs/manual/manual.adoc
evgenyz: The schema is somewhat documented within the schema file itself. Anyway, it is a different repo with its own issues and PRs.

utils/autotailor
evgenyz: Sure.

utils/autotailor
evgenyz: Sure.

evgenyz:
Latest changes include `groups` override in JSON tailoring.

utils/autotailor
evgenyz: Done.

jan-cerny (approved):
I have generated an XCCDF tailoring from the JSON tailoring using the `autotailor` tool and then I have successfully used the generated tailoring in `oscap`.

Issue #2076, opened 2024-01-25
The `collect_limit` test randomly fails in copr builds


evgenyz:
#### Description of Problem:

The test (or the code modifications) is unstable. We need to understand why and fix it before releasing the feature.

#### OpenSCAP Version:

maint-1.3

PR #2080, opened 2024-01-31
Improve counting collected items

jan-cerny:
The function SEXP_list_length turns out to be an unreliable provider of information about the length of the list of the collected items. It is unclear what this function returns. Instead of using SEXP_list_length we will have a counter variable that will count the amount of collected items by the probe.

Fixes: #2076


evgenyz (approved):
I see that tests are stable now. Thank you!

PR #2081, opened 2024-01-31
Fix Blueprint template to be self-contained

evgenyz:
Now the generated Blueprint file will be ready-to-use right after generation unless a custom data stream is used for hardening.

There are also instructions on how to adapt the Blueprint for a custom data stream.

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -1150,15 +1150,7 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
 			"profile_id = \"%s\"\n"
 			"# If your hardening data stream is not part of the 'scap-security-guide' package\n"
 			"# provide the absolute path to it (from the root of the image filesystem).\n"
-			"# datastream = \"/usr/share/xml/scap/ssg/content/ssg-xxxxx-ds.xml\"\n\n"
-			"# If your hardening data stream is not part of the 'scap-security-guide' package\n"
-			"# you don't need this package to be installed in the image (section can be removed).\n"
-			"[[packages]]\n"
-			"name = \"scap-security-guide\"\n"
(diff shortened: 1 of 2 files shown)


jan-cerny:
@evgenyz do you plan to do a self-review?

evgenyz:
> @evgenyz do you plan to do a self-review?

Nope. It was a glitch.

evgenyz:
Also, @comps, please take a look.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1143,10 +1143,23 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
...
+			"# provide the absolute path to it (from the root of the image filesystem).\n"
+			"# datastream = \"/usr/share/xml/scap/ssg/content/ssg-xxxxx-ds.xml\"\n\n"
+			"# If your hardening data stream is not part of the 'scap-security-guide' package\n"
+			"# you don't need this package to be installed in the image (section can be removed).\n"
+			"[[packages]]\n"
+			"name = \"scap-security-guide\"\n"
+			"version = \"*\"\n\n"
+			"[[packages]]\n"
+			"name = \"openscap-scanner\"\n"
+			"version = \"*\"\n\n",
```
comps, line 1161: I probably wouldn't do this.

The target system doesn't need `scap-security-guide` or `openscap-scanner` since the changes were already applied during image creation. It might be convenient in case the admin wants to run `oscap` from CLI to double-check, but it unnecessarily bloats the image.

Plus it might provide confusing results if the admin used `datastream = ` for `[customizations.openscap]` since the installed `scap-security-guide` RPM will likely have different content version.

evgenyz: I've removed these sections. By the time it hits downstream they won't be needed (they are implicit in the upstream of IB now).

jan-cerny (approved):

PR #2092, opened 2024-03-12
Build and test with NSS crypto back end

jan-cerny:
This commit adds a GitHub Actions CI gating job that runs on Fedora and verifies build with NSS as a crypto backend. Related to: https://github.com/OpenSCAP/openscap/pull/2074


evgenyz:
Why not Rawhide?

evgenyz (approved):
Actually, doesn't matter.

PR #2097, opened 2024-03-19
Remove CPE dictionary

evgenyz:
Fixes: #1683 

One dummy record is required by the scanner to be present.


jan-cerny:
> One dummy record is required by the scanner to be present.

@evgenyz  Do you know why is this required? Can we do something so that it isn't required and we can remove it completely?

evgenyz:
> > One dummy record is required by the scanner to be present.
> 
> @evgenyz Do you know why is this required? Can we do something so that it isn't required and we can remove it completely?

I was thinking about that. But then I realized that a dummy record, being fully valid at the same time, could be a good example for a maintainer that might want to adjust the dictionary. So I don't see it as particularly harmful.

evgenyz:
> But wouldn't removing the call to cpe_session_add_default_cpe from cpe_session_new?

Even if that's true, I'd probably still prefer to have this record as a demo of dictionary items.

evgenyz:
> there are some dictionary files that can be used as demo in the tests directory

How one would obtain that knowledge? Test suite is a rather enigmatict thing and we don't really have CPE dictionary well-documented. While this linux.* entry is 1) correct 2) right where it should be used 3) does not require any maintenance.

jan-cerny (approved):

PR #2100, opened 2024-04-03
Fix potential NULL pointer dereference before calling xmlXPathEvalExp…

icy17:
…ression

### DIFF
--- a/src/DS/sds.c
+++ b/src/DS/sds.c
@@ -791,7 +791,14 @@ static int ds_sds_compose_catalog_has_uri(xmlDocPtr doc, xmlNodePtr catalog, con
 	xpathCtx->node = catalog;
 
 	char *expression = oscap_sprintf("cat:uri[@uri = '%s']", uri);
+	if (expression == NULL)
+	{
+		oscap_seterr(OSCAP_EFAMILY_OSCAP, "Error: Unable to create XPath expression.");
+		xmlXPathFreeContext(xpathCtx);
 
+		return -1;
+	}
+	
 	xmlXPathObjectPtr xpathObj = xmlXPathEvalExpression(
 			BAD_CAST expression,
 			xpathCtx);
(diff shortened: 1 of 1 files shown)


src/DS/sds.c
```diff
@@ -791,7 +791,14 @@ static int ds_sds_compose_catalog_has_uri(xmlDocPtr doc, xmlNodePtr catalog, con
 	xpathCtx->node = catalog;
 
 	char *expression = oscap_sprintf("cat:uri[@uri = '%s']", uri);
+	if (expression == NULL)
+	{
+		oscap_seterr(OSCAP_EFAMILY_XML, "Error: Unable to create XPath expression.");
```
evgenyz, line 796: Possible problem's origin is not `libxml` or XML-related code. I believe error's affinity should be `OSCAP_EFAMILY_OSCAP`.

src/DS/sds.c
```diff
@@ -1012,6 +1019,13 @@ static int ds_sds_compose_has_component_ref(xmlDocPtr doc, xmlNodePtr datastream
 	xpathCtx->node = datastream;
 
 	char *expression = oscap_sprintf("*/ds:component-ref[@xlink:href = '#%s' and @id = '%s']", filepath, cref_id);
+	if (expression == NULL)
+	{
+		oscap_seterr(OSCAP_EFAMILY_XML, "Error: Unable to create XPath expression.");
```
evgenyz, line 1024: Ditto.

evgenyz (changes requested):
LGTM. Thank you!

src/DS/sds.c
icy17: I've changed `OSCAP_EFAMILY_XML` to `OSCAP_EFAMILY_OSCAP`

icy17:
Hi! Thanks for merging my pull requests!
I noticed these PR addresse potential security issues. I'm wondering if it would be appropriate to apply for CVEs for these bugs? Could you let me know if that's possible and what steps I should take?

PR #2111, opened 2024-05-29
Adding oscap-ssh fix_2107

avalanchein:


jan-cerny (approved):
The CI fail on Fedora Latest isn't caused by the contents of this PR and has been fixed by https://github.com/OpenSCAP/openscap/pull/2118

PR #2113, opened 2024-05-29
Remove DS command

### LINKED ISSUE #2086: Drop `oscap ds` feature
#### Description of Problem:

We don't like it, we don't need it. OpenSCAP is a scanner not an XML composer.

#### OpenSCAP Version:

1.3

Drop in 1.4

Mab879:
Keeping the code around as it used in other places in OpenSCAP.

Closes #2086


jan-cerny:
You will have to do something with failing CTest tests. Some of them will have to be removed, some of them will have to be changed so that they don't use the `oscap ds` subcommand. 

For example, I suggest changing `tests/bz2/test_bz2_datastream.sh` so that it doesn't compose the data stream during the test but instead it evaluates a static data stream stored in the repository (you can use the removed command `oscap ds sds-compose` to generate the data :hear_no_evil:  ).

Mab879:
Closing in favor of #2127

PR #2115, opened 2024-05-31
Document that you can't use -j for ctest

Mab879:
It causes issues


jan-cerny (approved):
It's unfortunate that they aren't working. Investigating them could help us to solve some concurrency issues or other mysteries.

jan-cerny:
The CI fail is unrelated to this PR because this PR only changes documentation.

PR #2116, opened 2024-05-31
Remove the option to build with PCRE2

### LINKED ISSUE #2022: Deprecate PCRE and switch the build system to PCRE2 by default
#### Description of Problem:

The PCRE library is no longer supported. All projects were encouraged to use PCRE2 successor.
We already have the option to build with PCRE2 since 1.3.9. For 1.4 it would make sense
to use it by default and deprecate or drop support for PCRE.

#### OpenSCAP Version:

1.3.x

Mab879:
The project will now only built with PCRE2 only.

Closes #2022


cmake/FindPCRE.cmake
```diff
@@ -1,37 +0,0 @@
...
-#		Searching headers and libraries is very simple and is NOT as powerful as scripts
-#		distributed with CMake, because LuaDist defines directories to search for.
-#		Everyone is encouraged to contact the author with improvements. Maybe this file
-#		becomes part of CMake distribution sometimes.
-
-# - Find pcre
-# Find the native PCRE headers and libraries.
-#
-# PCRE_INCLUDE_DIRS	- where to find pcre.h, etc.
-# PCRE_LIBRARIES	- List of libraries when using pcre.
```
jan-cerny, line 15: I have found `PCRE_LIBRARIES` in multiple places in the project, they should be removed as well.

docs/developer/developer.adoc
```diff
@@ -89,10 +89,8 @@ libxml-xpath-perl libperl-dev libbz2-dev librpm-dev g++ libyaml-dev \
 libxmlsec1-dev libxmlsec1-openssl
 ----
 
-Since version 1.3.9 OpenSCAP can use either PCRE or PCRE2 library to handle regular
-expressions. The default behaviour for 1.3.x line of versions is to try and link
-with PCRE. In order to switxh the build system to PCRE2 one must configure
-CMake with '-DWITH_PCRE2=True' parameter. Dependencies: Debian/Ubuntu — pcre2-dev,
+Since version 1.4.0 OpenSCAP uses PCRE2 library to handle regular
+expressions. Dependencies: Debian/Ubuntu — pcre2-dev,
 Fedora/RHEL — pcre2-devel.
```
jan-cerny, line 94: please update all the install commands above

jan-cerny (changes requested):

jan-cerny (approved):
I have built the project on Fedora 40 (using the default configuration) and I have verified that the built shared object links to /lib64/libpcre2-8.so.0.

LGTM. Can you mark it as ready to merge?

Mab879:
> I have built the project on Fedora 40 (using the default configuration) and I have verified that the built shared object links to /lib64/libpcre2-8.so.0.
> 
> LGTM. Can you mark it as ready to merge?

Done.

jan-cerny:
CI fail on Fedora Latest isn't caused by this PR and will be fixed in https://github.com/OpenSCAP/openscap/pull/2118 and Packit jobs on CS 8 fail because CS 8 has been sunset.

Issue #1720, opened 2021-04-06
Deprecate oscap_source_validate_schematron


jan-cerny:
#### Description of Problem:
During https://github.com/OpenSCAP/openscap/pull/1703 we have discovered that the API function oscap_source_validate_schematron is weird, It has `const char *outfile` argument which specifies the path to the output file. The schematron output is stored into this file. If outfile is NULL it is printed to stdout instead. In our code base we call this function only twice and both occurrences have it set to NULL. But the existence of the argument complicates the underlying static functions, they all need to have this outfile as an extra parameter. By deprecating the function and removing the parameter we would simplify them. An alternative view is that this is an improper way of communicating the results of the function and instead of writing into a file it should return a string.

#### OpenSCAP Version:
1.3.4

#### Operating System & Version:
all

#### Steps to Reproduce:

1. 
2. 
3. 

#### Expected Results:
oscap_source_validate_schematron is deprecated and a better function is invented

PR #2122, opened 2024-06-06
Simplify oscap_source_validate_schematron

jan-cerny:
The function `oscap_source_validate_schematron` is always called with the second parameter `outfile` set to NULL, therefore this parameter isn't needed and we can decide that the output will be printed to stdout. That allows further simplification of static functions in schematron.c.

Fixes: #1720


evgenyz (approved):
LGTM. Thanks!

Issue #771, opened 2017-07-03
Advertise path to SSG in remediation scripts


jan-cerny:
We have now a header in remediation scripts:

```
    3 # Bash remediation role for profile xccdf_org.ssgproject.content_profile_pci-dss
    4 # Profile Title:  PCI-DSS v3 Control Baseline for Red Hat Enterprise Linux 7
    5 # Profile Description:  This is a *draft* profile for PCI-DSS v3.
    6 #
    7 # Benchmark ID:  xccdf_org.ssgproject.content_benchmark_RHEL-7
    8 # Benchmark Version:  0.1.33
    9 #
   10 # XCCDF Version:  1.2
   11 #
   12 # This file was generated by OpenSCAP 1.2.15 using:
   13 #   $ oscap xccdf generate fix --profile xccdf_org.ssgproject.content_profile_pci-dss sds.xml 
   14 #
```
It would be nice to have a REAL path in the command on line 13, instead of sds.xml.

We think that people can download the scripts from the internet, so that way they can find that SSG exists on their system somewhere in `/usr/share`. (Of course this should work for any path.)

Unfortunately, it is not feasible to do this is maint-1.2 branch without breaking API. Thanks @kjankov  for investigating this.

I'm opening this issue so that we can implement this issue when we will work on OpenSCAP 1.3.

evgenyz:
@jan-cerny  Can we do it now without breaking API? Or it should be moved to `1.4`? Or dropped?

jan-cerny:
AFAIK this part of API hasn't changed so it should be either moved to future or dropped.

It would be a nice UX improvement, but it isn't causing any problems.

evgenyz:
Fixed in https://github.com/OpenSCAP/openscap/pull/2136.

PR #2136, opened 2024-07-10
Introduce ability to generate kickstarts

jan-cerny:
This PR introduces ability to generate kickstarts from SCAP content using:

```
oscap xccdf generate fix --fix-type kickstart ds.xml
```

The kickstart will be generated from kickstart snippets in XCCDF rules in the input SCAP content.
The kickstart snippets need to be stored in `<fix>` elements with `system` attribute set to `urn:xccdf:fix:script:kickstart`.

For more details, please read commit messages of each commit.

### Rationale:

We'd like to provide a lightweight system hardening for Anaconda, that is not dependent on UI and does not require user intervention. This should be pretty much similar to Image Builder's experience for users that can not use it for some reason.

### How to test:

1. Build RHEL 9 data stream from the ComplianceAsCode pull request master as of 2024-07-29 or newer.
2. Generate the kickstart `ks.cfg`:
```
./oscap_wrapper xccdf generate fix --profile stig --output ks.cfg --fix-type kickstart ~/work/git/scap-security-guide/build/ssg-rhel9-ds.xml
```
3. Add AppStream repository to the generated kickstart by adding this line (edit the URL)
```
repo --name=extra-repository --baseurl=http://DOWNLOAD_HOSTNAME/released/RHEL-9/9.4.0/AppStream/x86_64/os/
```
4. Serve the kickstart on a HTTP server
```
python3 -m http.server
```
5. Run `virt-manager`, click on Create a new virtual machine. Choose `Network Install`.
6. Provide installation URL for RHEL 9.4: ` http://DOWNLOAD_HOSTNAME/released/RHEL-9/9.4.0/BaseOS/x86_64/os/`
Under the installation URL field, click on `URL options` and insert there `inst.ks=http://192.168.124.1:8000/ks.cfg`. Complete next steps in the wizard.
7. VM will launch. Configure partitioning, continue with installation. Watch Anaconda progress.
8. After installation and reboot, log in to the installed VM and observe state of the system. Read Anaconda logs. Also, you can perform a oscap scan of the installed system.

### Review hints:

Start by reading the proposed user manual text in the PR.


comps:
I don't see it currently in the code, but note that an Anaconda kickstart can have multiple `%packages` sections, so we could have our own, independent of whatever the user wants to add in their own `%packages`.

This would allow us to use further options to `%packages` documented on https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#chapter-9-package-selection , like `%packages --exclude-weakdeps` (or `--excludeWeakdeps` to preserve RHEL-8 compatibility) to install the specified packages with minimum dependencies. Maybe something to consider.

Similarly, `%post` also has
```
%post [--erroronfail] [--interpreter /usr/bin/python] [--log LOG] [--nochroot]
```
and I imagine we'd definitely want `--erroronfail`, and perhaps `--nochroot` to use the Anaconda environment's tooling rather that the installed system tooling.
https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#chapter-6-post-installation-script

---

Also, what about the `bootloader` command? .. Presumably we don't want it just hardcoded in `common_kickstart_header`, but rather dynamic, so stuff like `fips=1` can be added there (alongside other kernel command line options).

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1286,7 +1395,204 @@ static int _xccdf_policy_generate_fix_other(struct oscap_list *rules_to_fix, str
...
+"\n"
+"# Set system keyboard type / layout (required)\n"
+"keyboard --vckeymap us\n"
+"\n"
+"# Configure network information for target system and activate network devices in the installer environment (optional)\n"
+"# --onboot	enable device at a boot time\n"
+"# --device	device to be activated and / or configured with the network command\n"
+"# --bootproto	method to obtain networking configuration for device (default dhcp)\n"
+"# --noipv6	disable IPv6 on this device\n"
+"network --onboot yes --device eth0 --bootproto dhcp --noipv6\n"
```
comps, line 1521: Just
```
network --onboot yes --bootproto dhcp
```
should be enough. The rest is autodetected by Anaconda.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1286,7 +1395,204 @@ static int _xccdf_policy_generate_fix_other(struct oscap_list *rules_to_fix, str
...
+"\n"
+"# The selected profile will restrict root login\n"
+"# Add a user that can login and escalate privileges\n"
+"# Plaintext password is: admin123\n"
+"user --name=admin --groups=wheel --password=$6$Ga6ZnIlytrWpuCzO$q0LqT1USHpahzUafQM9jyHCY9BiE5/ahXLNWUMiVQnFGblu0WWGZ1e6icTaCGO4GNgZNtspp1Let/qpM7FMVB0 --iscrypted\n"
+"\n"
+"# Configure firewall settings for the system (optional)\n"
+"# --enabled	reject incoming connections that are not in response to outbound requests\n"
+"# --ssh		allow sshd service through the firewall\n"
+"firewall --enabled --ssh\n"
```
comps, line 1537: I'm pretty sure we can remove this - our testing runs without the `firewall` command as Anaconda by default installs `firewalld`, and installing the `openssh-server` package adds port 22 to it seamlessly.

comps:
Do we really want a very verbose kickstart file as an output, giving the user many options to customize it (url/cdrom/harddrive), explaining every option?

Why not just point to https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html (or the RHEL version of it, ie. https://docs.redhat.com/en-us/documentation/red_hat_enterprise_linux/8/pdf/system_design_guide/Red_Hat_Enterprise_Linux-8-System_Design_Guide-en-US.pdf - APPENDIX J ) and let the user figure out what they need for their specific setup ?

jan-cerny:
@comps great ideas! thanks a lot!

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1286,7 +1395,204 @@ static int _xccdf_policy_generate_fix_other(struct oscap_list *rules_to_fix, str
...
+}
+
+static int _generate_kickstart_post(struct kickstart_commands *cmds, const char *profile_id, const char *input_path, int output_fd)
+{
+	_write_text_to_fd(output_fd, "%post\n");
+	char *dup = strdup(input_path);
+	char *basename = oscap_basename(dup);
+	free(dup);
+	char *oscap_command = oscap_sprintf(
+		"oscap xccdf eval --remediate --profile '%s' /usr/share/xml/scap/ssg/content/%s\n",
```
jan-cerny, line 1464: This will work sometimes, but sometimes it won't work. How do we ensure that the file is the same as the file that the user provided to `oscap` when generating the kickstart? And even a bigger problem, what if the user used a tailoring when generating the kickstart?

evgenyz: This is why we need to at least fix this: https://github.com/OpenSCAP/openscap/issues/771.

jan-cerny: Amazing! I didn't know that we have issue for that. Great news is that the issue is already "accidentally" fixed in this PR! The headers now contain actual file path of the file used as input to the `oscap xccdf generate fix` command.

The knowledge of the actual input file path is also reused in this place. Here, I take its base name and concatenate it with the well-known scap-security-guide directory path. This way I achieved that it will work sometimes. The sometimes specifically means that 1. user uses a data stream from the scap-security-guide package to generate the KS and 2. the version of the scap-security-guide package used to generate the KS is the same that will be installed on the installed system and 3. user doesn't use tailoring when generating the KS. If any of the three conditions isn't true then it won't work or will give surprising results. I think that's insufficient therefore I opened this comment.

jan-cerny: I have this proposal how to solve this problem. We will document that the recommended workflow is to serve the SCAP source data stream and the optional XCCDF tailoring file on a web server. Then, the generated kickstart will contain a command to download these files from the server. We will not install scap-security-guide package during the installation. This way it will make sure that the same files are used for both kickstart generation before installation and remediation during installation. The generated kickstart can contain placeholders for the content URI and comments highlighting it. Also, we can have a CLI option that would get propagated to the generated kickstart. The key thing would be to have the workflow well documented.

jan-cerny: Another proposal has been suggested by @dahaic : The kickstart would explicitly request the specific version of scap-security-guide RPM package to be installed. The package would be fetch from CDN. This way we would ensure that the same data stream is used for the remediation as is used for the kickstart generation. The tailoring file could be inserted in a form of base64 blob or in a form of autotailor command.

jan-cerny:
I have improved some of the hardcoded items.

docs/manual/manual.adoc
```diff
@@ -1120,6 +1120,45 @@ For example, to generate a blueprint remediation for RHEL 8 OSPP profile, run:
...
+* lines starting with something else than a supported command are dropped
+* excess whitespace are trimmed
+
+Supported commands:
+* `package install package_name` - adds `package_name` to `%packages` section in the kickstart
+* `package remove package_name` - adds `-package_name` to `%packages` section in the kickstart
+* `service enable service_name` - adds `service_name` to list in the `--enabled=` option in the `services` command in commands section in the kickstart
+* `service disable service_name` - adds `service_name` to list in the `--disabled=` option in the `services` command in commands section in the kickstart
+* `post command` - adds `command` to the `%post` section the kickstart
+* `logvol command` - adds `logvol command` to the commands section the kickstart
```
jan-cerny, line 1148: I think this needs to be split into specific fields for a better defined experience.

comps:
> I have improved some of the hardcoded items.

Note that you really do need `reqpart` like I mentioned, RHEL-10 will not install without it given the kickstart in this PR, because Anaconda uses GPT partitions even without UEFI / Secure Boot.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1478,6 +1486,44 @@ static int _generate_kickstart_post(struct kickstart_commands *cmds, const char
...
+"# Initialize (format) all disks (optional)\n"
+"zerombr\n"
+"\n"
+"# The following partition layout scheme assumes disk of size 20GB or larger\n"
+"# Modify size of partitions appropriately to reflect actual machine's hardware\n"
+"#\n"
+"# Remove Linux partitions from the system prior to creating new ones (optional)\n"
+"# --linux	erase all Linux partitions\n"
+"# --initlabel	initialize the disk label to the default based on the underlying architecture\n"
+"clearpart --linux --initlabel\n"
```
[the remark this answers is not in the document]
evgenyz: Default options for `/boot` are OK, oscap will remediate them in post section.

evgenyz:
> Do we really want a very verbose kickstart file as an output, giving the user many options to customize it (url/cdrom/harddrive), explaining every option?

We definitely do not want to do that. It should be bare-minimum, resembling the default minimal installation. We should not comment KS itself, only our stuff, if needed.

evgenyz:
> Also, what about the `bootloader` command? .. Presumably we don't want it just hardcoded in `common_kickstart_header`, but rather dynamic, so stuff like `fips=1` can be added there (alongside other kernel command line options).

We definitely don't want to enforce an encrypted boot loader, also we would need this section to be built from the content snippets exactly because of different kernel options.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1286,7 +1403,240 @@ static int _xccdf_policy_generate_fix_other(struct oscap_list *rules_to_fix, str
...
+"# Set system keyboard type / layout (required)\n"
+"keyboard --vckeymap us\n"
+"\n"
+"# Configure network information for target system and activate network devices in the installer environment (optional)\n"
+"# --onboot	enable device at a boot time\n"
+"# --bootproto	method to obtain networking configuration for device (default dhcp)\n"
+"network --onboot yes --bootproto dhcp\n"
+"\n"
+"# Set the system's root password (required)\n"
+"# Plaintext password is: server\n"
```
evgenyz, line 1569: I'd prefer not to put any passwords into generated KS files, but in this case maybe we could randomly generate the password every time (and write its readable form in the comment section) to avoid creating a potential security weakness?

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1286,7 +1403,240 @@ static int _xccdf_policy_generate_fix_other(struct oscap_list *rules_to_fix, str
...
+"# Modify size of partitions appropriately to reflect actual machine's hardware\n"
+"#\n"
+"# Remove Linux partitions from the system prior to creating new ones (optional)\n"
+"# --all	erase all partitions\n"
+"# --initlabel	initialize the disk label to the default based on the underlying architecture\n"
+"clearpart --all --initlabel\n"
+"reqpart\n"
+"\n"
+"# Create primary system partitions (required for installs)\n"
+"part /boot --fstype=xfs --size=512 --fsoptions=\"nodev,nosuid,noexec\"\n"
```
evgenyz, line 1504: Isn't it already a part of `reqpart`?

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1286,7 +1403,240 @@ static int _xccdf_policy_generate_fix_other(struct oscap_list *rules_to_fix, str
...
+		_write_text_to_fd(output_fd, command);
+		_write_text_to_fd(output_fd, "\n");
+	}
+	oscap_iterator_free(logvol_it);
+	_write_text_to_fd(output_fd, "logvol swap --name=swap --vgname=VolGroup --size=2016\n");
+	return 0;
+}
+
+const char *common_kickstart_header = (
+"# Specify installation method to use for installation\n"
```
evgenyz, line 1529: We should get rid of generic KS-related comments. Also, we should get rid of everything that is optional, our KS file should be the most simple thing to install the system as it would be installed by default, but hardened. Everything extra is an exercise for the advanced customer who would use the feature.

src/XCCDF_POLICY/xccdf_policy_remediate.c
evgenyz: 1. It'd be nice to actually have just a full path here for the content. We don't care much about non-packaged DSes. It's up to the advanced customer to figure out how they want to fetch, store and use it (we can give an example in the docs, though).
2. If we can somehow embed tailoring (when `--tailoring` option is provided along with `generate fix`) and generate proper KS that would use it without requiring the user to serve it separately — it would be very nice, but let's not concentrate on it before everything else is complete.

tests/API/XCCDF/unittests/test_remediation_kickstart.ds.xml
```diff
@@ -0,0 +1,102 @@
...
+          # the service commands will end up in the command section
+          service enable sshd
+          service disable telnet
+
+          post mkdir /etc/scap
+        </fix>
+      </Rule>
+    </Benchmark>
+  </ds:component>
+</ds:data-stream-collection>
```
evgenyz, line 102: No newline.

src/XCCDF_POLICY/xccdf_policy_remediate.c
comps: No, there is a difference between
```
reqpart
```
and
```
reqpart --add-boot
```
which should be pretty obvious. :smile:

Just `reqpart` adds stuff like
* 1 MB `biosboot` partition on GUID partition tables, for grub2 data
* 256 / 512 MB `efi` partition on UEFI systems
* `PReP` boot partition on ppc64 / ppc64le
* etc.

(Basically hardware-related boot dependencies.)

src/XCCDF_POLICY/xccdf_policy_remediate.c
comps: I assume the minimal remaining stuff should still be consumable by Anaconda as-is, correct?

Or would the user need to add some things to make it work in the first place?

(I'm thinking of an analogy with osbuild and including `name`, `description`, `version`, etc.)

src/XCCDF_POLICY/xccdf_policy_remediate.c
evgenyz: Oh, yeah, we are still missing `--add-boot`, okay.

src/XCCDF_POLICY/xccdf_policy_remediate.c
evgenyz: The `cdrom` option should work for a default RHEL image, right?

src/XCCDF_POLICY/xccdf_policy_remediate.c
evgenyz: If we could be able to bring tailoring into the remediation generator context it will help us to also generate self-contained tailored Blueprints.

src/XCCDF_POLICY/xccdf_policy_remediate.c
comps: Depends on what you mean by "RHEL image".

I would argue it's pretty uncommon to inject `inst.ks=` to an ISO image, but even in that case, Anaconda figures out what to do on its own and sets sources to local media (USB/CD/DVD/etc.), without any `cdrom` or `hd` option.

The autodetection is the default. So I would argue against using any one method as "default".

jan-cerny:
@evgenyz please revisit, I have updated the PR

comps:
* Per @evgenyz suggestion above, remove `part /boot` and change `reqpart` to `reqpart --add-boot`, the mount options should be remediated by `oscap` in `%post`
* I would rename `VolGroup` to something meaningful, ie. `system`, to differentiate it from other VGs added by a sysadmin for non-OS storage
* I would remove all `--fstype=xfs` and just use RHEL default - that's currently XFS, but may be some other filesystem on a future RHEL version
* I would change `/` size from 10240 to 2000 (2GB), current minimal install is ~200 MB and if profiles add partitions for `/usr` and/or `/var`, then
 we'll have just a few MBs used on `/`
* I would also resize swap to 1000 (1GB), using 2016 seems random

jan-cerny:
@comps It would be useful to report all your suggestions to ComplianceAsCode/content and fix it there as well.

jan-cerny:
I have changed the partitions defaults according @comps 's suggestion.

jan-cerny:
Great news is that in the current state of PR for RHEL 9 STIG profile I get only 13 rules failing in the scan after first log in, most of them because they have no remediation.

evgenyz:
> @evgenyz please revisit, I have updated the PR

I'm on it.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1286,7 +1442,212 @@ static int _xccdf_policy_generate_fix_other(struct oscap_list *rules_to_fix, str
...
+		_write_text_to_fd(output_fd, command);
+		_write_text_to_fd(output_fd, "\n");
+	}
+	oscap_iterator_free(post_it);
+	_write_text_to_fd(output_fd, "%end\n");
+	_write_text_to_fd(output_fd, "\n");
+	return 0;
+}
+
+static char *_remove_slash(const char *in)
```
evgenyz, line 1528: Isn't `oscap_strrm` beter for that? https://github.com/OpenSCAP/openscap/blob/8f06d23de1845c4aaec9cd6b931405c1d5948101/src/common/util.h#L322

evgenyz: Oh, okay, you actually need a copy. Well, I'd suggest a name like `_copy_path_without_slashes` then.

evgenyz (approved):
Besides a small nitpick with a function name I think we are good to go with the initial implementation.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1508,10 +1508,11 @@ static int _generate_kickstart_post(struct kickstart_commands *cmds, const char
 	char *basename = oscap_basename(dup);
 	free(dup);
 	char *oscap_command = oscap_sprintf(
-		"oscap xccdf eval --remediate --profile '%s' /usr/share/xml/scap/ssg/content/%s\n",
+		"oscap xccdf eval --remediate --results-arf /root/openscap_data/arf.xml --report /root/openscap_data/report.html --profile '%s' /usr/share/xml/scap/ssg/content/%s\n",
```
evgenyz, line 1511: How about we just write down results in `/root/oscap_arf_results.xml` and `/root/oscap_report.html` and skip creating a directory?

jan-cerny: It could be fine, but the directory separates and concentrates multiple related files in a single directory.

evgenyz: In IB we put them right in the `/root` by default. Also, `oscap_data` does not follow [XDG specs](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

What actually worries me is the requirement to create this directory at some point. As our Kickstart will become more complex we could run in race condition with it. The `/root` OTOH is always there.

comps: I would maybe recommend `/var/log/anaconda` instead (for IB as well if it creates files in `/root`) - storing files in `/root` might become problematic with bootc images and the overall move towards immutable systems where `/etc` and `/var` are the only dynamic places (aside from `*/tmp`).

I'm not sure if `/var/log/anaconda` is always created, or if `inst.nosave=logs` disables just storing `.log` files or the creation of the directory as well, so that would need to be checked.

jan-cerny: I have now tried to replace `/root` by `/var/log/anaconda` and it doesn't work, the installation doesn't run the `oscap xccdf remediate` and in the logs after installation it says `No such file or directory`.

comps:
Going through what is in the generated kickstart, we are missing a few mandatory commands. These (along with reboot) should probably be above other `%` sections:
```
lang en_US.UTF-8
keyboard --vckeymap us
network --onboot yes --bootproto dhcp
rootpw changeme
timezone --utc Europe/Prague
```
Yes, `rootpw` is unfortunately required. We can either encode it as crypt-encoded, or leave it as plaintext, but it needs to be there.

So the full kickstart should look like
```
###############################################################################
#
# Kickstart for DISA STIG for Red Hat Enterprise Linux 9
#
# Profile Description:
...
# Install the operating system using this kickstart.
#
###############################################################################

lang en_US.UTF-8
keyboard --vckeymap us
network --onboot yes --bootproto dhcp
rootpw changeme
timezone --utc Europe/Prague
reboot

# Configure boot loader options (required for security compliance)
bootloader --append="pti=on vsyscall=none page_poison=1 slub_debug=P audit=1 audit_backlog_limit=8192"

# Disable and enable systemd services (required for security compliance)
services --disabled=debug-shell,autofs,kdump --enabled=pcscd,rsyslog,systemd-journald,firewalld,fapolicyd,chronyd,sshd,usbguard,auditd

# Create partition layout scheme (required for security compliance)
zerombr
clearpart --all --initlabel
reqpart --add-boot
part pv.01 --grow --size=1
volgroup system pv.01
logvol / --name=root --vgname=system --size=2000 --grow
...

# Packages selection (required for security compliance)
%packages
...
%end

# Perform OpenSCAP hardening (required for security compliance)
%post --erroronfail
oscap xccdf eval --remediate --results-arf /root/oscap_arf.xml --report /root/oscap_report.html --profile 'xccdf_org.ssgproject.content_profile_stig' /usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml
[ $? -eq 0 -o $? -eq 2 ]
%end
```
(for `%post`, I put the comment above the section -- we want to encourage sysadmins to create their own `%post` sections, not reuse the `oscap` one).

jan-cerny:
I have changed the %post section to fail if the oscap remediation fails. I have added default values that ensure fully automated installation. The default values include also partition defaults which will be added to the kickstart if the SCAP profile doesn't specify any partition requirement, which means automated installation for these profiles as well.

comps:
> I have changed the %post section to fail if the oscap remediation fails. I have added default values that ensure fully automated installation. The default values include also partition defaults which will be added to the kickstart if the SCAP profile doesn't specify any partition requirement, which means automated installation for these profiles as well.

You might want to read my post more carefully - I mentioned that `lang` and others are "required" by documentation, but not actually required, and I provided a full example kickstart that works and doesn't need `lang` or `keyboard` or `timezone`, but you added them anyway -- was that intentional?

Also, for profiles without partitions, maybe `autopart` (default) would be better, ie.
```c
const char *fallback_partition = (
	"# Create partition layout scheme (optional)\n"
	"zerombr\n"
	"clearpart --all --initlabel\n"
	"autopart --type=lvm\n"
);
```

mildas:
```
%post --erroronfail
oscap xccdf eval --remediate ...
[ $? -eq 0 -o $? -eq 2 ]
%end
```
If I got `--erroronfail` correctly, it will terminate right at the failed command. Thus, if `oscap` returns 2 (something failed to be remediated or missing remediation), it will terminate installation at that `oscap` line and won't `[ $? -eq 0 -o $? -eq 2 ]`.
Shouldn't it be rather something like
```
%post --erroronfail
oscap xccdf eval --remediate ... || [ $? -eq 2 ]
%end
```
@comps ?

jan-cerny:
> You might want to read my post more carefully - I mentioned that lang and others are "required" by documentation, but not actually required, and I provided a full example kickstart that works and doesn't need lang or keyboard or timezone, but you added them anyway -- was that intentional?

The important thing here is to distinguish between a kickstart that works and kickstart that enables a fully automated installation. Without the options I added today, the kickstart works (it worked even yesterday), but the installation isn't fully automated, because the user needs to configure the options in the Anaconda during installation.

For example, if you remove the "lang" keyword from the kickstart, the Anaconda will ask you which language you want to use during the installation and you have to click and choose one to proceed. But if there are all the options I added, the installation doesn't require any user interaction in the whole process and goes fully automatically.

jan-cerny:
I have used the autopart

mildas:
> --erroronfail simply means "if the shell returned non-0, abort the installation"".

Okay, in that case your solution is fine. But then there's assumption that 
```
oscap xccdf eval --remediate ...
[ $? -eq 0 -o $? -eq 2 ]
```
will be the last 2 commands in `%post`. Are we sure it will be like that? Will `post command` remediation add commands BEFORE `oscap xccdf eval`? And what about customers adding more stuff to `%post`? Because for that I miss `exit 1` there.

comps:
> will be the last 2 commands in %post. Are we sure it will be like that? Will post command remediation add commands BEFORE oscap xccdf eval? And what about customers adding more stuff to %post? Because for that I miss exit 1 there.

A kickstart will typically have more `%post` sections, all of them run independently in separately executed shells (or `--interpreter`s).

I guess we can change it to
```sh
[ $? -eq 0 -o $? -eq 2 ] || exit 1
```
for extra safety in case someone adds more code to our `%post` section.

jan-cerny:
I have add exit 1, add ability to consume custom %pre and %post sections, add ability to disable kdump and solved the situation when --profile isn't provided. Then, I have improved the user manual changes.

evgenyz (approved):
Okay, there is nothing I would like to change immediately, although we might have some adjustments after getting some feedback from stakeholders (and more experiments with CI etc). I consider it good enough for the first implementation. 

Thank you!

Issue #2131, opened 2024-06-27
Compilation Error: incompatible integer to pointer conversion returning 'int' from a function with result type 'char *'


jaredledvina:
#### Description of Problem:
Trying to build `openscap` from the https://github.com/OpenSCAP/openscap/releases/download/1.3.10/openscap-1.3.10.tar.gz release but `make` fails with:

```
[ 79%] Building C object src/OVAL/CMakeFiles/oval_object.dir/oval_message.c.o
/tmp/openscap-20240627-8095-tmbe9s/openscap-1.3.10/src/common/util.c:332:9: error: incompatible integer to pointer conversion returning 'int' from a function with result type 'char *' [-Wint-conversion]
        return strerror_r(errnum, buf, buflen);
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

#### OpenSCAP Version:
1.3.10

#### Operating System & Version:
macOS Sonoma 14.5 (M1 arm64)

#### Steps to Reproduce:

1. Follow https://github.com/OpenSCAP/openscap/blob/main/docs/developer/developer.adoc#1-building-openscap-on-linux to get the release
2. When running any `make` command (`make all`, `make openscap`, or just `make`) it'll fail with the error above

#### Actual Results:

Error 2: 
```
/tmp/openscap-20240627-8095-tmbe9s/openscap-1.3.10/src/common/util.c:332:9: error: incompatible integer to pointer conversion returning 'int' from a function with result type 'char *' [-Wint-conversion]
        return strerror_r(errnum, buf, buflen);
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

#### Expected Results:

Binaries are compiled happily

#### Additional Information / Debugging Steps:

I'm actually trying to make a Homebrew formula to build and install the `openscap` binary. My hope is to be able to use `oscap xccdf generate` specifically less so the scanning features.

PR #2141, opened 2024-07-29
Fix CI

evgenyz:
Stop building/testing on c8s (we are not going to release 1.4 there).

Start building/testing on c10s.

Fix MacOS X build. (Fixes: https://github.com/OpenSCAP/openscap/issues/2131)

Throw away PCREv1 from spec file (we don't care about it in 1.4+).

Switch to NSS.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -16,16 +16,16 @@ jobs:
   metadata:
     targets:
     - fedora-all-x86_64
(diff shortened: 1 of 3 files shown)


jan-cerny:
@evgenyz Can you change the CS 10 job to use NSS instead of Gcrypt?

jan-cerny (approved):

Issue #1962, opened 2023-03-22
OpenSCAP can't process CPE AL platforms


jan-cerny:
#### Description of Problem:
OpenSCAP can't process CPE AL platforms in a SCAP source data stream if the SCAP source data stream does not contain at least one CPE dictionary component.

The popular SCAP source data streams provided by SCAP Security Guide upstream project use `<cpe-lang:check-fact-ref>` for the `<cpe-lang:platform>` definition for a large majority of `<xccdf-1.2:platform>` applicability checks. However, there still exists product-wide platforms that reference a CPE item within a CPE dictionary, for example `<xccdf-1.2:platform idref="cpe:/o:redhat:enterprise_linux:9"/>`. These platforms could be easily changed to also reference a `<cpe-lang:platform>` instead. If all the `<xccdf-1.2:platform>` would reference a `<cpe-lang:platform>`  then none of them would reference a CPE dictionary item and therefore the CPE dictionary wouldn't be needed. That means that the SCAP source data stream will not contain any CPE dictionary. 

A SCAP source data stream with no CPE dictionary is a valid SCAP source data stream. However, I have found that OpenSCAP can't find the OVAL checks referenced from  `<cpe-lang:check-fact-ref>`  if there isn't at least a minimalist CPE dictionary in that SCAP source data stream. Evaluation of SCAP data streams where solely `<cpe-lang:check-fact-ref>`  is used but no CPE dictionary is available terminates with errors, see Actual results below.

#### OpenSCAP Version:
openscap-1.3.7-1.fc37.x86_64

This issue is also reproducible from upstream maint-1.3 branch as of 2023-03-22 as of HEAD a6d6753f1d56b7f2090457af74ca0386a1bc08aa.

#### Operating System & Version:
F 37

#### Steps to Reproduce:

1.  Download and unzip the attached example SCAP source data stream  [ds_without_cpe_dict.tar.gz](https://github.com/OpenSCAP/openscap/files/11038705/ds_without_cpe_dict.tar.gz)
2.  oscap xccdf eval ds_without_cpe_dict.xml

#### Actual Results:
```
--- Starting Evaluation ---

Title   Test Rule
Rule    xccdf_moc.elpmaxe.www_rule_1
Result  notapplicable

OpenSCAP Error: Unable to open file: 'cpe-oval.xml' [/builddir/build/BUILD/openscap-1.3.7/src/source/oscap_source.c:288]
[...]

PR #2142, opened 2024-07-30
Process CPE AL platforms if CPE dict isn't part of DS

jan-cerny:
With this change, we will be able to process SCAP source data streams which use CPE AL platforms in XCCDF rules but at the same time there is no CPE dictionary present in the SCAP source data stream. Definition of a CPE dictionary isn't mandatory for platforms to be evaluated.

A small test for this situation is introduced as well.

Fixes: #1962


evgenyz (approved):
Cool, thanks!

Makes sense to port it to 1.3 as well.

PR #2143, opened 2024-07-30
Update User Manual

jan-cerny:
Update or remove some information from user manual that changed or are no longer relevant.


jan-cerny:
Question: What do you suggest doing with Section 13.2. Auditing Security Vulnerabilities of Red Hat Products?

Mab879:
> Question: What do you suggest doing with Section 13.2. Auditing Security Vulnerabilities of Red Hat Products?

I think with 1.4 we should remove this.

jan-cerny:
I have [Remove Section Auditing Security Vulnerabilities of Red Hat Products Then I have used a better profile name and

evgenyz (approved):

PR #2146, opened 2024-08-06
Enable autotailor to process multi-profile JSON Tailorings

evgenyz:
The script will now accept multiple profiles in JSON Tailorings and also will use command-line options to update existing profiles or will create a new profile in the XCCDF tailoring.

Also, I've added a synonym option `--tailoring-profile-id` for the existing rather confusing option `--new-profile-id`.

I've tried to keep backward-compatibility both in logic and in the CLI API.

### DIFF
--- a/utils/autotailor
+++ b/utils/autotailor
@@ -52,21 +52,18 @@ def is_valid_xccdf_id(string):
         string) is not None
 
 
-class Tailoring:
+class Profile:
     def __init__(self):
-        self.id = "xccdf_auto_tailoring_default"
         self.reverse_dns = DEFAULT_REVERSE_DNS
-        self.version = 1
         self._profile_id = None
         self.extends = ""
-        self.original_ds_filename = ""
         self.profile_title = ""
 
-        self.value_changes = []
-        self.rules_to_select = []
-        self.rules_to_unselect = []
-        self.groups_to_select = []
-        self.groups_to_unselect = []
+        self.value_changes = set()
+        self.rules_to_select = set()
+        self.rules_to_unselect = set()
+        self.groups_to_select = set()
+        self.groups_to_unselect = set()
         self._rule_refinements = collections.defaultdict(dict)
         self._value_refinements = collections.defaultdict(dict)
 
@@ -75,7 +72,7 @@ class Tailoring:
         if self._profile_id is not None:
             return self._profile_id
         else:
-            return self.extends + "_customized"
+            return self.extends + DEFAULT_PROFILE_SUFFIX
 
     @profile_id.setter
     def profile_id(self, new_profile_id):
@@ -100,7 +97,7 @@ class Tailoring:
         if not is_valid_xccdf_id(rule_id):
             msg = f"Rule id '{rule_id}' is invalid!"
             raise ValueError(msg)
-        enumeration = Tailoring._find_rule_enumeration(attribute)
+        enumeration = Profile._find_rule_enumeration(attribute)
         if value in enumeration:
             return
         allowed = ", ".join(map(quote, enumeration))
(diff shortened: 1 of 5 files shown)


utils/autotailor
```diff
@@ -339,7 +367,9 @@ def get_parser():
     parser.add_argument(
         "-j", "--json-tailoring", metavar="JSON_TAILORING_FILENAME", default="",
         help="JSON Tailoring (https://github.com/ComplianceAsCode/schemas/blob/main/tailoring/schema.json) "
-        "filename.")
+        "filename. When a JSON tailoring is accompanied with additional adjustments from the command-line"
```
jan-cerny, line 370: missing space at the end

utils/autotailor
```diff
@@ -339,7 +367,9 @@ def get_parser():
...
+        "options either a new profile would be added to the resulting XCCDF tailoring file or the profile"
```
jan-cerny, line 371: dtto

utils/autotailor.8
```diff
@@ -57,16 +57,22 @@ Specify the rule to select. The rule ID can be either full, or just the suffix,
...
+\fB--new-profile-id NEW_PROFILE_ID\fR
+.RS
+Synonym of --tailored-profile-id.
+.RE
+.TP
 \fB--json-tailoring JSON_TAILORING_FILE\fR
 .RS
 Import tailoring from a JSON file (https://github.com/ComplianceAsCode/schemas/tree/main/tailoring). This option makes BASE_PROFILE_ID positional argument optional.
 However, data passed in the command line options takes precedence over JSON contents, including the BASE_PROFILE_ID argument.
+When a JSON tailoring is accompanied with additional adjustments from the command-line options either a new profile would be added to the resulting XCCDF tailoring file or the profile with TAILORED_PROFILE_ID identifier, if present, will be updated accordingly.
```
jan-cerny, line 75: It doesn't seem to work the way it's documented here. For example:

```
utils/autotailor --id-namespace "com.example.www" --json-tailoring tests/utils/custom.json \
 --select R999 tests/utils/data_stream.xml.


Updating an existing profile with `--tailored-profile-id JSON_P1` seems to work.
```

should create a new profile but it doesn't create a new profile, the output is the same as if run without the `--select R999` option.

utils/autotailor
```diff
@@ -52,21 +52,18 @@ def is_valid_xccdf_id(string):
...
 
-        self.value_changes = []
-        self.rules_to_select = []
-        self.rules_to_unselect = []
-        self.groups_to_select = []
-        self.groups_to_unselect = []
+        self.value_changes = set()
+        self.rules_to_select = set()
+        self.rules_to_unselect = set()
+        self.groups_to_select = set()
```
jan-cerny, line 65: Wouldn't the lists be better to ensure deterministic output? With this change, the generated profiles are ordered randomly. The order or elements shouldn't matter for the functionality but could be unpleasant for users running diffs or our tests. I suggest using lists or keep sets but sort the items at serializing.

utils/autotailor
jan-cerny, line 370: I suggest improving the help text to clarify that the JSON is an input file that needs to be created and supplied by the user ie. it isn't an output file generated by autotailor.

utils/autotailor
```diff
@@ -296,32 +257,99 @@ class Tailoring:
...
+        if self.profile_title:
+            title.set("override", "true")
         else:
-            raise ValueError("JSON Tailoring does not define any profiles.")
+            title.set("override", "false")
+        title.text = self.profile_title
+
+        self._add_group_select_operations(profile)
+        self._add_rule_select_operations(profile)
+        self._add_value_overrides(profile)
```
jan-cerny, line 282: these methods should be renamed to make it clear that they convert something to XML

utils/autotailor.8
evgenyz: Well, okay, the thing is that if you actually want a new profile you absolutely must provide base profile id. I can't make it a required parameter if --json-tailoring is present since it is ambiguous what the user actually wants (and depends on the content of the JSON file). I don't want to ruin backward compatibility, so that's what I came up with. If you have any suggestions on how to work around this, I'm all ears.

utils/autotailor
evgenyz: Fixed.

utils/autotailor
evgenyz: Fixed.

utils/autotailor
evgenyz: Sets are now converted to sorted lists before being rendered as XML.

utils/autotailor
evgenyz: Text has been improved.

utils/autotailor
evgenyz: Done.

utils/autotailor.8
jan-cerny: My suggestion is to rephrase the doc text and help text. It needs to be clear that if the --json-tailoring option is set and other options are set at the same time, then the user needs to add `--tailored-profile-id`.

evgenyz: It's not about `--tailored-profile-id` (it could be auto-generated) but about BASE_PROFILE positional argument. Which we can't deduce.

jan-cerny: but `--tailored-profile-id` works for me as well

evgenyz: I've reworded the help message a bit.

jan-cerny (approved):
I have tried various other options together with the `--json-tailoring` option. I think that the autotailor tool now behaves according to the proposed documentation change.

PR #2147, opened 2024-08-06
Improve Kickstart remediations

evgenyz:
Add header-less variant via `--raw` switch.

The purpose of the `raw` mode is to be tools, integrations and customization-friendly. It is not meant to be directly consumable by a user.

For now the behavior is implemented only for Kickstart remediation type. We'll derefine the `raw` shape of others later.


jan-cerny:
@evgenyz Nice idea! I think that we can go this way. Please be aware that for some remediation types (Ansible, Kickstart) the "header" also contains essential parts of the output that shouldn't be removed. On the other hand, in some remediation types (Bash, Kickstart) additional comments are added also in a lower part of the generated output and these are not essential so we can remove these in the `--raw` mode.

evgenyz:
Come to think about it, if we are going to backport it we'd better not break the API in this change. Hmm.

evgenyz:
OTOH, we already broke it in the initial PR with the Kickstart remediation.

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1786,11 +1785,13 @@ static int _xccdf_policy_generate_fix_kickstart(struct oscap_list *rules_to_fix,
...
 	);
-	char *password = oscap_generate_random_string(24, NULL);
-	char *common = oscap_sprintf(common_template, password);
-	_write_text_to_fd(output_fd, common);
-	free(password);
-	free(common);
+	if (raw == 0) {
+		char *password = oscap_generate_random_string(24, NULL);
+		char *common = oscap_sprintf(common_template, password);
+		_write_text_to_fd(output_fd, common);
```
jan-cerny, line 1791: But without the items from the `common_template`, the generated kickstart won't be sufficient for automated installation of the operating system, in other words the installation will require user interaction.

tests/API/XCCDF/unittests/test_remediation_kickstart_expected_raw.cfg
```diff
@@ -0,0 +1,56 @@
+# Create partition layout scheme (required for security compliance)
```
jan-cerny, line 1: Why are these headers kept in the raw output?

tests/API/XCCDF/unittests/test_remediation_kickstart_expected_raw.cfg
```diff
@@ -0,0 +1,56 @@
...
+zerombr
+clearpart --all --initlabel
+reqpart --add-boot
+part pv.01 --grow --size=1
+volgroup system pv.01
+logvol / --name=root --vgname=system --size=2000 --grow
+logvol swap --name=swap --vgname=system --size=1000
+logvol /var/tmp --name=vartmp --vgname=system --size=1024
+
+# Configure boot loader options (required for security compliance)
```
jan-cerny, line 11: dtto

src/XCCDF_POLICY/xccdf_policy_remediate.c
```diff
@@ -1844,10 +1846,11 @@ int xccdf_policy_generate_fix(struct xccdf_policy *policy, struct xccdf_result *
 			return 1;
 		}
 
-		if (_write_script_header_to_fd(policy, result, sys, input_file_name, tailoring_file_name, output_fd) != 0) {
-			oscap_list_free(rules_to_fix, NULL);
-			return 1;
-		}
+		if (raw == 0)
+			if (_write_script_header_to_fd(policy, result, sys, input_file_name, tailoring_file_name, output_fd) != 0) {
```
jan-cerny, line 1850: When generating Ansible Playbooks the output is invalid because this part contains essential part of the output.

src/XCCDF_POLICY/public/xccdf_policy.h
```diff
@@ -520,7 +520,7 @@ OSCAP_API bool xccdf_policy_resolve(struct xccdf_policy * policy);
  * @param output_fd write prescription to this file descriptor
  * @returns zero on success, non-zero indicate partial (incomplete) output.
  */
-OSCAP_API int xccdf_policy_generate_fix(struct xccdf_policy *policy, struct xccdf_result *result, const char *sys, const char *input_file_name, struct oscap_source *tailoring, int output_fd);
+OSCAP_API int xccdf_policy_generate_fix(struct xccdf_policy *policy, struct xccdf_result *result, const char *sys, const char *input_file_name, struct oscap_source *tailoring, int output_fd, int raw);
```
jan-cerny, line 523: missing Doxygen text for the new `raw` parameter

evgenyz: Fixed.

src/XCCDF_POLICY/xccdf_policy_remediate.c
evgenyz: Fixed.

tests/API/XCCDF/unittests/test_remediation_kickstart_expected_raw.cfg
evgenyz: After another round of discussion with Satellite , I've changed the way we format sections in raw mode. It does not have comment headers now, but for partitioning scheme it has pseudo-sections `# %partitions`, `# %logvols` closed with `# %end` to make it easier for them to parse.

src/XCCDF_POLICY/xccdf_policy_remediate.c
evgenyz: The `raw` form is not meant to be easy-consumable, its purpose is to be convenient for parsing and customization using automation tools. It is going to be tools-friendly, not user-friendly.

jan-cerny (approved):
For current upstream scap-secutiy-guide I have generated various remediations (Ansible, Bash, Kickstart, Blueprint) both with and without the `--raw` option for multiple profiles and checked them.

PR #2148, opened 2024-08-07
Generate random passwords for Kickstart's rootpw section

evgenyz:
While the password is pseudo-random it is still better than accidentally having a server with 'changeme' password.


src/common/util.h
```diff
@@ -384,6 +384,18 @@ char *oscap_trim(char *str);
 /// Print to a newly allocated string using a va_list.
 char *oscap_vsprintf(const char *fmt, va_list ap);
 
+/**
+ * Generates a pseudorandom string of a given length.
+ * If charset string is not NULL and its length is greater than 0,
+ * it will be used as a dictionary, otherwize a default alphanumeric set
```
jan-cerny, line 390: otherwise

jan-cerny (approved):

PR #2154, opened 2024-08-20
Fix a cast

jan-cerny:
The size_t is implementation-dependent data type, it shouldn't be cast to unsigned int.
Fixes failing test probes/filehash58/test_probes_filehash58.sh on s390x architecture.


evgenyz (approved):
CI failure is unrelated. LGTM.

Issue #2153, opened 2024-08-19
`autotailor-unit-tests` unit test runs when pytest is not installed


mildas:
#### Description of Problem:
During ctest, autotailor test runs even when pytest is not installed on a system. As it is not openscap requirement, it should be optional and run only when pytest is installed.

#### OpenSCAP Version:
1.4.0

#### Operating System & Version:
RHEL 10

#### Steps to Reproduce:

1. `ctest --force-new-ctest-process --output-on-failure`

#### Actual Results:
```
276/277 Test #276: autotailor-unit-tests ..............................................................***Failed    0.01 sec
/usr/bin/python3: No module named pytest
```

#### Expected Results:
```
276/277 Test #276: autotailor-unit-tests ..............................................................***Skipped    0.01 sec
```

#### Additional Information / Debugging Steps:
`if(PY_PYTEST)` conditon for https://github.com/OpenSCAP/openscap/blob/main/tests/utils/CMakeLists.txt should ensure that

PR #2155, opened 2024-08-20
Check autotailor unit test dependencies

jan-cerny:
We will run autotailor unit tests only if pytest is installed on the system. CMake will check for the presence of pytest and enable or disable the unit tests test accordingly.

Fixes: #2153


Mab879 (changes requested):
Thanks for the PR.

Having a status message would nice. Everything else looks good.

Mab879 (approved):
Thanks!

Issue #2151, opened 2024-08-16
Don't hard require `/var/lib/rpm`


cgwalters:
This code is suboptimal:

https://github.com/OpenSCAP/openscap/blob/ecf658f69c8a54085ed2b032c126b4863d0991fb/src/OVAL/probes/unix/linux/rpminfo_probe.c#L298

As bootc images won't have that symlink by default because we want `/var` to start out empty as much as possible.

>         * Therefore, when running openscap on a Fedora >=36 system scanning another systems (such as RHEL, SLES, Fedora<36)
>        * openscap's librpm will try to read the rpm db from /usr/lib/sysimage/rpm which doesn't exist and therefore won't work.

In general this kind of "cross OS scanning" is going to get into tricky corner cases. It would seem to be more sustainable to simply fork `/usr/bin/rpm` in the target environment (as a container).

Alternatively, please at least check if `/usr/lib/sysimage/rpm` exists and set it as the dbpath first. Or, just try a query with the default config and accept that if it returns content.

cgwalters:
To be clear we *can* add that symlink into bootc images, and maybe we'll need to if we keep finding more issues like this...but again on the flip side you are hardcoding to read from a now deprecated path, which would make sense to stop doing.

PR #2156, opened 2024-08-22
Improve RPM database path in RPM probes

jan-cerny:
### Description

The assumption that /var/lib/rpm is always a symlink to /usr/lib/sysimage/rpm was wrong. In bootc images, it isn't the case. As a result, all rules were evaluated as notapplicable when scanning a bootc image or container.

We will fix it the following way: We will first try if the "new" location /usr/lib/sysimage/rpm exists, and use it only if it exists.  If it doesn't exist, we will fall back to the "old" location /var/lib/rpm.

Fixes: https://issues.redhat.com/browse/RHEL-55251
Fixes: https://github.com/OpenSCAP/openscap/issues/2151

### Rationale

This would be a first step to get OpenSCAP ready for  bootc experience.

### Review hints

First, reproduce the problem and verify that we fixed it.
sudo podman pull quay.io/centos-bootc/centos-bootc:stream9
sudo oscap-podman quay.io/centos-bootc/centos-bootc:stream9 oval eval --results /tmp/results.xml oval.xml
sudo oscap-podman --oscap=/home/jcerny/work/git/openscap/build/oscap_wrapper quay.io/centos-bootc/centos-bootc:stream9 oval eval --results /tmp/results.xml oval.xml

Then, verify we haven't broken scanning old systems on example of the ubi8 image.
sudo podman pull ubi8
sudo oscap-podman ubi8 oval eval --results /tmp/results.xml oval.xml
sudo oscap-podman --oscap=/home/jcerny/work/git/openscap/build/oscap_wrapper ubi8 oval eval --results /tmp/results.xml oval.xml

Also, verify that the probes still work locally.
./oscap_wrapper oval eval --results /tmp/local.xml oval.xml

In each command examine the produced OVAL result carefully and check if there is an item matching every test and check if the collected data comes from the scanned target and not form the host or from elsewhere.

### DIFF
--- a/tests/probes/rpm/rpm_common.sh
+++ b/tests/probes/rpm/rpm_common.sh
@@ -11,10 +11,7 @@ RPMBUILD="${RPMBASE}/build"
 
 # Since Fedora 36 RPM database location changed, see
 # https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr
-# However, /var/lib/rpm/ still works as it is a symlink to
-# the new path, /usr/lib/sysimage/rpm/, in Fedora >= 36
-# Therefore, always use /var/lib/rpm/ as it always works.
-RPMDB_PATH="/var/lib/rpm/"
+RPMDB_PATH="/usr/lib/sysimage/rpm/"
 
 function rpm_build {
     require "rpmbuild" || return 255
--- a/src/OVAL/probes/unix/linux/rpm-helper.c
+++ b/src/OVAL/probes/unix/linux/rpm-helper.c
@@ -51,3 +51,32 @@ void rpmLibsPreload()
 	const char* rcfiles = "";
 	rpmReadConfigFiles(rcfiles, NULL);
 }
+
+void set_rpm_db_path()
+{
+	/*
+	 * Fedora >=36 changed the default dbpath in librpm from /var/lib/rpm to /usr/lib/sysimage/rpm.
+	 * See: https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr
+	 *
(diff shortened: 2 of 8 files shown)


tests/probes/rpm/rpm_common.sh
```diff
@@ -11,10 +11,7 @@ RPMBUILD="${RPMBASE}/build"
 
 # Since Fedora 36 RPM database location changed, see
 # https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr
-# However, /var/lib/rpm/ still works as it is a symlink to
-# the new path, /usr/lib/sysimage/rpm/, in Fedora >= 36
-# Therefore, always use /var/lib/rpm/ as it always works.
-RPMDB_PATH="/var/lib/rpm/"
+RPMDB_PATH="/usr/lib/sysimage/rpm/"
```
matusmarhefka, line 14: We still need to use `/var/lib/rpm/` here because, for example, on CentOS Stream 9:
```
# ls /usr/lib/sysimage/rpm/
ls: cannot access '/usr/lib/sysimage/rpm/': No such file or directory
```

matusmarhefka: Or maybe we can do similar as in openscap source, check if `/usr/lib/sysimage/rpm/` exists, if not use `/var/lib/rpm/`.

evgenyz:
This fix will make CI more adequate: https://github.com/OpenSCAP/openscap/pull/2157.

jan-cerny:
I have rebased the PR on the top of the latest upstream maint-1.3 branch

matusmarhefka (approved):

PR #2157, opened 2024-09-02
Cleanup CI for maint-1.3 branch

evgenyz:
We don't need fresh Fedoras or Ubuntus. We keep that version only for older RHELs and Ubuntu LTSes.

I've also removed MacOS X.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,9 +6,9 @@ name: Gating
 # events but only for the main branch
 on:
   push:
-    branches: [ '*' ]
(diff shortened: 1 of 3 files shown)


Mab879:
I would keep the Windows build, so that we don't accidentally break it. We did commit making it at least build on Windows during the 1.3 lifetime.

jan-cerny:
I agree with keeping the Windows build.

evgenyz:
Okay, Windows is back.

jbjornivar:
Win is back

Mab879 (approved):

PR #2165, opened 2024-10-10
Allow preferring SCE

jan-cerny:
This commit introduces a new environment variable
`OSCAP_PREFERRED_ENGINE`. This variable allows users to set a preffered check engine for XCCDF rules. If a rule has multiple checks, the checks for the preffered check engine will be used. Allowed values: `SCE`, `OVAL`. If this variable is set to `SCE` and a rule has both SCE and OVAL checks the SCE check will be used. If this variable is set to `OVAL` and a rule has both SCE and OVAL checks the OVAL check will be used. If this environment variable isn't set, the standard XCCDF mechanism will be used for check selection.

This will allow us to explicitely prefer SCE checks when executing `oscap` in Containerfile in the `podman build` process when building hardened bootc images.

Also a small test is added to test this feature.

### DIFF
--- a/tests/sce/test_prefer_sce.sh
+++ b/tests/sce/test_prefer_sce.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+. $builddir/tests/test_common.sh
+
+set -e -o pipefail
+
+
+function test_prefer_sce_on () {
+    stdout=$(mktemp)
+    stderr=$(mktemp)
+    arf=$(mktemp)
+
+    OSCAP_PREFERRED_ENGINE="SCE" $OSCAP xccdf eval --verbose INFO --progress --profile common --results-arf "$arf" > "$stdout" 2> "$stderr" "$srcdir/test_prefer_sce.ds.xml" || ret="$?"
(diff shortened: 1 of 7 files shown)


tests/sce/test_prefer_sce.sh
```diff
@@ -0,0 +1,50 @@
...
+    grep -q "I: oscap: Evaluating definition 'oval:org.openscap.www:def:1': OVAL check for rule 1." "$stderr"
+    ! grep -q "I: oscap: Executing SCE check 'fedora/checks/sce/rule_1.sh'" "$stderr"
+
+    rm -rf "$stdout" "$stderr" "$arf"
+}
+
+
+test_prefer_sce_on
+test_prefer_sce_off
+test_prefer_oval_explicit
```
matusmarhefka, line 50: You can also consider adding a test for unknown value, e.g. `OSCAP_PREFERRED_ENGINE=FOO`.

jan-cerny:
I have add a test for the unknown value.
I have marked this PR as a ready for review.

matusmarhefka (approved):

PR #2169, opened 2024-10-21
Introduce oscap-bootc

jan-cerny:
The `oscap-bootc` tool is a convenience script that makes building hardened bootable container images easier.  This tool is designed to be used during the build of the bootable container image.  Include `oscap-bootc` in your `Containerfile` that will be used to build your bootable container image.


jan-cerny:
I have rebased this PR on the top of the latest upstream maint-1.3 branch.

utils/oscap-bootc.8
```diff
@@ -0,0 +1,29 @@
...
+
+The oscap-bootc tool can't be used anywhere else than in a Containerfile.
+
+.SH USAGE
+
+Usage of the tool mimics usage and options of oscap(8) tool.
+
+oscap-bootc OSCAP_ARGUMENT [OSCAP_ARGUMENT...]
+
+Refer to oscap(8) to learn about OSCAP_ARGUMENT options.
```
matusmarhefka, line 20: The script only supports certain options (specified in `parse_args` function) from `oscap(8)` so other options like `--rule`, `--skip-rule`, etc. won't work. Wouldn't it be better if we explicitly list only supported options in the man page?

So instead of
```
oscap-bootc OSCAP_ARGUMENT [OSCAP_ARGUMENT...]
```
we would have something like
```
oscap-bootc [--tailoring-file TAILORING_FILE]
    [--tailoring-id COMPONENT_REF_ID]
    [--results-arf FILE] [--report FILE]
    --profile PROFILE INPUT_DATA_STREAM
```

utils/oscap-bootc
```diff
@@ -0,0 +1,122 @@
...
+    try:
+        subprocess.run(oscap_cmd, env=env, check=True)
+    except subprocess.CalledProcessError as e:
+        if e.returncode not in [0, 2]:
+            print(e, file=sys.stderr)
+
+
+def main():
+    args = parse_args()
+    ensure_sce_installed()
```
matusmarhefka, line 116: It might be better if we do not install anything here and instead mention that `openscap-engine-sce` should be installed before using this script? Same as we mention `openscap-utils` and `scap-security-guide`. Another option is to make `openscap-engine-sce` a dependency of  `openscap-utils` package as without it the script won't work.

matusmarhefka (approved):

PR #2176, opened 2024-11-11
Merge maint-1.3 to main

jan-cerny:
This will bring useful changes from the maint-1.3 maintenance branch to the main branch. This includes oscap-bootc and changes supporting oscap-bootc.


.github/workflows/build.yml
```diff
@@ -6,9 +6,9 @@ name: Gating
 # events but only for the main branch
 on:
   push:
-    branches: [ '*' ]
+    branches: [ maint-1.3 ]
   pull_request:
-    branches: [ main, maint-1.3 ]
+    branches: [ maint-1.3 ]
```
matusmarhefka, line 11: I think this should be reverted to be as in the original file from `main`.

jan-cerny: yes!

matusmarhefka (approved):

8Z3737 (approved):

PR #2189, opened 2024-12-17
Pass OSCAP_BOOTC_BUILD variable to SCE checks

jan-cerny:
This change will cause that the `OSCAP_BOOTC_BUILD` environment variable will be passed from the external environment to the environment of SCE checks.

The outcome is that the `OSCAP_BOOTC_BUILD` environment variable can be used inside SCE checks to differentiate between code that is supposed to run only during building a bootable container image, the code that can't run  during building a bootable container image and code that can run in any environment.


jan-cerny:
After the latest changes in https://github.com/ComplianceAsCode/content/pull/12734 the PR https://github.com/ComplianceAsCode/content/pull/12734 no longer depends on this feature.

Issue #2110, opened 2024-05-24
Skipping a test, skips all subsequent tests in file


havardAasen:
#### Description of Problem:

If a file has multiple tests and one of them is skipped, all subsequent test in the file is skipped.

#### OpenSCAP Version:

Building from `main` with 74043f7.

#### Operating System & Version:

Debian Bookworm

#### Steps to Reproduce:

After a normal build, run `ctest -R test_ds_misc --verbose`
<details>
<summary>Log output</summary>

```
Test project /openscap/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 176
    Start 176: DS/test_ds_misc.sh

176: Test command: /openscap/tests/DS/test_ds_misc.sh
176: Working Directory: /openscap/build/tests/DS
176: Environment variables: 
176:  srcdir=/openscap/tests/DS
176:  top_srcdir=/openscap
176:  builddir=/openscap/build
176: Test timeout computed to be: 10000000
176: + sds_external_xccdf_in_ds                                    
176: TEST: sds_external_xccdf_in_ds
176: --- Starting Evaluation ---
176: 
176: Title
176: Rulxccdf_moc.elpmaxe.www_rule_1
176: Respass
176: 
176: OVAL Results are exported correctly.
176: XCCDF Results are exported correctly.
176: Result data stream exported correctly.
176: RESULT: PASSED
176: 
176: + sds_external_xccdf                                          
176: TEST: sds_external_xccdf
176: --- Starting Evaluation ---
176: 
176: Title
176: Rulxccdf_moc.elpmaxe.www_rule_1
176: Respass
176: 
176: OVAL Results are exported correctly.
176: XCCDF Results are exported correctly.
176: Result data stream exported correctly.
176: RESULT: PASSED
176: 
176: + sds_tailoring                                               
176: TEST: sds_tailoring
176: Document type: Source Data Stream
176: Imported: 2024-05-24T06:18:36
176: 
176: Stream: scap_com.example_datastream_with_checklist
176: Generated: 2016-09-26T15:00:00
176: Version: 1.2
176: Checklists:
176: 	Ref-Id: scap_com.example_cref_xccdf_01
176: 		Status: incomplete
176: 		Resolved: false
176: 		Profiles:
176: 		Referenced check files:
176: 			oval
[...]

PR #2195, opened 2025-01-11
tests: Don't set -e for compound tests managed by test_run

evgenyz:
The test_run wrapper should handle all return codes, we don't want to bail out early.

Fixes: https://github.com/OpenSCAP/openscap/issues/2110

### DIFF
--- a/tests/DS/test_ds_misc.sh
+++ b/tests/DS/test_ds_misc.sh
@@ -3,7 +3,7 @@
 # Author:
 #   Martin Preisler <mpreisle@redhat.com>
 
-set -e -o pipefail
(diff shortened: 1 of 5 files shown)


evgenyz:
@jan-cerny In this draft I've only changed a handful of compound tests, but we probably should do that to all of them. What do you think?

jan-cerny:
IIUC most of the test functions don't return any return value, they rely on that the `set -e` will cause that once any command in the function returns non-zero value the test will exit and fail. This is the case for all the `grep`, `assert_exists` and similar calls. So if you remove `set -e` the test will start passing permanently.

PR #2202, opened 2025-02-27
CMakeLists: add missing include dirs

chouquette:
This PR adds some more include dirs, in case the dependencies aren't all found in the same folder as the other ones

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -572,6 +572,12 @@ include_directories(
 	${XMLSEC_INCLUDE_DIRS}
 	${OPENSSL_INCLUDE_DIR}
 	${PCRE2_INCLUDE_DIRS}
+	${LIBXSLT_INCLUDE_DIR}
+	${RPM_INCLUDE_DIRS}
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -572,6 +572,12 @@ include_directories(
 	${XMLSEC_INCLUDE_DIRS}
 	${OPENSSL_INCLUDE_DIR}
 	${PCRE2_INCLUDE_DIRS}
+	${LIBXSLT_INCLUDE_DIR}
+	${RPM_INCLUDE_DIRS}
+	${GCRYPT_INCLUDE_DIRS}
+	${BLKID_INCLUDE_DIRS}
+	${POPT_INCLUDE_DIRS}
+	${SELINUX_INCLUDE_DIRS}
```
jan-cerny, line 580: It probably needs to be wrapped by a condition so that it doesn't add include directories if the libraries aren't found. See the output of the CI on Windows.

jan-cerny (approved):
Thank you for your contribution!

PR #2203, opened 2025-03-06
OPENSCAP-5235: Block remediation on deployed bootc system

jan-cerny:
OpenSCAP remediation is supposed to be used only at bootc container image build. Deployed bootc system is immutable, it can't be remediated with OpenSCAP and trying to do so would result in errors and bad user experience.

We will update OpenSCAP to print error message for users in case they try to run remediation on an already deployed bootc system, informing them that it is not possible and that the openscap remediation must be performed during container build.

### DIFF
--- a/src/XCCDF/xccdf_session.c
+++ b/src/XCCDF/xccdf_session.c
@@ -1906,32 +1906,6 @@ struct xccdf_rule_result_iterator *xccdf_session_get_rule_results(const struct x
 	return xccdf_result_get_rule_results(session->xccdf.result);
 }
 
-static bool _system_is_in_bootc_mode(void)
-{
-#ifdef OS_WINDOWS
-	return false;
-#else
-	#define BOOTC_PATH "/usr/bin/bootc"
-	struct stat statbuf;
-	if (stat(BOOTC_PATH, &statbuf) == -1) {
-		return false;
-	}
-	FILE *output = popen(BOOTC_PATH " status --format json 2>/dev/null", "r");
-	if (output == NULL) {
-		return false;
-	}
-	char buf[1024] = {0};
-	int c;
-	size_t i = 0;
-	while (i < sizeof(buf) && (c = fgetc(output)) != EOF) {
-		buf[i] = c;
-		i++;
-	}
-	pclose(output);
-	return *buf != '\0' && strstr(buf, "\"booted\":null") == NULL;
-#endif
-}
-
 int xccdf_session_remediate(struct xccdf_session *session)
 {
 	int res = 0;
@@ -1941,14 +1915,6 @@ int xccdf_session_remediate(struct xccdf_session *session)
 		return 1;
 	if(getenv("OSCAP_PROBE_ROOT") != NULL) {
 		oscap_seterr(OSCAP_EFAMILY_OSCAP, "Can't perform remediation in offline mode: not implemented");
-		return 1;
-	}
-	if (_system_is_in_bootc_mode()) {
-		oscap_seterr(OSCAP_EFAMILY_OSCAP,
-			"Detected running Image Mode operating system. OpenSCAP can't "
-			"perform remediation of this system because majority of the "
-			"system is read-only. Please apply remediation during bootable "
-			"container image build using 'oscap-im' instead.");
 		return 1;
 	}
 	xccdf_policy_model_unregister_engines(session->xccdf.policy_model, oval_sysname);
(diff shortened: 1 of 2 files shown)


src/XCCDF/xccdf_session.c
```diff
@@ -1906,6 +1906,27 @@ struct xccdf_rule_result_iterator *xccdf_session_get_rule_results(const struct x
 	return xccdf_result_get_rule_results(session->xccdf.result);
 }
 
+static int system_is_in_bootc_mode(void)
+{
+#ifdef OS_WINDOWS
+	return 0;
+#else
+	FILE *output = popen("bootc status --format json 2>/dev/null | jq \".status.booted\" 2>/dev/null", "r");
```
matusmarhefka, line 1914: Do we really want to rely on `jq` being available on the system?

jan-cerny: Great question! I don't know. The `jq` tool is a part of base images for both RHEL and CentOS 9 and 10. That means it should work for our use case. But I haven't checked other systems.

jan-cerny: I will explore using libyaml and bootc YAML output.

jan-cerny: Using YAML would require refactoring the current yamlfilecontent_probe code so that it could be reused here.

Instead, we can search for the booted key using substring search, which will be easier to implement.

src/XCCDF/xccdf_session.c
```diff
@@ -1917,6 +1943,14 @@ int xccdf_session_remediate(struct xccdf_session *session)
 		oscap_seterr(OSCAP_EFAMILY_OSCAP, "Can't perform remediation in offline mode: not implemented");
 		return 1;
 	}
+	if (_system_is_in_bootc_mode()) {
```
matusmarhefka, line 1946: I don't like the current behavior when user needs to wait for the initial scan results just to see the error:
```
# oscap xccdf eval --progress --remediate --profile stig /usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml
xccdf_org.ssgproject.content_rule_package_aide_installed:pass
xccdf_org.ssgproject.content_rule_aide_build_database:pass
xccdf_org.ssgproject.content_rule_aide_check_audit_tools:pass
xccdf_org.ssgproject.content_rule_aide_periodic_cron_checking:pass
xccdf_org.ssgproject.content_rule_aide_scan_notification:pass
xccdf_org.ssgproject.content_rule_aide_use_fips_hashes:fail
xccdf_org.ssgproject.content_rule_aide_verify_acls:pass
xccdf_org.ssgproject.content_rule_aide_verify_ext_attributes:pass
...
OpenSCAP Error: Detected running Image Mode operating system. OpenSCAP can't perform remediation of this system because majority of the system is read-only. Please apply remediation during bootable container image build using 'oscap-im' instead. [/builddir/build/BUILD/openscap-1.3.12/src/XCCDF/xccdf_session.c:1947]
```

Please move the error so it is printed right away and the initial scan is not even performed.

utils/oscap-xccdf.c
```diff
@@ -586,6 +586,43 @@ int xccdf_set_profile_or_report_bad_id(struct xccdf_session *session, const char
...
+	struct stat statbuf;
+	if (stat(BOOTC_PATH, &statbuf) == -1) {
+		return false;
+	}
+	FILE *output = popen(BOOTC_PATH " status --format json 2>/dev/null", "r");
+	if (output == NULL) {
+		return false;
+	}
+	size_t buf_size = CHUNK_SIZE;
+	char *buf = calloc(buf_size, sizeof(char));
```
matusmarhefka, line 606: Let's check for `buf == NULL` after `calloc` call.

utils/oscap-xccdf.c
```diff
@@ -586,6 +586,43 @@ int xccdf_set_profile_or_report_bad_id(struct xccdf_session *session, const char
...
+			if (new_buf == NULL) {
+				pclose(output);
+				return false;
+			}
+			buf = new_buf;
+		}
+		buf[i++] = c;
+	}
+	pclose(output);
+	return *buf != '\0' && strstr(buf, "\"booted\":null") == NULL;
```
matusmarhefka, line 622: We are leaking memory here, what about something like this:
```
bool rv = (*buf != '\0' && strstr(buf, "\"booted\":null") == NULL);
free(buf);
return rv;
```

matusmarhefka (changes requested):

matusmarhefka (approved):

PR #2210, opened 2025-03-24
Fix textfilecontent54_probe behaviour

Xeicker:
## Description

Reading OVAL docs: https://oval-community-guidelines.readthedocs.io/en/latest/oval-schema-documentation/independent-definitions-schema.html?highlight=textfilecontent54_object#textfilecontent54-object

The `instance` tag inside `textfilecontent54_object` should be able to take negative numbers. And these negative numbers should count from the last element matched. This is was not the case in oscap implementation.


evgenyz:
Hey! Thanks for the fix. Please add a test for this behavior.

tests/probes/textfilecontent54/test_negative_instance.sh
```diff
@@ -0,0 +1,37 @@
...
+
+    probecheck "textfilecontent54" || return 255
+	
+    local ret_val=0;
+    local DF="${srcdir}/test_negative_instance.xml"
+    local RF="results.xml"
+    
+    [ -f $RF ] && rm -f $RF
+
+    local FILE_A="/tmp/test_negative_instance.tmp_file"
```
evgenyz, line 15: Please use `$(mktemp)` for that.

src/OVAL/probes/independent/textfilecontent54_probe.c
```diff
@@ -235,27 +228,49 @@ static int process_file(const char *prefix, const char *path, const char *file,
...
 	} while (substr_cnt > 0 && ofs < buf_used);
 
+	probe_ent_getvals(pfd->instance_ent, &instance_value_list);
+	instance_value = SEXP_list_first(instance_value_list);
+	negative_instance_value = SEXP_number_geti_64(instance_value) < 0;
+	SEXP_free(instance_value_list);
+	SEXP_free(instance_value);
+
+	for(cur_inst = 0; cur_inst < instance_count; cur_inst++){
+		if(negative_instance_value)
```
evgenyz, line 250:
```suggestion
		if (negative_instance_value)
```

src/OVAL/probes/independent/textfilecontent54_probe.c
```diff
@@ -235,27 +228,49 @@ static int process_file(const char *prefix, const char *path, const char *file,
...
+			next_inst = SEXP_number_newi_32(cur_inst + 1);
+
+		if (probe_entobj_cmp(pfd->instance_ent, next_inst) == OVAL_RESULT_TRUE)
+			want_instance = 1;
+		else
+			want_instance = 0;
+
+		SEXP_free(next_inst);
+
+		if (want_instance){
```
evgenyz, line 263:
```suggestion
		if (want_instance) {
```

evgenyz (changes requested):
Some nitpicks, otherwise LGTM.

Xeicker:
Yes it was intentional to put this into 1.3, that's what is used in Oracle Linux 8 and 9, so would be good to have that fix in those distros

evgenyz (approved):

PR #2222, opened 2025-04-15
Exit oscap-im if oscap fails

jan-cerny:
This commit will cause that oscap-im will exit with exit code 1 if any of the called oscap calls will fail. This means that building hardened bootable container images will terminate if oscap crashes or doesn't work.

Resolves: https://issues.redhat.com/browse/OPENSCAP-5415

### DIFF
--- a/utils/oscap-im
+++ b/utils/oscap-im
@@ -111,9 +111,22 @@ def pre_scan_fix(args):
             "--output", remediation_script.name]
         add_common_args(args, gen_fix_cmd)
         gen_fix_cmd.append(args.data_stream)
-        subprocess.run(gen_fix_cmd, check=True)
-        subprocess.run(["bash", remediation_script.name], check=True)
-
+    try:
+        subprocess.run(gen_fix_cmd, check=True, capture_output=True)
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(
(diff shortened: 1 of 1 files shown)


utils/oscap-im
```diff
@@ -111,9 +111,22 @@ def pre_scan_fix(args):
...
-        subprocess.run(["bash", remediation_script.name], check=True)
-
+    try:
+        subprocess.run(gen_fix_cmd, check=True, capture_output=True)
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(
+            f"OpenSCAP generate fix failed with return code {e.returncode}.\n"
+            f"Output: {e.stderr.decode()}")
+    try:
+        subprocess.run(
```
matusmarhefka, line 121: The remediation script will install packages and sometimes it might be useful to see the whole dnf output. So here I would remove `capture_output=True` so we see both stdout and stderr combined together as originally outputted by dnf and in `RuntimeError` exception we can just print the failed with return code message (current first line).

jan-cerny:
I have removed delete=False

matusmarhefka (approved):

matusmarhefka:
I have tested the changes and it works as expected, thanks!

PR #2223, opened 2025-04-15
tests: OVAL/API/skip_paths

evgenyz:
Modify the test so it could catch the regression with environment variable modified during execution. See #2168.


evgenyz:
Fedora 42 becomes the latest and now there are more warnings in gcc.

Mab879 (approved):

PR #2224, opened 2025-04-15
tests: OVAL/API/skip_paths

evgenyz:
Modify the test so it could catch the regression
with environment variable modified during execution. See #2168.

### DIFF
--- a/tests/API/OVAL/skip_paths/test_skip_paths.sh
+++ b/tests/API/OVAL/skip_paths/test_skip_paths.sh
@@ -15,8 +15,8 @@ cp "$srcdir/test.xml" "$root/b/"
 mkdir -p "$root/c"
 touch "$root/c/z"
 cp "$srcdir/test.xml" "$root/c/"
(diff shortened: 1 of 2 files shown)


tests/API/OVAL/skip_paths/test_skip_paths.sh
```diff
@@ -16,7 +16,7 @@ mkdir -p "$root/c"
 touch "$root/c/z"
 cp "$srcdir/test.xml" "$root/c/"
 # oscap probes will skip directories "b" and "c"
```
Mab879, line 18:
```suggestion
# oscap probes will skip directories "n" and "c"
```

evgenyz: Fixed.

evgenyz:
Fedora 42 becomes the latest and now there are more warnings in gcc.

Mab879 (approved):

Mab879 (approved):

PR #2226, opened 2025-04-17
Handle errors returned by msg_handler

0intro:
This change handles errors returned by msg_handler. It prevents the probe_worker_runfn function from crashing, when msg_handler returns NULL and the probe_res pointer is dereferenced.

For example, the probe_worker function could return NULL when calling chroot without root privileges.


jan-cerny (approved):
The CI fail will be fixed by https://github.com/OpenSCAP/openscap/pull/2228 and isn't caused by the changes of this PR.

Issue #2221, opened 2025-04-15
The PATH variable is altered when using the --remediate option


vojtapolasek:
I observe this in openscap-1.3.11-1.el9.x86_64.

Steps to reproduce:
1. prepare a RHEL 9 environment, login as root and note contents of $PATH variable
2. clone this branch and build the rhel9 product: https://github.com/vojtapolasek/content/tree/add_remediation_accounts_root_path_dirs_no_write
3. Move the datastream to the rhel9 environment and make the system incompliant, for example chmod 777 /usr/bin
4. run oscap xccdf eval --remediate --report report.html --rule xccdf_org.ssgproject.content_rule_accounts_root_path_dirs_no_write ssg-rhel9-ds.xml
5. Examine the report. The remediation is modified so that it prints the $PATH variable and the output is in the report

Notice that the PATH variable is different than the one you noted at the begining.

jan-cerny:
yes, the remediation scripts don't inherit the parent environment
it's here:
https://github.com/OpenSCAP/openscap/blob/maint-1.3/src/XCCDF_POLICY/xccdf_policy_remediate.c#L467

PR #2227, opened 2025-04-17
Inherit OpenSCAP environment when executing Bash remediations with `--remediate` option

evgenyz:
Make the Bash remediation environment consistent with other types of remediation.

Fixes: #2221.


jan-cerny (approved):
I have add `set -x` to the test and run it locally.

PR #2228, opened 2025-04-17
Fix `unterminated-string-initialization` errors

evgenyz:
Add more space for `low_density_name` elements in `compat/dev_to_tty.c`

```
dev_to_tty.c:170:71: error: initializer-string for array of 'char'
truncates NUL terminator but destination lacks 'nonstring' attribute
(7 chars into 6 available)
[-Werror=unterminated-string-initialization].
```

Add more space for `xbase64.c` alphabets

```
xbase64.c:155:42: error: initializer-string for array of 'char'
truncates NUL terminator but destination lacks 'nonstring' attribute
(65 chars into 64 available)
[-Werror=unterminated-string-initialization]
```

### DIFF
--- a/compat/dev_to_tty.c
+++ b/compat/dev_to_tty.c
@@ -142,7 +142,7 @@ static int driver_name(char *restrict const buf, unsigned maj, unsigned min){
 }
 
 // major 204 is a mess -- "Low-density serial ports"
-static const char low_density_names[][6] = {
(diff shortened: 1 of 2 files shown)


jan-cerny (approved):
CI fail on Rawhide is in the RPM probes and isn't caused by this PR.

PR #2236, opened 2025-04-25
Fix codeql dependencies

Mab879:
The codeql job was failing due to dependencies not installing. Hopefully this should fix it.


evgenyz (approved):
Okay, let's try it.

PR #2248, opened 2025-07-19
fix: update bash_completion

darukutsu:
Was experiencing small error msg, change was while ago https://github.com/scop/bash-completion/commit/19a3798a0bf46ad9e39a055c4113eb5e9e9cb98b.
Also added few missing flags according to man page.

offtopic: I'm not sure, but it seems to me that [contributing](https://github.com/OpenSCAP/openscap/blob/main/docs/contribute/contribute.adoc) is a bit outdated. Like I wasn't sure because 1.2 is default but i can clearly see there is also 1.3 branch and most activity is in main, so pushed there.


evgenyz (approved):
LGTM. Thank you!

PR #2255, opened 2025-08-18
Handle virtual packages in dpkginfo probe

0intro:
This change handles virtual packages in dpkginfo probe by parsing the "Provides" field of /var/lib/dpkg/status.

Instead of relying on the "Package" field exclusively, this change also matches the requested package name with the package names listed in the "Provides" field.

It should allow to query virtual package names like "system-log-daemon" to actually installed packages likes "rsyslog" and so on.

See deb-control(5).

This PR is currently a work-in-progress.

### DIFF
--- a/src/OVAL/probes/unix/linux/dpkginfo-helper.c
+++ b/src/OVAL/probes/unix/linux/dpkginfo-helper.c
@@ -68,7 +68,7 @@ err:
 struct dpkginfo_reply_t* dpkginfo_get_by_name(const char *name, int *err)
 {
 	FILE *f;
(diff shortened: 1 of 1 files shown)


evgenyz:
Is the PR still WIP?

I see that we don't really have a test for `dpkginfo` probe except for `test_mitre_linux_probes.sh`, which is not enabled by default (ENABLE_MITRE). Do you think it'd make sense to do something about that?

evgenyz:
BTW if enabled `test_mitre_linux_probes.sh` fails on Fedora :(

a-skr:
I just checked, and the issue is solved.

Thank you.

PR #2259, opened 2025-09-23
Bump OVAL_SUPPORTED to 5.11.3

dodys:
5.11.2 was added back in commit [141f6d2](https://github.com/OpenSCAP/openscap/commit/141f6d2)

I wonder if we should remove 5.11.3 though, as I believe it was never an official supported OVAL version.

### DIFF
--- a/src/OVAL/oval_agent_api_impl.h
+++ b/src/OVAL/oval_agent_api_impl.h
@@ -35,7 +35,7 @@
 
 #define OVAL_ENUMERATION_INVALID (-1)
 
-#define OVAL_SUPPORTED "5.11.1"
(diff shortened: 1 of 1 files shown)


dodys:
Let me see if I understood this correctly, openscap has an OVAL that is not valid and that adds schemas that were never made into OVAL itself?

evgenyz:
> Let me see if I understood this correctly, openscap has an OVAL that is not valid and that adds schemas that were never made into OVAL itself?

Yep. Only the "invalid" content is part of ComplianceAsCode not OpenSCAP itself, but otherwise — correct.

evgenyz (approved):
LGTM, thanks!

PR #2262, opened 2025-10-06
Add new STIG URL

### LINKED ISSUE #2261: Can no longer import my Oscap results into STIG Viewer
I am having trouble importing XCCDF into my STIG Viewer, I am using the --stig-viewer option in my oscap command and the error is "Error parsing XCCDF". It will let me import an XCCDF .xml from a month ago before we updated our Rhel repo. I'm assuming something was updated causing this not to import anymore, system is RHEL 8.10 and tried using SSG Benchmark and DISA STIG Benchmark. Thank you for any help!

Mab879:
Fixes #2261

### DIFF
--- a/src/XCCDF/result.c
+++ b/src/XCCDF/result.c
@@ -91,7 +91,8 @@
 
 // References containing STIG Rule IDs can be found by their href attribute, it must match the following url
 static const char *DISA_STIG_VIEWER_HREF[] = { "http://iase.disa.mil/stigs/Pages/stig-viewing-guidance.aspx",
-					       "https://public.cyber.mil/stigs/srg-stig-tools/" };
+						"https://public.cyber.mil/stigs/srg-stig-tools/",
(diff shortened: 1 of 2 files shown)


evgenyz:
At this point with DISA mood for naming things swinging all over the place (4-th URL to match) I wonder if we should make it a config file or something that'd allow users to amend things without rebuilding OpenSCAP.

evgenyz (commented):
But okay, let's just have it for now.

BTW, what about https://github.com/OpenSCAP/openscap/blob/main/xsl/xccdf-references.xsl#L85?

jan-cerny:
> At this point with DISA mood for naming things swinging all over the place (3-rd URL to match) I wonder if we should make it a config file or something that'd allow users to amend things without rebuilding OpenSCAP.

Great idea! You can have an environment variable that allow users to override the hard coded URL if a need arises in future. Using the environment variable they will be able to work around their problems until we update the package.

PR #2268, opened 2025-10-14
Show rule details in output

jan-cerny:
Introduce new option `--show-rule-details` to the `oscap xccdf eval` module. This option will cause that rule description and texts will be printed to the stdout during the run.

Resolves: https://issues.redhat.com/browse/RHEL-104651

Example:

```
Title   Disallow Configuration to Bypass Password Requirements for Privilege Escalation
Rule    xccdf_org.ssgproject.content_rule_disallow_bypass_password_sudo
Description
        Verify the operating system is not configured to bypass password requirements for privilege
        escalation. Check the configuration of the "/etc/pam.d/sudo" file with the following command:
        $ sudo grep pam_succeed_if /etc/pam.d/sudo
        If any occurrences of "pam_succeed_if" is returned from the command, this is a finding.
Rationale
        Without re-authentication, users may access resources or perform tasks for which they do not
        have authorization. When operating systems provide the capability to escalate a functional
        capability, it is critical the user re-authenticate.
Ident   CCE-85967-8
Result  notapplicable
```

#### Review hints
```
./oscap_wrapper xccdf eval --profile stig --show-rule-details /usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml
```

### DIFF
--- a/tests/common/test_oscap_util.c
+++ b/tests/common/test_oscap_util.c
@@ -156,7 +156,7 @@ int test_oscap_remove_excess_whitespace()
 	char *str3 = "  \n  \n  \n";
 	result = oscap_remove_excess_whitespace(str3);
 	if (strcmp(result, "") != 0)
-		return 2;
+		return 3;
 	free(result);
 
 	char *str4 = "";


jan-cerny:
I have add test for functions oscap_indent and oscap_remove_excess_whitespace, I have renamed function xccdf_rule_get_readable_warnings, I have prevented empty description/rationale and I have fixed return code.

PR #2269, opened 2025-10-15
utils/oscap-im: Inherit environment for scanning and remediating

evgenyz:
Bash remediations require PATH to be inherited by the oscap process to work correctly.

@jan-cerny in `main` we still have `"OSCAP_BOOTC_BUILD": "YES"` in the environment: https://github.com/OpenSCAP/openscap/blob/8bc18a7ed6daca6c73cb49c86597c72b23e3a6e2/utils/oscap-im#L129

So I've put it back here. Should I instead remove it from both branches?

### DIFF
--- a/utils/oscap-im
+++ b/utils/oscap-im
@@ -17,6 +17,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
 import argparse
+import os
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


jan-cerny:
@evgenyz

SCAP content in ComplianceAsCode no longer depends on the OSCAP_BOOTC_BUILD environment variable. It has been removed from content in https://github.com/ComplianceAsCode/content/pull/12988.

In OpenSCAP, we have also removed this variable, by https://github.com/OpenSCAP/openscap/pull/2201. The PR is at the maint-1.3 branch therefore the patch will be propagated to main branch by merge before the next release.

PR #2283, opened 2025-12-03
Always check the current lists before inserting on blueprints

### LINKED ISSUE #2282: Blueprints do not correctly deduplicate
Seems we only checking kernel opts, seems like a bug

https://github.com/OpenSCAP/openscap/blob/a9c51af35578a2e2d4da5a4925763dbb9039a84e/src/XCCDF_POLICY/xccdf_policy_remediate.c#L745-L749

Mab879:
Fixes #2282

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -742,7 +742,7 @@ static inline int _parse_blueprint_fix(const char *fix_text, struct blueprint_cu
 			memcpy(val, &fix_text[ovector[2]], ovector[3] - ovector[2]);
 			val[ovector[3] - ovector[2]] = '\0';
 
-			if (!oscap_list_contains(customizations->kernel_append, val, (oscap_cmp_func) oscap_streq)) {
+			if (!oscap_list_contains(tab[i].list, val, (oscap_cmp_func) oscap_streq)) {
 				oscap_list_prepend(tab[i].list, val);
 			} else {
 				free(val);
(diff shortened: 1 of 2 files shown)


tests/API/XCCDF/unittests/test_duplicate_blueprint_service.toml
```diff
@@ -0,0 +1,52 @@
...
+# Blueprint for Profile title on one line
+#
+# Profile Description:
+# Profile description
+#
+# Profile ID:  xccdf_moc.elpmaxe.www_profile_common
+# Benchmark ID:  xccdf_moc.elpmaxe.www_benchmark_test
+# Benchmark Version:  1.0
+# XCCDF Version:  1.2
+#
```
evgenyz:
```
# $ oscap xccdf generate fix --profile xccdf_moc.elpmaxe.www_profile_common --fix-type blueprint xccdf-file.xml
```

evgenyz: The `1.4` and `1.3` are different in the file name: `TEST_XCCDF_FILE_NAME` vs `xccdf-file.xml`.

evgenyz:
Okay, I believe that you over-complicated things. In order to get the desired behavior you just have to duplicate `<Rule selected="true" id="xccdf_moc.elpmaxe.www_rule_9">` in the existing test (`test_remediation_blueprint.xccdf.xml` DS) and that's it.

With that backporting the change will be super simple.

evgenyz:
The `205 - probes/fwupdsecattr/test_probes_fwupdsecattr_mock.sh (Failed)` failure on Ubuntu is irrelevant to the PR changes.

evgenyz (approved):

PR #2328, opened 2026-04-01
Backport 2324

Mab879:


jan-cerny (approved):
It's the same as #2324 and it looks safe

PR #2329, opened 2026-04-02
Backport 2317 - 1.3

Mab879:
Backport #2317


jan-cerny (approved):
LGTM because it's the same as #2317 and tests utils/autotailor_integration_test.sh and autotailor-unit-tests passed.

PR #2331, opened 2026-04-03
openscap-1.4.4

Mab879:
We somehow need this commit to be last commit before merging.


jan-cerny:
Merged manually in 4ac56ce0e6dfdea7215b5ad202b3db20855d2507

PR #2335, opened 2026-04-09
Version bump after release. - 1.3.15

Mab879:
Next release from the maint-1.3 branch will be 1.3.15


jan-cerny (changes requested):
The version number needs to be changed also in CMakeLists.txt.

jan-cerny (approved):

PR #2336, opened 2026-04-09
Version bump after release. - 1.4.5

Mab879:
Next release from the main branch will be 1.4.5


jan-cerny (changes requested):
The version number needs to be changed also in CMakeLists.txt.

jan-cerny (approved):

PR #2337, opened 2026-04-10
autotailor: Add --local-path option for layered product compatibility

ggbecker:
Add a new --local-path flag that generates benchmark references using relative paths instead of absolute file:// URIs. This enables compatibility with Red Hat Satellite and other layered products that cannot resolve local filesystem paths.

When --local-path is specified:
- Absolute paths are converted to basename only
- Relative paths are preserved as provided by the user

The default behavior remains unchanged for backward compatibility.

Resolves: https://redhat.atlassian.net/browse/RHEL-143616

### DIFF
--- a/utils/autotailor
+++ b/utils/autotailor
@@ -303,6 +303,7 @@ class Tailoring:
         self.id = "xccdf_auto_tailoring_default"
         self.version = 1
         self.original_ds_filename = ""
+        self.use_local_path = False
 
         self.profiles = []
 
@@ -318,12 +319,31 @@ class Tailoring:
         self.profiles.append(profile)
         return profile
 
+    def _get_datastream_uri(self):
+        """
+        Determine the datastream URI based on the use_local_path setting.
+
+        Returns:
+            str: The URI/path to use in the benchmark href attribute
+        """
+        if self.use_local_path:
+            # Preserve relative paths as-is, convert absolute paths to basename
+            ds_path = pathlib.Path(self.original_ds_filename)
+            if ds_path.is_absolute():
+                # Convert absolute path to basename for compatibility with layered products
+                return ds_path.name
+            else:
+                # Keep relative paths as provided by the user
+                return self.original_ds_filename
+        else:
+            # Use absolute URI (default behavior for backward compatibility)
+            return pathlib.Path(self.original_ds_filename).absolute().as_uri()
+
     def to_xml(self, root):
         root.set("id", self.id)
 
         benchmark = ET.SubElement(root, "{%s}benchmark" % NS)
-        datastream_uri = pathlib.Path(
-            self.original_ds_filename).absolute().as_uri()
+        datastream_uri = self._get_datastream_uri()
         benchmark.set("href", datastream_uri)
 
         version = ET.SubElement(root, "{%s}version" % NS)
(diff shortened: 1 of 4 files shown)


ggbecker:
To which openscap versions do we want to include this change?

utils/autotailor
```diff
@@ -436,22 +445,25 @@ def get_parser():
         "-o", "--output", default="-",
         help="Where to save the tailoring file. If not supplied, write to "
         "standard output.")
+    parser.add_argument(
+        "--relative-path", action="store_true",
```
jan-cerny: Please document the new option in the man page.

utils/autotailor
```diff
@@ -322,8 +320,19 @@ class Tailoring:
         root.set("id", self.id)
 
         benchmark = ET.SubElement(root, "{%s}benchmark" % NS)
-        datastream_uri = pathlib.Path(
-            self.original_ds_filename).absolute().as_uri()
+        if self.use_relative_path:
```
jan-cerny: The `to_xml` method starts getting long. I suggest extracting the code that determines the URI to a separate method eg. `_get_datastream_uri()`

utils/autotailor
```diff
@@ -436,22 +445,25 @@ def get_parser():
         "-o", "--output", default="-",
         help="Where to save the tailoring file. If not supplied, write to "
         "standard output.")
+    parser.add_argument(
+        "--relative-path", action="store_true",
+        help="Use relative path (basename only) for the benchmark href instead of "
```
jan-cerny: Please add a test that covers this feature.

jan-cerny (changes requested):

jan-cerny:
> To which openscap versions do we want to include this change?

Good question. The answer isn't clear. The RHEL ticket is reported against RHEL 10. That'd mean we shall merge it only to the main branch and it will be a part of OpenSCAP version 1.4.5. However, in the RHEL ticket description, they say they use it for RHEL 9 content, which would suggest that they will need the fix also in RHEL 9, which contains OpenSCAP 1.3.xx and therefore we would need it merge it also to the maint-1.3 branch so that it would become part of OpenSCAP 1.3.15.

In general, we try to do the backporting only for bugfixes, in this case it's difficult to say whether this is a bugfix or a new feature, but I feel this would be more of a bugfix.

If you agree, then please after this PR is merged create another PR with the same contents but to the maint-1.3 branch.

utils/autotailor
```diff
@@ -322,8 +320,19 @@ class Tailoring:
         root.set("id", self.id)
 
         benchmark = ET.SubElement(root, "{%s}benchmark" % NS)
-        datastream_uri = pathlib.Path(
-            self.original_ds_filename).absolute().as_uri()
+        if self.use_relative_path:
+            # Preserve relative paths as-is, convert absolute paths to basename
+            ds_path = pathlib.Path(self.original_ds_filename)
+            if ds_path.is_absolute():
+                # Convert absolute path to basename for compatibility with layered products
```
jan-cerny, line 333: The option name is `-relative-path` but if this branch is executed there will be a basename which doesn't have to be a relative path. Consider some better name for the command line option.

jan-cerny:
@ggbecker This CI error looks legit:

```
Failed: expected count: 1, real count: 0, xpath: '/Tailoring/benchmark[@href="data_stream.xml"]'
```

jan-cerny (approved):
I have used the autotailor both with and without the --local-path option and verified that the output contains the expected path/URI in the "benchmark" element. I have run the unit and integration tests locally.

jan-cerny:
@ggbecker Now please create a PR with the same content but targeting the maint-1.3 branch.

PR #2342, opened 2026-04-22
autotailor: Add validation for variable names, rule IDs, and profile IDs 

ggbecker:
Fixes: https://redhat.atlassian.net/browse/RHEL-143568

### DIFF
--- a/utils/autotailor
+++ b/utils/autotailor
@@ -26,10 +26,12 @@ import pathlib
 import xml.etree.ElementTree as ET
 import xml.dom.minidom
 import json
+import difflib
 
 
 NS = "http://checklists.nist.gov/xccdf/1.2"
 NS_PREFIX = "xccdf-1.2"
+DS_NS = "http://scap.nist.gov/schema/scap/source/1.2"
 DEFAULT_PROFILE_SUFFIX = "_customized"
 DEFAULT_REVERSE_DNS = "org.ssgproject.content"
 ROLES = ["full", "unscored", "unchecked"]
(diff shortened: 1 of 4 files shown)


utils/autotailor
```diff
@@ -53,12 +55,115 @@ def is_valid_xccdf_id(string):
...
+        self._parse_datastream()
+
+    def _parse_datastream(self):
+        """Parse the datastream to extract all valid IDs."""
+        try:
+            tree = ET.parse(self.datastream_path)
+            root = tree.getroot()
+
+            # Register namespaces
+            namespaces = {
```
jan-cerny: This can be a global variable. It could be used all over the script.

utils/autotailor
```diff
@@ -53,12 +55,115 @@ def is_valid_xccdf_id(string):
...
+
+    def __init__(self, datastream_path):
+        self.datastream_path = datastream_path
+        self.profile_ids = set()
+        self.value_ids = set()
+        self.rule_ids = set()
+        self.group_ids = set()
+        self._parse_datastream()
+
+    def _parse_datastream(self):
```
jan-cerny, line 106: Split this method to more smaller methods.

utils/autotailor
```diff
@@ -460,6 +608,11 @@ def get_parser():
         "--local-path", action="store_true",
         help="Use local path for the benchmark href instead of absolute file:// URI. "
         "Absolute paths are converted to basename, relative paths are preserved.")
+    parser.add_argument(
+        "--no-validate", action="store_true",
+        help="Skip validation of IDs against the datastream. This significantly speeds up "
+        "execution on large datastreams but may produce invalid tailoring files if incorrect "
```
jan-cerny:
```suggestion
        help="Skip validation of IDs against the data stream. This significantly speeds up "
        "execution on large data streams but may produce invalid tailoring files if incorrect "
```

utils/autotailor
```diff
@@ -53,12 +55,115 @@ def is_valid_xccdf_id(string):
...
+
+        if not valid_ids:
+            msg += f"\n\nNo {id_type.lower()}s found in the datastream."
+        else:
+            msg += f"\n\nAvailable {id_type.lower()}s can be listed by examining the datastream file."
+
+        return msg
+
+    def validate_profile(self, profile_id):
+        """Validate a profile ID exists in the datastream."""
```
jan-cerny: It's always a "data stream", never a "datastream" or even something worse like "DataStream".

utils/autotailor
```diff
@@ -473,27 +626,55 @@ if __name__ == "__main__":
...
+
+    t = Tailoring(validator=validator)
     t.original_ds_filename = args.datastream
     t.reverse_dns = args.id_namespace
     t.use_local_path = args.local_path
 
     if args.json_tailoring:
-        t.import_json_tailoring(args.json_tailoring)
+        try:
+            t.import_json_tailoring(args.json_tailoring)
```
jan-cerny, line 663: Is the JSON tailoring also validated?

utils/autotailor
```diff
@@ -460,6 +608,11 @@ def get_parser():
         "--local-path", action="store_true",
         help="Use local path for the benchmark href instead of absolute file:// URI. "
         "Absolute paths are converted to basename, relative paths are preserved.")
+    parser.add_argument(
+        "--no-validate", action="store_true",
```
jan-cerny, line 631: Add this option to the man page.

utils/autotailor
```diff
@@ -53,12 +55,115 @@ def is_valid_xccdf_id(string):
...
+            msg += f"\n\nAvailable {id_type.lower()}s can be listed by examining the datastream file."
+
+        return msg
+
+    def validate_profile(self, profile_id):
+        """Validate a profile ID exists in the datastream."""
+        if profile_id not in self.profile_ids:
+            raise ValueError(self._create_validation_error("Profile", profile_id, self.profile_ids))
+
+    def validate_value(self, value_id):
```
jan-cerny, line 149: It would be amazing if the tool would also verify validity of selectors used in the `-V` CLI option.

For example:

```
utils/autotailor -V xccdf_org.ssgproject.content_value_var_selinux_state=OOO /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml cis
```

This command should error because the XCCDF Value doesn't contain selector `OOO`. But currently it isn't checked and the command doesn't error and creates invalid tailoring.

jan-cerny (changes requested):

ggbecker:
@jan-cerny  I've addressed the feedback on [ef1d59d](https://github.com/OpenSCAP/openscap/pull/2342/commits/ef1d59dd8b801860ce0afd92c3a37cf5fba05b3a)

jan-cerny (approved):
I have tried various combinations of parameters (selections, unselections, variables, selectors) and verified that they are validated against the data stream. I also used the `--no-validate` option. Tests passed both locally and in CI.

PR #2346, opened 2026-05-04
process58: bound sscanf perm field in get_exec_shield_status

Mab879:
Use %7s and larger buffer; read maps lines with fgets(buf, sizeof(buf)).

Fixes https://github.com/OpenSCAP/openscap/security/code-scanning/1589


jan-cerny (approved):
It's a good fix.

The fail of build on Fedora isn't caused by contents of this PR and  will be fixed by https://github.com/OpenSCAP/openscap/pull/2347

PR #2347, opened 2026-05-04
selinuxsecuritycontext: const-correct strchr in split_level

Mab879:
Fixes build failure on main.


jan-cerny (approved):
Good fix, it fixed the failing build on Fedora

PR #2326, opened 2026-03-20
macos/freebsd portability

edlitmus:
## Summary

This PR improves OpenSCAP portability for macOS and FreeBSD, expands targeted regression coverage, and tightens the CTest/test-documentation workflow while preserving Linux behavior.

### Build and configure

- Reset stale cached `PERL_INCLUDE_PATH` and `PERL_LIBRARY` entries during CMake configure when cached paths no longer exist. This makes reused build trees more resilient after Perl upgrades or dependency path changes.
- Detect `fgetpwent(3)` at configure time and use that signal for portable password-probe behavior.
- Gate Linux-specific compiler/linker behavior so non-Linux builds do not inherit unsupported settings.
- Accept `asciidoctor` as a docs backend in addition to `asciidoc`.

### Portability and probe behavior

- Add macOS implementations for memory usage reporting in `memusage`.
- Extend XCCDF system info collection on macOS to populate target MAC data from local interfaces.
- Add a portable password-probe fallback parser for platforms that do not provide `fgetpwent(3)`.
- Mark the runlevel probe as unsupported on macOS and FreeBSD instead of attempting Linux/SysV-specific behavior there.
- Mark shadow offline mode as unsupported on platforms where the Linux-style shadow path does not apply.
- Rework the BSD/macOS `sysctl` probe path to parse `sysctl -ae` output safely, including multiline values, by treating only valid `name=value` headers as new items.
- Fix FreeBSD build issues by declaring `environ` where needed and guarding `IFF_NOGROUP` usage when that flag is not defined.

### Test infrastructure and regression coverage

- Add `add_oscap_ctest()` and complete a broader CTest label rollout so shell and direct CTest/pytest tests can be filtered consistently by suite and platform.
- Register more probe suites and Python-based tests with meaningful labels such as `api`, `probes`, `unix`, `linux_only`, `macos`, and `freebsd`.
- Add targeted regression coverage for the portability work, including:
  - macOS/XCCDF system info platform facts
  - macOS memory reporting
  - password offline fallback behavior
  - unsupported runlevel behavior on macOS/FreeBSD
  - unsupported shadow offline behavior on non-Linux platforms
  - BSD/macOS sysctl collection behavior
- Update the `sysctl` parity test so BSD/macOS comparisons are based on value-bearing `sysctl -ae` entries and stable sorted output.

### Documentation

- Document cross-platform build expectations and label-based test execution in the developer manual.
- Document the stale-Perl-cache reconfigure behavior.
[...]

PR #2348, opened 2026-05-04
Set install RPATH on macOS so binaries find libopenscap

edlitmus:
## Summary

After `make install` on macOS, `oscap` referenced `@rpath/libopenscap.<ver>.dylib` but the binary had no `LC_RPATH` entries, so `dyld` failed at startup:

```
dyld[…]: Library not loaded: @rpath/libopenscap.33.dylib
  Referenced from: /usr/local/bin/oscap
  Reason: no LC_RPATH's found
```

The library was installed correctly to `/usr/local/lib/libopenscap.33.dylib`; the binary just had no way to locate it without `DYLD_LIBRARY_PATH`. CMake on macOS strips the build-tree RPATH at install time and provides nothing in its place unless `CMAKE_INSTALL_RPATH` is set.

This adds, inside the existing `if(APPLE)` block in the root `CMakeLists.txt`:

- `CMAKE_MACOSX_RPATH ON`
- `CMAKE_INSTALL_NAME_DIR "@rpath"`
- `CMAKE_INSTALL_RPATH "@loader_path/../lib;${CMAKE_INSTALL_PREFIX}/lib"`

`@loader_path/../lib` keeps the install relocatable; the absolute prefix path is a fallback. Linux and Windows are unaffected — the block is `APPLE`-only.

This is a follow-on to the macOS/FreeBSD portability work in #2326.

## Test plan

- [x] Configure + build cleanly on macOS (Apple Silicon, Homebrew toolchain)
- [x] Staged install (`DESTDIR=…  make install`): `otool -l .../bin/oscap` shows both `LC_RPATH` entries (`@loader_path/../lib` and `/usr/local/lib`)
- [x] Real install to `/usr/local`: `oscap --version` and `oscap xccdf generate report …` run without `DYLD_LIBRARY_PATH`
- [x] Linux unaffected (block is gated on `APPLE`; no CMake variables Linux uses are touched)

No new automated test — existing CI (`.github/workflows/`) runs on Ubuntu and Fedora only, so an RPATH check wouldn't execute. The fix is verifiable manually with `otool -l` on macOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -564,6 +564,11 @@ endif()
 if(APPLE)
 	#full Single Unix Standard v3 (SUSv3) conformance (the Unix API)
 	add_definitions(-D_DARWIN_C_SOURCE)
+
+	# So that installed binaries can find libopenscap.dylib without DYLD_LIBRARY_PATH.
+	set(CMAKE_MACOSX_RPATH ON)
+	set(CMAKE_INSTALL_NAME_DIR "@rpath")
+	set(CMAKE_INSTALL_RPATH "@loader_path/../lib;${CMAKE_INSTALL_PREFIX}/lib")
 endif()
 
 include_directories(


Mab879 (approved):
Thanks!

The failure on rawhide will be fixed #2347.

PR #2349, opened 2026-05-04
xinetd probe: bound paths and strans keys; export oscap_path_join

Mab879:
Use snprintf and length checks for stack buffers; build includedir paths with oscap_path_join. Mark oscap_path_join OSCAP_API for embedded tests. Add regression test for oversized name+protocol key.

Fixes various code issues in this file.

### DIFF
--- a/src/common/util.h
+++ b/src/common/util.h
@@ -396,17 +396,6 @@ char *oscap_vsprintf(const char *fmt, va_list ap);
  */
 char *oscap_generate_random_string(size_t len, char *charset);
 
-/**
- * Join 2 paths in an intelligent way.
- * Both paths are allowed to be NULL.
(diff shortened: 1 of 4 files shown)


src/common/util.h
```diff
@@ -405,7 +405,7 @@ char *oscap_generate_random_string(size_t len, char *charset);
  * @return Join of path1 and path2. The first path is separated by the second
  * path by exactly 1 slash separator.
  */
-char *oscap_path_join(const char *path1, const char *path2);
+OSCAP_API char *oscap_path_join(const char *path1, const char *path2);
```
jan-cerny: Function prototypes marked with `OSCAP_API` need to be put in a public header file. Please move it for example to `src/common/public/oscap.h`.

PR #2352, opened 2026-05-08
fix(oval): use %d in CIM datetime sscanf for int fields

Mab879:
Match sscanf specifiers to int storage; zero-initialize parsed fields.

Fixes
* https://github.com/OpenSCAP/openscap/security/code-scanning/1484
* https://github.com/OpenSCAP/openscap/security/code-scanning/1483
* https://github.com/OpenSCAP/openscap/security/code-scanning/1482
* https://github.com/OpenSCAP/openscap/security/code-scanning/1481
* https://github.com/OpenSCAP/openscap/security/code-scanning/1479


jan-cerny (approved):
This helps reduce undefined behavior and makes the code more defensive.

PR #2353, opened 2026-05-11
Generate raw partitioning mode in Image Builder blueprints

### LINKED ISSUE #2334: Update OpenSCAP to generate Image Builder blueprints with `partitioning_mode = "raw"`
The `partitioning_mode = "raw"` (https://osbuild.org/docs/user-guide/blueprint-reference/#partitioning-mode) makes sure that `/` filesystem is grown to fill any left over space on the partition table which is not the default behavior with `auto-lvm` and when filesystem customizations are used in blueprint (in which case `lvm` is used). With `partitioning_mode = lvm` the `/` partition is set to be of minimal hard-coded size, based on https://osbuild.org/docs/user-guide/partitioning/ these are:
```
These are 1 GiB for / and 2 GiB for /usr. This means that, if /usr is not on a separate partition, the root filesystem will always be at least 3 GiB.
```

This size is not enough for some profiles like STIG where the 3 GiB size for `/` was causing `No space left on device` errors, see https://github.com/ComplianceAsCode/content/issues/14551

Therefore, we should update OpenSCAP to generate the blueprint with:
```
[customizations]
partitioning_mode = "raw"
```

This was already done in our tests by https://github.com/RHSecurityCompliance/contest/pull/586, once this issue is resolved the customization in contest should be removed.

officialasishkumar:
## Summary
- emit `[customizations] partitioning_mode = "raw"` in generated Image Builder blueprints
- update the blueprint remediation golden fixture

Fixes #2334.

## Testing
- `cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_PROBES=OFF -DENABLE_PYTHON3=OFF -DENABLE_PERL=OFF -DENABLE_DOCS=OFF -DENABLE_OSCAP_UTIL_DOCKER=OFF -DENABLE_OSCAP_UTIL_AS_RPM=OFF ..`
- `make -j2 oscap`
- `ctest -R 'API/XCCDF/unittests/test_remediation_blueprint.sh' --output-on-failure`\n- `git diff --check`

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy_remediate.c
+++ b/src/XCCDF_POLICY/xccdf_policy_remediate.c
@@ -1405,6 +1405,8 @@ static int _write_script_header_to_fd(struct xccdf_policy *policy, struct xccdf_
 			"name = \"hardened_%s\"\n"
 			"description = \"%s\"\n"
 			"version = \"%s\"\n\n"
+			"[customizations]\n"
+			"partitioning_mode = \"raw\"\n\n"
 			"[customizations.openscap]\n"
 			"profile_id = \"%s\"\n"
 			"%s\n",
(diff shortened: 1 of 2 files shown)


Mab879 (approved):
Thanks!

PR #2357, opened 2026-05-14
Sanitize strings to prevent UTF errors

jan-cerny:
If a string is collected by a probe, we will sanitize this string before converting it to SEXP. This will ensure that no errors related to UTF enconding can happen during the whole processing. This way we will fix multiple errors emitted by pcre2 or libxml2 libraries that are caused by providing non-UTF strings.

Resolves: https://redhat.atlassian.net/browse/RHEL-171005

Review hints:
1. `touch /mnt/$(printf "evil_filename_\334_non_utf8_character")`
2. `oscap xccdf eval --remediate --profile xccdf_org.ssgproject.content_profile_cis --rule xccdf_org.ssgproject.content_rule_audit_rules_privileged_commands /usr/share/xml/scap/ssg/content/ssg-rhel10-ds.xml`
3. `oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_cis --rule xccdf_org.ssgproject.content_rule_audit_rules_privileged_commands /usr/share/xml/scap/ssg/content/ssg-rhel10-ds.xml` - this returns `error` with latest released openscap
4. install the build by Packit from COPR from this PR
5. repeat step 3 - it will return `pass`


src/common/oscap_utf8.c
```diff
@@ -0,0 +1,130 @@
+/*
+ * Copyright 2026 Red Hat Inc., Durham, North Carolina.
```
Mab879:
```suggestion
 * Copyright 2026 Red Hat LLC, Raleigh, North Carolina.
```

src/common/oscap_utf8.h
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2026 Red Hat Inc., Durham, North Carolina.
```
Mab879:
```suggestion
 * Copyright 2026 Red Hat LLC, Raleigh, North Carolina.
```

Mab879 (commented):
* If you do make changes please correct the spelling of encoding in the commit message.

Please take look at the Sonar findings might be worth fixing.

src/common/oscap_utf8.c
```diff
@@ -0,0 +1,130 @@
...
+
+	if (i > 0)
+		memcpy(out, in, i);
+	size_t o = i;
+
+	while (i < input_len) {
+		int clen = _utf8_char_len(in + i, input_len - i);
+		if (clen < 0) {
+			if (o + 3 > alloc) {
+				alloc = alloc * 2;
```
Mab879, line 102: Add guard for overflow?

src/common/oscap_utf8.c
```diff
@@ -0,0 +1,130 @@
...
+		int clen = _utf8_char_len(in + i, input_len - i);
+		if (clen < 0)
+			break;
+		i += clen;
+	}
+
+	if (i == input_len)
+		return NULL;
+
+	size_t alloc = input_len + 64;
```
Mab879, line 85: We might want to add a guard to make sure we don't overflow alloc. 

We need add one byte here for the null byte we add below on this check.

Mab879:
I was able to pass the tests above using Fedora 44, but I had use a nightly content build. I believe that is expected. If that is not. something to look into.

jan-cerny:
I wanted to reply that it isn't expected, but after I dove deeper I found that it's more complicated and it matters if you test it on Fedora or RHEL 10.

On RHEL 10, the review hints steps should work fine with the released version of content. It works for me in a fresh RHEL 10 VM using content from scap-security-guide-0.1.80-1.el10_1.noarch.

You actually need to do the review hints steps with a version of content before https://github.com/ComplianceAsCode/content/pull/14706 got merged because that PR modified the rule audit_rules_privileged_commands in a way the situation couldn't be reproduced using the rule. So ideally you should run the review hints steps with content from the 0.1.80 or older release.

However, a different situation is on Fedora. I get a "error" or "fail" on Fedora 44 with scap-security-guide-0.1.80-2.fc44.noarch. However, the reason is that the OVAL doesn't pass after the remediation because of the detection of whether we use audit with augenrules or auditctl. The content checks for `^ExecStart=(\/usr|)?\/sbin\/augenrules.*$` but  the file `/usr/lib/systemd/system/audit-rules.service` contains `ExecStart=/usr/bin/augenrules --load` on Fedora. This has been fixed in https://github.com/ComplianceAsCode/content/pull/14620 in upstream. So actually to be able to follow the review hints steps on Fedora you would need to have a version of content after #14620 was merged but before #14706 was merged.

jan-cerny:
So to simplify this: for testing you need to use an OVAL that does a regular expression operation on the offending not-UTF string. The simplest OVAL that does that is actually part of this PR as the unit test content in `tests/API/OVAL/unittests/test_pcre_nonutf_characters.xml`.

Mab879:
> So to simplify this: for testing you need to use an OVAL that does a regular expression operation on the offending not-UTF string. The simplest OVAL that does that is actually part of this PR as the unit test content in `tests/API/OVAL/unittests/test_pcre_nonutf_characters.xml`.

The tests pass, so I'm fine with my testing.

Thoughts my other comments?

src/common/oscap_utf8.c
jan-cerny: I believe that a byte for the null byte is added in the malloc and realloc calls below by the `+ 1` in the parameters:

```
uint8_t *out = malloc(alloc + 1);
...
uint8_t *tmp = realloc(out, alloc + 1);
```

Mab879 (approved):

PR #2361, opened 2026-06-03
XCCDF/OVAL/CPE/DS: fix NULL-pointer dereferences on malformed input

edznux-dd:
Hey :)

I was looking at some of our logs and noticed that `openscap` was crashing on memory corruption issues, so I decided to dig a bit into it and write some fuzz test (I didn't have the exact source of the crashes).

This is the 1st of 4 fix PR (+ one for the fuzz test), that I tried to split into manageable diff.

(There will be one PR that contains the fuzz-harnesses that I used. They are not comprehensive for all the code path

This current PR mostly ensure that all references are guarded by null checks.

Notes:
- I've read your contribution guideline, and I _believe_ I'm targeting the correct branch, but I might be mistaken?
- I've not created associated individual issues as I believe it creates unnecessary toil when I had to fix the bugs to continue fuzzing anyway (a lot of these bugs where found iteratively, run fuzz -> bug fix -> run fuzz -> bug fix ...)
- I've let all the fuzz test run for 24h once all bug were fixed (not on a huge amount of cores, sure, but still, seems fairly robust now, under the current harnesses.) 

For some context on this specific PR:
- the missing XML attributes/namespaces/text make libxml return NULL, which was then passed to strcmp/atoi/atof or dereferenced.
- common/util.h: oscap_str_startswith()/oscap_str_endswith() return false on a NULL string (matching oscap_strcmp/oscap_streq).
- OVAL: guard the optional `version` attribute before atoi() in oval_state, oval_definition, oval_object, oval_variable, oval_resultDefinition; guard the `reported` attribute (oval_directives); guard a NULL frame from a duplicate variable id (oval_varModel).
- XCCDF: NULL-safe atof() for empty lower/upper-bound and score elements (value.c, result.c); reject a NULL <TestResult> (benchmark.c); skip a <platform> with no @idref before strlen() (item.c); reject a NULL profile in tailoring (tailoring.c).
- DS: NULL-safe id comparisons in sds_index / rds; reject a component with no id as a hash key (ds_rds_session.c).

Other associated PRs:
- https://github.com/OpenSCAP/openscap/pull/2362
- https://github.com/OpenSCAP/openscap/pull/2363
- https://github.com/OpenSCAP/openscap/pull/2364
- https://github.com/OpenSCAP/openscap/pull/2365


src/DS/sds_index.c
```diff
@@ -459,13 +462,16 @@ int ds_sds_index_select_checklist_by_benchmark_id(struct ds_sds_index* s,
 
 	int ret = 1;
 
+	if (s == NULL)
```
Mab879: Shouldn't this be placed before line 457?

src/XCCDF/value.c
```diff
@@ -144,13 +144,15 @@ struct xccdf_item *xccdf_value_parse(xmlTextReaderPtr reader, struct xccdf_item
 		case XCCDFE_LOWER_BOUND:
 			if (type == XCCDF_TYPE_NUMBER) {
 				val = _xccdf_value_find_or_create_instance(XVALUE(value), selector, type);
-				val->lower_bound = atof(oscap_element_string_get(reader));
+				const char *lb = oscap_element_string_get(reader);
+				val->lower_bound = lb ? atof(lb) : 0.0;
```
Mab879: Defaulting 0.0 might not be right thing here. Something like below might be better.

```c
        if (lb == NULL) {
            dW("Empty <lower-bound> element is invalid, rejecting <Value>.");
            xccdf_value_free(value);
            return NULL;
        }
        val = _xccdf_value_find_or_create_instance(XVALUE(value), selector, type);
        val->lower_bound = atof(lb);
```

src/XCCDF/value.c
```diff
@@ -144,13 +144,15 @@ struct xccdf_item *xccdf_value_parse(xmlTextReaderPtr reader, struct xccdf_item
...
 			}
 			break;
 		case XCCDFE_UPPER_BOUND:
 			if (type == XCCDF_TYPE_NUMBER) {
 				val = _xccdf_value_find_or_create_instance(XVALUE(value), selector, type);
-				val->upper_bound = atof(oscap_element_string_get(reader));
+				const char *ub = oscap_element_string_get(reader);
+				val->upper_bound = ub ? atof(ub) : 0.0;
```
Mab879: Same as lower bound. Something like 

```c
   const char *ub = oscap_element_string_get(reader);
        if (ub == NULL) {
            dW("Empty <upper-bound> element is invalid, rejecting <Value>.");
            xccdf_value_free(value);
            return NULL;
        }
        val = _xccdf_value_find_or_create_instance(XVALUE(value), selector, type);
        val->upper_bound = atof(ub);
```

src/XCCDF/result.c
```diff
@@ -1453,7 +1453,8 @@ static struct xccdf_score *xccdf_score_new_parse(xmlTextReaderPtr reader)
 	if (xccdf_attribute_has(reader, XCCDFA_MAXIMUM))
 		score->maximum = xccdf_attribute_get_float(reader, XCCDFA_MAXIMUM);
 	else score->maximum = XCCDF_SCORE_MAX_DAFAULT;
-	score->score = atof(oscap_element_string_get(reader));
+	const char *score_str = oscap_element_string_get(reader);
+	score->score = score_str ? atof(score_str) : 0.0;
```
Mab879: Something like might be better here.

```c
if (score_str == NULL) {
    dW("Empty <score> element is invalid, rejecting.");
    xccdf_score_free(score);
    return NULL;
}
score->score = atof(score_str);
```

Mab879 (changes requested):
First of all, thank you for this series of PRs.

Please consider adding regression tests.

src/DS/sds_index.c
edznux-dd: oops, yep, thanks 😅  (fixed in 906022e8472c17b3d23fc3eddbc103741773f26c)

src/XCCDF/value.c
edznux-dd: Yea, I wasn't sure if it was preferable to default or to fail. sounds good to me :)
Applied your suggestion in  906022e8472c17b3d23fc3eddbc103741773f26c as well

edznux-dd:
> Please consider adding regression tests.

Ah, and forgot to reply to that part:
I've put reproducer in the PR https://github.com/OpenSCAP/openscap/pull/2365 that could/should work as regression tests (if hooked in CI), but sure, I can also put them in your normal testing suite!

This is the following running locally:

```
Start  13: API/OVAL/unittests/test_null_ptr_regression.sh
1/3 Test  #13: API/OVAL/unittests/test_null_ptr_regression.sh ....   Passed    0.19 sec
    Start 168: API/XCCDF/unittests/test_null_ptr_regression.sh
2/3 Test #168: API/XCCDF/unittests/test_null_ptr_regression.sh ...   Passed    0.20 sec
    Start 176: DS/test_null_ptr_regression.sh
3/3 Test #176: DS/test_null_ptr_regression.sh ....................   Passed    0.21 sec

100% tests passed, 0 tests failed out of 3

Label Time Summary:
api      =   0.40 sec*proc (2 tests)
ds       =   0.21 sec*proc (1 test)
shell    =   0.61 sec*proc (3 tests)

Total Test time (real) =   0.62 sec
```

I believe I've followed your `cmake` + `sh` + `test file` pattern!

Mab879 (approved):

PR #2364, opened 2026-06-03
XCCDF_POLICY: bound recursion on cyclic Profile @extends

edznux-dd:
(see https://github.com/OpenSCAP/openscap/pull/2361 for more context)

This is the 4th and last fix-PR of the series (the next one will be the fuzz harnesses)
This one feels more like hardening but it fixes a stack overflow issue on cyclic/deeply nested `@extends` chain.

I've set `XCCDF_POLICY_MAX_EXTENDS_DEPTH` to 64, as an arbitrary value that seems conservative but can be changed if you have stronger opinion. (similar to https://github.com/OpenSCAP/openscap/pull/2363)

### DIFF
--- a/src/XCCDF_POLICY/xccdf_policy.c
+++ b/src/XCCDF_POLICY/xccdf_policy.c
@@ -1880,9 +1880,10 @@ static void _xccdf_policy_add_profile_selectors(struct xccdf_policy* policy, str
 	if (!profile)
 		return;
 
-	if (_xccdf_extends_recursion_depth > XCCDF_POLICY_MAX_EXTENDS_DEPTH) {
+	if (_xccdf_extends_recursion_depth >= XCCDF_POLICY_MAX_EXTENDS_DEPTH) {
+		const char *profile_id = xccdf_profile_get_id(profile) ? xccdf_profile_get_id(profile) : "(unknown)";
 		oscap_seterr(OSCAP_EFAMILY_XCCDF, "Profile @extends chain is too deep or cyclic; aborting selector resolution for '%s'.",
-				xccdf_profile_get_id(profile));
+				profile_id);
 		return;
 	}
 	_xccdf_extends_recursion_depth++;
--- a/src/XCCDF_POLICY/xccdf_policy.c
+++ b/src/XCCDF_POLICY/xccdf_policy.c
@@ -1864,15 +1864,34 @@ xccdf_policy_add_select(struct xccdf_policy *policy, struct xccdf_select *sel)
 	return _xccdf_policy_add_selector_internal(policy, benchmark, sel, true);
 }
 
+// Upper bound on the Profile @extends chain length. Real content nests at most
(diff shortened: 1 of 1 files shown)


src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -1864,15 +1864,34 @@ xccdf_policy_add_select(struct xccdf_policy *policy, struct xccdf_select *sel)
...
+// -> xccdf_policy_new -> _add_profile_selectors -- which would reset a per-call
+// argument and let a cyclic @extends overflow the stack.
+static unsigned _xccdf_extends_recursion_depth = 0;
+
 static void _xccdf_policy_add_profile_selectors(struct xccdf_policy* policy, struct xccdf_benchmark *benchmark, struct xccdf_profile *profile) {
 	if (!profile)
 		return;
 
+	if (_xccdf_extends_recursion_depth > XCCDF_POLICY_MAX_EXTENDS_DEPTH) {
+		oscap_seterr(OSCAP_EFAMILY_XCCDF, "Profile @extends chain is too deep or cyclic; aborting selector resolution for '%s'.",
```
Mab879, line 1885: It might be better to assign `xccdf_profile_get_id(profile) ? xccdf_profile_get_id(profile) : "(unknown)"` to a variable and use it so that we don't reference a null here.

src/XCCDF_POLICY/xccdf_policy.c
```diff
@@ -1864,15 +1864,34 @@ xccdf_policy_add_select(struct xccdf_policy *policy, struct xccdf_select *sel)
...
+// policy creation -- _add_profile_selectors -> xccdf_policy_model_get_policy_by_id
+// -> xccdf_policy_new -> _add_profile_selectors -- which would reset a per-call
+// argument and let a cyclic @extends overflow the stack.
+static unsigned _xccdf_extends_recursion_depth = 0;
+
 static void _xccdf_policy_add_profile_selectors(struct xccdf_policy* policy, struct xccdf_benchmark *benchmark, struct xccdf_profile *profile) {
 	if (!profile)
 		return;
 
+	if (_xccdf_extends_recursion_depth > XCCDF_POLICY_MAX_EXTENDS_DEPTH) {
```
Mab879:
```suggestion
	if (_xccdf_extends_recursion_depth >= XCCDF_POLICY_MAX_EXTENDS_DEPTH) {
```

If we are be pedantic about the limit.

Mab879 (commented):
Thanks again for the PRs. 

Just a couple of small notes.

edznux-dd:
> Thanks again for the PRs.
> 
> Just a couple of small notes.

Thanks @Mab879 

sorry for the delay, took a bit of a break :)
Should be fixed now!

Mab879 (approved):
No worries! Thanks for the PR!

PR #2368, opened 2026-06-06
test: use a unique temp dir in test_variable_in_filter

### LINKED ISSUE #1924: API/OVAL/unittests/test_variable_in_filter.sh uses /tmp/key_file
#### Description of Problem:
The `API/OVAL/unittests/test_variable_in_filter.sh` test writes to (and then removes) a `/tmp/key_file` file, which might conflict with other logic running on the same OS, which is the same OS (chroot) used for building the RPM.

Other files seem to correctly use `mktemp`.

If the `test_variable_in_filter.xml` cannot be dynamically sed-modified with the name of a temporary file, consider at least changing the name to something like `/tmp/openscap_unittests_test_variable_in_filter`.

Thanks!

#### OpenSCAP Version:
`maint-1.3` as of 2023-02-03

eran132:
## Problem

`tests/API/OVAL/unittests/test_variable_in_filter.sh` wrote its fixture to a
fixed path, `/tmp/key_file`, and the OVAL definition hardcoded `/tmp` as the
`file_object` / `file_state` path (#1924). A predictable, world-writable path
in a shared directory:

- races with parallel test runs (and with other users) on the same machine, and
- lets a pre-existing, unrelated `/tmp/key_file` left by someone else affect the
  result.

## Fix

Adopt the isolation pattern already used by `test_pcre_nonutf_characters.sh`:

- The OVAL definition now uses a `TEMP_DIR_PLACEHOLDER` token instead of `/tmp`.
- The test copies the definition to a temp file, `sed`-substitutes a per-run
  `mktemp -d` directory, and creates the `key_file` fixture inside it.
- The temporary directory is removed on completion.

The filename pattern (`^key_file$`) and the evaluation logic are unchanged, so
the test asserts exactly the same OVAL behavior — just in an isolated location.

## Verification

This is a test-only change. I verified mechanically (the full suite runs in CI):

- `bash -n` passes on the script.
- The `sed` substitution leaves zero `TEMP_DIR_PLACEHOLDER` occurrences and
  rewrites both `<ns3:path>` elements to the unique temp dir.
- The resulting OVAL definition is still well-formed XML.
- The `key_file` fixture is created in the temp dir as expected.

Fixes #1924

### DIFF
--- a/tests/API/OVAL/unittests/test_variable_in_filter.sh
+++ b/tests/API/OVAL/unittests/test_variable_in_filter.sh
@@ -4,11 +4,18 @@
 set -e
 set -o pipefail
 
-oval_def=`mktemp`
-result=`mktemp`
-stdout=`mktemp`
-stderr=`mktemp`
-temp_dir=`mktemp -d`
+oval_def=$(mktemp)
+result=$(mktemp)
+stdout=$(mktemp)
+stderr=$(mktemp)
+temp_dir=$(mktemp -d)
+
+cleanup() {
+	rm -f "$oval_def" "$result" "$stdout" "$stderr"
+	rm -rf "$temp_dir"
+}
+trap cleanup EXIT
+
 cp "$srcdir/test_variable_in_filter.xml" "$oval_def"
 sed -i "s;TEMP_DIR_PLACEHOLDER;$temp_dir;" "$oval_def"
 echo "secret_key" > "$temp_dir/key_file"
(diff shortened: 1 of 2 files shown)


eran132:
Thanks for the review. Addressed in 47338aaab:

- **Cleanup on failure:** added a `cleanup()` handler registered with `trap ... EXIT`, so the temp files and `mktemp -d` directory are removed even if an `assert_exists` fails under `set -e`.
- **Backticks → $(...):** switched, matching the sibling `test_pcre_nonutf_characters.sh`.
- **`sed -i` portability:** kept as-is intentionally — the sibling `test_pcre_nonutf_characters.sh` uses the identical `sed -i "s;...;...;"` form and these OVAL unit tests run on GNU/Linux CI, so this stays consistent with the existing convention rather than introducing a divergent in-place-edit approach.

jan-cerny (approved):

PR #2390, opened 2026-07-27
Fix build without crypto support

Eljees:
## Summary

- always build and link the CRAPI object on platforms with mmap support
- compile only the no-op `crapi.c` implementation when no digest backend is available
- keep the full CRAPI source set when Gcrypt or NSS is enabled

This lets non-crypto builds provide `crapi_init` without compiling digest code that requires a crypto backend.

Fixes #2310.

## Testing

Reproduced the failure on the unmodified commit: building the `oscap` target without Gcrypt fails to link with `undefined reference to 'crapi_init'`.

The patched tree builds successfully with:

```sh
cmake -DCMAKE_DISABLE_FIND_PACKAGE_GCrypt=TRUE \
  -DENABLE_OSCAP_UTIL_DOCKER=OFF \
  -DENABLE_TESTS=OFF ..
cmake --build . -j2
```

This builds `libopenscap`, `libopenscap_sce`, and the `oscap` CLI.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -59,7 +59,7 @@ if (ENABLE_PROBES)
 		)
 	endif()
 endif()
-if (HAVE_MMAN_H AND (GCRYPT_FOUND OR NSS_FOUND))
+if (HAVE_MMAN_H)
 	list(APPEND OBJECTS_TO_LINK_AGAINST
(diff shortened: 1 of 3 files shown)


jan-cerny (approved):
builds successfully using the provided command

Eljees:
Thank you!, i'm glad could have helped! ;)

PR #2391, opened 2026-07-28
Replace removed Python 2 C API macros for SWIG 4.5.0 compatibility

Mab879:
Backport from https://src.fedoraproject.org/rpms/openscap/pull-request/35

### DIFF
--- a/swig/openscap.i
+++ b/swig/openscap.i
@@ -48,8 +48,6 @@
 {
     if (PyLong_Check($input))
         $1 = (time_t) PyLong_AsLong($input);
-    else if (PyInt_Check($input))
-        $1 = (time_t) PyInt_AsLong($input);
     else if (PyFloat_Check($input))
         $1 = (time_t) PyFloat_AsDouble($input);
     else {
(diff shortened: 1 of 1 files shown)


swig/openscap.i
```diff
@@ -48,8 +48,8 @@
 {
     if (PyLong_Check($input))
         $1 = (time_t) PyLong_AsLong($input);
-    else if (PyInt_Check($input))
-        $1 = (time_t) PyInt_AsLong($input);
+    else if (PyLong_Check($input))
+        $1 = (time_t) PyLong_AsLong($input);
```
jan-cerny: This is the same as the previous branch.

Mab879: Can you please elaborate, this is (now) rebased on latest main so this shouldn't be the case?

jan-cerny: The `if` condition and statement is the same as the `else if` condition and statement. The `else if` is superflous and lines 51-52 should be removed.

```
    if (PyLong_Check($input))
        $1 = (time_t) PyLong_AsLong($input);
    else if (PyLong_Check($input))
        $1 = (time_t) PyLong_AsLong($input);
```

havardAasen:
Is this fix planned to be merged soon? This is starting to be a problem in most distros packaging this project.

Is there also a planned release date of 1.4.5? Which hopefully includes this fix.

swig/openscap.i
Mab879: Updated. Please take a look.

jan-cerny (approved):
The test fails on Fedora Rawhide aren't caused by the contents of this PR and will be fixed by https://github.com/OpenSCAP/openscap/pull/2410.

The fails of testing farm jobs are infrastructure problems (failed to setup the test environment) and aren't caused by the contents of this PR.

jan-cerny:
@havardAasen PR merged now. At this moment we don't have any specific plan for releasing the 1.4.5 version.

PR #2403, opened 2026-08-18
Do not use gcrypt cmake variables if gcrypt is not found

Tropicao:
Commit d12d820a9493 ("Fix build without crypto support") enabled building the crapi library without a crypto backend, but left the target_include_directories and target_compile_definitions calls unguarded. When no crypto library is available, the GCrypt find module leaves GCRYPT_INCLUDE_DIRS unset, so CMake fails at the generate step with:

  CMake Error: The following variables are used in this project, but
  they are set to NOTFOUND: GCRYPT_INCLUDE_DIR

The test command provided at this time made the build successful because it provided -DCMAKE_DISABLE_FIND_PACKAGE_GCrypt=TRUE, but if no crypto backend is provided, it does not make sense to provide those cmake configuration options preventing dependency search manually.

Guard both calls with a CRYPTO_FOUND check so they are only applied when an actual digest backend is present, even without providing -DCMAKE_DISABLE_FIND_PACKAGE_foo


Tropicao:
The failed jobs do not look related to the proposed change. I guess those are parts of what https://github.com/OpenSCAP/openscap/pull/2391 is aiming to fix ?

jan-cerny (approved):

Tropicao:
Hi @jan-cerny, I was actually about to ask for guidance about the remaining failure, as I suspect it to be completely unrelated, but you merged while I was typing my question, so... thanks :)

PR #2407, opened 2026-08-31
Prevent denial of service in filehash and filehash58 probes

jan-cerny:
The filehash and filehash58 probes call `open(pbuf, O_RDONLY)` without checking the file type. If the target is a FIFO (named pipe), `open()` blocks indefinitely waiting for a writer. The user simply runs `mkfifo ~/somefile` and when the compliance scan traverses their home directory, the entire oscap process hangs.

Additionally, a symlink to `/dev/zero` causes an infinite read loop in `crapi_mdigest_fd()`.

Both problems will be avoided by checking whether the file isn't blocking and is a regular file. The function `crapi_mdigest_fd` isn't modified by the commit but the issue is avoided because the function is guarded by the checks.

### Review hints
- Run Ctest test test_probes_filehash58_dos
- Create OVAL content that computes a hash of a `somefile`, create `somefile` named pipe using the `mkfifo somefile` command. Then run `oscap oval eval` with the OVAL content and verify that the `oscap` command isn't stuck but completes successfully. A similar OVAL is used in the test introduced in this PR.

### DIFF
--- a/src/OVAL/probes/independent/filehash58_probe.c
+++ b/src/OVAL/probes/independent/filehash58_probe.c
@@ -57,6 +57,7 @@
 #include "oscap_helpers.h"
 
 #define FILE_SEPARATOR '/'
+#define __ERRBUF_SIZE 128
 
 /* List of hash types listed in OVAL specification */
 static const char *OVAL_FILEHASH58_HASH_TYPES[] = {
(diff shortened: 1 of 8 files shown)


tests/probes/filehash58/test_probes_filehash58_dos.sh
```diff
@@ -0,0 +1,37 @@
...
+mkfifo /tmp/oscap-test-fifo
+
+# Burn CPU forever
+rm -f /tmp/oscap-test-symlink
+ln -s /dev/zero /tmp/oscap-test-symlink
+
+result="$(mktemp)"
+
+$OSCAP oval eval --results "$result" "$srcdir/dos.oval.xml"
+[[ $? -eq 0 ]]
```
Mab879: Isn't this unreachable to due `set -e`?

src/OVAL/probes/independent/filehash58_probe.c
```diff
@@ -186,6 +190,45 @@ static int filehash58_cb(const char *prefix, const char *p, const char *f, const
...
 
+	/*
+	 * Make sure we only try to hash regular files. Hashing a FIFO, socket,
+	 * device or other special file makes no sense and could block the read.
+	 * The type is checked on the already-opened descriptor (fstat) to avoid
+	 * a TOCTOU race with the preceding open().
+	 */
+	struct stat st;
+	if (fstat(fd, &st) != 0) {
+		char errbuf[__ERRBUF_SIZE] = {0};
```
Mab879, line 201: Claude is flagging that `__ERRBUF_SIZE` may be undefined here. Should also define it at the top of the file like the other probe.

Mab879 (approved):

PR #2411, opened 2026-09-03
Fix groff warning for nested "tagged paragraph"

havardAasen:
Warning can be reprodused with this command:
LC_ALL=C.UTF-8 MANROFFSEQ='' MANWIDTH=80 \
    man --warnings -E UTF-8 -l -Tutf8 -Z utils/oscap.8 >/dev/null

One paragraph is indented differently after this fix is implemented.


jan-cerny (approved):
the CI fails aren't caused by contents of this PR

